Issue #3, opened 2021-01-28
FermionicSumOp without primitive


manoelmarques:
Migrated from Enterprise Github.
Creator: @ikkoham

### Summary

Jupyter note.
https://gist.github.ibm.com/Ikko-Hamamura1/27c70e3d6c7e8c9b383dd04c1a0bd1b0

![classes](https://media.github.ibm.com/user/281886/files/5aac1c80-617c-11eb-866d-37ab0f599a6c)

manoelmarques:
Creator: @ikkoham

I've separated this Pull Request to #23. I will rebase once #23 is merged.

Issue #23, opened 2021-01-28
Finish the BosonicOp implementation


manoelmarques:
Migrated from Enterprise Github.
Creator: @mrossinek

The current `BosonicOp` is a mere copy of the `FermionicOp` with a search-and-replace of `fermionic` for `bosonic`. The actual mathematics of the implementation need to adapted.

- [ ] adapt the mathematics of the `BosonicOp`
- [ ] unittest all methods

manoelmarques:
Creator: @mrossinek

Pauline and I will look into getting the mathematics of this operator sorted out in early February.

mrossinek:
We had some more discussions offline with regards to the implementation of the `BosonicOp` and `SpinOp` (#22).
Here, I will try to summarize some of these thoughts.

The most general representation of a `BosonicOp` is very similar to that of the `FermionicOp` but differs in the commutation relation with respect to the represented particles. Any additional details of the representation such as a differentiation of modes and modals are imposed upon the operator by the specific problem one is interested in.
Therefore, it is also the task of the problem (which constructs the `BosonicOp` as part of the factory method `second_quantized_operators`) to handle such differentiation during the construction.

That being said, there is no longer a need for the `SpinOp` since it is a mere wrapper of the more general `BosonicOp`. We are therefore likely to remove it in favor of extracting the wrapper logic into the specific problem instances.

However, we also noticed that the `BosonicOp` cannot be constructed from a label which is similar to the style of the `FermionicOp` since it is possible to apply multiple `+` or `-` operations on the same mode. Even if we were to come up with characters for multiple such operations on a single index, the number of these is not really bound and therefore this approach does not scale.
Therefore, we would have to go for an approach such as `[2+]III` where we combine an operator character and a count. However, in doing so we loose all of the benefits which came with a 1D-label representation of our operator. The most important one being that we loose the information of the number of modes given by the length of the label.

This is means, that we need to come up with a better language (and internal representation) for the operators. A basic example could look like the following:
```
BosonicOp('+^2[0]', length=4)
```
which would construct an operator with 4 modes and applies a `+` on the 0th index, twice.
The list of possible operations should include: `+`, `-`, `X`, `Y` and `Z`. The identity no longer needs to be accounted for explicitly (since the language is sparse).

Here are some more requirements and thoughts we should keep in mind:
- the internal representation should be efficient for multiplication of two operators
- it should preserve the order in which operators are chained together

[...]

Issue #18, opened 2021-01-28
Implement the QubitMappers


manoelmarques:
Migrated from Enterprise Github.
Creator: @mrossinek

The state of the QubitMappers is the following:

- [x] Implementation
  - [x] JordanWignerMapper (fermionic)
  - [x] ParityMapper (fermionic)
  - [x] BravyiKitaevMapper (fermionic)
  - [x] BravyiKitaevSuperFastMapper (fermionic)
  - ~DirectMapper (bosonic)~
  - [x] LinearMapper (spin)
  - [x] LogarithmicMapper (spin)
- [x] Unittests
  - [x] JordanWignerMapper (fermionic)
  - [x] ParityMapper (fermionic)
  - [x] BravyiKitaevMapper (fermionic)
  - [x] BravyiKitaevSuperFastMapper (fermionic)
  - ~DirectMapper (bosonic)~
  - [x] LinearMapper (spin)
  - [x] LogarithmicMapper (spin)

Note: this issue may be broken down into several issues as work starts to progress.

manoelmarques:
Creator: @mrossinek

You can find an example of the new Jordan-Wigner mapping in action here:
https://gist.github.ibm.com/oss-zurich/a2b75f58d4cd2eae7df8e1f1e9dcd6b0

manoelmarques:
Creator: @woodsp-ibm

Can we look at avoiding `isinstance` and using methods implemented on a base class/interface to do things like this. If someone wants to add another operator that is of type Fermionic then an instance check like this is problematic. There is already some type field that seems to be passed to superclass - can we not have some method query based off that, or an extension thereof, in some way.

```
        if isinstance(particle_type, FermionicOperator):
            return True
        return False
```

manoelmarques:
Creator: @pbark

@woodsp I think that the idea behind this is that not all mappings work with all type of operators. e.g. if one provides a BosonicOperator in the JW mapping this is not going to work. Sure we can try to think ways of spliting the mappings (e.g. FermionicMappings VS BosonicMappings) but I am not sure how much complicated this can be.

manoelmarques:
Creator: @mrossinek

I partially agree with @woodsp here. The `isinstance` call is not the nicest solution here but I went for it because it was the quickest route to a prototype.
Could you explain what you mean by
> There is already some type field that seems to be passed to superclass - can we not have some method query based off that, or an extension thereof, in some way.

[...]

PR #25, opened 2021-01-29
FermionicOp

ikkoham:
### Summary

This is the PR for discussion.
Related to #3 

Jupyter notebook
https://gist.github.com/ikkoham/3526cbb006af17cb5c97315467e31190

TODO in this PR:
- [x] Add unittest
- [x] More documentation

Future work
- SpinOp, BoconicOp, SecondQuantizedOp
- Mapper
- Performance improvement (e.g. improve internal data structure using NumPy)


mrossinek (changes requested):
Thanks @ikkoham for the great work on this! I think this is definitely a step in the right direction :+1: I checked the code and validated the `JordanWignerMapper` as a basic test case (see also comments therein). Here are some more general observations:

- so far, only the `FermionicOp` is migrated, correct?
- the documentation strings will need some updates (can be done in the end)
- I did not yet validate all methods but this can probably done the easiest by starting to add some unittests

All in all, I really like where this PR is going! Please go ahead and continue the work on this by adding:
- unittests for the existing functionalities
- migrating also the Bosonic and Spin cases (bosonic logic does not need to be finished, that I will do later with @paulineollitrault (see also #23))
- documentation can be postponed until the end

I will now work on finalizing/updating the interface for the `QubitMapper` classes (#18). Once that is done we can figure out where to place the logic which you moved from the `FermionicOperator` into the `JordanWignerMapper`. I will update you here when that happens.

mrossinek:
I have some follow-up questions also referring to the UML from #3.

- Can you please provide some reasoning for the naming choice of `StarAlgebraMixin`?
- `SecondQuantizedOp` also implements that interface. Am I correct in the assumption that you therefore use this to ensure that all operators implement the same set of basic mathematical operations (+ dagger, equality comparison, etc.)?
- How far did you think about the further unification of the `ParticleOp` implementations? I see that we can extract more requirements into this interface such as `to_list`, `reduce`, etc.

ikkoham:
Thank you for the review. 

> so far, only the FermionicOp is migrated, correct?

Yes. I will do it for SpinOp and BosonicOp after your check. (So I start just now.)

> the documentation strings will need some updates (can be done in the end)
> I did not yet validate all methods but this can probably done the easiest by starting to add some unittests

Sure. I will add docs and unittests.

Please see my jupyter notebook in #3 for some tests including JW transformaiton.

ikkoham:
> Can you please provide some reasoning for the naming choice of StarAlgebraMixin?

I will rename to StarAlgebra and I want move this into opflow module. Star algebra means a ring which has adjoint.
So StarAlgebra does not have `__eq__`.

If possible, ParticleOp should have more methods as abstract method, but I'm not sure what method BosonicOp has. So I did not do it now.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -0,0 +1,304 @@
...
+        elif isinstance(data, list):
+            if not data and register_length:
+                self._labels = ["I" * register_length]
+                self._coeffs = [0]
+            elif not data:
+                raise QiskitNatureError(
+                    "Empty data requires register_length parameter."
+                )
+            else:
+                self._register_length = len(data[0][0])
```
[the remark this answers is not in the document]
ikkoham: If you frequently create operators with the same coefficients, I would like to create this or similar.
How often do you expect this?

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -0,0 +1,304 @@
...
+
+    def reduce(self) -> "FermionicOp":
+        """
+        Reduce
+
+        Returns:
+            The reduced `FermionicOp`
+        """
+        # TODO: atol, rtol
+        label_list, indexes = np.unique(self._labels, return_inverse=True, axis=0)
```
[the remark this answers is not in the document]
ikkoham: I think add operation is commutative since fermion follows CAR algebra.

mrossinek:
> Please see my jupyter notebook in #3 for some tests including JW transformaiton.

Thank you! I saw that shortly after I did my own test already :sweat_smile: But I also liked the overview of all the mathematical operations! They should be a good starting point for the unittests :+1: 

Feel free to only finalize the `FermionicOp` for now (including docs and unittests). Then we can iterate on that until we are happy and then we can create the `BosonicOp` (at least the framework) and `SpinOp` based on what we converged on with the `FermionicOp`.

> I will rename to StarAlgebra and I want to move StarAlgebra into opflow module. Star algebra means a ring which has adjoint.

Ah I see. Are there other classes in opflow which can also leverage this `StarAlgebra` interface? If yes, I am open for the move to Terra. Otherwise, I am not sure it will fit there.

> If possible, ParticleOp should have more methods as abstract method, but I'm not sure what method BosonicOp has. So I did not do it now.

Yes, let's wait until we know what the other classes will look like and then extract anything common in the end :+1:

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -0,0 +1,304 @@
...
+    def __len__(self):
+        return len(self._labels)
+
+    def to_opflow(self, method: str = "JW") -> PauliSumOp:
+        # TODO: other mappings
+        # if method == "JW":
+
+        # pylint: disable=cyclic-import
+        from qiskit_nature.mappings.jordan_wigner_mapping import JordanWignerMapping
+        return JordanWignerMapping().map(self)
```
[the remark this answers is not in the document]
ikkoham: I see. I think all logic for mapper should be in the mapping. So duplicated logic should be moved into parent class (QubitMapper). Please look at @t-imamichi 's suggestion.

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
mrossinek: I cannot really say how often this may occur but I think it is reasonable to assume that this occurs often enough to be supported. We can provide a limited version in which we allow either a 2D-list or a 1D-list with either a single coeff or a list of coeffs with matching length? I think this should be doable and maintainable with the least amount of edge cases.. :thinking:

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
mrossinek: Thanks! I had not seen that gist yet. I think it is going in the right direction but I am not a fan of the hard-coded class dependencies. I.e. `QubitMapper` needs to know about the classes which implement it as their interface..
I have some ideas on how to resolve and neatly implement all of this which I will present in our meeting on Wednesday :+1:

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
pbark: This is a quite common occurrence in the case of Lattice models (e.g. think of the hopping and on site interaction terms in the Fermi-Hubbard model). It is not so common in molecular problems.

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
pbark: @ikkoham you are right. The sum ordering is not important. The only point I see in @mrossinek comment is that preserving the order will be possibly easier to read the operator later on.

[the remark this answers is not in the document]
mrossinek: That is only the case if the operator is constructed in some meaningful order. Otherwise the sorted list may actually be easier to read later.. I don't have a strong opinion on whether we need to preserve the order (since mathematically it makes no difference). I don't think it causes much overhead to keep it sorted and it is probably what the end-user expects. So personally, I would sort it again, but we can also leave it be.

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
ikkoham: How about `0.25 * sum(FermionicOp(label) for label in ['NIII', 'INII', 'IINI', 'IIIN'])` for this?

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
ikkoham: OK. `Reduce()` is costly, so don't do it inside add. When users want to reduce, use reduce() explicitly.

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
mrossinek: I am fine with an approach like this. But if we choose to do so we need to document this properly. I.e. we need to carefully explain the kind of list (i.e. a 2D list) which is required for the constructor and we should add an example like the one above to show how a user can get around it.

mrossinek:
Hi @ikkoham. @stefan-woerner, @pbark and I discussed this a little bit more offline and we think it would be great to get this PR merged with only the `FermionicOp` ready for now, so that we can continue working on the other aspects of the workflow. The `BosonicOp` and `SpinOp` can follow later. Therefore, I think if the following checkpoints are fulfilled this PR should be in a good place to be merged later this week:

- [ ] fix the remaining comments of my review from yesterday
- [ ] add unittests based on your gist linked to in #3 
- [ ] update documentation to a degree where it's no longer wrong or out-of-date

We can also postpone improvements to the internal data storage for now.

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
ikkoham: Yes, I'll leave the design of the Mapper. However, the operator is not responsible for any of the mapping processing. (Single-responsibility principle)
You can move the processing of mapping anywhere but the operator. (super class or mixin or some new class)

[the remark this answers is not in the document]
mrossinek: Yes I actually had some more thoughts on where the mapper logic needs to go. You can remove it fully from the operator and we also probably won't need the `to_opflow()` method any more. More on that in our meeting later :+1:

qiskit_nature/operators/second_quantization/star_algebra_mixin.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class StarAlgebraMixin(ABC):
+    """The star algebra mixin class.
+
+    Star algebra means algebra with adjoint (dagger).
+    """
+
+    @abstractmethod
+    def mul(self, other: complex):
```
mrossinek, line 26: I think this splitting of the actual logic and the operator overloads is a good idea. However, in this specific case `complex` is quite restrictive. We should check what the opflow in Terra uses as the type here.

test/operators/test_fermionic_op.py
```diff
@@ -0,0 +1,162 @@
...
+        """Test __init__ of empty list"""
+        fer_op = FermionicOp([], register_length=length)
+        self.assertEqual(fer_op.to_list(), [("I" * length, 0)])
+
+    def test_init_multiterm(self):
+        """Test __init__ with multi terms"""
+        labels = [("N", 2), ("-", 3.14)]
+        self.assertListEqual(FermionicOp(labels).to_list(), labels)
+
+    def test_init_invelid(self):
```
mrossinek, line 52:
```suggestion
    def test_init_invalid(self):
```

mrossinek (commented):
I will go through this code again more carefully later but these are some minor comments I found will scrolling through

qiskit_nature/operators/second_quantization/star_algebra_mixin.py
ikkoham: `float` and `int` is a virtual subclass of `complex` in mypy. So I think this code is valid.

test/operators/test_fermionic_op.py
ikkoham: Thanks.

ikkoham:
I have changed some APIs.
- remove coeff from init
- dagger is property (if you don't like I can revert this)

mrossinek:
> I have changed some APIs.
> 
> * remove coeff from init

So by doing so there are the following possibilities to construct a `FermionicOp`:
```python
# empty operator; requires register_length to be set
FermionicOp([], register_length=n)

# from a single label
FermionicOp('NIII')

# with a 2D list
FermionicOp([('NIII', 0.5), ('INII', 0.5)])

# you can add a coefficient through multiplication:
op = 0.5 * FermionicOp('NIII')
```

To me, this looks like you can do everything necessary. @pbark what do you think?
However, I still think we need to clarify the documentation of the 2D list case.

> * dagger is property (if you don't like I can revert this)

I am fine with this.

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
mrossinek: This can be resolved since you removed `reduce()`, right?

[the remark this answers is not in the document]
ikkoham: Yes!

qiskit_nature/operators/second_quantization/star_algebra.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        raise NotImplementedError
+
+    def __add__(self, other):
+        return self._add(other)
+
+    def __sub__(self, other):
+        return self._add(other._multiply(-1))
+
+    def __mul__(self, other: complex):
```
woodsp-ibm, line 39: Why does this have a type on it. From what I can tell for add etc the type would end up being defined by the class using the mixin.

Also should we not have `__rmul__` etc here

qiskit_nature/operators/second_quantization/star_algebra.py
```diff
@@ -0,0 +1,86 @@
...
+    def _add(self, other):
+        r""" Return Operator addition of self and other, overloaded by ``+``.
+        """
+        raise NotImplementedError
+
+    def __add__(self, other):
+        return self._add(other)
+
+    def __sub__(self, other):
+        return self._add(other._multiply(-1))
```
woodsp-ibm, line 37: Would it be preferable to have an _sub method defined, like _add, but where is has the implementation here so you do not need to override by default. But if you had a better way to do sub then you could. Any why would you not say `other * -1` or even `-other` rather than directly calling upon some method here to do it.

qiskit_nature/operators/second_quantization/star_algebra.py
```diff
@@ -0,0 +1,86 @@
...
+
+"""The Sum Operator base interface."""
+
+from abc import ABC, abstractmethod
+
+from .multiply import MultiplyMixin
+from .tolerances import TolerancesMixin
+
+
+class StarAlgebraMixin(MultiplyMixin, TolerancesMixin, ABC):
```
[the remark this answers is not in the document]
ikkoham: I want to use quantum_info's mixin as possible. Common mixin is better. Please see [`LinearOp`](https://github.com/Qiskit/qiskit-terra/blob/ac15bd1870eee159e65e28deb26cb05f88c50e10/qiskit/quantum_info/operators/linear_op.py) for example

qiskit_nature/operators/second_quantization/star_algebra.py
ikkoham: Thanks. You are right. This type hint is unnecessary.

quantum_info's multiplication is a bit strange from fermionic_op and opflow. `*` is for scalar multiplication and dot, `@` is for compose(). Note that this compose is a product of opposite direction. (i.e. A * B = B @ A) This property is incompatible for opflow and seocnd_quantized_op, so I made StarSlgebraMixin. After this PR, I am planning to move this mixin to opflow as a part of opflow refactoring.

qiskit_nature/operators/second_quantization/star_algebra.py
ikkoham: Indeed, if you want to override `__sub__`, it is better to prepare `_sub` and override it. Adding a method does not involve deprecation, so it is easy to do even after release.

> Any why would you not say other * -1 or even -other rather than directly calling upon some method here to do it.

There is no deep reason for this, but I feel that it is better to have a shallow hierarchy of calls as possible.

qiskit_nature/operators/second_quantization/star_algebra.py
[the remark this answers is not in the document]
mrossinek: Okay I see. And the reason for currently duplicating these interfaces here is that https://github.com/Qiskit/qiskit-terra/pull/5617 is not merged yet, right?
Once that is part of Terra, we can replace the imports of our Mixin's with the `quantum_info`-ones?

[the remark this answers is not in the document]
ikkoham: Yes. That is the plan.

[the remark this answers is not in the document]
mrossinek: Alright I also just saw your other comment to Steve's question below. Therefore there are two aspects to this:
1. once the mixin's are merged in Terra we can use those for `Multiply` etc.
2. you will PR the `StarAlgebra` one into Terra because that is currently not supported there.

Since you mentioned `LinearMixin`: why don't you use that here, too?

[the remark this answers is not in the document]
ikkoham: In `LinearMixin`, there are an parameter `qargs` which our operators don't support. I think we can modify our operator to use it.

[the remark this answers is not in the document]
mrossinek: I see. `qargs` is used to specify which subset is being acted on, right? While the name does not fit our needs at all, do you see an example where this could be useful? I can't see a use in our addition but maybe in our multiplication if we want to do something like the following:
```python
op = FermionicOp('I' * 4)
plus = FermionicOp('+', qargs=0)
new_op = op @ plus
# new_op.label should now be '+III'
```
Is this what this could look alike?

qiskit_nature/operators/second_quantization/star_algebra.py
mrossinek: So far I cannot think of a reasonable scenario in which the existence of `_sub` would be useful. And even if it does occur, as @ikkoham has pointed out, we can still add it later on, if needed.
However, I would like to see `-other` here as I think it easier to read. And as you point out yourself, it should not change the performance :+1: 
```suggestion
        return self._add(-other)
```

PS: this is also what the `__sub__` method looks like in the currently open PR of Terra.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -23,17 +23,84 @@
...
+    **Label**
+
+    Allowed characters for the label are `I`, `-`, `+`, `N`, and, `E`.
+
+    .. list-table::
+        :header-rows: 1
+
+        * - Label
+          - Mathematical Representation
+          - Mean
```
mrossinek, line 38:
```suggestion
          - Meaning
```

ikkoham:
I have added the operation: `0 + FermionicOp` equals to `FermionicOp`.
This is convenient because it has the following usage:
```python
fer_op = 0

for idx in itertools.product(range(len(one_body_ints)), repeat=2):
    coeff = one_body_ints[idx]
    if not coeff:
        continue
    label = 'I' * register_length
    base_op = FermionicOp(label) * coeff
    for i, op in [(idx[0], '+'), (idx[1], '-')]:
        label_i = label[:i] + op + label[i+1:]
        base_op @= FermionicOp(label_i)
    fer_op += base_op
```

And `sum(list or generator of FermionicOp)` is also allowed now.

If you feel it's odd that left addition of int isn't allowed (i.e. FermionicOp + 0 is NOT allowed), I'll fix it. But I haven't added it yet because I don't see a use case for it other than the above, and I don't think it should be used for anything other than the above.

qiskit_nature/operators/second_quantization/star_algebra.py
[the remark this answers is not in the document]
ikkoham: Yeah. I agree. `qargs` might be useful. However, we don't have enough time in this PR. I may enhance FermionicOp after this PR.

mrossinek:
> I have added the operation: `0 + FermionicOp` equals to `FermionicOp`.
> And `sum(list or generator of FermionicOp)` is also allowed now.

I think the `sum(...)` is really important! Thanks for adding support for this case!

> If you feel it's odd that left addition of int isn't allowed (i.e. FermionicOp + 0 is NOT allowed), I'll fix it. But I haven't added it yet because I don't see a use case for it other than the above, and I don't think it should be used for anything other than the above.

I also think that this is fine. But maybe add that note in the documentation of the right addition function. Just so that it is written somewhere. If anyone ever needs this it should be trivial to add though, right?

> In addition, I think that the empty operator is no longer necessary since the above is added. Can I remove it?

Please keep it. I think there could be cases where it can pop up and where we would want it to be of the same type rather than using `0` which is an int and therefore not of the same type. It does not hurt to keep it, or do you see a problem with something?

qiskit_nature/operators/second_quantization/star_algebra.py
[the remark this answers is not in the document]
mrossinek: Sounds good to me :+1: I'll resolve this for now then.

ikkoham:
> It does not hurt to keep it, or do you see a problem with something?

Pros: We can remove `register_length` parameter. So, the initialization becomes simple. Moreover, we can write like `FermionicOp((I * register_length, 0))`. This is same with `FermionicOp([], register_length=register_length)`. So, removal of empty list op doesn't hurt the usability.

mrossinek:
> Pros: We can remove `register_length` parameter from `__init__`. So, the initialization becomes simple. Moreover, we can write like `FermionicOp((I * register_length, 0))`. This is same with `FermionicOp([], register_length=register_length)`. So, removal of empty list op doesn't hurt the usability.

Okay I am open to have it removed then. @pbark Any objections?

ikkoham:
I decided to remove register_length once. But if you feel it is necessary, please feel free to revert commit cb29270.

pbark:
I also think that removing the `register_length` a good idea! Thanks for doing it!

ikkoham:
Done for this PR. I will improve the FermionicOp another PR.

This is current class diagram.
classes

qiskit_nature/operators/second_quantization/particle_op.py
```diff
@@ -12,44 +12,48 @@
...
-        self._particle_type = particle_type
-
-    @property
-    def particle_type(self):
-        """Return the particle type"""
-        return copy.deepcopy(self._particle_type)
-
     @property
     @abstractmethod
     def register_length(self) -> int:
```
mrossinek, line 32: ~~I guess we can only remove this once we have also removed it from `BosonicOp` and `SpinOp`, right?~~

EDIT: sorry, I didn't realize we only removed it from `__init__`. Do we actually still need it though? The length of the label should be sufficient, right?

qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -0,0 +1,91 @@
...
+from .spin_op import SpinOp
+from .star_algebra import StarAlgebraMixin
+
+
+class SecondQuantizedOp(StarAlgebraMixin):
+    """A general SecondQuantizedSumOp.
+
+    This class represents sums of mixed operators, i.e. linear combinations of
+    SecondQuantizedOperators with identical particle type registers.
+    """
```
mrossinek, line 31:
```suggestion
    """A general second-quantized operator.

    This class is used to combine operators of different particle type.
    """
```

qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -0,0 +1,91 @@
...
+
+        operator_list = []
+        if self._fermion is not None:
+            operator_list.append(other * self._fermion)
+
+        return SecondQuantizedOp(operator_list)
+
+    def compose(self, other):
+        if not isinstance(other, SecondQuantizedOp):
+            raise TypeError("Unsupported operand type(s) for @: 'SecondQuantizedSumOp' and "
```
mrossinek, line 65:
```suggestion
            raise TypeError("Unsupported operand type(s) for @: 'SecondQuantizedOp' and "
```

Please also check the other occurences of the old class name.

qiskit_nature/operators/second_quantization/particle_op.py
```diff
@@ -12,44 +12,48 @@
...
         """Getter for the length of the particle register that the SumOp acts on."""
         raise NotImplementedError
 
-    @property
     @abstractmethod
-    def operator_list(self) -> List[Any]:
-        """Getter for the operator_list of the `SumOp`"""
+    def to_opflow(self, method) -> PauliSumOp:
+        """TODO"""
         raise NotImplementedError
```
mrossinek, line 39: Please remove this method from the interface and all subclasses.
We won't need it in this form.

qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        Mapping from SecondQuantizedOp to PauliSumOp
+        """
+        raise NotImplementedError
+
+    def transform(self):
+        """
+        Transformation from SecondQuantizedOp to SecondQuantizedOp
+        """
+        raise NotImplementedError
```
mrossinek, line 91: Please remove these method stubs for now. If we really choose that we want them, we can add them later.

qiskit_nature/operators/second_quantization/star_algebra.py
```diff
@@ -0,0 +1,113 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""The Sum Operator base interface."""
```
mrossinek, line 13: Please update docstring.

qiskit_nature/operators/second_quantization/star_algebra.py
```diff
@@ -0,0 +1,113 @@
...
+from abc import ABC, abstractmethod
+
+from .multiply import MultiplyMixin
+# TODO: Move import path after Terra#5617 is merged.
+
+
+class StarAlgebraMixin(MultiplyMixin, ABC):
+    """The star algebra mixin class.
+
+    Star algebra is an algebra with adjoint (dagger).
```
mrossinek, line 24:
```suggestion
    Star algebra is an algebra with an adjoint.
    The :meth:`adjoint` can also be accessed via the `dagger` property for convenience.
```

Is there some special tag for `property` or would one also use `:meth:` there?

test/operators/test_fermionic_op.py
```diff
@@ -0,0 +1,159 @@
...
+        fer_op = FermionicOp([("+N", 3), ("N-", 1)]) ** 2
+        self.assertListEqual(fer_op.to_list(), [("+-", -3)])
+
+        fer_op = (3 * FermionicOp("IIII")) ** 3
+        self.assertListEqual(fer_op.to_list(), [("IIII", 27)])
+
+        fer_op = FermionicOp([("+N", 3), ("E-", 1)]) ** 0
+        self.assertListEqual(fer_op.to_list(), [("II", 1)])
+
+    def test_dagger(self):
```
mrossinek, line 144: I think this should rather test `adjoint` directly and have another special case which asserts that `dagger` works too.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -0,0 +1,368 @@
...
+
+from typing import Dict, List, Optional, Tuple, Union, cast
+
+import numpy as np
+from qiskit.opflow import PauliSumOp
+
+from qiskit_nature import QiskitNatureError
+
+from .particle_op import ParticleOp
+# TODO: Remove this after Terra#5617 is merged.
```
mrossinek, line 23:
```suggestion
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -0,0 +1,368 @@
...
+          - :math:`c`
+          - Annihilation operator
+        * - `+`
+          - :math:`c^\dagger`
+          - Creation operator
+        * - `N`
+          - :math:`n = c^\dagger c`
+          - Number operator
+        * - `E`
+          - :math:`I - n = c c^\dagger`
```
mrossinek, line 53:
```suggestion
        * - `-`
          - :math:`a`
          - Annihilation operator
        * - `+`
          - :math:`a^\dagger`
          - Creation operator
        * - `N`
          - :math:`N = a^\dagger a`
          - Number operator
        * - `E`
          - :math:`I - N = a a^\dagger`
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -0,0 +1,368 @@
...
+    The FermionicOp can be initialized in several ways:
+
+        `FermionicOp(label)`
+          The labels consists of the characters above.
+
+        `FermionicOp(tuple)`
+          The tuples are the form `(label, coeff)`. Coeff is int, float, or complex.
+
+        `FermionicOp(list)`
+          The lists are the form `[(label, coeff)]`. That is, lists of above tuples are allowed.
```
mrossinek, line 67:
```suggestion
        `FermionicOp(label)`
          A label consists of the permitted characters listed above.

        `FermionicOp(tuple)`
          Valid tuples are of the form `(label, coeff)`. `coeff` can be either `int`, `float`, or `complex`.

        `FermionicOp(list)`
          The list must be a list of valid tuples as explained above.
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -0,0 +1,368 @@
...
+
+    **Algebra**
+
+    FermionicOp supports some basic algebras: addition, subtraction, scalar multiplication,
```
mrossinek, line 71:
```suggestion
    `FermionicOp` supports the following basic arithmetic operations: addition, subtraction, scalar multiplication,
```

woodsp-ibm:
Do we want to have a name string that can be set on the operator - maybe entered on constructor with some default so you don;t need to. It would then be printed from any string representation of class. This was done in the past as it was convenient to be able to name the operators for users to see what it was (analagous to what is done in QuantumCircuit and circuit library where you can give a readable name to the circuit) and be used to name it when needed - e.g. if one computes the expectation values and gives some user friendly algorithm result the name can be part of the result.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -0,0 +1,368 @@
...
+      print("Sum")
+      print(0.25 * sum(FermionicOp(label) for label in ['NIII', 'INII', 'IINI', 'IIIN']))
+      print("Operator multiplication")
+      print(FermionicOp("+-") @ FermionicOp("E+"))
+      print("Dagger")
+      print(FermionicOp("+").dagger)
+
+    Basically, addition is defined between the :class:`FermionicOp`, but only `0 + FermionicOp` is
+    allowed as an exception. This is because it makes sum (in above example) valid, and because it
+    is useful in the following use cases.
```
mrossinek, line 90:
```suggestion
    In principle, you can also add :class:`FermionicOp` and integers, but the only valid case is the addition of `0 + FermionicOp`.
    This makes the `sum` operation from the example above possible and it is useful in the following scenario:
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -0,0 +1,368 @@
...
+                    conjugated_coeff *= (-1) ** (
+                        permute_through.count("+") + permute_through.count("-")
+                    )
+
+            label_list.append(daggered_label)
+            coeff_list.append(conjugated_coeff)
+
+        return FermionicOp(list(zip(label_list, coeff_list)))
+
+    def reduce(self, atol: Optional[float] = None, rtol: Optional[float] = None) -> "FermionicOp":
```
mrossinek, line 332: Here we are missing a docstring.

mrossinek (changes requested):
Alright in general this PR looks quite good already. I did some nitpicking but we should be able to get this merged tomorrow. :+1:

qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -0,0 +1,91 @@
...
+        # TODO: implement
+
+    def adjoint(self):
+        daggered_operator_list = []
+        if self._fermion is not None:
+            daggered_operator_list.append(self._fermion.dagger)
+
+        return SecondQuantizedOp(daggered_operator_list)
+
+    def map(self):
```
woodsp-ibm, line 81: This was something I mentioned in the design review as I had not seen it in the UML. It seemed the present thinking that a mapping conversion was another object that would take a 2nd quant Op and map it to an opflow op.

qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        Mapping from SecondQuantizedOp to PauliSumOp
+        """
+        raise NotImplementedError
+
+    def transform(self):
```
woodsp-ibm, line 87: I thought we did not have any transformations at this level anymore

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -0,0 +1,368 @@
...
+
+    def to_opflow(self, method: str = "JW") -> PauliSumOp:
+        # TODO: other mappings
+        # if method == "JW":
+
+        # pylint: disable=cyclic-import
+        from qiskit_nature.mappings.jordan_wigner_mapping import \
+            JordanWignerMapping
+
+        return JordanWignerMapping().map(self)
```
woodsp-ibm, line 364: See also the ParticleOp 'map' method comment below. If this is here in my mind it should be map(self, Mapper) where a given mapper instance is passed. Or at the very least if there is a shortcut string that is wanted for convenience maybe we can do that via a Mapper Factory or something. I had the impression though from the review that having a mapping function in the operator was being pushed back against.

qiskit_nature/operators/second_quantization/star_algebra.py
```diff
@@ -0,0 +1,113 @@
...
+    def __radd__(self, other):
+        if other == 0:
+            return self
+        return self.add(other)
+
+    def __sub__(self, other):
+        return self.add(-other)
+
+    def _add(self, other):
+        return self.add(other)
```
woodsp-ibm, line 71: I guess I was expecting `__add__` to call `self._add()` where `_add()` was the abstract method that would be implemented by a subclass. But there is `add()` which seems to have this role and its public so I guess you can use an operator or call .add(). I have not figured the role for _add()

qiskit_nature/operators/second_quantization/second_quantized_op.py
mrossinek: Yes this is a left-over. I asked to remove it above.

qiskit_nature/operators/second_quantization/particle_op.py
ikkoham: Basically, if you take the length of the label, you can get it as information. However, the shape (#terms, lem (label)) of the operator is often used, so I think it is better to have the method or property. I think there is room for renaming, but would be better to have such a function.

qiskit_nature/operators/second_quantization/star_algebra.py
ikkoham: Thanks. Sorry, I forgot to remove this private method. (I have added this when I try to use LinearMixin but now I decided the implementation is future work.) I will remove it.

qiskit_nature/operators/second_quantization/star_algebra.py
ikkoham: Thanks. It is pending whether to put the mixin class in the Qiskit docs, so adding the sphinx markup will be in another PR when we need it.

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Super class has the docstring. (maybe not enough)

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: I would like to use `c` for fermion annihilation and `b` for boson. I would like to save `a` for an usual harmonic oscillator. This may be the notation of physicists. What do you think @pbark?

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Sorry for my mistake. The doc is not generated correctly. I'll fix it.

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Thank you. I've removed `to_opflow()`. When the design of Mapper is close to completed, we may revive it, but let's put it on hold for now.
I think (self, Mapper) is very nice because it can avoid the cyclic-dependency.

ikkoham:
As for name, it's not obvious about arithmetic operations. Should the name of A + B be A.name + B.name or A.name or B.name? Anyway, we don't have time to discuss the design in this PR. If it becomes necessary, we can add it.

qiskit_nature/operators/second_quantization/fermionic_op.py
mrossinek: Sure no problem :+1: If you have a clear plan for a consistent naming that is fine by me :slightly_smiling_face:

mrossinek: Could we still make `n` and `e` capital letters, though?

test/operators/test_fermionic_op.py
```diff
@@ -0,0 +1,164 @@
...
+        self.assertListEqual(fer_op.to_list(), [("+-", -3)])
+
+        fer_op = (3 * FermionicOp("IIII")) ** 3
+        self.assertListEqual(fer_op.to_list(), [("IIII", 27)])
+
+        fer_op = FermionicOp([("+N", 3), ("E-", 1)]) ** 0
+        self.assertListEqual(fer_op.to_list(), [("II", 1)])
+
+    def test_adjoint(self):
+        """Test dagger"""
```
mrossinek, line 145:
```suggestion
        """Test adjoint method and dagger property"""
```

mrossinek (commented):
Thanks @ikkoham for the great work here! I think this looks almost good to go :+1: There are only two or three minor comments left.

As per the addition of a `name` attribute. I agree, that it is non-trivial for the arithmetics. However, I also see Steve's use case. How about we go for a middle ground:
We can add the `name` attribute to the `SecondQuantizedOp` (not the `ParticleOp` and not in this PR anyways). And if we perform arithmetic with it, we keep the name if it is identical and otherwise reset it to `None` (I think this is an unlikely event anyways).
But the benefit of having such a `name` is there, but I think it is sufficient on the `SecondQuantzedOp`-interface.

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: I feel that `N` is total number operator (N = \sum_i n_i). So I use small letter `n` here and I didn't introduce `e` for mathematical expression.

qiskit_nature/operators/second_quantization/multiply.py
```diff
@@ -0,0 +1,47 @@
...
+        return self._multiply(-1)
+
+    @abstractmethod
+    def _multiply(self, other):
+        """Return the {cls} other * self.
+        Args:
+            other (complex): a complex number.
+        Returns:
+            {cls}: the {cls} other * self.
+        """.format(cls=type(self).__name__)
```
mrossinek, line 47: Okay just one final question: in the `StarAlgebraMixin` you now went with `add` as a public method. Is there a reason for sticking to a private one here? Otherwise we could probably replace `_multiply` with `mul`, right?

ikkoham: Yes. But MultiplyMixin is quantum_info's class in the future. I wrote a related comment in the Terra PR. Since Terra's PR is not merged now, the specification here may be changed. If the Terra's PR is merged I will change here.

(Ane the name `mul` is compatible with opflow.)

mrossinek: Alright sounds good to me. Once the Terra PR is merged we can directly use the code there rather than adapting the interface here though :+1:

mrossinek (approved):
Thanks for the great work, @ikkoham!

ikkoham:
Thank you very much for your review @mrossinek @woodsp-ibm. Thanks to your review, the quality of PR has improved a lot. This is the beginning of development, and I would like to continue to improve it.

pbark (approved):
Everything looks good to me. I think we can merge this 👍

pbark (changes requested):
I just noticed that we also need a reno.

ikkoham:
I propose to write one release note for the operator class (or Qiskit Nature) because this is the first release of Qiskit-nature.

pbark:
Ok, let’s go with this then.

pbark (approved):

woodsp-ibm:
In regards of the release note I guess it needs to be coordinated in some way with the migration guide that will be done around the deprecation of qiskit-aqua and the new/moved/refactored code in Terra and the domain repos that replaces it.

Issue #22, opened 2021-01-28
Finish the SpinOp implementation


manoelmarques:
Migrated from Enterprise Github.
Creator: @mrossinek

- [ ] implement the `to_opflow()` methods for both, `primitives.SpinOperator` and `SpinSumOp`
- [ ] unittest all methods

manoelmarques:
Creator: @mrossinek

Below is Simon's original code for the `to_qiskit()` mapping which needs to be rewritten to use the `PaulliSumOp` (rather than the legacy `WeightedPauliOperator`) and can be used as a starting point for the `to_opflow()` method.
```python
    def to_qiskit(self, mode=default_spin_mode, precomputed_XYZI_transform=None):
        """
        Creates a qiskit.aqua.operators.WeightedPauliOperator from `self` that maps the spinS
        system operator to a qubit operator.

        Args:
            mode (str):
                The mode with which the spin operators should be mapped to qubits. Should be one of
                ['lin_encoding', 'log_encoding'].

            precomputed_XYZI_transform (list):
                This list must contain exactly 4 elements corresponding to the transformations for the X, Y, Z and
                the identity operator. Each of these 4 elements must be a qiskit.aqua.operators.WeightedPauliOperator
                in `pauli` representation (i.e. a list of Pauli strings)

        Returns:
            qiskit.aqua.operators.WeightedPauliOperator
        """
        # 1. Setup
        # Get the dimension of the internal spin Hilbertspace per spin system
        dim_S = int(2 * self.S + 1)

        # Initialize an empty list, used to store the embedded qubit operators
        # used to construct the final aqua operator.
        operatorlist = []

        # 2. Generate the mapping (spin S system --> qubits)
        # Map the X, Y, Z operators of an individual spin system to the `nqubit` system.
        # The final operator will be the sum of a tensor product of these operators with
        # complex coefficients.
        if self.transformed_XYZI is None and precomputed_XYZI_transform is None:
            # if neither is given, calculate a specified transform and store it in self.transformed_XYZI
            print('Calculating encoding for operator')
            if mode == 'log_encoding':
                self.transformed_XYZI = self._logarithmic_encoding()
            elif mode == 'lin_encoding':
                self.transformed_XYZI = self._linear_encoding()
            else:
                raise UserWarning("`mode` for BaseSpinOperator to qubit transform must be one of "
                                  "['log_encoding', lin_encoding'].")

        if precomputed_XYZI_transform is None:
[...]

PR #35, opened 2021-02-08
SpinOp

ikkoham:
### Summary

Issue #22.

Jupyter notebook: https://gist.github.com/ikkoham/9c6604b482a165d2731620465cc59ca5

Question:

1. Is compose (matrix multiplication) which cannot be found in [the original](https://github.com/Qiskit/qiskit-nature/blob/34d460f06c9bad858aa413f2821becb2c5b33752/qiskit_nature/operators/second_quantization/primitives/spin_operator.py) necessary? 

TODO:
- [x] Implement reduce()
- [x] Unit test
- [x] label
- [x] plus and minus label
- [x] Documentation

Future Work (not included in this PR)
- Implementation of compose


qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -18,111 +18,168 @@
...
+        for pos, (n_x, n_y, n_z) in enumerate(zip(self.x[i], self.y[i], self.z[i])):
+            label = []
+            if n_x > 0:
+                label.append(f'X^{n_x}')
+            if n_y > 0:
+                label.append(f'Y^{n_y}')
+            if n_z > 0:
+                label.append(f'Z^{n_z}')
+            if not label:
+                label.append(f'I')
```
manoelmarques, line 172:
```suggestion
                label.append('I')
```

manoelmarques (approved):
Suggestion to fix lint error

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: Thanks!

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -18,111 +18,497 @@
...
+            [
+                [(self.spin - i) if i == j else 0 for j in range(self._dim)]
+                for i in range(self._dim)
+            ],
+            dtype=np.complex128,
+        )
+
+    def _parse_sparse_label(self, labels):
+        xyz_dict = {"I": 0, "X": 1, "Y": 2, "Z": 3}
+        re_index = re.compile(r"(?<=[IXYZ]_)\d+((?=\^)|$)")
```
mrossinek, line 422: I generally think that complex regex expressions like this ones should be documented in a comment.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -18,111 +18,497 @@
...
+                    raise ValueError("Label must be XYZ order.")
+                if (
+                        datum[0] == 2
+                        and self._spin_array[i][3 * self._register_length - datum[1] - 1] > 0
+                ):
+                    raise ValueError("Label must be XYZ order.")
+
+                if self._spin_array[i][datum[0] * self._register_length - datum[1] - 1] != 0:
+                    raise ValueError("Duplicate label.")
+                self._spin_array[i][datum[0] * self._register_length - datum[1] - 1] = datum[2]
```
mrossinek, line 476: Please add some comments here. Spotting the differences between the different checks is rather tedious and a comment would help with that.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -18,111 +18,497 @@
...
+        for i, label in enumerate(labels):
+            for pos, char in enumerate(label):
+                if char == "X":
+                    self._spin_array[i][pos] = 1
+                elif char == "Y":
+                    self._spin_array[i][self._register_length + pos] = 1
+                elif char == "Z":
+                    self._spin_array[i][2 * self._register_length + pos] = 1
+
+    def _parse_ladder(self, data):
```
mrossinek, line 491: Please ensure to document somewhere, that this method *must* be called before parsing a sparse or dense label because those respective methods cannot handle ladder operators themselves. It may even be a good idea to assert that neither label contains a `+` or `-` in the beginning of the actual parse methods.

test/operators/test_spin_op.py
```diff
@@ -0,0 +1,137 @@
...
+import numpy as np
+from ddt import data, ddt
+
+from qiskit.quantum_info import Pauli
+from qiskit_nature.operators import SpinOp
+
+
+def spin_labels(length):
+    """Generate list of fermion labels with given length."""
+    return ["".join(label) for label in product(["I", "X", "Y", "Z"], repeat=length)]
```
mrossinek, line 28: Please also test `+` and `-`

test/operators/test_spin_op.py
```diff
@@ -0,0 +1,137 @@
...
+        """Test reduce"""
+        actual = (self.heisenberg - self.heisenberg).reduce()
+        self.assertListEqual(actual.to_list(), [("I_1 I_0", 0)])
+
+    def test_to_matrix(self):
+        """Test to_matrix()"""
+        actual = SpinOp([("X_2 Y_1 Z_0", 1)]).to_matrix()
+        desired = Pauli("XYZ").to_matrix() / 8
+        np.testing.assert_array_almost_equal(actual, desired)
+
```
mrossinek, line 134: Please add tests to separately test the parsing logic. That is very complex and we should make sure it does not break.

mrossinek (changes requested):
Thanks @ikkoham for the great work, once again! :+1: 
There are still some minor things which need to be adapted but I did (as usual) already go into the nitty-gritty details with this review.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -18,111 +18,497 @@
...
+                Tuple[np.ndarray, np.ndarray],
+            ],
+            spin: float = 1 / 2,
+            label_mode: str = "sparse",
+    ):
+        r"""Initialize ``SpinOp``.
+
+        Args:
+            data: label string or list of labels and coefficients. See documentation of SpinOp for
+                  more details.
```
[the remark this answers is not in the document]
ikkoham: OK. I added to docs.

test/operators/test_spin_op.py
ikkoham: The test is [here](https://github.com/Qiskit/qiskit-nature/pull/35/files/564d221c2c4a288ba967673aaccce54369a1228b#diff-e1fad0f243ef03387da83f3f9c9b006212b413a01addd08569d3d78048469dd1R65-R80) since the logic of test is different.

test/operators/test_spin_op.py
ikkoham: end2end `parsing_logic` was tested [here](https://github.com/Qiskit/qiskit-nature/pull/35/files#diff-e1fad0f243ef03387da83f3f9c9b006212b413a01addd08569d3d78048469dd1R60-R98).
The parse function cannot be used independently. What tests do you think is necessary?

test/operators/test_spin_op.py
mrossinek: Alright, but please also add a variant where you test `+` and `-` in the dense format.

test/operators/test_spin_op.py
mrossinek: I was thinking that it may make sense to explicitly test the parsing methods like `_flatten_ladder_ops`, `_parse_dense_labels` and `_parse_sparse_labels`. Currently, all of these are always tested within the complete flow but if that test fails for some reason we have multiple places to track down. Thus, if we have single tests for each of these methods we can more easily see which part fails :+1:

test/operators/test_spin_op.py
ikkoham: OK, thanks

test/operators/test_spin_op.py
ikkoham: I see. I have added a test for `_flatten_ladder_ops`, but I cannot make tests for `_parse_dense_labels` and `_parse_sparse_labels ` because it's not pure functions.

mrossinek: I am sorry but I don't fully understand what you mean by "not pure functions"

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -18,111 +18,497 @@
...
+            minus_indices = [m.start() for m in pattern_minus.finditer(label)]
+            len_plus = len(plus_indices)
+            len_minus = len(minus_indices)
+            pm_indices = plus_indices + minus_indices
+            label_list = list(label)
+            for indices in product(["X", "Y"], repeat=len_plus+len_minus):
+                for i, index in enumerate(indices):
+                    label_list[pm_indices[i]] = index
+                phase = indices[:len_plus].count("Y") - indices[len_plus:].count("Y")
+                new_data.append((''.join(label_list), coeff * 1j ** phase))
```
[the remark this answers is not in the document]
mrossinek: @ikkoham please also address this. Other than that and the last discussion point on the tests I think this PR is almost ready to go :+1:

test/operators/test_spin_op.py
ikkoham: They don't have return value but side effects.
https://en.wikipedia.org/wiki/Pure_function

mrossinek: Aha I see. But surely you can just test that the side effects are what you expect? I don't see a problem with that

qiskit_nature/operators/second_quantization/spin_op.py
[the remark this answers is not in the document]
ikkoham: I refactored the code and it becomes easier to read. I have added the comment a bit, but it may not be enough.

test/operators/test_spin_op.py
ikkoham: It cannot be called by itself. They only have side effects (assign to instance attributes). 
Since `self` is required to call them, they are necessary to call `__init__`, and they are essentially same with the `__init__` tests.

mrossinek: How about the following:
1. construct an empty operator: `spin_op = SpinOp(<whatever needs to go here in that case>)`
2. `spin_op._parse_dense_label(['XX'])`
3. assert the new contents of `spin_op`

And then something analogous for the sparse label parser

ikkoham: In general, tests are not necessary for private methods.

In this case, these methods cannot be called by itself. The purpose of these methods is to improve readability. Even if tests are not enough we should add tests for public methods. If tests are needed here, I'll move them to `__init__`.

mrossinek: I think it doesn't hurt to test private methods but if you feel so strong about this, we can leave it as is. I agree that in principle these methods will never be called by a user as is, but since we do have them split up for readability in my mind it makes sense to attempt to test them separately as well, as it allows a more fine-grained testing of the code.

But you can leave it as is.

qiskit_nature/operators/second_quantization/spin_op.py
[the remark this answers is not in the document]
mrossinek: I'll continue this discussion on the updated code snippet.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,502 @@
...
+            plus_indices = [i for i, char in enumerate(label) if char == "+"]
+            minus_indices = [i for i, char in enumerate(label) if char == "-"]
+            len_plus = len(plus_indices)
+            len_minus = len(minus_indices)
+            pm_indices = plus_indices + minus_indices
+            label_list = list(label)
+            for indices in product(["X", "Y"], repeat=len_plus + len_minus):
+                for i, index in enumerate(indices):
+                    label_list[pm_indices[i]] = index
+                phase = indices[:len_plus].count("Y") - indices[len_plus:].count("Y")
```
mrossinek, line 510: I think the refactoring helped a lot in terms of understandability of the code. But, this line should really have a comment which explains why the phase is computed like this. It may be clear to us because we have been writing this code but for someone in the future it may not be as obvious. Other than that I think this function is good now.

ikkoham: Thank you. I have added these lines.

https://github.com/Qiskit/qiskit-nature/pull/35/commits/8694aca8cb76a14d2d96b38fc4e13d0c4c5a783b#diff-e5323a89985f1020886e74b7a8971f8c80a4329667e279daeac59c53cade653bR473-R476

ikkoham:
I've improved the codes quite a bit. I could probably improve it forever, but there is no end, so let's finish it.

qiskit_nature/operators/second_quantization/spin_op.py
mrossinek: Perfect thanks :+1:

ikkoham:
@mrossinek @t-imamichi(offline) The codes are improved significantly from the first version. Thank you for your review.

pbark (approved):
I think that the SpinOp now is fine and looks good to merge. Thanks  a lot @ikkoham  and @t-imamichi and @mrossinek for the reviews.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,469 @@
...
+            if n_x == n_y == n_z == 0:
+                labels_list.append(f"I_{rev_pos}")
+        return " ".join(labels_list)
+
+    @lru_cache()
+    def to_matrix(self) -> np.ndarray:
+        """Convert to dense matrix
+
+        Returns:
+            The matrix (numpy.ndarray with dtype=numpy.complex128)
```
woodsp-ibm, line 385: So the data type here is not influenced by the data type on the constructor?

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -148,53 +148,46 @@ class SpinOp(ParticleOp):
...
+        self,
+        data: Union[
+            str,
+            List[Tuple[str, complex]],
+            Tuple[np.ndarray, np.ndarray],
+        ],
+        spin: float = 1 / 2,
+        dtype: Union[type, str] = np.complex128,
     ):
         r"""Initialize ``SpinOp``.
```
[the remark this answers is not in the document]
ikkoham: I see. Hou about this?
image

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -148,53 +148,46 @@ class SpinOp(ParticleOp):
...
 
         Args:
             data: label string or list of labels and coefficients. See documentation of SpinOp for
```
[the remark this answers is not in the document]
ikkoham: I see. Hou about this?
image

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -148,53 +148,46 @@ class SpinOp(ParticleOp):
...
                   more details.
             spin: positive integer or half-integer which represents spin.
-            label_mode: The mode of label. (`sparse` (default) or `dense`)
+            dtype: data type of coefficients.
```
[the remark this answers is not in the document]
ikkoham: Sorry, I forget to remove this line.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -148,53 +148,46 @@ class SpinOp(ParticleOp):
...
+        ],
+        spin: float = 1 / 2,
+        dtype: Union[type, str] = np.complex128,
     ):
         r"""Initialize ``SpinOp``.
 
         Args:
             data: label string or list of labels and coefficients. See documentation of SpinOp for
                   more details.
             spin: positive integer or half-integer which represents spin.
```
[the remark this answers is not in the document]
ikkoham: I agree with this is not good for Python. But, I think physics users are familiar with half integer as a spin value.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,469 @@
...
+        Raises:
+            ValueError: invalid data is given.
+            QiskitNatureError: invalid spin value
+        """
+        self._coeffs: np.ndarray
+        self._spin_array: np.ndarray
+        dtype = np.complex128  # TODO: configurable data type. mixin?
+
+        if (round(2 * spin) != 2 * spin) or (spin <= 0):
+            raise QiskitNatureError("spin must be a positive integer or half-integer")
```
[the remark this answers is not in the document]
ikkoham: Thanks. I've added it.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,469 @@
...
     def register_length(self):
         return self._register_length
 
-    def __repr__(self):
-        full_str = ''
-        for operator in self._operator_list:
-            full_str += '{1} \t {0}\n'.format(operator.coeff, operator.label)
-        return full_str
+    @property
+    def spin(self) -> float:
```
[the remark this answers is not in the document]
ikkoham: How about returning `Fraction ` here? (but I think `float` is well.)

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,469 @@
...
+        I.e. [0, 4, 2] corresponds to X_2^0 \\otimes X_1^4 \\otimes X_0^2, where X_i acts on the
+        i-th spin system in the register.
         """
-        if isinstance(other, numbers.Number):
-            # Create copy of the SpinOp in which every SpinOperator is multiplied by
-            # `other`.
-            new_operatorlist = [copy.deepcopy(base_operator) * other
-                                for base_operator in self.operator_list]
-            return SpinOp(new_operatorlist)
+        return self._spin_array[0]
```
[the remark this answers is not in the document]
ikkoham: Thank you. You are right. This spoils the immutability. I make internal ndarrays immutable by using `ndarray.flag.writeable` since copy is costly.

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: Sorry for confusing. Configurable data type is not implemented now.

ikkoham:
Thank you for your review @woodsp-ibm.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -21,7 +21,7 @@
 import re
```
dlasecki, line 21: Typo in line 16: an -> on.

ikkoham: Wow, you found it. Thanks.

mrossinek (approved):

test/operators/test_spin_op.py
woodsp-ibm: The goal of unit test is to test the module via its public API. When doing so it should cover testing the private methods via use of appropriate test values to drive it across the set of paths to cover all the code so they are used/tested in the context of being called by the module code via the public API since that is how the code will be used in practice in the system. Normally one does not directly test private methods, but of course the code needs to be covered by unit tests. You will find discussions on testing private methods out there so I am not going to say more

qiskit_nature/operators/second_quantization/spin_op.py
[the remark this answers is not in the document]
woodsp-ibm: What I am saying is that this is the documentation of SpinOp module - the Sphinx docstrings inside this file. So they will be reading the API doc for SpinOp when they read 'see documentation of SpinOp for more details' and I imagine their first reaction would be well that is exactly what I am reading! Maybe we can find a way to link to the part of the documentation here that has the description of these labels - that I think is earlier in the module. That was what I meant by a hyperlink to go directly to the more detailed description of the labels. Worst case, if a hyperlink does not seem doable, we could say 'See above for more details about the labels.' or something like that

qiskit_nature/operators/second_quantization/spin_op.py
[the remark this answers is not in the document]
woodsp-ibm: I meant we end up with text like you show there which would come just after the docstring from the class and these two words like 'Initialise SpinOp' or 'SpinOp Constructor' etc really add no value and often clutter the text. It should be self evident since this is the documentation for the init method, i,e the constructor method of a class, such that text like this is not needed.

qiskit_nature/operators/second_quantization/spin_op.py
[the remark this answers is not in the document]
woodsp-ibm: * There were two aspects in my comment above - how to represent the spin as a value:
What I am saying is that could this just be a multiplier value for 1/2. So you are stating how many half-spins you have in the system rather than the total spin. So instead of typing 3.5 or 7/2 it would be simply 7. Being a multiplier you read it as 7 * 1/2 i.e. 7/2. 

* And the docstring there - which is of course related
If its kept like this maybe the text `Value of the integer or half-integer spin. Must be positive.` When I start out reading 'Positive integer...' I think this is why I made the comment since it makes me think of an integer to start with. Hopefully 'Value of the...' breaks that association a bit - at least that was my attempt.

qiskit_nature/operators/second_quantization/spin_op.py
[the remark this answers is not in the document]
woodsp-ibm: Since this is always an integral number of half-spins I guess that is why I suggesting the multiplier. So in my example above, if you gave it 7 in the constructor (for 7/2) you would get 7 back here. This could then be an integer without any rounding/precision issues. Maybe since this is dividing by 2 there would not be any precision error. But seeing round() and things in the code leads me to believe such precision issues are being catered to. Maybe its not a big deal and float is fine.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,479 @@
...
+        """
+        self._coeffs: np.ndarray
+        self._spin_array: np.ndarray
+        dtype = np.complex128  # TODO: configurable data type. mixin?
+
+        if (round(2 * spin) != 2 * spin) or (spin <= 0):
+            raise QiskitNatureError(
+                f"spin must be a positive half-integer (integer or half-odd-integer), not {spin}."
+            )
+        self._dim = int(round(2 * spin)) + 1
```
mrossinek, line 205: Since we decided to go with `Fraction` we should actually leverage that and get rid of `round`. I propose something along the lines of the following:
```suggestion
        spin = Fraction(spin)
        
        if (spin.denominator not in (0, 1)) or (spin <= 0):
            raise QiskitNatureError(
                f"spin must be a positive, either integral or half-integral value, not {spin}."
            )
        self._dim = int(2 * spin + 1)
```

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,479 @@
...
+            Tuple[np.ndarray, np.ndarray],
+        ],
+        spin: Union[float, Fraction] = Fraction(1, 2),
+    ):
+        r"""``SpinOp``.
+
+        Args:
+            data: label string or list of labels and coefficients. See documentation of
+                  :class:`SpinOp` for more details.
+            spin: positive half-integer (integer or half-odd-integer) that represents spin.
```
mrossinek, line 191:
```suggestion
            spin: positive half-integral or integral value.
```

mrossinek (changes requested):

qiskit_nature/operators/second_quantization/spin_op.py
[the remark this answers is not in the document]
mrossinek: @woodsp-ibm Please take a look at my proposed changes below

qiskit_nature/operators/second_quantization/spin_op.py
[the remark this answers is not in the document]
mrossinek: I don't think going to an integer value is a user-friendly decision here. Instead, we discussed offline that the `Fraction`-approach is probably best suited. Please see also my proposed changes below.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,479 @@
...
+
+        Raises:
+            ValueError: invalid data is given.
+            QiskitNatureError: invalid spin value
+        """
+        self._coeffs: np.ndarray
+        self._spin_array: np.ndarray
+        dtype = np.complex128  # TODO: configurable data type. mixin?
+
+        if (round(2 * spin) != 2 * spin) or (spin <= 0):
```
dlasecki, line 201: Might extract the logical formula to a spearate function and give a friendly name, e.g. is_spin_value_correct.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,479 @@
...
+        I.e. [0, 4, 2] corresponds to X_2^0 \\otimes X_1^4 \\otimes X_0^2, where X_i acts on the
+        i-th spin system in the register.
         """
-        if isinstance(other, numbers.Number):
-            # Create copy of the SpinOp in which every SpinOperator is multiplied by
-            # `other`.
-            new_operatorlist = [copy.deepcopy(base_operator) * other
-                                for base_operator in self.operator_list]
-            return SpinOp(new_operatorlist)
+        return self._spin_array[0]
```
dlasecki, line 264: Would it make sense to create a separate class Spin that would have x, y, z fields? It would make the code more readable - here we need to know the convention that 0 is x, 1 is y etc. Usually, having data in an array and accessing it by indices is a bad pattern. Easy to make a mistake, more difficult to spot it and you can also get an index out of bounds.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,479 @@
...
-        """operator list"""
-        return self._operator_list
+    def z(self) -> np.ndarray:
+        """A np.ndarray storing the power i of (spin) Z operators on the spin system.
+        I.e. [0, 4, 2] corresponds to Z_2^0 \\otimes Z_1^4 \\otimes Z_0^2, where Z_i acts on the
+        i-th spin system in the register.
+        """
+        return self._spin_array[2]
+
+    def __repr__(self) -> str:
```
dlasecki, line 282: Do we have a convention in the project for the order of methods in a class? Usually, it might make sense to put magic methods right after the __init__ magic method.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,479 @@
...
+        return [(self._generate_label(i), coeff_list[i]) for i in range(len(self))]
+
+    def _generate_label(self, i):
+        """Generates the string description of `self`."""
+        labels_list = []
+        for pos, (n_x, n_y, n_z) in enumerate(self._spin_array[:, i].T):
+            rev_pos = self.register_length - pos - 1
+            if n_x > 1:
+                labels_list.append(f"X_{rev_pos}^{n_x}")
+            if n_x == 1:
```
dlasecki, line 376: Seems like we might use elif.

dlasecki, line 376: Seems like we might use elif.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,479 @@
...
+                labels_list.append(f"X_{rev_pos}")
+            if n_y > 1:
+                labels_list.append(f"Y_{rev_pos}^{n_y}")
+            if n_y == 1:
```
dlasecki, line 380: Elif

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,479 @@
...
+                labels_list.append(f"Y_{rev_pos}")
+            if n_z > 1:
+                labels_list.append(f"Z_{rev_pos}^{n_z}")
+            if n_z == 1:
```
dlasecki, line 384: Elif

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,480 @@
...
+                labels_list.append(f"Z_{rev_pos}")
+            if n_x == n_y == n_z == 0:
```
dlasecki, line 387: Might move it to the first if and put the rest in else.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,480 @@
...
+            (self._dim, self._dim),
+            dtype=np.complex128,
+        )
+        z_mat = np.fromfunction(
+            lambda i, j: np.where(i == j, (self._dim - 2 * i - 1) / 2, 0),
+            (self._dim, self._dim),
+            dtype=np.complex128,
+        )
+
+        return cast(
```
dlasecki, line 422: This might be too long and therefore difficult to understand. Would it make sense to make it a separate method and make it more readable?

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: https://en.wikipedia.org/wiki/Half-integer

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: Thanks, I adopt Max's idea here.

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: It cannot use NumPy methods such as transpose and stack so it loses the performance.
If you can make it without losing the performance, please make PR after merged.

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: I am not sure for Qiskit, but you are right. I will change it.

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: Why?

dlasecki: Currently, when n_x = 2 for example, Python will still evaluate 2 conditions to see if they are true or not. If the second one was elif, Python would only evaluate the first one.

qiskit_nature/operators/second_quantization/spin_op.py
dlasecki: Ok

qiskit_nature/operators/second_quantization/spin_op.py
mrossinek: I know that technically "half-integer" is correct. But I agree with @woodsp-ibm that it may be confusing to read about integers while the code does not accept `int`. If you feel strongly about leaving "half-integer" written as such, please make it a link to exactly this wikipedia entry as this should clear up any confusion for the reader.

qiskit_nature/operators/second_quantization/spin_op.py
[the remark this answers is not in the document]
ikkoham: Ah, I understood it. Thanks.

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: `int` is also acceptable.

woodsp-ibm: Yes, `int` is a valid type to give when its says `float` for a typehint. Supporting `Fraction` seems fine too as a better way to pass in a 1/2 etc.

qiskit_nature/operators/second_quantization/spin_op.py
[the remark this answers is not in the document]
woodsp-ibm: Chemists are used to multiplicity etc. but I understand having a multiplier is not quite as nice as specifying the actual spin value. Fraction will let you specify things so you avoid any precision/rounding etc that might occur/be needed.

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: You are right. But , it doesn't have a big impact, and the ternary operator (conditional operator) is faster.
image
image

dlasecki: Ok, great job with time tests!

mrossinek (approved):
LGTM. Thanks again, @ikkoham!

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,478 @@
...
+        if isinstance(data, str):
+            data = [(data, 1)]
 
-        # 3. Set the operators particle type to 'spin S' with S the spin value (as float with 1
-        # decimal).
-        # SumOp.__init__(self, particle_type='spin {0:.1f}'.format(self.spin))
+        if isinstance(data, list):
+            invalid_labels = [label for label, _ in data if self._VALID_LABEL_PATTERN.match(label)]
+            if not invalid_labels:
+                raise ValueError(f"Invalid labels: {invalid_labels}")
```
dlasecki, line 217: `invalid_labels` seems to be storing valid labels here. Then, if there is any valid label and some invalid ones, the list is not empty and not valid_labels resolves to False and no error is thrown. Seems like we have not enough unit tests with invalid labels that check this behaviour.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,478 @@
...
-            assert len(elem) == self._register_length, \
-                'Cannot sum operators acting on registers of different length'
-            assert elem.spin == self.spin, \
-                'Cannot sum operators with different spins.'
+    _XYZ_DICT = {"X": 0, "Y": 1, "Z": 2}
+    _VALID_LABEL_PATTERN = re.compile(
+        r"^([IXYZ\+\-]_\d(\^\d)?\s)*[IXYZ\+\-]_\d(\^\d)?(?!\s)$|^[IXYZ\+\-]+$"
+    )
+    _SPARSE_LABEL_PATTERN = re.compile(r"^([IXYZ]_\d(\^\d)?\s)*[IXYZ]_\d(\^\d)?(?!\s)$")
+    _DENSE_LABEL_PATTERN = re.compile(r"^[IXYZ]+$")
```
dlasecki, line 174: Shall we also have `+` and `-` here?

ikkoham: No, this is for after `flatten_ladder_ops`, so ladder operators are not acceptable.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -12,117 +12,478 @@
...
+                0,
+            ),
+            (self._dim, self._dim),
+            dtype=np.complex128,
+        )
+        z_mat = np.fromfunction(
+            lambda i, j: np.where(i == j, (self._dim - 2 * i - 1) / 2, 0),
+            (self._dim, self._dim),
+            dtype=np.complex128,
+        )
```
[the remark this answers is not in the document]
ikkoham: I feel this part is easy to read.

Which functionality do you want to extract? 
If you don't like lambda function,
```
def _x_element(self, i, j):
    return np.where(i == j, (self._dim - 2 * i - 1) / 2, 0)
```

I don't think it make sense.

qiskit_nature/operators/second_quantization/spin_op.py
dlasecki: Oh, I see.

pbark (approved):
This LGTM and we can merge it asap. I see the point in the comments of @dlasecki but I think that the lambda used is fine. I am fine either way, although I am not sure if we really need to change it right now.

PR #41, opened 2021-02-10
Re-organize unit tests along folder structure as per code

### LINKED ISSUE #40: Re-organize unit tests along folder structure as per code
Re-organize unit tests such that instead of a flat structure they are in a folder structure akin to the code.

We may want to adopt the same structure as Terra under test where they have different folders with the units tests under a `python` folder.

manoelmarques:
### Summary

Closes #40


test/algorithms/ground_state_solvers/test_swaprz.py
```diff
@@ -44,7 +44,8 @@ def setUp(self):
     def test_excitation_preserving(self):
```
woodsp-ibm, line 44: Since this test was originally done as a unit test for the SwapRZ variational form as it was in Aqua, and since its moved to Terra where it ought to be unit tested there, It maybe that we can think about getting rid of this later. Its more an integration/system level test here in this context to make sure things work as expected when using it

woodsp-ibm (approved):
Thanks for the speedy re-org of the tests @manoelmarques As the code gets refactored/moved/deleted it should address the files that are for now left in the root.

Issue #42, opened 2021-02-12
Implement the MolecularProblem


mrossinek:
### What is the expected feature?

Implement the `MolecularProblem` class.
It should be located in the `qiskit_nature.problems.second_quantization.molecular` module.

A non-exhaustive list of tasks is:
- [x] port the old `FermionicTransformation` to the `MolecularProblem`
- [ ] move all `mapping` logic out of the problem and into the `GroundStateCalculation` interface
- [x] add logic for handling `transformers`
- [x] add logic for constructing a list of `SecondQuantizedOp` out of the `QMolecule` as generated by the driver
- [x] unittest all methods

![screenshot_1613123296](https://user-images.githubusercontent.com/21973473/107753040-d59a3900-6d1f-11eb-82c9-f3972b118595.png)

mrossinek:
The last TODO of this issue will be addressed by #83.
Since that is decoupled from this, I am closing this issue.

PR #49, opened 2021-02-15
Molecular problem implementation

dlasecki:
### Summary

Issue #42.

### Details and comments

This is a draft PR. Some unit tests should be made more detailed.


qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def _fill_ferm_op_two_body_ints(fermionic_op: FermionicOp, two_body_integrals):
+    for idx in itertools.product(range(len(two_body_integrals)), repeat=4):
+        coeff = two_body_integrals[idx]
+        if not coeff:
+            continue
+        label = ['I'] * len(two_body_integrals)
+        base_op = coeff * FermionicOp(''.join(label))
+        for i, op in [(idx[0], '+'), (idx[2], '+'), (idx[3], '-'), (idx[1], '-')]:
```
mrossinek, line 82: This code assumes that the 2-electron integrals are ordered in chemist notation as outlined in the old `FermionicOperator` class (see here: https://github.com/Qiskit/qiskit-nature/blob/master/qiskit_nature/fermionic_operator.py#L55-L73). Since that class is going to be removed, we should add this documentation somewhere here.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
```diff
@@ -0,0 +1,87 @@
...
+        FermionicOp: FermionicOp built from a QMolecule object.
+    """
+
+    one_body_ints = q_molecule.one_body_integrals
+    two_body_ints = q_molecule.two_body_integrals
+
+    return build_ferm_op_from_ints(one_body_ints, two_body_ints)
+
+
+def build_ferm_op_from_ints(one_body_integrals, two_body_integrals=None) -> FermionicOp:
```
mrossinek, line 38: I assume that you made `two_body_integrals` optional because this was also the case in `FermionicOperator`. I am actually not sure why this is necessary. @woodsp-ibm @pbark do you have any insight on this?

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
```diff
@@ -0,0 +1,111 @@
...
+from qiskit_nature.problems.second_quantization.molecular.fermionic_op_builder import \
+    build_fermionic_op, \
+    build_ferm_op_from_ints
+from qiskit_nature.problems.second_quantization.molecular.integrals_calculators \
+    .magnetization_integrals_calculator \
+    import calc_total_magnetization_ints
+from qiskit_nature.problems.second_quantization.molecular.integrals_calculators \
+    .particle_number_integrals_calculator \
+    import \
+    calc_total_particle_num_ints
```
mrossinek, line 34: You can make all of these relative imports

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
```diff
@@ -0,0 +1,111 @@
...
+from qiskit_nature.transformations.second_quantization import BaseTransformer
+
+
+class MolecularProblem:
+    """Molecular Problem"""
+
+    def __init__(self, fermionic_driver: FermionicDriver,
+                 second_quantized_transformations: Optional[List[BaseTransformer]]):
```
mrossinek, line 42: We need to adapt the naming of this variable as these are no longer linked to the `second_quantization` modules which we have scattered across Qiskit Nature.
For now, you can call them `q_molecule_transformers` or something alike.

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+
+        Args:
+            fermionic_driver: A fermionic driver encoding the molecule information.
+            second_quantized_transformations: A list of second quantized transformations to be
+            applied to the molecule.
+        """
+        self.driver = fermionic_driver
+        self.transformers = second_quantized_transformations
+        self._q_molecule = self.driver.run()
```
mrossinek, line 52: I don't think you should call `driver.run()` during the construction. Rather, do so in `second_q_ops` and store the result. If the operators are queried multiple times you can re-use the stored `q_molecule`.

qiskit_nature/transformations/fermionic_transformation.py
```diff
@@ -21,18 +21,18 @@
...
-from qiskit_nature import QiskitNatureError, QMolecule
+
 from qiskit_nature.fermionic_operator import FermionicOperator
 from qiskit_nature.drivers import BaseDriver
 from qiskit_nature.results import DipoleTuple, EigenstateResult, ElectronicStructureResult
 from qiskit_nature.components.variational_forms import UCCSD
-
 from .transformation import Transformation
+from .. import QiskitNatureError
+from ..drivers.qmolecule import QMolecule
```
mrossinek, line 35:
```suggestion
from qiskit_nature.drivers.qmolecule import QMolecule
```

I usually use relative imports within a linear packaging hierarchy and absolute ones when diverting from a linear path in the tree structure. I only do this based on personal preference though so if anyone has a strong opinion for or against either method, please feel free to chime in here

qiskit_nature/transformations/second_quantization/active_space_transformer.py
```diff
@@ -14,12 +14,11 @@
 
 from typing import Optional
 
-from qiskit_nature.operators.second_quantization.particle_op import ParticleOp
+from ...drivers.qmolecule import QMolecule
```
mrossinek, line 17:
```suggestion
from qiskit_nature.drivers.qmolecule import QMolecule
```

test/problems/second_quantization/molecular/integrals_calculators/test_magnetization_integrals_calculator.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def test_calc_total_magnetization_ints(self):
+        """Tests that one- and two-body integrals for total magnetization are calculated
+        correctly."""
+        num_modes = 1
+
+        h_1, h_2 = calc_total_magnetization_ints(num_modes)
+
+        assert isinstance(h_1, np.ndarray)
+        assert isinstance(h_2, np.ndarray)
```
mrossinek, line 34: Please also compare against some expected values.

test/problems/second_quantization/molecular/integrals_calculators/test_angular_momentum_integrals_calculator.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def test_calc_total_ang_momentum_ints(self):
+        """Tests that one- and two-body integrals for total angular momentum are calculated
+        correctly."""
+        num_modes = 1
+
+        h_1, h_2 = calc_total_ang_momentum_ints(num_modes)
+
+        assert isinstance(h_1, np.ndarray)
+        assert isinstance(h_2, np.ndarray)
```
mrossinek, line 34: Please also compare against some expected values.

test/problems/second_quantization/molecular/integrals_calculators/test_particle_number_integrals_calculator.py
```diff
@@ -0,0 +1,35 @@
...
+    def test_calc_total_particle_num_ints(self):
+        """Tests that one- and two-body integrals are calculated correctly."""
+        num_modes = 1
+        expected_h_1 = np.eye(num_modes, dtype=complex)
+        expected_h_2 = np.zeros((num_modes, num_modes, num_modes, num_modes))
+
+        h_1, h_2 = calc_total_particle_num_ints(num_modes)
+
+        assert h_1 == expected_h_1
+        assert h_2 == expected_h_2
```
mrossinek, line 35:
```suggestion
        assert np.allclose(h_1, expected_h_1)
        assert np.allclose(h_2, expected_h_2)
```

mrossinek (changes requested):
I think this is definitely going in the right direction! Great work @dlasecki :+1: 

I left some comments and some nit-picking suggestions, too. I also noticed some more general things:

- you sometimes use type hints correctly and other times not at all. Please address this.
- the unittests will need to assert the actual results and not only the object types
- I made some import change suggestions w.r.t. relative vs. absolute. However, all of these suggestions are based on personal preference so feel free to ignore them if you have a strong opinion here.

All in all this is on a good track :slightly_smiling_face:

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
woodsp-ibm: Operators for dipole, spin, num particles etc were all just h1 (no h2).

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
woodsp-ibm: If the driver can be updated, which I think can be done via Molecule class, then asking for the 2nd q op should given back something consistent with the driver. Hence calling the run and transforms etc (maybe transforms can be updated too via some setter/getter) in the method to get the 2nd q op seems like the right thing to do. Given the driver can logically be updated without this class knowing I am not sure storing/caching anything here is a good idea.

Also since the integrals are affected by the transforms storing some self._num_modes upfront is problematic,

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
mrossinek: Yes, but looking at the code they still always specify `h2 = np.zeros((modes, modes, modes, modes))`.
So I don't think we actually used the fact that `h2` is optional.

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
mrossinek: That is a good point! I still think that storing the `q_molecule` can have a very significant performance benefit in some cases. But we could probably invalidate any stored values based on whether the driver was updated. But let's leave that to the driver refactoring.

dlasecki: You are right that there is a danger here. But how about we make driver and transformers immutable in `MolecularProblem`? This way, a user would create a new instance of this class for each `driver` and/or `transformers`. This might be a cleaner solution that allowing things to change.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
dlasecki: Currently, dipole case uses h1 only in the code. 
To be honest, I didn't realise that `h2 = np.zeros((modes, modes, modes, modes))` is a dummy value - thought it is somehow relevant for calculations. In such a case, I will delete this variable for num particles and spin so that we benefit fully from having an optional parameter.

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
mrossinek: I think this may conflict with some of the plans for the PES samplers. @pbark should know more about this.

woodsp-ibm: In the overall scheme of things the amount of time it has taken so far in the classical computation with the driver is but a fraction of the overall time. 

As far as making them immutable, since you can change Molecule, which is given to a driver, making the driver itself not be able to be changed in the problem will make no difference. It looks at the Molecule when asked to run and has no hooks to know things were changed in it ahead of that point.

I would just do the simple thing and run the driver and transforms when needed when creating the 2QOp. Later performance tests can be done to look at places where things are worth improving that will make a significant difference.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
woodsp-ibm: I do not know why those operators that FermionicOperator could build out specified the same value as the default. I guess in the docstring it doesn't say what h2=None really does, you only get to see by looking at code. Something hopefully we can improve upon with the new code.

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
```diff
@@ -13,24 +13,17 @@
...
-from qiskit_nature.drivers.qmolecule import QMolecule
-from qiskit_nature.drivers import FermionicDriver
-from qiskit_nature.operators import FermionicOp
-from qiskit_nature.operators.second_quantization import SecondQuantizedOp
-from qiskit_nature.problems.second_quantization.molecular.integrals_calculators \
-    .angular_momentum_integrals_calculator import \
+from ....drivers.qmolecule import QMolecule
+from ....drivers import FermionicDriver
+from ....operators import FermionicOp
+from ....operators.second_quantization import SecondQuantizedOp
```
mrossinek, line 19:
```suggestion
from qiskit_nature.drivers.qmolecule import QMolecule
from qiskit_nature.drivers import FermionicDriver
from qiskit_nature.operators import FermionicOp
from qiskit_nature.operators.second_quantization import SecondQuantizedOp
```

These, I would leave absolute since you are traversing into another branch of the import tree.

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
```diff
@@ -13,24 +13,17 @@
...
-from qiskit_nature.problems.second_quantization.molecular.integrals_calculators \
-    .particle_number_integrals_calculator \
-    import \
-    calc_total_particle_num_ints
-from qiskit_nature.transformations.second_quantization import BaseTransformer
+from ....problems.second_quantization.molecular.fermionic_op_builder import \
+    build_fermionic_op, build_ferm_op_from_ints
+from .integrals_calculators.magnetization_integrals_calculator import calc_total_magnetization_ints
+from .integrals_calculators.particle_number_integrals_calculator import calc_total_particle_num_ints
+from ....transformations.second_quantization import BaseTransformer
```
mrossinek, line 26:
```suggestion
from .fermionic_op_builder import build_fermionic_op, build_ferm_op_from_ints
from .integrals_calculators.magnetization_integrals_calculator import calc_total_magnetization_ints
from .integrals_calculators.particle_number_integrals_calculator import calc_total_particle_num_ints
from qiskit_nature.transformations.second_quantization import BaseTransformer
```

And here, the first one can be simplified a lot whereas the last one should again be absolute imo.

qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -31,31 +30,31 @@ class SecondQuantizedOp(StarAlgebraMixin):
     def __init__(self, operator_list: List[ParticleOp]):
         """
         """
-        # TOOD: validation of operator_list
+        # TODO: validation of operator_list
 
-        self._fermion = None
-        self._boson = None
-        self._spin: Dict[int, SpinOp] = {}
+        self.fermion = None
```
woodsp-ibm, line 35: I guess you have made these public to allow them to be part of public API and settable/gettable directly and did not want to add getter/setters as is done elsewhere in qiskit. Documentation is a concern here since how Sphinx is setup for Qiskit the setters/getters are documented but instance vars are not even if they have comments on them. I would most likely vote for maintaining consistency and having these are properties to expose what were the private marked vars. Things can be changed throughout Qiskit later once Sphinx etc are done to cater for public instance vars. It should be a simple code code with no impact to end users code to go from properties to public vars.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
```diff
@@ -0,0 +1,105 @@
...
+
+def build_ferm_op_from_ints(one_body_integrals: np.ndarray,
+                            two_body_integrals: np.ndarray = None) -> FermionicOp:
+    """
+    Builds a fermionic operator based on 1- and/or 2-body integrals.
+    This method requires the integrals stored in the '*chemist*' notation
+             h2(i,j,k,l) --> adag_i adag_k a_l a_j
+    and the integral values are used for the coefficients of the second-quantized
+    Hamiltonian that is built. The integrals input here should be in block spin
+    format and also have indexes reordered as follows 'ijkl->ljik'
```
woodsp-ibm, line 48: The required to be stored in chemist notation has been confusing in the past. FermionicOperator had similar language. Certainly if we are building out the block spin format, from integrals that were originally in chemist notation, then to have them compatible with the way they were used as coeffs for input for the operator buildout the indexes needed the reordering. But the input here does not care how the integrals were originally stored as long as they are in the right order when presented here.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
```diff
@@ -0,0 +1,105 @@
...
+    fermionic_op = _fill_ferm_op_one_body_ints(fermionic_op, one_body_integrals)
+    if two_body_integrals is not None:
+        fermionic_op = _fill_ferm_op_two_body_ints(fermionic_op, two_body_integrals)
+
+    fermionic_op = fermionic_op.reduce()
+
+    return fermionic_op
+
+
+def _fill_ferm_op_one_body_ints(fermionic_op: FermionicOp,
```
woodsp-ibm, line 76: Not that I want to get too hung up over names but would something like `_add_one_body_terms` and `_add_two_body_terms` be "better". 

I am wondering on the efficiency of building the FermionicOp out this way. Would it be more efficient just to build the list of labels+coeffs and then make the FermionicOp from that list in a single go? Again I don't want to get hung up over early optimization though...

qiskit_nature/problems/second_quantization/molecular/integrals_calculators/angular_momentum_integrals_calculator.py
```diff
@@ -0,0 +1,114 @@
...
+
+""" Calculator of 1- and 2-body integrals for a total angular momentum."""
+
+import itertools
+from typing import Tuple
+
+import numpy as np
+
+
+def calc_total_ang_momentum_ints(num_modes: int) -> Tuple[np.ndarray, np.ndarray]:
```
woodsp-ibm, line 21: Are we going to directly expose function via the init files - I see a lot of empty init files (just copyright I mean)

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
```diff
@@ -0,0 +1,92 @@
...
+from qiskit_nature.drivers.qmolecule import QMolecule
+from qiskit_nature.drivers import FermionicDriver
+from qiskit_nature.operators import FermionicOp
+from qiskit_nature.operators.second_quantization import SecondQuantizedOp
+from qiskit_nature.transformers import BaseTransformer
+from .integrals_calculators.angular_momentum_integrals_calculator import \
+    calc_total_ang_momentum_ints
+from .fermionic_op_builder import build_fermionic_op, build_ferm_op_from_ints
+from .integrals_calculators.magnetization_integrals_calculator import calc_total_magnetization_ints
+from .integrals_calculators.particle_number_integrals_calculator import calc_total_particle_num_ints
```
woodsp-ibm, line 25: Goes back to my comment above about init files this could be from .integrals_calculators import...

test/problems/second_quantization/molecular/test_fermionic_op_builder.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+# TODO add more detailed tests
+class TestFermionicOperatorBuilder(QiskitNatureTestCase):
+    """Tests Fermionic Operator builder."""
+
+    def setUp(self):
+        super().setUp()
+        try:
+            self.driver = PySCFDriver(atom='Li .0 .0 .0; H .0 .0 1.595',
```
woodsp-ibm, line 28: We should be able to use a save HDF5 file for this - we have some that already exist. And its easy to create a new one should it be needed. This has the advantage that the tests can be run on any platform and does not require PySCF to be installed which is not possible for instance on Windows. Arguably though here to test the build out - can we not simply use an h1/h2 we define here and ensure the FermionicOp builds out as we expect it to be with the right amount of terms have the expected labels+coeffs. Maybe this is what you have in mind by adding more detailed tests - at which point maybe even the use of an HDF5 driver is not needed unless you want to test with say a larger set of integrals as you would have from LiH.

test/problems/second_quantization/molecular/test_molecular_problem.py
```diff
@@ -0,0 +1,71 @@
...
+class TestMolecularProblem(QiskitNatureTestCase):
+    """Tests Molecular Problem."""
+
+    def test_second_q_ops_without_transformers(self):
+        """Tests that the list of second quantized operators is created if no transformers
+        provided."""
+        expected_num_of_electr_terms = 185
+        expected_num_of_sec_quant_ops = 7
+
+        driver = HDF5Driver(hdf5_input=self.get_resource_path('H2_631g.hdf5', 'transformers'))
```
woodsp-ibm, line 33: I see here you have used HDF5Driver.... (I made the comment above before I saw this, but this is much better if a driver is needed as an integrals source for the test.)

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
dlasecki: I like names you suggested!

test/problems/second_quantization/molecular/test_fermionic_op_builder.py
dlasecki: To be honest, I copy-pasted this driver declaration from other tests in Qiskit-nature. I agree that it would be better to make it platform-independent so I might change it to HDF5 as suggested. I just learned about HDF5 yesterday because Max used it in his tests. Perhaps we could open a new issue to address numerous driver inits like this in the repo? Of course if both ways achieve same goals in those contexts.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
dlasecki: Do you suggest any specific change in this case? Here I mostly migrated what was already done elsewhere in the project.

qiskit_nature/operators/second_quantization/second_quantized_op.py
dlasecki: Generally, we will need to access these specific fields in various arguments (as far as I understand). I can make them properties for the sake of being compliant.

dlasecki:
Here are some observations to justify merging this feature prior to integrating it with algorithms (e.g. GSE). Generally the suggestion is to first decompose the class FermionicTransformation into separate classes as separate issues because they are needed when integrating this feature to GSE for example.

1) `interpret` is needed for interpreting results of Eigensolvers ([code](https://github.com/Qiskit/qiskit-nature/blob/269a8340e44ddc4d50f8bf871b0fc05b988a28c1/qiskit_nature/transformations/fermionic_transformation.py#L522-L604)), could be extracted as an Interpreter class

2) `molecule_info` dictionary is populated in methods of `FermionicTransformation` responsible for applying transformations. This logic was migrated to the `MolecularProblem` as per UML that actually does not mention what to do with `molecule_info` [(code)](https://github.com/Qiskit/qiskit-nature/blob/269a8340e44ddc4d50f8bf871b0fc05b988a28c1/qiskit_nature/transformations/fermionic_transformation.py#L360-L375). The `molecule_info` refers to spin orbitals for example. Should it be placed somewhere in the `MolecularProblem`?

3) `build_hopping_operator` is needed in `qeom` that accesses it through `FermionicTransformation` stored in the GSE that is passed as an argument. ([code](https://github.com/Qiskit/qiskit-nature/blob/269a8340e44ddc4d50f8bf871b0fc05b988a28c1/qiskit_nature/transformations/fermionic_transformation.py#L653-L746)). It depends on the `molecule_info` and also `UCCSD.compute_excitation_lists` which is supposed to be significantly refactored, too. Code could be extracted as a HoppingOperatorBuilder class.

4) There are also methods like `_process_z2symmetry_reduction` [(code)](https://github.com/Qiskit/qiskit-nature/blob/269a8340e44ddc4d50f8bf871b0fc05b988a28c1/qiskit_nature/transformations/fermionic_transformation.py#L385-L457), `get_default_filter_criterion` ([code](https://github.com/Qiskit/qiskit-nature/blob/269a8340e44ddc4d50f8bf871b0fc05b988a28c1/qiskit_nature/transformations/fermionic_transformation.py#L479-L498)).

I invite you to a discussion and hopefully we could create issues related to the points above with a more detailed specification and perhaps reflected in our UML diagram.

test/problems/second_quantization/molecular/test_fermionic_op_builder.py
woodsp-ibm: Outside of testing the drivers themselves it would be good to improve matters. Hopefully too we can make these tests much more focused on just testing the modules themselves without too much reliance on other code modules here like drivers.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
woodsp-ibm: I think this text that it states the it **requires** the integrals stored in chemist notation. Its more like if we have a source that is two body integrals stored in chemist notation then it needs the ijkl->ljik  reordering as well as altering to be in this block spin format. Also since we treated these more as coeffs for the build out of the 2nd quantized operator there is a 0.5 factor applied to the coeff. For me its the **requires** where it is - what we require is the final block spin with the ordering we expect etc. The source we do not require chemist notation ordered integrals, but of course you can start from there, which is in fact what we have in our normal path.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
woodsp-ibm: I am not sure. Since operations on the FermionicOp like adding a new term make a new one with the combined set of terms, I was thinking that building out the terms as a complete list and contructing a FermionicOp once with the complete list would be more efficient.

dlasecki: Ok, now I can see how `+` and `@` are overloaded and fully see what you mean.

mrossinek (changes requested):
Thanks for the great work @dlasecki! I left some more comments but this is definitely looking like we can get it out soon :+1:

As per your other comment I have inlined some replies below:

> 1) `interpret` is needed for interpreting results of Eigensolvers ([code](https://github.com/Qiskit/qiskit-nature/blob/269a8340e44ddc4d50f8bf871b0fc05b988a28c1/qiskit_nature/transformations/fermionic_transformation.py#L522-L604)), could be extracted as an Interpreter class

I agree! I think initially we can also straight forwardly integrate that into the `MolecularProblem` and then refactor it out during a separate PR. But for this to happen we need to address the steps below, first.

> 2) `molecule_info` dictionary is populated in methods of `FermionicTransformation` responsible for applying transformations. This logic was migrated to the `MolecularProblem` as per UML that actually does not mention what to do with `molecule_info` [(code)](https://github.com/Qiskit/qiskit-nature/blob/269a8340e44ddc4d50f8bf871b0fc05b988a28c1/qiskit_nature/transformations/fermionic_transformation.py#L360-L375). The `molecule_info` refers to spin orbitals for example. Should it be placed somewhere in the `MolecularProblem`?

In terms of design I think this should go into the `MolecularProblem`. We will have to see how the refactoring of UCCSD progresses and how we pass the required runtime information into that. Once we know what that should look like, we can come up with a solution here.
Then, we will have two scenarios:
1. A user does want to use the full stack and storing the data in `MolecularProblem` is sufficient.
2. A user wants to use the code in smaller chunks and pieces, then we should be able to expect that he can also handle this runtime information himself.

> 3) `build_hopping_operator` is needed in `qeom` that accesses it through `FermionicTransformation` stored in the GSE that is passed as an argument. ([code](https://github.com/Qiskit/qiskit-nature/blob/269a8340e44ddc4d50f8bf871b0fc05b988a28c1/qiskit_nature/transformations/fermionic_transformation.py#L653-L746)). It depends on the `molecule_info` and also `UCCSD.compute_excitation_lists` which is supposed to be significantly refactored, too. Code could be extracted as a HoppingOperatorBuilder class.

As you say, this depends on the refactoring of UCCSD (WIP: #57).

> 4) There are also methods like `_process_z2symmetry_reduction` [(code)](https://github.com/Qiskit/qiskit-nature/blob/269a8340e44ddc4d50f8bf871b0fc05b988a28c1/qiskit_nature/transformations/fermionic_transformation.py#L385-L457), `get_default_filter_criterion` ([code](https://github.com/Qiskit/qiskit-nature/blob/269a8340e44ddc4d50f8bf871b0fc05b988a28c1/qiskit_nature/transformations/fermionic_transformation.py#L479-L498)).

I am going to address parts of this this shortly over in my review of #60.

qiskit_nature/operators/second_quantization/second_quantized_op.py
mrossinek: I have to agree with Steve here. Please stick to using properties for now and keep the instance variables private.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
```diff
@@ -0,0 +1,121 @@
...
+        base_ops_list.append(base_op)
+    return base_ops_list
+
+
+def _calc_coeffs_with_ops_one_body(idx):
+    return [(idx[0], '+'), (idx[1], '-')]
+
+
+def _calc_coeffs_with_ops_two_body(idx):
+    return [(idx[0], '+'), (idx[2], '+'), (idx[3], '-'), (idx[1], '-')]
```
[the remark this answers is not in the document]
dlasecki: I pass these functions as arguments so that I only have one `_create_base_ops_labels` function:
`_create_base_ops_labels(one_body_integrals, repeat_num, _calc_coeffs_with_ops_one_body)`
`_create_base_ops_labels(two_body_integrals, repeat_num, _calc_coeffs_with_ops_two_body)`
It helps me to avoid code duplication. Note how I pass them as functions, not as values they return.

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
```diff
@@ -0,0 +1,91 @@
...
+        y_dipole_operator = build_ferm_op_from_ints(q_molecule.y_dipole_integrals)
+        z_dipole_operator = build_ferm_op_from_ints(q_molecule.z_dipole_integrals)
+
+        return x_dipole_operator, y_dipole_operator, z_dipole_operator
+
+    def _create_total_magnetization_operator(self, num_modes) -> FermionicOp:
+        return build_ferm_op_from_ints(calc_total_magnetization_ints(num_modes))
+
+    def _create_total_angular_momentum_operator(self, num_modes) -> FermionicOp:
+        return build_ferm_op_from_ints(*calc_total_ang_momentum_ints(num_modes))
```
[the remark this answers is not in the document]
dlasecki: It returns a tuple and others just an `ndarray`. I use `*` to unpack the tuple and provide it as 2 arguments to the builder function.

test/problems/second_quantization/molecular/test_molecular_problem.py
```diff
@@ -0,0 +1,71 @@
...
+        assert expected_electr_sec_quant_op.__eq__(electr_sec_quant_op)
+        # TODO test QMolecule itself if it is ever a field in MolecularProblem
+
+    @staticmethod
+    def _calc_expected_sec_quant_op(driver, trafo):
+        q_molecule = driver.run()
+        if trafo:
+            q_molecule = trafo.transform(q_molecule)
+        electronic_fermionic_op = build_fermionic_op(q_molecule)
+        return SecondQuantizedOp([electronic_fermionic_op])
```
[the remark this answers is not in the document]
dlasecki: Yes. It might feel a bit redundant because MolecularProblem does sth similar inside. 

The other option would be copy-pasting the actual content (values) of the `SecondQuantizedOp` that I observe when running the test and assuming it is correct, setting it as an expected behaviour and if any changes in the future will result in a different behaviour, we will see that the test fails, something changed and we might want to assess if we expected such a change. 

I am not sure if I am up to date with all the physics-related knowledge in this particular topic to be able to tell if the observed values actually make sense. Same for building the `FermionicOp` and that's why I was reluctant with asserting actual values. Perhaps I can rely on your feedback with that. Let me know.

test/problems/second_quantization/molecular/integrals_calculators/test_angular_momentum_integrals_calculator.py
```diff
@@ -0,0 +1,34 @@
...
+
+"""Tests total angular momentum integrals calculator."""
+from test import QiskitNatureTestCase
+import numpy as np
+
+from qiskit_nature.problems.second_quantization.molecular.integrals_calculators import \
+    calc_total_ang_momentum_ints
+
+
+# TODO add more detailed tests
```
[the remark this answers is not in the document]
dlasecki: I agree, hence my comment `# TODO add more detailed tests`. Will add more tests today.

qiskit_nature/problems/second_quantization/molecular/integrals_calculators/angular_momentum_integrals_calculator.py
```diff
@@ -0,0 +1,114 @@
...
+    x_h1, x_h2 = _calc_s_x_squared_ints(num_modes)
+    y_h1, y_h2 = _calc_s_y_squared_ints(num_modes)
+    z_h1, z_h2 = _calc_s_z_squared_ints(num_modes)
+    h_1 = x_h1 + y_h1 + z_h1
+    h_2 = x_h2 + y_h2 + z_h2
+
+    return h_1, h_2
+
+
+# TODO eliminate code duplication below
```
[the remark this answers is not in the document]
dlasecki: I might. Just need to think of a smart and clear way to do it.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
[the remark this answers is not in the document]
mrossinek: Okay but I still don't understand why you need to wrap these lists into functions. You could just pass the list as the argument in each case.

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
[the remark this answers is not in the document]
mrossinek: I think these methods should have a unified return type. Even if that means that the other two cases return `(smth, None)`

test/problems/second_quantization/molecular/test_molecular_problem.py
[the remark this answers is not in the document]
mrossinek: I have to think about this. I will get back to you :+1:

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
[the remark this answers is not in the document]
dlasecki: `idx` does not exist at the moment of passing the list as an argument.
There is an alternative way, namely passing the whole iterator as an argument and getting rid of functions that you address but not sure if one way is clearly better than the other.

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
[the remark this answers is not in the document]
dlasecki: That's an option, too.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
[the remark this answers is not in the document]
mrossinek: Oh sorry I completely overlooked that `idx` is an input argument here. My bad

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
```diff
@@ -71,51 +72,65 @@ def build_ferm_op_from_ints(one_body_integrals: np.ndarray,
...
-    one_body_base_ops = _create_one_body_base_ops(one_body_integrals)
-    two_body_base_ops = _create_two_body_base_ops(
+    one_body_base_ops_labels = _create_one_body_base_ops(one_body_integrals)
+    two_body_base_ops_labels = _create_two_body_base_ops(
         two_body_integrals) if two_body_integrals is not None else []
-    base_ops = one_body_base_ops + two_body_base_ops
+    base_ops_labels = one_body_base_ops_labels + two_body_base_ops_labels
+    initial_label_with_ceoff = ('I'*len(one_body_integrals), 1)
+    base_ops_labels.append(initial_label_with_ceoff)
+    fermionic_op = FermionicOp(base_ops_labels)
```
mrossinek, line 81: Is it even necessary to have this initial label? Or can the following suffice:
```suggestion
    fermionic_op = FermionicOp(base_ops_labels)
```

dlasecki: Not sure. I might test it on Monday. It was present in the code snippet that you provided me with.

test/problems/second_quantization/molecular/test_molecular_problem.py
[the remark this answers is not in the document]
mrossinek: I think I would rather copy-paste the actual contents into the test. It seems less prone to errors in the test suite. @woodsp-ibm @manoelmarques any thoughts on this?

qiskit_nature/__init__.py
```diff
@@ -31,7 +31,7 @@
 by means of a molecule specification, along with other configuration such as basis set and method
 (RHF, ROHF, UHF). This configuration may include custom settings for the specific driver for more
 custom control over the driver's behavior. When the driver is run the output is a mostly driver
-independent :class:`~.QMolecule`. This contains various quantities that were
+independent :class:`qiskit_nature.drivers.QMolecule`. This contains various quantities that were
```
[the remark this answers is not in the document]
dlasecki: Thank you!

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
dlasecki: Actually, deleting this dummy label failed some tests. Here is why:
I use the file H2_631g for testing. It looks like dipole integrals are matrices full of 0s.
QMolecule checks for the presence of dipole integrals in the following way
```
def has_dipole_integrals(self):
    """ Check if dipole integrals are present. """
    return self.x_dip_mo_ints is not None and \
        self.y_dip_mo_ints is not None and \
        self.z_dip_mo_ints is not None
```
So it resolves as True for the file given.
3. Then, in the MolecularProblem, I do the following
```
if q_molecule_transformed.has_dipole_integrals():
    x_dipole_operator, y_dipole_operator, z_dipole_operator = self._create_dipole_operators(
        q_molecule_transformed)
```
4. The code for building fermionic op labels has the following clause
```
if not coeff:
    continue
```
Which resolves to True because it processes these dipole matrices with 0s only. As a result, no labels are built which results in an empty list of labels.
5. FermionicOp looks at that empty list and crashes when initializing variables:
`self._register_length = len(data[0][0])`  because data is an empty list. It did not crash with a dummy labels because then there was only a dummy label in the list so it was not empty.

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        self.driver = fermionic_driver
+        self.transformers = q_molecule_transformers
+
+    def second_q_ops(self) -> List[SecondQuantizedOp]:
+        """Returns a list of `SecondQuantizedOp` created based on a driver and transformations
+        provided.
+
+        Returns:
+            A list of `SecondQuantizedOp`.
```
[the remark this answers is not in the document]
dlasecki: Added the order to docs.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
dlasecki: As discussed with Max, we currently leave the dummy label as it is.

test/problems/second_quantization/molecular/integrals_calculators/test_angular_momentum_integrals_calculator.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class TestAngularMomentumIntegralsCalculator(QiskitNatureTestCase):
+    """Tests total angular momentum integrals calculator."""
+
+    num_modes_list = [1, 2, 3]
+    expected_h_1_list = [[[0.]], [[0.75, 0.],
+                                  [0., 0.75]], [[0.75, 0., 0.],
+                                                [0., 0.75, 0.],
+                                                [0., 0., 0.]]]
```
mrossinek, line 28: Please reformat this list and the one below to be more readable. E.g. in the following style:
```suggestion
    # A list of expected h1 integrals. Each entry in the list is the expected
    # result for a different number of `num_modes`.
    expected_h_1_list = [
        [[0.]],
        [[0.75, 0.],
         [0., 0.75]],
        [[0.75, 0., 0.],
         [0., 0.75, 0.],
         [0., 0., 0.]]
    ]
```

test/problems/second_quantization/molecular/integrals_calculators/test_magnetization_integrals_calculator.py
```diff
@@ -0,0 +1,48 @@
...
+
+class TestMagnetizationIntegralsCalculator(QiskitNatureTestCase):
+    """Tests total magnetization integrals calculator."""
+
+    num_modes_list = [1, 2, 3]
+    expected_h_1_list = [[[-0.5 + 0.j]], [[0.5 + 0.j, 0. + 0.j],
+                                          [0. + 0.j, -0.5 + 0.j]],
+                         [[0.5 + 0.j, 0. + 0.j, 0. + 0.j],
+                          [0. + 0.j, -0.5 + 0.j, -0. + 0.j],
+                          [0. + 0.j, -0. + 0.j, -0.5 + 0.j]]]
```
mrossinek, line 29: Please do a similar reformatting as per my previous comment.

test/problems/second_quantization/molecular/integrals_calculators/test_magnetization_integrals_calculator.py
```diff
@@ -0,0 +1,31 @@
...
+"""Tests total magnetization integrals calculator."""
+from test import QiskitNatureTestCase
+import numpy as np
+
+from qiskit_nature.problems.second_quantization.molecular.integrals_calculators import \
+    calc_total_magnetization_ints
+
+
+# TODO add more detailed tests
+class TestMolecularProblem(QiskitNatureTestCase):
```
woodsp-ibm: I think test maybe needs renaming (assuming it was copied here and edited)

test/problems/second_quantization/molecular/test_fermionic_op_builder.py
```diff
@@ -0,0 +1,48 @@
...
+
+"""Tests Fermionic Operator builder."""
+from test import QiskitNatureTestCase
+from qiskit_nature.operators import FermionicOp
+from qiskit_nature.problems.second_quantization.molecular import fermionic_op_builder
+from qiskit_nature.drivers import HDF5Driver
+
+
+# TODO add more detailed tests
+class TestFermionicOperatorBuilder(QiskitNatureTestCase):
```
woodsp-ibm, line 20: I guess we have no way of doing some exact diagonalization on the operator to get a eigenvalue(s) that might help in testing to determine if it appears correct.

test/problems/second_quantization/molecular/test_molecular_problem.py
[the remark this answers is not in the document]
woodsp-ibm: Since it returns a list of operators, as it public API, I guess its a question of how to determine that the list is correct. Evidently, as it done the length and the type of each in the list is basic. If we named the operators that could be tested. Aside from storing the terms as a list and checking them, I don't know how else to assure its building the set (list) of operators we expect.  If the operator had some exact diagonalization capability, as mentioned above, that might work too.

test/problems/second_quantization/molecular/test_molecular_problem.py
```diff
@@ -0,0 +1,71 @@
...
+
+        second_quantized_ops = molecular_problem.second_q_ops()
+        electr_sec_quant_op = second_quantized_ops[0]
+        assert electr_sec_quant_op.boson is None
+        assert electr_sec_quant_op.spin == {}
+        assert len(second_quantized_ops) == expected_num_of_sec_quant_ops
+        for second_quantized_op in second_quantized_ops:
+            assert isinstance(second_quantized_op, SecondQuantizedOp)
+        assert len(electr_sec_quant_op.fermion) == expected_num_of_electr_terms
+        assert expected_electr_sec_quant_op.__eq__(electr_sec_quant_op)
```
woodsp-ibm, line 47: Can you do these tests using the with self.subTest construction - otherwise it will stop at the first failure and it may be better to see whether the other tests here pass or not too. The subTest will do that and you can put a string in there saying testing length, testing types etc as best describes the aspect being checked.

test/problems/second_quantization/molecular/integrals_calculators/test_magnetization_integrals_calculator.py
dlasecki: Was fixed earlier (outdated tag).

woodsp-ibm:
For the molecular integrals tests, where its seems you want to run the same test with different parameters, and have a static method and subTest to do this, consider using ddt. You can write a single test and have ddt expand out sets of different test values for the same test. Check elsewhere in the tests here and in Aqua etc for examples/use of ddt,

test/problems/second_quantization/molecular/test_fermionic_op_builder.py
mrossinek: Do you mean as an alternative to the explicit list checking done here or in addition to that?
I think for now, the way it is done is fine, but this is definitely something we can keep in mind for the future :+1:

mrossinek (dismissed):
Thanks @dlasecki for the great work here! This LGTM now :+1: 
Let's wait for @woodsp-ibm's review and once he approves we can merge this :tada:

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
```diff
@@ -0,0 +1,128 @@
...
+    return fermionic_op
+
+
+def _build_fermionic_op(one_body_integrals: np.ndarray,
+                        two_body_integrals: np.ndarray) -> FermionicOp:
+    one_body_base_ops_labels = _create_one_body_base_ops(one_body_integrals)
+    two_body_base_ops_labels = _create_two_body_base_ops(
+        two_body_integrals) if two_body_integrals is not None else []
+    base_ops_labels = one_body_base_ops_labels + two_body_base_ops_labels
+    initial_label_with_ceoff = ('I' * len(one_body_integrals), 1)
```
mrossinek, line 79: I just did some calculations locally which are sort-of related to this and I noticed that we should ensure that the coefficient here is actually 0 rather than 1:
```suggestion
    initial_label_with_ceoff = ('I' * len(one_body_integrals), 0)
```

dlasecki: My bad! I looked at this piece of code from the `FermionicOp` constructor in the context of the original code snippet that passed the dummy label in the constructor.
```
        elif isinstance(data, str):
            label = data
            if not self._validate_label(label):
                raise QiskitNatureError(
                    "Label must be a string consisting only of "
                    f"['I','+','-','N','E'] not: {label}"
                )
            self._register_length = len(label)
            self._labels = [label]
            self._coeffs = [1]
```

mrossinek: Yes no worries at all! I think I also made that mistake in my original snippet but I noticed yesterday that this leads to an energy offset of `1.0` further down in the stack.

mrossinek: But it's a good point that you bring up that snippet! We should discuss whether to change the default coeff to `0`. Let's do that offline and do that in another PR which is likely to happen on the `FermionicOp` anyways.

pbark (approved):

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
mrossinek: This will be addressed in #68 :+1:

mrossinek (dismissed):
Thanks for the great work @dlasecki :+1:

pbark:
provided that @woodsp-ibm agrees we can merge this today!

test/problems/second_quantization/molecular/test_fermionic_op_builder.py
woodsp-ibm: I mean it could be used perhaps for the unit test but was thinking more for something on operator - hence if it were there it could be used by the unit test too.

qiskit_nature/problems/second_quantization/molecular/integrals_calculators/angular_momentum_integrals_calculator.py
```diff
@@ -0,0 +1,129 @@
...
+def _modify_s_x_squared_ints_eq(h_2: np.ndarray, p_ind: int, num_modes_2: int) -> np.ndarray:
+    indices = [(p_ind, p_ind + num_modes_2, p_ind, p_ind + num_modes_2),
+               (p_ind + num_modes_2, p_ind, p_ind + num_modes_2, p_ind),
+               (p_ind, p_ind, p_ind + num_modes_2, p_ind + num_modes_2),
+               (p_ind + num_modes_2, p_ind + num_modes_2, p_ind, p_ind)]
+    values = [-1, -1, -1, -1]
+    return _modify_s_squared_ints(h_2, indices, values)
+
+
+def _modify_s_y_squared_ints_neq(h_2: np.ndarray, p_ind: int, q_ind: int,
```
woodsp-ibm, line 89: It might be helpful to have comments in there here to help the reader as 'modify' is such a generic term. For my part, even though you might have said there was duplication in the code, it seems to me that the original was easier to follow/read - but maybe that's partly because I am more used to seeing it.

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def _build_fermionic_op(one_body_integrals: np.ndarray,
+                        two_body_integrals: np.ndarray) -> FermionicOp:
+    one_body_base_ops_labels = _create_one_body_base_ops(one_body_integrals)
+    two_body_base_ops_labels = _create_two_body_base_ops(
+        two_body_integrals) if two_body_integrals is not None else []
+    base_ops_labels = one_body_base_ops_labels + two_body_base_ops_labels
+    initial_label_with_ceoff = ('I' * len(one_body_integrals), 0)
+    # TODO the initial label should be eliminated once QMolecule is refactored (currently
```
woodsp-ibm, line 80: In general the question is what to do with an empty operator it seems. If you do some operations on the operator and end up with an empty one do you have an instance with a label like this. In the stack for us there was a difference between evaluating an empty operator to 0 and having no operator to evaluate i.e. None in that results in None out.

woodsp-ibm (approved):

qiskit_nature/problems/second_quantization/molecular/fermionic_op_builder.py
mrossinek: Yes we need to be careful about this differentiation. We discussed the case of the empty operator a little bit here: https://github.com/Qiskit/qiskit-nature/pull/25#issuecomment-773404891
If we need to go back and handle that a little bit differently we can discuss this in #68 :+1:

qiskit_nature/problems/second_quantization/molecular/integrals_calculators/angular_momentum_integrals_calculator.py
mrossinek: I have to agree :+1: @dlasecki if you could add a few comments (or even proper docstrings) to these methods that would be great :slightly_smiling_face:

dlasecki: Sure. Hopefully the situation is improved now.

mrossinek (approved):

qiskit_nature/problems/second_quantization/molecular/integrals_calculators/angular_momentum_integrals_calculator.py
pbark: I think this is resolved with the comments added.

pbark (approved):

Issue #45, opened 2021-02-12
Implement the ActiveSpaceTransformer


mrossinek:
### What is the expected feature?

Add the `ActiveSpaceTransformer`.

- [x] actual implementation
- [x] unittests
- [x] documentation

(I will set this to high priority because we need a first transformer working, in order to perform some initial testing. I will hopefully have a PR for this ready by the mid of next week.)

Issue #13, opened 2021-01-28
Naming and packaging fixes


manoelmarques:
Migrated from Enterprise Github.
Creator: @mrossinek

#### Naming
In the remainder of Qiskit we are consistently using verbs for class names, rather than nouns.
Thus, we should follow that approach and rename the classes and modules like:
- [x] `Mapping` -> `Mapper`
- [x] `Transformation` -> `Transformer`
etc.

#### Location
Currently, we plan a differentiation between first and second quantization in the `problems`, `operator` and `transformer` modules. However, the `mappers` also require this differentiation because (at least the currently existing ones) only work in that regime. Thus, we should move all current `mappers`:
- [x] `qiskit_nature.mappers.*` -> `qiskit_nature.mappers.second_quantization`

mrossinek:
Closed by #60.

PR #50, opened 2021-02-16
Implement the ActiveSpaceTransformer

mrossinek:
### Summary

Implements #45.

This also includes part of #13

### Details and comments

Two changes are also made to the `QMolecule`:
- addition of the `mo_occ` information (currently only set by `PySCFDriver` but easily extendable to the others)
- addition of the `energy_shift` attribute: I propose, that we use this as a generalized attribute to gather energy shifts which are caused by transforms. I.e. this will also replace the old `FermionicOperator.ph_trans_shift` attribute.


qiskit_nature/drivers/pyscfd/integrals.py
```diff
@@ -135,21 +135,21 @@ def _calculate_integrals(mol, hf_method='rhf', conv_tol=1e-9, max_cycle=50, init
     if isinstance(m_f.mo_coeff, tuple):
         mo_coeff = m_f.mo_coeff[0]
         mo_coeff_b = m_f.mo_coeff[1]
-        # mo_occ   = m_f.mo_occ[0]
-        # mo_occ_b = m_f.mo_occ[1]
+        mo_occ = m_f.mo_occ[0]
+        mo_occ_b = m_f.mo_occ[1]
```
woodsp-ibm, line 139: Do we really need these - the orbitals are ordered in energies and since we know the number of particles the occupied orbitals are easily known no?

qiskit_nature/qmolecule.py
```diff
@@ -73,6 +73,10 @@ def __init__(self, filename=None):
         self.mo_coeff_b = None  # v2
         self.orbital_energies = None
         self.orbital_energies_b = None  # v2
+        self.mo_occ = None  # v3
+        self.mo_occ_b = None  # v3
+
+        self.energy_shift = 0.0  # v3
```
woodsp-ibm, line 79: See comment in driver around occupied.

I think I would rather keep the QMolecule as some driver raw result and put this inside another class that has the integrals in block (or paired) spin format. My current assumption is that we have 'default' transform somewhere, maybe as method in the class rather than in QMolecule. where they are now, which converts them from the driver form into block spin. I guess it should be possible to go directly to paired spin format too, but if the block spin is more the default/buil-in transform then, at that is what all the transforms work on, then the SeniortiyZero is a transform across the formats and my need to be the last one in sequence (assuming we don't do FreezeCore equivalent etc on the paired spin format, which presumably is doable, but more effort.

The energies can then be accumulated in this extra class. As well as things like num particles tracked as they are removed from the problem too. This way it may be easier to split out as separate properties later too.

qiskit_nature/transformers/active_space_transformer.py
```diff
@@ -0,0 +1,193 @@
...
+from qiskit_nature import QMolecule, QiskitNatureError
+from .base_transformer import BaseTransformer
+
+logger = logging.getLogger(__name__)
+
+
+class ActiveSpaceTransformer(BaseTransformer):
+    """The Active-Space reduction."""
+
+    def __init__(self, num_electrons: int, num_orbitals: int, num_alpha: Optional[int] = None,
```
woodsp-ibm, line 28: I think whatever this acts on should maintain the num electrons etc and orbital indexes currently present in the system that may already have been transformed. Here in my mind I should just be specifying orbitals to be removed - whether they be occupied or unoccupied. If any orbitals do not exist at transform time then raise an error.

I was more thinking we would transform the quantities that were interested in an leave the raw driver result untouched. Otherwise reducing its AOs etc etc seems like unneccessry work. So if you consider this extra data in this class that contains the driver result as what is being transformed, then yes the raw result would be untransformed but this can be stated.

qiskit_nature/transformers/active_space_transformer.py
```diff
@@ -0,0 +1,193 @@
...
+                             space. This argument must match with the remaining arguments and should
+                             only be used to enforce an active space that is not chosen purely
+                             around the Fermi level.
+        """
+        self.num_electrons = num_electrons
+        self.num_orbitals = num_orbitals
+        self.num_alpha = num_alpha
+        self.active_orbitals = active_orbitals
+
+    def transform(self, q_molecule: QMolecule) -> QMolecule:
```
woodsp-ibm, line 46: I was more considering  like these transforms to act on an h1/h2+num_particles and return say an extracted energy where h1/h2/num_particles would be updated according. Transforming the set of properties is then done by the class holding the set of them (as I described it here the class that is holding the raw driver result and has these integrals). This transformation needs to get applied to dipoles etc. And what is holding the extracted energies can maintain a list of them along with say some transform name so that the various extracted energies can be printed in some understandable way and correlated to the transforms

woodsp-ibm:
I removed the changelog label as I am assuming that we generate a release note of some kind detailing what is a first release from this repo that goes along with some migration quide for those currently using Aqua. Changelog would be then for changes after this initial release, which is when the labels would be used

qiskit_nature/drivers/pyscfd/integrals.py
mrossinek: In general this does not need to be the case. It is true, that for all drivers which supply us with ground states this is so, but also only because we only take single-determinant calculations as the input.
I don't think it hurts to have this additional raw data from the driver readily available, which is why I added it.

qiskit_nature/qmolecule.py
mrossinek: Since we decided to postpone the refactoring of everything which ranges from drivers to the transforms I think it is okay to add this here in an intermediate step. I need to talk to Panos to clarify some aspects of the SeniorityZeroTransformer because (if I recall correctly) the last discussion we had on this topic made everything seem more complicated than it really needs to be.

qiskit_nature/transformers/active_space_transformer.py
mrossinek: It does not make sense to specify the active space based on the orbitals you would like to remove. As an example you can think of a large metal complex which has hundreds of orbitals. If I want to select an active space (of say 10 orbitals) I want to specify those and not the hundreds of inactive ones.

The most simplistic choice of active space is `CAS(#nelec, #norb)` which stands for "complete active space" with `#nelec` electrons in `#norbs` active orbitals. More sophisticated choices can be done by manually selecting orbital indices (which still need to match the number of electrons and orbitals, though) and for those we need to figure out how to keep track of possible index shifting due to other transforms. However, that part of the code can only be added once we have more than one working transformer.

As per AOs: I do have a TODO there but I don't think it really makes sense to transform them at all. Although if a user specifies and active space we may want to be able to specify in which basis this is supposed to happen. But we can add that in the future.

qiskit_nature/transformers/active_space_transformer.py
mrossinek: As also mentioned above, I think this aspect will have to be dealt with in the later refactoring of the driver and `QMolecule` classes. Right now, I don't think it makes sense to add an auxiliary container with only the `h1` and `h2` integrals (which we are likely to refactor) if we can simply write all transformers to use the `QMolecule` directly. It does contain all of the information we need (just not in `h1` and `h2` fields) but readily exposes that format in the `one/two_body_integrals`.

qiskit_nature/drivers/pyscfd/integrals.py
woodsp-ibm: Ok, if it useful for future purposes. I am not sure whether its supplied - if not it will have to be built under whatever assumption holds.

qiskit_nature/transformers/active_space_transformer.py
woodsp-ibm: Sure, I have no problem in having different mechanisms to specify it. I guess whether you do it on what you want to keep active or throw away depends on what is the larger quantity for active space, and exactly how you want to specify it whether things are contiguous spaces or need to be stated separately, And it would be nice whatever is done here is consistent with UCCSD.

I don't think it makes sense to transform AOs either - which is why I was proposing to keep the aspects that would be transformed more separate. That way the raw result stays as the raw driver output and is self-consistent in the set of data it contains with the other integrals we need being maintained elsewhere.

qiskit_nature/transformers/active_space_transformer.py
woodsp-ibm: QMolecule currently does not contain the other operator/integral logic for spin, num particles etc, my thought was to have this in this additional class along with storing energies, h1/h2 etc. Yes it would be new and might have to be refactored later too - but its new stuff where ever it goes

qiskit_nature/drivers/pyscfd/integrals.py
mrossinek: I think in most cases it should be available but in the others we can construct it based on your assumption above :+1:

qiskit_nature/transformers/active_space_transformer.py
mrossinek: The biggest difference between this transform and the old orbital reduction method is that here we go through the "effort" of correcting for the exclusion of the reduced orbitals. This means, that we incorporate an effective potential describing the excluded electrons into the active-space Hamiltonian and therefore don't introduce an approximation which may not be valid in certain edge cases.
Since this is a mathematically more correct approach which does not pose too much of a computational overhead, I would argue that this can replace the original orbital reduction fully.

However, if we find that it does indeed introduce an overhead which we do not want, we can always add a more simplistic orbital reduction transformer using the logic we had previously. However, then we will have to make sure to clearly differentiate the two and make sure that the "active space" notation is not mixed up and is consistent across the two transformers.

qiskit_nature/transformers/active_space_transformer.py
mrossinek: I kind of like what @dlasecki did over in #49 with regards to the additional operators. He created "calculators" which construct them based on the most minimal input. This is something I would have put into the "properties" or "observables" which we discussed at another point (but which we will leave open for the driver refactoring).
The `QMolecule` does contain all of the information which is necessary to construct all of the operators, it just does not include all of them in the `h1`/`h2` format.

qiskit_nature/transformers/active_space_transformer.py
```diff
@@ -0,0 +1,193 @@
...
+        mo_occ_inactive = mo_occ_full[inactive_orbs_idxs]
+        density_inactive = np.dot(mo_coeff_inactive*mo_occ_inactive,
+                                  np.transpose(mo_coeff_inactive))
+        density_inactive_b = None
+        if beta:
+            mo_occ_inactive_b = mo_occ_full_b[inactive_orbs_idxs]
+            density_inactive_b = np.dot(mo_coeff_inactive_b*mo_occ_inactive_b,
+                                        np.transpose(mo_coeff_inactive_b))
+
+        # compute inactive Fock matrix
```
dlasecki, line 122: I would extract computing the Fock matrix to a helper function for improved readibility and compactness. Similarly for other fragments annotated with comments.

qiskit_nature/transformers/active_space_transformer.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # determine active and inactive orbital indices
+        if self.active_orbitals is None:
+            norbs_inactive = nelec_inactive // 2
+            if norbs_inactive + self.num_orbitals > q_molecule.num_orbitals:
+                raise QiskitNatureError("More orbitals requested than available.")
+            inactive_orbs_idxs = list(range(norbs_inactive))
+            active_orbs_idxs = list(range(norbs_inactive, norbs_inactive+self.num_orbitals))
+        else:
+            if self.num_orbitals != len(self.active_orbitals):
```
dlasecki, line 95: I would extract such checks to helper functions, e.g. `validate_num_orbitals()`.

qiskit_nature/transformers/active_space_transformer.py
woodsp-ibm: Sounds fine to me if this is an improvement over what existed that retains the same level of reduction in problem complexity over what we finally compute.

I guess part of my initial comment was around the configuration where before we just specified orbitals that were wanted to be discarded (we did not have any option to state those to keep.) As such it was just a list of orbitals so you could pick stuff out wherever you wanted and it was frozen or deleted depending on whether said orbital was occupied or not.

One thing that would help, at least for me, is to improve the docstring to better describe the parameters and any interaction. In my mind all we have before is akin to the active_orbitals parameter that is defaulted to None (we specified the inactive ones to be discarded so those that were not to be kept as active). I guess since we know what is occupied or not from that list the num_electrons and num_orbitals seems confusing to me. Maybe its a simpler alternative way or specifying things but they are mandatory  parameters unlike the orbital list. Overall I was trying to figure what these do, as such, maybe the num_alpha is in there to designate which of the orbitals hold only alpha electrons out of the total number provided in the first param. Hence I had pointed out that all we did before was have a list that we deleted (maybe the deleted part took your thoughts in a different direction) . But I meant we just had a list of orbitals and could figure things based on known occupancy. Here I imagine you have more flexibility to define the active space and maybe all it needs for me is improved docstrings for the parameters - maybe we an inline example or two of configurations.

woodsp-ibm: Another thought on the orbital reduction that we currently have - is the extra info that is needed for the active space transform available in all case from the set of drivers. If not it might be good reason to have the simpler version too even if it is not as accurate?

qiskit_nature/transformers/active_space_transformer.py
woodsp-ibm: I did not believe h1/h2 would contain all the aspects - since num particles is altered and orbitals re-indexed etc. For me there is the set of things that are being acted upon, ie the above, and whether or not more info is needed, such as what atoms comprise the qmolecule in order that the core orbitals can be figured out for freeze core. Having access to data is one thing - maybe some of that data needed by transforms needs to be transformed too. I guess I was already thinking about keeping things apart since I was predicating the thought on all the transforms working on the block spin formatted h1/h2 which are not part of QMolecule. It seems the active space transform you have done on the integrals from which we computed the block spin form - ie. the integrals straight out the driver. I will note the transformation code we have today, that is in FermionicOperator or in the ParticleHole file works on blocked spin (well actually the particle hole logic internally was done on interleaved spin and the block->interleave and back to block was done internal to that code. Working on the original ints would be a little more efficient, its just not what has been done in the existing code.

woodsp-ibm:
>addition of the energy_shift attribute: I propose, that we use this as a generalized attribute to gather energy shifts which are caused by transforms. I.e. this will also replace the old FermionicOperator.ph_trans_shift attribute.
As mentioned elsewhere it may be preferable to gather these as list that can be summed, along with descriptions, so in a result the act of the transforms can be seen as has been done todate.
Not only was the PH energy stored externally and added back at the end, but that filed you refer to was used internally in the FermionicOperator when it built out a qubit operator during the mapping process that it added another Pauli term. I do recall asking in the past, I cannot recall the reasoning though of answer other than it was stated it was needed too.

qiskit_nature/qmolecule.py
mrossinek: I have made this a dictionary now. This allows us to keep track of where an energy shift came from while also easily summing over the `dict.values()` to get the total energy shift. We can then refactor this into a nicer handling when looking into the drivers and `QMolecule` for a later release.

qiskit_nature/transformers/active_space_transformer.py
mrossinek: I have improved the class-level docstring to hopefully clarify the method a bit more. Let's discuss in a comment of a new review if we need to improve it further.

I will comment separately below about the freeze core/orbital reduciton method.

qiskit_nature/transformers/active_space_transformer.py
mrossinek: I think we can see how things develop on a case-by-case basis for the different transformers. If it is more efficient to operate on the raw driver integrals directly we can do that for now and see how to integrate everything nicely in the refactoring of the `QMolecule` class.

mrossinek:
I pushed some updates to address your comments, thanks for the inputs @woodsp-ibm @dlasecki!

With regards to the freeze core/orbital reduction method which we previously had in Qiskit Aqua: in the end, this active space transformer is a generalized version of that technique which does not introduce an assumption that can potentially break in certain edge cases. Therefore, I see this as an improvement of said method and would argue that it is sufficient for the future of Qiskit Nature.
Nonetheless, I do see the benefit in terms of usability to provide a simpler interface to a user who (for example) is only interested in freezing the core orbitals. What we can do, is provide another `FreezeCoreTransformer` which would essentially built up on the `ActiveSpaceTransformer` but use the `core_orbtials` as stored in the `QMolecule` class to specify the list of orbitals to be frozen (rather than having to specify the ones which are left to be active).
If you think, that another class just for this purpose is not necessary, we could even add a keyword argument to this class to achieve the same behavior. However, I would prefer the former option, as an active-space technically has nothing to do with a frozen core approximation.

What do you think?

woodsp-ibm:
> If you think, that another class just for this purpose is not necessary, we could even add a keyword argument to this class to achieve the same behavior. However, I would prefer the former option, as an active-space technically has nothing to do with a frozen core approximation.

In the prior orbital reduction if you manually included the core orbitals it was equivalent to setting freeze_core to True. Stated another way freeze_core defined a set of orbitals to be removed and the user could define a custom set. What was removed was the union of the two. As it was done in one set of code the orbital indexing, as orbitals were removed was tracked accordingly ,since removal of unoccupied and occupied was done be separate methods.

If in the active space I configure things manually, so that the core orbitals are omitted, is this not similar? Maybe doing this in the active space logic ends up more like doing ECP? Are we talking about doing freeze core as it has been done - either using old code or adjusting new code if that is what is required. As separate logic there is either an ordering issue or a need to track orbitals if that is not done (the latter being desired for more generally to allow other transforms to be added where ordering may not be a solution).

I have not directly answered your question I guess, maybe the discussion helps though.

mrossinek:
> If in the active space I configure things manually, so that the core orbitals are omitted, is this not similar?

Yes this is similar. But again, the ASTransformer is mathematically more rigorous than the orbital reduction we used to have. In the core orbital limit this may not be as important (because there the approximation done during the orbital freezing is fine).

> Maybe doing this in the active space logic ends up more like doing ECP?

I am not familiar enough with the internal details of an ECP but I presume it is rather similar, yes.

> Are we talking about doing freeze core as it has been done - either using old code or adjusting new code if that is what is required.

What I wanted to say is:
- what we had before is mathematically different to the ASTransformer
- *but* the ASTransformer can be used to achieve an identical outcome in terms of problem size reduction

I was then proposing, that we add another class which uses the ASTransformer logic but provides the means to define the orbitals which are set to be inactive (as we had in the old code). Obviously, combining this other transformer with the AS-one does not make sense so we would have to make sure that the user does not use both at the same time.

> As separate logic there is either an ordering issue or a need to track orbitals if that is not done

I omitted the tracking of orbital indices on purpose for now. We can add that when we have a PR for a new transformer and actually start needing this.

pbark:
This looks good to me. I do not have further comments here. There are a couple of things that we may need in the freeze core but as pointed we can implement them straight in the new transformer or add them to the Active Space Transformed later. If @woodsp-ibm feels that his comments are adressed we can merge this.

pbark (dismissed):

woodsp-ibm:
> but provides the means to define the orbitals which are set to be inactive (as we had in the old code). 

I have no-preference over whether one defines things as inactive or active (other than maybe if it mostly all active it may be simpler to define inactive in which case maybe we can support either way). My reference to it, as I have re-iterated was more around that was simple config before, in that all we did was to denote orbitals, here there is much more complexity which the updated docstring helps. We could have done active versus inactive before, we just picked one since mostly were were taking less away.

qiskit_nature/drivers/pyscfd/integrals.py
```diff
@@ -135,21 +135,21 @@ def _calculate_integrals(mol, hf_method='rhf', conv_tol=1e-9, max_cycle=50, init
     if isinstance(m_f.mo_coeff, tuple):
         mo_coeff = m_f.mo_coeff[0]
         mo_coeff_b = m_f.mo_coeff[1]
-        # mo_occ   = m_f.mo_occ[0]
-        # mo_occ_b = m_f.mo_occ[1]
+        mo_occ = m_f.mo_occ[0]
```
woodsp-ibm, line 138: If this can be inferred for now with the information in QMolecule (which is what we have done to date) can we do that to avoid refactoring the drivers to accommodate this now, and do it later along with whatever other driver refactoring is wanted. If we infer it in the QMolecule say, for now, this will then work with all the existing drivers without any need to go change them all.

qiskit_nature/transformers/active_space_transformer.py
```diff
@@ -0,0 +1,352 @@
...
+        # invalidate AO basis integrals
+        q_molecule_reduced.hcore = None
+        q_molecule_reduced.hcore_b = None
+        q_molecule_reduced.kinetic = None
+        q_molecule_reduced.overlap = None
+        q_molecule_reduced.eri = None
+        # invalidate dipole integrals
+        q_molecule_reduced.x_dip_ints = None
+        q_molecule_reduced.y_dip_ints = None
+        q_molecule_reduced.z_dip_ints = None
```
woodsp-ibm, line 181: So this method does not allow the dipole ints to be transformed accordingly meaning we cannot measure dipole when using active space reduction. In which case maybe we want to have the approximate reduction too since that allowed it.

qiskit_nature/transformers/active_space_transformer.py
```diff
@@ -0,0 +1,352 @@
...
+
+        e_inactive = self._compute_inactive_energy(hcore, density_inactive, fock_inactive,
+                                                   mo_coeff_inactive)
+
+        hij, hijkl = self._compute_active_integrals(mo_coeff_active, fock_inactive, eri)
+
+        # construct new QMolecule
+        q_molecule_reduced = copy.deepcopy(q_molecule)
+        # Energies and orbits
+        q_molecule_reduced.energy_shift['inactive_energy'] = e_inactive
```
woodsp-ibm, line 157: If we do this as string does it make sense to make it something more like `Active Space` in this case then when a result is generated if we want the extracted energies can be printed along the lines of `key + " energy extracted = " + value` for each entry and the naming allows user to correlate the extracted part (however we want to call it) to the transform. That way we can see the `Particle Hole` part etc etc and better know what they are.

test/transformers/test_active_space_transformer.py
```diff
@@ -0,0 +1,150 @@
...
+        trafo = ActiveSpaceTransformer(num_electrons=2, num_orbitals=2)
+        q_molecule_reduced = trafo.transform(q_molecule)
+
+        assert np.allclose(q_molecule_reduced.mo_onee_ints, q_molecule.mo_onee_ints)
+        assert np.allclose(q_molecule_reduced.mo_eri_ints, q_molecule.mo_eri_ints)
+        assert np.isclose(q_molecule_reduced.energy_shift['inactive_energy'], 0.0)
+
+    def test_minimal_active_space(self):
+        """Test a minimal active space manually."""
+        driver = PySCFDriver(atom='H 0 0 0; H 0 0 0.735', basis='6-31g')
```
woodsp-ibm, line 43: Hopefully the ints don't change over time with any PySCF changes. Normally we had created an HDF5 file of the molecule and used that since it was guaranteed to not change (unless we updated it ourselves or course).

test/transformers/test_active_space_transformer.py
```diff
@@ -0,0 +1,150 @@
...
+        expected_mo_eri_ints = np.asarray([[[[0.65209847, 0.16790822], [0.16790822, 0.53250905]],
+                                            [[0.16790822, 0.10962908], [0.10962908, 0.11981429]]],
+                                           [[[0.16790822, 0.10962908], [0.10962908, 0.11981429]],
+                                            [[0.53250905, 0.11981429], [0.11981429, 0.46345617]]]])
+
+        assert np.allclose(q_molecule_reduced.mo_onee_ints, expected_mo_onee_ints)
+        assert np.allclose(q_molecule_reduced.mo_eri_ints, expected_mo_eri_ints)
+        assert np.isclose(q_molecule_reduced.energy_shift['inactive_energy'], 0.0)
+
+    def test_error_raising(self):
```
woodsp-ibm, line 121: Can we either use subTest here or better yet use ddt to supply parameters since its the same test with different parameterization.

qiskit_nature/drivers/pyscfd/integrals.py
mrossinek: Rather than removing this working logic I am now handling (and testing) the case where `mo_occ is None` inside of the transformer.

qiskit_nature/transformers/active_space_transformer.py
mrossinek: I think `Active Space` is a bit misleading since it would suggest the energy corresponds to the active part (although it's the opposite). I changed it to `ActiveSpaceTransformer` instead.

qiskit_nature/transformers/active_space_transformer.py
mrossinek: I agree. But I will introduce another transformer implementation which does this in a separate PR.

test/transformers/test_active_space_transformer.py
mrossinek: I produced some HDF5 files to ensure that we don't run into such a problem. I produced new files in order to be able to test the existence (or lack of) the `mo_occ` attribute of the `QMolecule`.

pbark (approved):

mrossinek:
@woodsp-ibm I addressed your comments from last week this morning. If you could have another look at this PR today we can go ahead and merge this after your approval :slightly_smiling_face:

woodsp-ibm (approved):

PR #60, opened 2021-02-20
Add parity and bk mappings. Refactor jw mapping

woodsp-ibm:
Added parity and BK mappings as per JW mapping. JW mapping was refactored to such that the pauli_table (mode table) was separated from the logic that used the mode table, and put as a common method in base class to be used by any mapper that has a mode table to do its conversion.


mrossinek:
I am not sure whether I am missing something but I only see the parity and bravyi-kitaev mappings here. *Not* the BKSF mapping. Is that correct? If yes, please adapt the title.

I also proposed a different style of implementation [here](https://github.com/Qiskit/qiskit-nature/issues/18#issuecomment-773494151) when we were discussing how to migrate the BKSF mapping. I think if we go with a "label-translation"-like approach we ought to do something similar for the JW, parity and BK mappings, too.

woodsp-ibm:
Yes, my bad, should have been BK not BKSF, as its just Parity and BK mappings (it does not include BKSF) - the former two had the same structure in the past code as JW and were done using the JW code that had already been brought over.

I would think BKSF would be a separate PR in order to have these merged and available sooner. Even if there is consideration of changing over away from using the mode (pauli) table as it has been done in the past.

mrossinek (changes requested):
Alright, given the discussion above, let's keep this as a simple migration of the mode-based legacy code and refactor the mappers into label translators once we look into the implementation of BKSF. In that process we may find other common aspects anyways.

Nonetheless, before approving this PR I would like to see the addition of the `QubitOpConverter` class which we discussed offline. This should wrap a `QubitMapper` and the symmetry reduction arguments for the Z2 symmetries as implemented in Terra now.

mrossinek:
Ah and please also incorporate the following things:

- [ ] add unittests for JW, parity and BK mappers
- [ ] rename `Mapping` to `Mapper`
- [ ] relocate the module as explained in #13

woodsp-ibm:
This could be merged and the QubitOp Convertor class done in another PR. I understood that a skeleton that just contained the mapping was going to be done for the algorithm flow. I can extend that with symmetry logic once that is merged. It should be fine for now with the the mapping to test out the whole stack.

Unit tests. Yes, I started looking, but I need a FermionicOp as input and the logic to create that from ints, which was going to be at least my starting point to verify they work as before, is in the MolecularProblem. JW was already there with no unit test, so these could be added too and this PR merged and unit tests added afterwards. But yes of course they need unit tests

mrossinek:
> This could be merged and the QubitOp Convertor class done in another PR. I understood that a skeleton that just contained the mapping was going to be done for the algorithm flow.

I assumed that you can provide the skeleton of that class to the algorithm PR since you also designed that class. I will try to get #49 merged this week (i.e. before this PR) and then we can bundle the unittests as well as at least a basic skeleton into this PR.

qiskit_nature/mappers/second_quantization/bravyi_kitaev_mapper.py
```diff
@@ -0,0 +1,180 @@
...
+from qiskit.opflow import PauliSumOp
+from qiskit.quantum_info.operators import Pauli
+
+from qiskit_nature.operators.second_quantization.particle_op import ParticleOp
+from qiskit_nature.operators.second_quantization.fermionic_op import FermionicOp
+
+from .qubit_mapper import QubitMapper
+
+
+class BravyiKitaevMapper(QubitMapper):
```
mrossinek, line 26: As discussed offline, please add a `FermionicQubitMapper` base class (or `FermionicMapper` is also fine imo)

qiskit_nature/mappers/second_quantization/direct_mapper.py
```diff
@@ -10,17 +10,17 @@
...
 
 from qiskit_nature.operators.second_quantization.spin_op import SpinOp
 from qiskit_nature.operators.second_quantization.particle_op import ParticleOp
 
-from .qubit_mapping import QubitMapping
+from .qubit_mapper import QubitMapper
 
 
-class DirectMapping(QubitMapping):
+class DirectMapper(QubitMapper):
```
mrossinek, line 23: Following the same idea as above, this one will get a `Spin(Qubit)Mapper` base class. Although we need to address which mappers will actually be needed here because `DirectMapper` and `LinearMapper` might be equivalent.

qiskit_nature/mappers/second_quantization/qubit_mapper.py
```diff
@@ -35,40 +39,42 @@ def supports_particle_type(self, particle_type: ParticleOp) -> bool:
...
-            second_q_op: the `SecondQuantizedOp` to be mapped.
+            second_q_op: the `ParticleOp` to be mapped.
+
         Returns:
             The `PauliSumOp` corresponding to the problem-Hamiltonian in the qubit space.
+        """
+        raise NotImplementedError()
+
+    @staticmethod
+    def mode_based_mapping(second_q_op: ParticleOp,
```
mrossinek, line 57: I am okay with leaving things as they are here now. But I would like to address my point about a "label-translation"-based implementation of these mappers. I agree, that this is not a high priority now, but when we will refactor BKSF, we might as well refactor these ones too, because it will likely help with the implementation of the "more difficult" BKSF mapper, if we first did these simple ones. No need to do it now, I just wanted to put this out here.

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -0,0 +1,111 @@
...
+
+from qiskit.opflow import PauliSumOp
+from qiskit.opflow.primitive_ops import Z2Symmetries
+
+from qiskit_nature import QiskitNatureError
+from qiskit_nature.mappers.second_quantization import QubitMapper
+
+from . import SecondQuantizedOp
+
+# TODO two qubit reduction flag for parity mapping; special case it here? Or add some flag to
```
mrossinek, line 24: I would add it here, as I think it is a reasonable user-convenience feature to have. Since it is a special case of the Z2 symmetries (which are handled here), this case should also be handled here and not on the mappers.
@pbark any comments?

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -0,0 +1,111 @@
...
+
+        if self.z2symmetry_reduction is None:
+            reduced_qubit_ops = qubit_ops
+            self._z2symmetries = Z2Symmetries([], [], [], None)
+        else:
+            # If we are given the symmetry sector we can use it to reduce all operators.
+            # Otherwise if we are given a callable that can determine the right sector we
+            # will use that. Note this callable will most likely need problem related data
+            # such as info contained within MolecularProblem, so we need access in some way
+            # which needs to be determined.
```
mrossinek, line 107: Yes, problem-specific information needs to be passed into here. Thus, this needs to happen at runtime. We will discuss later offline and I will update here afterwards :+1:

woodsp-ibm: The thing about the callable was an initial thought. I dropped that idea for just being provided the sector either when the converter is created or when the 2nd quant ops are converted to qubit_ops are created (the latter would override anything that had been passed in. This way an end user can manually configure the convertor for the sector as well as the stack (algorithms) using it can auto determine a sector and ask for that to be used. Since I was not focussing on Z2Symmetries I had thought about it and made these comments but forgot to delete them since I did not go back to that yet.

qiskit_nature/mappers/second_quantization/qubit_mapper.py
woodsp-ibm: Sure, the current set (what has been done) is based on sites. The BKSF is based on interactions between sites. I do not have enough understanding to know how much the former would inform the latter. As you have seen in the current codebase they are two very different things even though they are based on the same integrals.

qiskit_nature/mappers/second_quantization/direct_mapper.py
woodsp-ibm: Maybe that is a view to some earlier commit? In the code that is there now it is DirectMapper.

mrossinek: I was just saying that it may turn out that we can remove `DirectMapper` because it might be equivalent to the `LinearMapper`. But that will become clear when we implement the latter :+1: (elsewhere obviously..)

mrossinek (approved):
I fixed the copyright of the new files but other than that this LGTM!

Although some functionality is not fully there yet (Z2Symmetries) we should merge this asap in order to continue with this skeleton in other tasks that depend on this. @woodsp-ibm already said offline that he will open another PR to add the final details of these mappers later.

@pbark I am waiting for your approval, too, and I had already checked with Steve yesterday that we can merge this today in the morning (our timezone) :+1:

qiskit_nature/mappers/second_quantization/direct_mapper.py
woodsp-ibm: Ah ok my bad, too much going on! I read it as the name had not been updated!

mrossinek:
Merging now. I got @pbark's approval offline.

PR #61, opened 2021-02-22
Comply with pylint 2.7.0

mrossinek:
### Summary

pylint 2.7.0 just got released and introduced a new check as well as a new "problem" which are being fixed in this PR.

### Details and comments

- the new check suggests generators whenever possible. A few cases had to be adapted in our code base accordingly.
- a "problem" occurs with the type detection in the `QMolecule.load` method. There, a lot of variables are supposedly of the `h5py.Group` type although they should be either numpy arrays or `h5py.Dataset`s. I added the `Group` class to the ignored pylint classes for the time being.


manoelmarques (approved):

pbark (approved):
Thanks for taking care of this.

PR #63, opened 2021-02-23
Fix FermionicOp.reduce

t-imamichi:
### Summary

`FermionicOp.reduce` casts complex coefficients to real values.
This PR fixes the issue and add a unit test.

```python

op = FermionicOp(('I', 1j))
print('before', op)
print('after ', op.reduce())
```

```
# (master)
before I * 1j
/path/qiskit/nature/qiskit_nature/operators/second_quantization/fermionic_op.py:334: ComplexWarning: Casting complex values to real discards the imaginary part
  coeff_list[i] += val
after  I * 0
```

```
# this PR
before I * 1j
after  I * 1j
```


mrossinek (dismissed):
Thanks @t-imamichi for finding and fixing this!

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -329,19 +329,19 @@ def reduce(self, atol: Optional[float] = None, rtol: Optional[float] = None) ->
             rtol = self.rtol
 
         label_list, indexes = np.unique(self._labels, return_inverse=True, axis=0)
-        coeff_list = np.zeros(len(self._coeffs))
```
woodsp-ibm, line 332: I guess another way might have been to set the data type on the np.zeros array as complex. But then it seems the coeff is can be passed in as a Union of int, float and complex (typehint of  complex implies this) and internally its left as the original type, so this would have made them all complex which perhaps was not desired.

woodsp-ibm (dismissed):

test/operators/test_fermionic_op.py
```diff
@@ -166,6 +166,9 @@ def test_reduce(self):
         reduced_op = fer_op.reduce()
         self.assertSetEqual(frozenset(reduced_op.to_list()), frozenset([("N", 2), ("E", 1)]))
 
+        fer_op = FermionicOp(("+", 1)) + FermionicOp(("-", 1j)) + FermionicOp(("+", 1j))
```
woodsp-ibm, line 169: Ideally the unit tests, where they have multiple asserts, where testing will stop on the first one that fails, rather than completing all the parts of that test, which may be useful to narrow down any problem, that these should either be parameterized by `ddt` or use the `with self.subTest` construct to separate the parts if the test is not so easily parameterized or makes sense say to have different unit tests

qiskit_nature/operators/second_quantization/fermionic_op.py
t-imamichi: If all original coefficients are `List[float]` or `List[int]`, I thought  we need to make a zero array of float or int. Otherwise, the resulting coefficients are converted into `List[complex]` even if there is no complex values. If you are ok with it, I can change it as `np.zeros(..., dtype=complex)`. https://github.com/Qiskit/qiskit-nature/blob/52b0861d686ca5767626b809e7463257ae1bf95c/qiskit_nature/operators/second_quantization/fermionic_op.py#L116

t-imamichi: Yes, the internal keeps the original types float or int. So, I made this code to keep the original type. If we can convert the internal coeffs into complex, we can simply change this line as `np.zeros(..., dtype=complex)`.

test/operators/test_fermionic_op.py
t-imamichi: Yes, ddt is better. Other unit tests in this file should be revised with ddt maybe in another PR.

mrossinek (approved):

woodsp-ibm (approved):

pbark (approved):

PR #66, opened 2021-02-26
UCCSD Refactoring

### LINKED ISSUE #51: same_spin_doubles works incorrectly in UCCSD
### Information

- **Qiskit Aqua version**: qiskit-master
- **Python version**: 3.8

`same_spin_doubles = True` in UCCSD variational form does not work correctly, when only beta same spin double excitations are required.

I faced this problem with triplet O2 molecule. For alpha orbitals there is only one unoccupied spin-orbital, while for beta there are three. I'm expecting to have same spin double excitations only for beta orbitals. However, there are no such excitations when `same_spin_doubles = True`.

### Suggested solutions
Change condition [here](https://github.com/Qiskit/qiskit-aqua/blob/4768635d9a4f467dff035999f4734edb601b1a2c/qiskit/chemistry/components/variational_forms/uccsd.py#L620). When this condition is False due to alpha orbitals, beta сondition is omitted. My solution is:
```
        if same_spin_doubles:
            if len(active_occ_list_alpha) > 1 and len(active_unocc_list_alpha) > 1:
                for i, occ_alpha in enumerate(active_occ_list_alpha[:-1]):
                    for j, unocc_alpha in enumerate(active_unocc_list_alpha[:-1]):
                        for occ_alpha_1 in active_occ_list_alpha[i + 1:]:
                            for unocc_alpha_1 in active_unocc_list_alpha[j + 1:]:
                                double_excitations.append([occ_alpha, unocc_alpha,
                                                           occ_alpha_1, unocc_alpha_1])

            up_active_occ_list = active_occ_list_beta
            up_active_unocc_list = active_unocc_list_beta

            if len(active_occ_list_beta) > 1 and len(active_unocc_list_beta) > 1:        
                for i, occ_beta in enumerate(up_active_occ_list[:-1]):
                    for j, unocc_beta in enumerate(up_active_unocc_list[:-1]):
                        for occ_beta_1 in up_active_occ_list[i + 1:]:
                            for unocc_beta_1 in up_active_unocc_list[j + 1:]:
                                double_excitations.append([occ_beta, unocc_beta,
                                                           occ_beta_1, unocc_beta_1])
```

### LINKED ISSUE #55: Hamiltonian Variational Form (TASP)
See Qiskit/qiskit-aqua#703 which was a draft PR to added a Hamiltonian Variational Form (or Trotterized Adiabatic State Preparation, TASP) .

What might be nice to have is a base class for an ansatz that is built from evolving operators. Maybe this base class ends up in qiskit circuit library. It could potentially be used as base for QAOA ansatz, UCCSD etc. And also to provide an ansatz as per the linked draft PR.

### LINKED ISSUE #57: Refactor UCCSD

`UCCSD` needs to be refactored for several reasons:
- it is incompatible with the `VQE` implementation provided by Qiskit Terra
- it uses the (soon-to-be-removed) `FermionicOperator` class
- it uses the legacy `WeightedPauliOperator`

The following requirements should be met by the new implementation:
- it should derive `BlueprintCircuit` and, thus, be compatible with Terra
- the hopping operator factory logic should be moved to an external place as it can also be of use elsewhere (currently `FermionicTransformation` actually uses the logic inside of UCCSD)
- it should make use of the new `FermionicOp` and opflow operators
- ideally, the hopping operator factory should be generalized to work for any kind of exctitation (`S`, `D`, `T`, `Q`, ...)
- ideally, we would have a `UCC` class in which the user can freely specify the types of excitations he would like to have
- (we may provide ease-of-use subclasses like `UCCSD`, `UCCD`, etc.)

Once `UCCSD` has been refactored we should also be able to replace all `qiskit.aqua` imports with their new counterparts. Thus, the `critical` status of this issue.

mrossinek:
### Summary

Closes #57, closes #51, closes #55 

### Details and comments

For more details see the discussion there. I will update this description as we go.


qiskit_nature/components/variational_forms/ucc.py
```diff
@@ -0,0 +1,63 @@
...
+                    - `d` for doubles
+                    - `t` for triples
+                    - `q` for quadruples
+                - a single, positive `int` which denotes the number of excitations (1 == `s`, etc.)
+                - a list of positive integers
+                - a list of `callable` objects which are used to generate the excitations. Each
+                  `callable` must take the following arguments:
+                      - `excitation_type`: the type of excitation to generate
+                      - `num_particles`: the same as above
+                      - `num_spin_orbitals`: the same as above
```
mrossinek, line 52: About this possibility for specifying the excitations: after discussing with @Cryoris we believe that this is probably the best way to go about introducing flexibility in terms of method variants like what we used to have through the `method_singles` and `method_doubles` parameters. You would then simply have different implementations of the method in this file: https://github.com/Qiskit/qiskit-nature/blob/35fb89c3a579796bf625e8c5dec72a3e1911ee80/qiskit_nature/components/variational_forms/excitation_op_builder.py

While writing this I actually noticed that I did not update that file after the aforementioned discussion. I had originally planned the excitation operator builders as very simple classes rather than standalone functions. I am open for either option but we should discuss what we prefer here. The docstring above kind-of assumes the latter.

@woodsp-ibm @pbark @stefan-woerner please comment :slightly_smiling_face:

Cryoris:
Can we put the circuits in `qiskit_nature/circuit/library` analogous to the other application stacks and Qiskit core?

mrossinek:
> Can we put the circuits in qiskit_nature/circuit/library analogous to the other application stacks and Qiskit core?

All locations and namings are *very* temporary. I just wanted to get this out there. But yes, this is definitely something we can do!

qiskit_nature/components/variational_forms/adaptive_ansatz.py
```diff
@@ -0,0 +1,42 @@
...
+from abc import ABC, abstractmethod
+
+import logging
+
+from qiskit.circuit.library import BlueprintCircuit
+
+logger = logging.getLogger(__name__)
+
+
+class AdaptiveAnsatz(BlueprintCircuit, ABC):
```
woodsp-ibm, line 25: Do you think we need a separate class for this. I was thinking, since the evolved operator class takes a set of operators that these could be programmatically swapped in and out via the base class. Of course some algos may never use that function, but by having it there it would be available if needed.

qiskit_nature/components/variational_forms/excitation_op_builder.py
```diff
@@ -0,0 +1,83 @@
...
+
+from typing import Tuple
+
+import itertools
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class ExcitationOpBuilder:
```
woodsp-ibm, line 24: Maybe this can be an ExcitationBuilder (no Op) and have it be able to build/return simple excitation lists too.

qiskit_nature/components/variational_forms/adapt_ucc.py
```diff
@@ -0,0 +1,34 @@
...
+
+import logging
+
+from .adaptive_ansatz import AdaptiveAnsatz
+from .ucc import UCC
+
+logger = logging.getLogger(__name__)
+
+
+class AdaptUCC(UCC, AdaptiveAnsatz):
```
woodsp-ibm, line 24: If these are all going to be Blueprint circuits then the should so to `circuit` and some sub-folder there `ansatzes` ? (The base class will evidently migrate over to Terra)

qiskit_nature/components/variational_forms/ucc.py
woodsp-ibm: The QubitConvertor needs either to be the same instance that was used for the problem conversion, so the hopping operators can be symmetry reduced the same way etc, or configured given the same mapper type and the symmetry defined here based on the symmetry sector selected found from the problem. Which means that info needs to be be made available - this used to be put into what was called molecule_info that held this and other values that were needed elsewhere like in UCCSD  ( e.g num particles, orbitals...)

mrossinek: I don't see a problem with the `QubitConverter` being the same instance. As per information availability: yes, this is a similar discussion which popped up in #49 at some point. I think it should be sufficient that we have this information available inside of the problem instance. If a "power user" wants to work without the problem instance, it can be reasonably expected that he knows what he is doing and is able to extract the information from the respective places himself.

qiskit_nature/components/variational_forms/adaptive_ansatz.py
mrossinek: It is true that the interface of the `EvolvedOperatorAnsatz` currently takes a list of operators. However, I don't think that it makes sense that this ansatz also has about adaptiv behavior (single-responsibility). By splitting this into a separate class/interface, we can also reuse this concept outside of the evolved operators.
@Cryoris what do you think?

qiskit_nature/components/variational_forms/excitation_op_builder.py
mrossinek: Yes, we can decide on the naming at the end of this PR.

qiskit_nature/components/variational_forms/adapt_ucc.py
mrossinek: Yes I agree. This was also already mentioned above by @Cryoris

qiskit_nature/components/variational_forms/adaptive_ansatz.py
woodsp-ibm: A class that is responsible for managing a set operators as an ansatz seems fine to me as an encapsulated functionality. Do you have an example of where controlling the set of active operators seems like it would be used elsewhere. While I can see the function could be split out, it seemed to me it could be part of the core function of having/managing a set of operators to evolve as an ansatz. I guess we can see what things look as the design/implementation progresses

qiskit_nature/components/variational_forms/ucc.py
woodsp-ibm: I was not saying anything about whether I considered any way problematic or not, merely that the information needed to get there one way or another and we need to ensure through the design that we have the information, that was all. This is true for UCCSD and also true for HartreeFock initial state that needs consideration/reworking again given the new design.

qiskit_nature/components/variational_forms/adaptive_ansatz.py
Cryoris: It would probably be very easy to implement the adaptive behaviour only for the EvolvedOperatorAnsatz, since it already keeps a list of operators and is already able to rebuild itself when the list changes. If we only want this adaptive behaviour for the EvolvedOperatorAnsatz I think we could add it there. We could also add a intermediate `AdaptiveEvolvedOperatorAnsatz` to stick to the single-responsibility design.

If we want a more general adaptive behaviour though, an additional base class seems sensible to me.

qiskit_nature/circuit/library/ansaetze/ucc.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @property
+    def num_spin_orbitals(self) -> int:
+        """The number of spin orbitals."""
+        return self._num_spin_orbitals
+
+    @num_spin_orbitals.setter
+    def num_spin_orbitals(self, n: int) -> None:
+        """Sets the number of spin orbitals."""
+        self._num_spin_orbitals = n
```
Cryoris, line 88: If changing the `num_spin_orbitals` changes the circuit structure, the setter must call `self._invalidate` to tell the circuit to rebuild itself 🙂

qiskit_nature/circuit/library/ansaetze/ucc.py
```diff
@@ -0,0 +1,173 @@
...
+
+        if self.excitations is None:
+            if raise_on_failure:
+                raise ValueError('The excitations cannot be `None`.')
+            return False
+
+        return True
+
+    def _build(self) -> None:
+        excitation_ops = self.excitation_ops()
```
Cryoris, line 129:
```suggestion
        if self._data is not None:
            return 

        excitation_ops = self.excitation_ops()
```

qiskit_nature/circuit/library/ansaetze/evolved_operator_ansatz.py
```diff
@@ -0,0 +1,135 @@
...
+                raise ValueError('The operators are not set.')
+            return False
+
+        if self.reps < 0:
+            if raise_on_failure:
+                raise ValueError('The reps cannot be smaller than 0.')
+            return False
+
+    @property
+    def reps(self) -> int:
```
woodsp-ibm, line 64: Don;t we normally have a setter for reps too. Maybe we should have it for evolution too.

qiskit_nature/circuit/library/ansaetze/evolved_operator_ansatz.py
```diff
@@ -0,0 +1,135 @@
...
+        """The evolution converter used to compute the evolution."""
+        return self._evolution
+
+    @property
+    def operators(self) -> List[OperatorBase]:
+        """The operators that are evolved in this circuit."""
+        return self._operators
+
+    @operators.setter
+    def operators(self, operators: Union[OperatorBase, List[OperatorBase]]) -> None:
```
woodsp-ibm, line 79: Shouldn't setting the operators invalidate the built data?

At this point one can manage the set of operators being used a whole set. I guess we can see what the sort of code looks like for one where the set of operators can be managed over an active/inactive list. Maybe as has been stated this is enough functionality to warrant another class deriving from this. I just wonder would we derive all our present evolved operator ansatzes from that and never directly from this. Can see how it pans out I guess.

qiskit_nature/circuit/library/ansaetze/evolved_operator_ansatz.py
```diff
@@ -0,0 +1,135 @@
...
+from qiskit.circuit import Parameter, ParameterVector, QuantumRegister
+from qiskit.circuit.library import BlueprintCircuit
+from qiskit.opflow import OperatorBase, EvolutionBase, PauliTrotterEvolution
+
+
+class EvolvedOperatorAnsatz(BlueprintCircuit):
+    """The evolved operator ansatz."""
+
+    def __init__(self,
+                 operators: Union[OperatorBase, List[OperatorBase]],
```
woodsp-ibm, line 26: Should this be optional to allow them to be passed later and maybe default to None. Currently the code checks if they are still None but neither here nor the setter state that it is Optional

qiskit_nature/circuit/library/ansaetze/ucc.py
```diff
@@ -0,0 +1,173 @@
...
+    """The Unitary Coupled-Cluster Ansatz."""
+
+    EXCITATION_TYPE = {
+        's': 1,
+        'd': 2,
+        't': 3,
+        'q': 4,
+    }
+
+    # TODO: expose evolution and reps here, too?
```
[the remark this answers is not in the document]
woodsp-ibm: You mean we should add reps here - I would agree.

qiskit_nature/circuit/library/ansaetze/ucc.py
```diff
@@ -0,0 +1,173 @@
...
+    def __init__(self, qubit_op_converter: QubitOpConverter,
```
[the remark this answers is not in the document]
woodsp-ibm: Convertor is needed since the excitation operators are created as Fermionic and they need to be mapped consistently with the the main Hamiltonian operator for which this is the state. Whether it could be optional here I guess is choice over how to set this up - since num_particles and num_spin_orbitals seems to be allowed to be Optional here maybe the convertor can be in that regard too. Ultimately before use they need to be set of course.

qiskit_nature/circuit/library/ansaetze/ucc.py
```diff
@@ -0,0 +1,173 @@
...
+                 num_particles: Optional[Tuple[int, int]] = None,
+                 num_spin_orbitals: Optional[int] = None,
+                 excitations: Optional[Union[str, int, List[int], List[Callable]]] = None):
```
woodsp-ibm, line 50: We should be able to improve the Callable here to specify arguments and return types. Why would we want a list here - would a single callable not suffice? If a user wants to break down their code to have separate functions for single etc they can make one overall callable that calls the parts and builds the complete list.

qiskit_nature/circuit/library/ansaetze/ucc.py
```diff
@@ -0,0 +1,173 @@
...
+        """Sets the number of spin orbitals."""
+        self._num_spin_orbitals = n
+
+    @property
+    def num_particles(self) -> Tuple[int, int]:
+        """The number of particles."""
+        return self._num_particles
+
+    @num_particles.setter
+    def num_particles(self, n: Tuple[int, int]) -> None:
```
woodsp-ibm, line 96: Generally setting any property here should invalidate the data so any existing circuit (data) get rebuilt next time its wanted

mrossinek:
Marking this as ready-for-review. I am leaving the WIP though because we are still missing some docstrings and unittests for more parameter variations. These will follow soon :+1:

qiskit_nature/circuit/library/ansaetze/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Ansatz circuits and utilities thereof."""
```
woodsp-ibm, line 13: I am not sure about `/library/ansaetze/` - we used to use the `ansatzes` as the plural - its listed as the plural of ansatze and is more easily recognizable I think to English speakers,

qiskit_nature/circuit/library/ansaetze/evolved_operator_ansatz.py
```diff
@@ -0,0 +1,149 @@
...
+    @operators.setter
+    def operators(self, operators: Union[OperatorBase, List[OperatorBase]]) -> None:
+        """Set the operators to be evolved."""
+        if not isinstance(operators, list):
+            operators = [operators]
+
+        if len(operators) > 1:
+            num_qubits = operators[0].num_qubits
+            if any(operators[i].num_qubits != num_qubits for i in range(1, len(operators))):
+                raise ValueError('All operators must act on the same number of qubits (for now).')
```
woodsp-ibm, line 101:
```suggestion
                raise ValueError('All operators must act on the same number of qubits.')
```
I would remove the `(for now)` as that seems to imply we will change. We may well change this going forwards but I think its best left out of the message

qiskit_nature/circuit/library/ansaetze/evolved_operator_ansatz.py
```diff
@@ -0,0 +1,149 @@
...
+        # use setter to set operators
+        self.operators = operators
+
+    def _check_configuration(self, raise_on_failure: bool = True) -> bool:
+        if self.operators is None:
+            if raise_on_failure:
+                raise ValueError('The operators are not set.')
+            return False
+
+        if self.reps < 0:
```
woodsp-ibm, line 58: Isn't 1 the minimum number of reps. If we use the same notion as the other NLocal ansatzes then 0 is an invalid value. This is the total number of layers, 1 being the min and the default.

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
```diff
@@ -0,0 +1,115 @@
...
+from qiskit_nature.operators.second_quantization import FermionicOp, SecondQuantizedOp
+
+logger = logging.getLogger(__name__)
+
+
+class ExcitationBuilder:
+    """A factory class to construct excitation operators."""
+
+    @staticmethod
+    def build_excitation_ops(num_excitations: int,
```
woodsp-ibm, line 30: Should the method name include `fermionic` in it, or `molecular`. Maybe could build ops for lattices too etc.

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
```diff
@@ -0,0 +1,115 @@
...
+
+        This method assumes block-ordered spin-orbitals.
+
+        Args:
+            num_excitations: number of excitations per operator (1 means single excitations, etc.).
+            num_spin_orbitals: number of spin-orbitals.
+            num_particles: number of alpha and beta particles.
+            alpha_spin: boolean flag whether to include alpha-spin excitations.
+            beta_spin: boolean flag whether to include beta-spin excitations.
+            pure_spin: boolean flag whether to include only pure-spin excitations.
```
woodsp-ibm, line 48: Can you say what pure spin here means. Does it mean preserve spins i.e. alpha->alpha and not flipping them?

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
```diff
@@ -0,0 +1,115 @@
...
+        operators = []
+        for exc in excitations:
+            label = ['I'] * num_spin_orbitals
+            for occ in exc[0]:
+                label[occ] = '+'
+            for unocc in exc[1]:
+                label[unocc] = '-'
+            op = FermionicOp(''.join(label))
+            # TODO: this is UCC specific. Do we want to keep this here and make this a UCC-specific
+            # excitation builder?
```
woodsp-ibm, line 108: I guess this somewhat related to my comment above about the method name

woodsp-ibm (commented):
So for AdaptVQE we still need the managed operator aspect for UCCSD right.

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
mrossinek: I was going to talk to Pauline about the UVCC Ansatz and see how much of this method we can potentially reuse. If we can reuse none of it, I will make it a fermionic version.
If, however, we can reuse the excitation generation logic, I will extract the conversion to the operator in order to not imply the fermionic property.

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
mrossinek: Pure spin means that we don't have mixed excitations. E.g. if you consider double excitations, you will only get `alpha,alpha` or `beta,beta` excitations, but no excitations where one is an `alpha` and another is a `beta` particle.

As far as I know in the past (and also the way the new code works now) we always preserve the spin during excitations and never allow flipping.

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
mrossinek: Please see my reply there.

qiskit_nature/circuit/library/ansaetze/__init__.py
mrossinek: What do you mean by "its listed as the plural of ansatze"?
I don't think the recognition factor is really decreased here..

mrossinek:
In order to do some proper testing with the correct integration etc. it would be great to get #85 merged. There are also fewer open tasks left in that PR so I don't think it slows us too much down if we wait for that and then do proper tests with the new `HartreeFock` class here.

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
```diff
@@ -0,0 +1,115 @@
...
+
+import itertools
+import logging
+
+from qiskit_nature.operators.second_quantization import FermionicOp, SecondQuantizedOp
+
+logger = logging.getLogger(__name__)
+
+
+class ExcitationBuilder:
```
Cryoris, line 26: If this only provides a single staticmethod, can we make it a callable instead of a class? I think this will be simpler to use and users are not exposed to even more classes 🙂

qiskit_nature/circuit/library/ansaetze/ucc.py
```diff
@@ -0,0 +1,213 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+"""
+TODO.
```
Cryoris, line 13: Also this one

qiskit_nature/circuit/library/ansaetze/ucc.py
```diff
@@ -0,0 +1,213 @@
...
+
+import logging
+
+from qiskit_nature import QiskitNatureError
+from qiskit_nature.operators.second_quantization import SecondQuantizedOp
+from qiskit_nature.operators.second_quantization.qubit_converter import QubitConverter
+from .evolved_operator_ansatz import EvolvedOperatorAnsatz
+from .excitation_builder import ExcitationBuilder
+
+logger = logging.getLogger(__name__)
```
Cryoris, line 27: Same as above -- do we need the logger?

qiskit_nature/circuit/library/ansaetze/ucc.py
```diff
@@ -0,0 +1,213 @@
...
+            reps: The number of times to repeat the evolved operators.
+        """
+        self._qubit_converter = qubit_converter
+        self._num_particles = num_particles
+        self._num_spin_orbitals = num_spin_orbitals
+        self._excitations = excitations
+        self._alpha_spin = alpha_spin
+        self._beta_spin = beta_spin
+        self._pure_spin = pure_spin
+        # TODO: Added to pass lint, need change
```
Cryoris, line 82: What change is needed?

qiskit_nature/circuit/library/ansaetze/uccsd.py
```diff
@@ -0,0 +1,53 @@
...
+"""
+
+from typing import Optional, Tuple
+
+import logging
+
+from qiskit_nature.operators.second_quantization.qubit_converter import QubitConverter
+from .ucc import UCC
+
+logger = logging.getLogger(__name__)
```
Cryoris, line 23:
```suggestion
from qiskit_nature.operators.second_quantization.qubit_converter import QubitConverter
from .ucc import UCC

```

qiskit_nature/circuit/library/ansaetze/uccsd.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __init__(self, qubit_converter: Optional[QubitConverter] = None,
+                 num_particles: Optional[Tuple[int, int]] = None,
+                 num_spin_orbitals: Optional[int] = None,
+                 reps: int = 1):
+        """
+
+        Args:
+            qubit_converter: the QubitConverter instance which takes care of mapping a
+            :code:`~.SecondQuantizedOp` to a :code:`~.PauliSumOp` as well as performing all
```
Cryoris, line 40: What's the difference of `:code:` to using 
```
`
```
?

qiskit_nature/circuit/library/ansaetze/ucc.py
```diff
@@ -0,0 +1,213 @@
...
+from qiskit_nature.operators.second_quantization import SecondQuantizedOp
+from qiskit_nature.operators.second_quantization.qubit_converter import QubitConverter
+from .evolved_operator_ansatz import EvolvedOperatorAnsatz
+from .excitation_builder import ExcitationBuilder
+
+logger = logging.getLogger(__name__)
+
+
+class UCC(EvolvedOperatorAnsatz):
+    """The Unitary Coupled-Cluster Ansatz."""
```
Cryoris, line 31: It would be nice to have some examples on how to call this class in the docstring

test/circuit/library/ansaetze/test_uccsd.py
```diff
@@ -0,0 +1,34 @@
...
+    def test_ucc_ansatz(self):
+        """Tests the UCCSD Ansatz."""
+        converter = QubitConverter(JordanWignerMapper())
+
+        ansatz = UCCSD(qubit_converter=converter, num_particles=[1, 1], num_spin_orbitals=4)
+
+        ansatz._build()
+        print(vars(ansatz))
+
+        assert ansatz.num_qubits == 4
```
Cryoris, line 34:
```suggestion
        self.assertEqual(ansatz.num_qubits, 4)
```

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
woodsp-ibm: Do people/chemists know what pure-spin is. That it seems to be, from what you sau, to bes what was called same spin before, where the text was more understandable to me since it explained things and did not rely on understanding some  term - I did not hence I thought impure (ie not pure) might be flipping the spin. So when I said 'Can you say here what pure spin means' I meant could you improve the docstring to clarify things - maybe like you explained here in the comments

qiskit_nature/circuit/library/ansaetze/uccsd.py
woodsp-ibm: In cases like the above, since its referring to a class we always used `:class:`. `:meth:` when referring to a method etc.

test/circuit/library/ansaetze/test_uccsd.py
```diff
@@ -0,0 +1,34 @@
...
+    """Tests for the UCCSD Ansatz."""
+
+    def test_ucc_ansatz(self):
+        """Tests the UCCSD Ansatz."""
+        converter = QubitConverter(JordanWignerMapper())
+
+        ansatz = UCCSD(qubit_converter=converter, num_particles=[1, 1], num_spin_orbitals=4)
+
+        ansatz._build()
+        print(vars(ansatz))
```
woodsp-ibm, line 32: And can we drop any print() statements - I assume this was added for debugging.

qiskit_nature/circuit/library/ansaetze/__init__.py
woodsp-ibm: > What do you mean by "its listed as the plural of ansatze"?

Sorry I meant if go look up ansatze then it will 'list' ansatzes as the plural of that word.

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
mrossinek: Ah I see :sweat_smile: I will update the code later :+1: I can also call it `same_spin` again, in order to be consistent with what we had before. I didn't check what it used to be and simply went with something that popped into my head.

qiskit_nature/circuit/library/ansaetze/ucc.py
mrossinek: I am no longer sure tbh. But that `[]` in `__init__` looks weird to me. Will check later

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
mrossinek: I see that the old `same_spin_doubles` argument actually did exactly the opposite of `pure_spin`: it was used to disable the alpha,alpha and beta,beta excitations while always including alpha,beta ones.
`pure_spin` does the opposite: enable/disable alpha,beta.

I will update the code to use `same_spin_doubles` as this was the original intent (support an equivalent set of configuration options).

mrossinek: This was actually a little bit more complicated than I initially assumed. I update the code now so it is worth another check.

qiskit_nature/circuit/library/ansaetze/excitation_builder.py
```diff
@@ -0,0 +1,115 @@
...
+"""
+
+from typing import List, Tuple
+
+import itertools
+import logging
+
+from qiskit_nature.operators.second_quantization import FermionicOp, SecondQuantizedOp
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
mrossinek: I consider this an open TODO: add logging :wink:

mrossinek:
A quick overview of open tasks which I plan to tackle before the end of the week [estimated dates]:
- [ ] add logging [_Thursday_]
- [ ] add/update docstrings [_Thursday_]
- [ ] write proper unittests [_Thursday_]
- [ ] migrate stack to use new `UCCSD` rather than old one [_Friday_]
- [ ] check excitation generator compatibility with `UVCC` [_Friday_]

I will take care of the `UCC` and `UCCSD` classes.
@Cryoris please take care of the `EvolvedOpAnsatz`

mrossinek:
Some updates and open tasks until the next Wednesday meeting:
- [x] `PUCCD`
- [x] `SUCCD`
- [ ] vibrational mode excitation generator (WIP)
- [ ] `UVCC`

I may also include here:
- [ ] `VSCF`
- [ ] `CHC`
Both of these should be pretty format to migrate.

mrossinek:
Alright, I gave this some more thoughts. As you can see, I also added a WIP version of the vibration excitation list generator.
However, in order to also implement `UVCC` and `VSCF` we need to have access to the `VibrationalSpinOp` which is not yet merged (#79). Thus, I suggest we continue without this for now.

I will also suggest waiting with `CHC` (although it is straight forward to migrate) just because we need the `VibrationalProblem` in order to test it properly.

mrossinek:
Open questions:

- [ ] Can we already migrate the stack to use UCCSD without #83 resolved?
- [ ] What will happen to `qiskit_nature.mp2info`? That is coupled to the old `UCCSD` code? Do we need that for this release? (@pbark)
- [ ] The adaptivity of the `UCC` Ansatz still needs to be looked at. I will do this with @Cryoris early next week.
- [ ] The old `UCCSD` code had the `succ_full` option. There is an open `TODO` for this variant in the `SUCC` class. This also requires discussion with @Cryoris

mrossinek:
Some updates:

- `QuantumCircuit` does not acutally need to be tested because internally everything builds up from the already tested evolutions in Terra
- adaptivity should be supported now. Once we look at the algorithm stack I will update `AdaptVQE` accordingly
- `succ_full`: it is fine if we only support `succ` for now. I know how to implement `succ_full` but it requires more time and a slightly different approach to `succ` because it not only needs to adapt the `excitation_list` but also the way the operators are constructed from them.

pbark:
Regarding `mp2info` I think it is straightforward to calculate from the excitation builder and the integrals. It should be kept attached to the UCCSD since the mp2 coeffs can be good approximations for initial guess of the cluster coefficients.

mrossinek:
> Regarding `mp2info` I think it is straightforward to calculate from the excitation builder and the integrals. It should be kept attached to the UCCSD since the mp2 coeffs can be good approximations for initial guess of the cluster coefficients.

To clarify: I did not mean to outright remove it, merely to postpone it until a feature release in the future.
We discussed offline and agreed on doing exactly that.

qiskit_nature/circuit/library/ansaetze/evolved_operator_ansatz.py
Cryoris: Updated 👍🏻

qiskit_nature/circuit/library/ansaetze/evolved_operator_ansatz.py
```diff
@@ -0,0 +1,149 @@
...
+        if len(operators) > 1:
+            num_qubits = operators[0].num_qubits
+            if any(operators[i].num_qubits != num_qubits for i in range(1, len(operators))):
+                raise ValueError('All operators must act on the same number of qubits.')
+
+        self._invalidate()
+        self._operators = operators
+
+    @property
+    def qregs(self):
```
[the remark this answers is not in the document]
Cryoris: It's not but should be 🙂

qiskit_nature/circuit/library/ansatzes/evolved_operator_ansatz.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class EvolvedOperatorAnsatz(BlueprintCircuit):
+    """The evolved operator ansatz."""
+
+    def __init__(self,
+                 operators: Optional[Union[OperatorBase, List[OperatorBase]]] = None,
+                 reps: int = 1,
+                 evolution: Optional[EvolutionBase] = None,
+                 insert_barriers: bool = False,
```
woodsp-ibm, line 29: Shouldn't this have a initial state (QuantumCircuit) here too in the base class as this should be directly usable if you have your own set of operators for this.

qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -0,0 +1,356 @@
...
+
+    @excitations.setter
+    def excitations(self, exc: Union[str, int, List[int], Callable]) -> None:
+        """Sets the excitations."""
+        self._invalidate()
+        self._excitations = exc
+
+    # TODO: the `preferred_init_points`-implementation can (and should!) be improved!
+    @property
+    def preferred_init_points(self):
```
mrossinek, line 208: Do we want this to be on the `EvolvedOpAnsatz` level now that we moved `initial_state` down there, too?

woodsp-ibm: I think that makes sense right to have it there since it should apply to any evolved op ansatz that if you want the output state to reflect the input one, then the parameters needed to be set are all 0.

pbark (approved):

mrossinek:
@woodsp-ibm I know that you did not have time to review this yesterday but we need to continue with the algorithms where we rely on this work. Thus I merged it.
If you find something that needs to change, please comment here and tag me. Then I can address it in a follow-up PR :+1:

PR #68, opened 2021-03-03
Improve FermionicOp and SpinOp

ikkoham:
### Summary

1. Fix the but about an empty string (`FermionicOp("")` is not allowed now.)
1. Improve test with assertFemionEqual 
2. ~Deprecate initialization with tuple (`FermionicOp(("I+", coeff))`). Users should use `coeff * FermionicOp("I+")` or `FermionicOp([("I+", coeff)])` instead.~
3. Support sparse label for FermionicOp
4. Sparse label requires register_length.
4. Make internal `FermionicOp._coeff_list` np.ndarray
6. ~Implement `FermionicOp.to_normal_order_list()`~ Move to #88.


woodsp-ibm:
In regards of 3 above, often we have APIs that take a List[x] (where x in this case is a Tuple[str, complex]) and to simplify things for a user, where they have only one element, so they do not have to make a List to pass it in, we support x directly as a type. In this case you had that and have removed it. Instead requiring the user to make it into a List or to break the Tuple apart to use a different style to initialize it where presumably by passing a str the coeff is implied to be 1 and then multiply it afterwards. Having it support the Tuple directly seems consistent with the usability we do elsewhere. What is the reason to remove it?

ikkoham:
Thanks. For 3, I thought it is just adding square brackets to make a list. The reason for deprecating it is that it would be confusing if there were various initialization methods, and I wanted to save tuple for initializing directly from internal data in the future. Of course, this (latter) isn't necessary because we can detect the type from the element of tuple. If you want to use `Tuple[label, coeff]`, I will revert it.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -96,69 +128,55 @@ class FermionicOp(ParticleOp):
 
     """
 
+    _SPARSE_LABEL_PATTERN = re.compile(r"^([I\+\-NE]_\d\s)*[I\+\-NE]_\d(?!\s)$")
+    _DENSE_LABEL_PATTERN = re.compile(r"^[I\+\-NE]+$")
+
     def __init__(
-            self,
-            data: Union[str, Tuple[str, complex], List[Tuple[str, complex]]],
```
mrossinek, line 101: I agreee with @woodsp-ibm and would like to keep the `Tuple` option available here :+1:

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -50,19 +52,49 @@ class FermionicOp(ParticleOp):
...
+    For example,
+
+    .. code-block:: python
+
+        "+_0"
+        "-_2"
+        "+_10 -_4 +_1 I_0 +_0"
+
+    are possible labels.
+    The :code:`index` must be in descending order.
```
mrossinek, line 85: Is there a particular reason for this limitation?

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: OK

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Yes. Because "+_1 -_0" is not equal to "-_0 +_1".

mrossinek: Exactly. So why can't we allow the latter?

ikkoham: Ah, I got what you said.

It is possible to implement it. But, I think it's non-trivial behavior to initialize an operator that is different from the input label.
Above case, `FermionicOp([('+-', -1)])` is generated.
If user wants to do that,  write explicitly
```
FermionicOp("I_1 -_0") @ FermionicOp("+_1")
```
for now. ("I_1" is not so cool, I want to solve this in the future...)

pbark: Although for the purpose of generality it would be good to support any ordering, I think that respecting the ordering here is not bad in view of the commutators algebra that need to be developed further (in another PR). I believe that keeping the indices in descending order will make the reading of the operator easier. Solving the "I_1" would be nice, as @ikkoham pointed.

pbark:
This LGTM. This should also be reviewed by @woodsp-ibm and after the review we can merge it.

qiskit_nature/operators/second_quantization/fermionic_op.py
mrossinek: What exactly needs to happen here to resolve `I_1`? Is that just required in order to make sure that the `register_length` is correct?

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -12,12 +12,13 @@
...
 """The Fermionic-particle Operator."""
 
+import re
 from typing import List, Optional, Tuple, Union
 
 import numpy as np
 
 from qiskit_nature import QiskitNatureError
-from .particle_op import ParticleOp
+from qiskit_nature.operators.second_quantization.particle_op import ParticleOp
```
mrossinek, line 21: Why did you change the import here?

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -329,23 +369,54 @@ def reduce(self, atol: Optional[float] = None, rtol: Optional[float] = None) ->
...
+            return False
         return set(label).issubset({"I", "+", "-", "N", "E"})
+
+    def _from_sparse_label(self, list_label):
+        num_terms = len(list_label)
+        parsed_data = []
+        max_index = 0
+        for term, label in enumerate(list_label):
+            prev_index = -1
+            list_label = label.split()
```
mrossinek, line 397: You are reusing the `list_label` variable here. Please change that.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -329,23 +369,54 @@ def reduce(self, atol: Optional[float] = None, rtol: Optional[float] = None) ->
...
+                if 0 <= prev_index <= index:
+                    raise ValueError("Indices of labels must be in descending order.")
+
+                max_index = max(max_index, index)
+
+                if op_label in {"+", "-", "N", "E"}:
+                    parsed_data.append((term, op_label, index))
+                prev_index = index
+
+        self._register_length = max_index + 1
```
mrossinek, line 411: I am not sure how I had missed this during the review of the `SpinOp` PR but I assume that this is what you refer to with the `I_1` problem above.
I think it ought to be necessary that the user can simply specify/overwrite the `register_length` himself when using sparse labels. Otherwise it is rather unintuitive that you need to add `I_{register_length - 1}` in order to encode the register length into a sparse label.
If we do not want to add a constructor argument then we need to document this more properly.

test/operators/test_fermionic_op.py
```diff
@@ -18,73 +18,93 @@
...
-        "".join(label) for label in product(["I", "+", "-", "N", "E"], repeat=length)
-    ]
+    return ["".join(label) for label in product(["I", "+", "-", "N", "E"], repeat=length)]
 
 
 @ddt
 class TestFermionicOp(QiskitNatureTestCase):
     """FermionicOp tests."""
 
+    def assertFermionEqual(self, first: FermionicOp, second: FermionicOp):
```
mrossinek, line 33: I think it may be useful to implement `__eq__` in the `FermionicOp` class instead of this.

mrossinek (changes requested):

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: absolute import is better than relative import

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: oh my god. thanks.

test/operators/test_fermionic_op.py
ikkoham: This is not true equality. Users will be confused.

qiskit_nature/operators/second_quantization/fermionic_op.py
woodsp-ibm: I am surprised that unit tests did not catch this

woodsp-ibm:
FYI: In commenting on unit test above I will note that in #60 I added a unit test for qubit_converter, which is in operators/second_quant.. and in doing so I created second_quant... folder in that PR and moved the 2 existing unit tests there (which includes the one modified here). That PR is not yet merged so we will have to reconcile things between these.

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: We can choose:

1. Explicit: `register_length`
2. Implicit: automatically expansion to larger register_length operator

I prefer the explicit one from the perspective of the Zen of Python.

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: But this works correctly in Python and we can not make a test to catch this.

This is same with
```python
num_list = [0, 1, 1, 2, 3]
for num in num_list:
    num_list = [1000]
    print(num)
```
outputs 
```
0
1
1
2
3
```. 
Not 1000.

Of course, I will fix it because readability is bad.

File: test/operators/second_quantization/test_fermionic_op.py [hunk 7]
```diff
@@ -18,73 +18,102 @@
...
+            ("II+IIIIIII-II", 1 + 2j),
+            ("-IIIIIIIIIIII", 56),
+        ]
+        self.assertListEqual(actual.to_list(), desired)
 
     def test_neg(self):
         """Test __neg__"""
         fer_op = -FermionicOp("+N-EII")
-        self.assertListEqual(fer_op.to_list(), [("+N-EII", -1)])
+        targ = FermionicOp([("+N-EII", -1)])
```
mrossinek, line 81: What does `targ` stand for?

test/operators/second_quantization/test_spin_op.py
```diff
@@ -66,45 +67,51 @@ def assertSpinEqual(first: SpinOp, second: SpinOp):
     @data(*spin_labels(1))
     def test_init_label(self, label):
```
mrossinek, line 68: Please add some `test_init_from_dense`

test/operators/second_quantization/test_spin_op.py
```diff
@@ -66,45 +67,51 @@ def assertSpinEqual(first: SpinOp, second: SpinOp):
...
         """Test __init__"""
-        spin = SpinOp(f"{label}_0")
+        spin = SpinOp(f"{label}_0", register_length=1)
         self.assertListEqual(spin.to_list(), [(f"{label}_0", 1)])
 
     @data(*spin_labels(2))
```
mrossinek, line 73: I also don't see the `Tuple`-init case being covered. In fact, I think it may be broken. Please check.

mrossinek (changes requested):

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -172,17 +211,25 @@ def __str__(self) -> str:
         if len(self) == 1:
             label, coeff = self.to_list()[0]
             return f"{label} * {coeff}"
-        return "  " + "\n+ ".join(
-            [f"{label} * {coeff}" for label, coeff in self.to_list()]
-        )
+        return "  " + "\n+ ".join([f"{label} * {coeff}" for label, coeff in self.to_list()])
+
+    def __len__(self):
+        return len(self._labels)
```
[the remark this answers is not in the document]
ikkoham: These are different. `len` is the number of terms.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -201,7 +248,7 @@ def compose(self, other: "FermionicOp") -> "FermionicOp":
                     new_data.append((new_label, cf1 * cf2 * sign))
 
             if not new_data:
-                return FermionicOp(("I" * self._register_length, 0))
+                return FermionicOp([("I" * self._register_length, 0)])
```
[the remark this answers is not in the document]
ikkoham: Yes. Thanks. This change is introduced when tuple is deprecated. I forgot to revert it.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -172,17 +211,25 @@ def __str__(self) -> str:
...
+        return self._register_length
 
     def mul(self, other: complex) -> "FermionicOp":
         if not isinstance(other, (int, float, complex)):
             raise TypeError(
                 f"Unsupported operand type(s) for *: 'FermionicOp' and '{type(other).__name__}'"
             )
         return FermionicOp(
-            list(zip(self._labels, [coeff * other for coeff in self._coeffs]))
+            list(zip(self._labels, [coeff * other for coeff in self._coeffs.tolist()]))
```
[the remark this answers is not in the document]
ikkoham: Thnaks. And tolist() is necessary here.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -285,27 +332,21 @@ def add(self, other: "FermionicOp") -> "FermionicOp":
         if self.register_length != other.register_length:
             raise TypeError("Incompatible register lengths for '+'.")
 
-        label1, coeffs1 = zip(*self.to_list())
-        label2, coeffs2 = zip(*other.to_list())
-
-        return FermionicOp(list(zip(label1 + label2, coeffs1 + coeffs2)))
+        return FermionicOp(
+            list(
+                zip(self._labels + other._labels, np.hstack((self._coeffs, other._coeffs)).tolist())
```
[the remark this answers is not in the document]
ikkoham: if tolist() is called, the type of coeffs is complex, but if not called, the type is np.complex128.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -329,23 +370,32 @@ def reduce(self, atol: Optional[float] = None, rtol: Optional[float] = None) ->
...
-        new_labels = label_list[non_zero].tolist()
-        new_coeffs = np.array(coeff_list)[non_zero].tolist()
-        return FermionicOp(list(zip(new_labels, new_coeffs)))
+            return FermionicOp([("I" * self.register_length, 0)])
+        return FermionicOp(list(zip(label_list[non_zero].tolist(), coeff_list[non_zero])))
 
-    def __len__(self):
-        return len(self._labels)
+    def _from_sparse_label(self, labels):
+        list_label = [["I"] * self._register_length for _ in labels]
```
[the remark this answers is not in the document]
ikkoham: No. We can't. For example, 
```python
ls = [["I"] * 2] * 3
ls[2][1] = 2
print(ls)
```
[['I', 'X'], ['I', 'X'], ['I', 'X']]
```python
ls = [["I"] * 2 for _ in range(3)]
ls[2][1] = 2
print(ls)
```
[['I', 'I'], ['I', 'X'], ['I', 'I']]

test/operators/second_quantization/test_fermionic_op.py
ikkoham: target. In terra, this is used (https://github.com/Qiskit/qiskit-terra/search?q=targ)

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
mrossinek: Why is that necessary here?

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
mrossinek: Ah right.. We don't get new instances but the same instance `n` times.. Thanks :+1:

qiskit_nature/operators/second_quantization/fermionic_op.py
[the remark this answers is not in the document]
mrossinek: Alright, this also explains my question above :+1:

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -212,33 +207,51 @@ def __init__(
...
-
             data = self._flatten_ladder_ops(data)
 
             labels, coeffs = zip(*data)
             self._coeffs = np.array(coeffs, dtype=dtype)
 
-            if all(self._SPARSE_LABEL_PATTERN.match(label) for label in labels):
-                self._from_sparse_label(labels)
-            elif all(self._DENSE_LABEL_PATTERN.match(label) for label in labels):
+            if register_length is None:  # Dense label
```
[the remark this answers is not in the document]
ikkoham: `register_length` is valid only for the sparse label as a required parameter, as explained in the documentation.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -54,7 +55,7 @@ class FermionicOp(ParticleOp):
     There are two types of label modes for :class:`FermionicOp`.
     The label mode is automatically detected.
 
-    1. Dense Label (if underscore `_` does not exist in the label)
+    1. Dense Label (default, `register_length = None`)
 
     Dense labels are strings with allowed characters above.
     This is similar to Qiskit's string-based representation of qubit operators.
```
woodsp-ibm, line 61: In this case register_length is inferred from the label length. I guess its a mistake to state it explicitly, when they are dense labels, as its not going to work?

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -437,35 +450,30 @@ def to_matrix(self) -> np.ndarray:
...
+            for splitted_label in label.split():
+                xyz, nums = splitted_label.split("_", 1)
+
+                if xyz not in xyz_dict:
+                    continue
+
+                xyz_num = xyz_dict[xyz]
+                index, power = map(int, nums.split("^", 1)) if "^" in nums else (int(nums), 1)
+                if index >= self._register_length:
+                    raise ValueError("index must be smaller than register_length.")
```
woodsp-ibm, line 467: Its nice/helpful to have the value(s) being referred to be in the message as well ie index and register_length, though it might be more meaningful to a end user to have the whole label there so they can see it in context better

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -437,35 +450,30 @@ def to_matrix(self) -> np.ndarray:
...
+                register = self._register_length - index - 1
+                # Check the order of X, Y, and Z whether it has been already assigned.
+                if self._spin_array[range(xyz_num + 1, 3), term, register].any():
+                    raise ValueError("Label must be XYZ order.")
```
woodsp-ibm: Similalrly here (and below) the failing label value in the message would be helpful

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -203,50 +195,80 @@ def __init__(
...
+            sparse = r"([IXYZ]_\d+(\^\d+)?|[\+\-]_\d+?)"
+            # space (\s) separated sparse label or dense label (repeat of [IXYZ+-])
+            label_pattern = re.compile(rf"^({sparse}\s)*{sparse}(?!\s)$|^[IXYZ\+\-]+$")
+            invalid_labels = [label for label, _ in data if not label_pattern.match(label)]
             if invalid_labels:
                 raise ValueError(f"Invalid labels: {invalid_labels}")
 
+            if register_length is not None:
+                data = self._flatten_raising_lowering_ops(data, register_length)
             data = self._flatten_ladder_ops(data)
```
mrossinek, line 224: This does not work as intended, as far as I can tell. Some examples which fail:
```
SpinOp('+_0 i_0')
SpinOp('+_0')  -> this will not be affected by `_flatten_raising_lowering_ops` but then goes through `_flatten_ladder_ops` (which only knows how to handle dense ops)
```

ikkoham: > SpinOp('+_0 i_0')

This must be error. What is i_0?

> SpinOp('+_0')

Error (register_length is required)

> SpinOp('+_0', register_length=1)

Correct.
`SpinOp([('X_0', (1+0j)), ('Y_0', 1j)], spin=1/2, register_length=1)`

mrossinek: Ah my apologies. The `register_length` fixes it.
But the error was kind of unexpected because it complains about the invalid labels for dense labels (In the `+_0` case) *after* converting it into `X_0 Y_0`. Maybe we could do an assertion of the labels before that.

ikkoham: OK I will add the error message somewhere.

ikkoham:
Please merge this PR if normal_order_list() is ok because this PR blocks other PRs. I will add additional tests and improvement of error messages in follow up PR.

mrossinek:
> Please merge this PR if normal_order_list() is ok because this PR blocks other PRs. I will add additional tests and improvement of error messages in follow up PR.

I understand you want to get this out. But, let us wait for the offline decision with regards to the endianess convention.
Which PRs are blocked by this one? I just closed the p/h transformer and I know that `VibrationalProblem` needs this. Anything else?

ikkoham:
Yes. PH and VibrationalProblem.

mrossinek (approved):
Let's merge this PR today in order to unlock the further development of features which are blocked by this.

*Warning*: the sparse labels are still using little endian. A follow-up PR will fix this!

@ikkoham please also address the remaining open comments of this PR (mainly improvements to error messages as well as additional unittests) in another follow-up PR.

ikkoham:
> Warning: the sparse labels are still using little endian. A follow-up PR will fix this!

Dense label is also little endian now. I've implemented all operators as little endian from #25 and #35.

ikkoham:
Anyway thanks a million. @mrossinek

mrossinek:
> Dense label is also little endian now. I've implemented all operators as little endian from #25 and #35.

Interesting I had not noticed.. I would have assumed that the mapper unittests would fail in this case :thinking:
But let's still proceed for now. The switch to big endian should follow-up early this week anyways :+1:

ikkoham:
Sorry. I was wrong. #25 does not assume endian (endian does not influence) and #35 is little e.g. https://github.com/Qiskit/qiskit-nature/pull/35/files#diff-e1fad0f243ef03387da83f3f9c9b006212b413a01addd08569d3d78048469dd1R131-R132.

pbark (approved):

pbark:
I think now this LGTM. We need to add the normal ordering in a different PR for future reference and resolve the big endian discussion we had offline (different PR)

PR #73, opened 2021-03-04
Remove Aqua dependencies

manoelmarques:
### Summary

Remove all Aqua dependencies and use Terra algorithms/opflow


qiskit_nature/components/variational_forms/chc.py
```diff
@@ -20,8 +20,7 @@
 
 from qiskit.circuit import ParameterVector, Parameter
 
-from qiskit.aqua.components.variational_forms import VariationalForm
-from qiskit.aqua.components.initial_states import InitialState
+from qiskit.algorithms.variational_forms import VariationalForm
```
mrossinek, line 23: Why does this class exist in Terra? This should *not* be the case and instead we should refactor all classes in Nature which depend on it to become a `BluePrintCircuit`.

@Cryoris tagging to verify :+1:

qiskit_nature/core/__init__.py
```diff
@@ -19,7 +19,7 @@
 **DEPRECATED** See :mod:`qiskit_nature.transformations` which replace this.
```
mrossinek, line 19: As seen here, this module is deprecated. Please remove it entirely :+1:

qiskit_nature/algorithms/eigen_solvers/q_eom_ee.py
```diff
@@ -27,7 +27,7 @@
 class QEomEE(NumPyMinimumEigensolver):
```
mrossinek, line 27: I am actually surprised to see this class as well as `QEomVQE` and `QEquationOfMotion`. All of these classes should have been deprecated when we released `qiskit.chemistry.algorithms.excited_state_solvers.QEOM`.

@paulineollitrault Please verify the above

If this is correct, @manoelmarques please remove the entire `qiskit_nature.algorithms.eigen_solvers` module :+1:

qiskit_nature/fermionic_operator.py
```diff
@@ -12,6 +12,7 @@
 
 """ Fermionic Operator """
```
mrossinek, line 13: Same as with the `BosonicOperator`, you can remove the `FermionicOperator` entirely.

qiskit_nature/transformations/bosonic_transformation.py
```diff
@@ -22,9 +22,9 @@
 import numpy as np
```
mrossinek, line 23: You can remove the entire `qiskit_nature.transformations` module as all of this is being refactored :+1:

test/algorithms/eigen_solvers/test_qeom_ee.py
```diff
@@ -18,8 +18,8 @@
 from test import QiskitNatureTestCase
 import numpy as np
```
mrossinek, line 20: Corresponding to my earlier comment, the module `test.algorithms.eigen_solvers` can be removed :+1:

test/applications/test_app_mgse.py
```diff
@@ -16,13 +16,12 @@
 import unittest
```
mrossinek, line 17: Since `qiskit_nature.applicartions` can be removed, so can `test.applications`.

test/core/test_core_hamiltonian.py
```diff
@@ -16,7 +16,7 @@
 import unittest
```
mrossinek, line 17: The `test.core` module can be removed

test/test_bksf_mapping.py
```diff
@@ -15,8 +15,7 @@
 import unittest
 from test import QiskitNatureTestCase
 import numpy as np
-from qiskit.quantum_info import Pauli
-from qiskit.aqua.operators import WeightedPauliOperator
+from qiskit.opflow import PauliSumOp
 from qiskit_nature.bksf import edge_operator_aij, edge_operator_bi
```
mrossinek, line 20: Since I proposed to remove `qiskit_nature.bksf`, you can also remove this test file

test/test_bosonic_operator.py
```diff
@@ -17,7 +17,7 @@
 
 import numpy as np
```
mrossinek, line 19: Please remove in accordance with the removal of `BosonicOperator`

test/test_fermionic_operator.py
```diff
@@ -16,8 +16,7 @@
 import unittest
```
mrossinek, line 16: Please remove in accordance with the removal of `FermionicOperator`

test/test_particle_hole.py
```diff
@@ -16,7 +16,7 @@
```
mrossinek, line 16: I think we can remove both, this test file as well as `qiskit_nature.particle_hole`. A PR is on the way by @pbark anyways, correct?

test/transformations/test_bosonic_transformation.py
```diff
@@ -17,9 +17,9 @@
 from test import QiskitNatureTestCase
```
mrossinek, line 18: Please remove `test.transformations` entirely :+1:

mrossinek (changes requested):
Alright I did a first pass of looking at the changes which you made here.
Once you had the time to address some of the comments below, I will also check out the branch locally and ensure that we are not overlooking anything (since the unchanged places are rather hard to review in the web interface :wink:)

In any case, thanks a lot @manoelmarques for the speedy PR :+1:

qiskit_nature/algorithms/eigen_solvers/q_equation_of_motion.py
```diff
@@ -21,25 +21,23 @@
...
 from qiskit_nature import FermionicOperator
+from qiskit_nature.exceptions import QiskitNatureError
 
 logger = logging.getLogger(__name__)
 
 
 class QEquationOfMotion:
     """ QEquationOfMotion algorithm """
-    def __init__(self, operator: LegacyBaseOperator,
+    def __init__(self, operator: OperatorBase,
```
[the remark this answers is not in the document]
manoelmarques: I agree, in fact I was passing None in the init and later setting the operator private member variable in the new versions of the algorithms. I will change it.

qiskit_nature/algorithms/excited_states_solvers/qeom.py
```diff
@@ -409,9 +414,67 @@ def _compute_excitation_energies(m_mat: np.ndarray, v_mat: np.ndarray, q_mat: np
         return excitation_energies_gap, res[1]
 
 
-class QEOMResult(AlgorithmResult):
+class QEOMResult(AlgorithmResult, collections.UserDict):
```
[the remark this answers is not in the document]
manoelmarques: I agree. We should be inline with the Qiskit result classes. I added the UserDict temporarily because there were so many other changes that I didn't want to add more stuff to this PT. I waqs going to change it in a follow-up PR but can go ahead and fix it here.

qiskit_nature/algorithms/eigen_solvers/q_equation_of_motion.py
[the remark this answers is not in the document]
manoelmarques: Removed from init and added to calculate_excited_states

qiskit_nature/algorithms/eigen_solvers/q_eom_ee.py
woodsp-ibm: The lack of deprecation was most likely just an oversight given all the work that went on. It was my understanding that the excited state algorithms were being used now and were the only ones that were wanted.

qiskit_nature/components/variational_forms/chc.py
woodsp-ibm: My guess it was maintained there pending its complete removal from Qiskit, VQE still supports the present UCCSD based on VariationalForm. I think once the UCCSD refactor is done we should eliminate this from Terra. If we don't mind breaking things in advance, then of course it could be removed from Terra now.

woodsp-ibm:
I imagine that trying to remove some of the files, like FermionicOperator ends up with a ton of code not loading such that it would all need to be changed. Hence in switching over to Terra opflow, and removing dependency on Aqua, you did some changes to use the PauliSumOp to mitigate the extent of the overall changes. I guess this leaves the algorithms working with the old stack where the old stack has been switched to use Terra. Now the algorithms need to be changed to use the new stack which is still work in progress. But I guess this way things are more working and can be changed over more incrementally. As soon as the algorithms are changed to use the new stack then FermionicOperator etc can be deleted. At that point in time deleting them should be straightforward since it should be a set of code that is self-dependent but no longer used elsewhere,

mrossinek:
I agree with your reasoning and also with @manoelmarques (talked offline). I think we should keep the old `Fermionic/BosonicOperator` and the `transformations` around for now, in order to leave the old stack mostly working.

Nonetheless, I think we can remove the following things:
- [ ] `qiskit_nature.applications
- [ ] `qiskit_nature.core`
- [ ] `qiskit_nature.algorithms.eigen_solvers`

qiskit_nature/components/variational_forms/chc.py
mrossinek: I think we should remove it from Terra right away. All imports of the `VaritionalForm` class are based on the Aqua module.

woodsp-ibm (approved):
Looks good to me. I see that switching we lost a significant percentage of unit test coverage due to the current breakage. As the logic gets switched over and test fixed and/or new ones are created this should get corrected.

qiskit_nature/algorithms/__init__.py
```diff
@@ -11,31 +11,18 @@
 # that they have been altered from the originals.
 
 """
-Chemistry specific Aqua algorithms (:mod:`qiskit_nature.algorithms`)
+Nature Sciences specific algorithms (:mod:`qiskit_nature.algorithms`)
```
mrossinek, line 14: Shouldn't it rather by "Natural sciences" or just "Nature" if you want to reflect the repo's name?

mrossinek:
Ah, looks like I was too slow on this one..

qiskit_nature/algorithms/__init__.py
woodsp-ibm: I think there is going to be quite a lot of documentation cleanup that needs to be done. Manoel really just got rid of Aqua from the text and changed things to build. But it needs going over - since this is all under Nature Sciences now, and are different from Aqua algorithms (before all we had were QuamtumAlgorithm based algos like Aqua, but we did not put them there as they needed params (or were built out internally) from logic that was here.)
Since all these domain repos now have Algorithms etc we can try and make this text consistent across them when we do this work later on the docs.

Issue #64, opened 2021-02-25
Disambiguate `num_orbitals`


mrossinek:
### What is the expected enhancement?

@Cryoris brought up the very good suggestion, that we should make an effort to disambiguate the commonly used `num_orbitals` keyword argument. This is used throughout a large part of the stack with either of the two meanings:
1. number of *molecular* orbitals
2. number of *spin* orbitals

Before releasing Nature, we should attempt an effort to clearly disambiguate these cases by replacing the argument with (e.g.)
1. `num_molecular_orbitals`
2. `num_spin_orbitals`

This will clearly tell the user what is expected without having to look into the docstring of each class.

woodsp-ibm:
Yes, agree. Closer to the user problem where the end user was more likely to configure things we had leaned towards allowing them to think etc in MOs. As soon as we get to the operator then its block spin. Arguably MOs is a single number whereas perhaps spin ought to be a tuple, like the num particles, so if that were the case the type would help too, though of course naming is front and center

mrossinek:
Do you mean you would like `num_spin_orbitals` to be a tuple in order to easily differentiate the case of seniority zero or do you see another reason here?

woodsp-ibm:
In the prior FermionicOperator it would be possible to remove say the spin-down orbital but not the corresponding spin-up one so there are more spin up than down orbitals - I do not know if one needs to know, but if one did then a tuple would help ie (num spin up, num spin down). I know we do the particles this way, just wondering about doing it for the spin orbitals too. I do not know how seniority zero affects this other than the orbitals are doubly occupied so I guess this does not really apply in this case so maybe the single number is the only common notion - though with seniority zero I am not sure they are spin orbitals in the way we normally think of them

mrossinek:
> In the prior FermionicOperator it would be possible to remove say the spin-down orbital but not the corresponding spin-up one so there are more spin up than down orbitals

I cannot think of an application where this makes sense physically. Thus, I'd argue that the `int` for the number of orbitals is sufficient. @pbark what do you think?

PR #77, opened 2021-03-09
ActiveSpaceTransformer generalization

### LINKED ISSUE #47: Implement the FreezeCoreTransformer
### What is the expected feature?

Add the `FreezeCoreTransformer`.

- [x] actual implementation
- [x] unittests
- [x] documentation

mrossinek:
### Summary

Closes #47

As discussed in the linked issue above, the `ActiveSpaceTransformer` can actually be generalized in order to provide the full functionality which was previously available through the `freeze_core` option. This PR implements this generalization.

### Details and comments

I am opening this PR in a *very early draft* stage in order to allow discussion from the very beginning. I plan to do the following things:

- [x] extend the MO space reduction to the dipole operators
- [x] refactor the code to be more general (i.e. make clear that the methods are *no longer* tailored to the energy case)
- [x] add quick access keyword arguments for commonly usage scenarios:
   - `freeze_core`
   - and maybe one more option to simplify cutting of a large number of inactive orbitals (e.g. only applicable to unoccupied orbitals). The idea here is to simplify usage when one would otherwise have a very large active space. But in general, the approach of specifying the active space is more scalable than having to specify everything that is inactive (as it used to be the case in Aqua)
- ~(maybe) refactor the code to also permit scenarios in which a `QMolecule` does not have any AO information (but only MO)~ Edit: postponed until further notice


mrossinek:
This is ready for review now :+1: 
There is an optional open TODO left in the description but I am not sure whether it will be of use for now. Below is a little bit more context:

Currently, the `ActiveSpaceTransformer` requires `mo_coeff`, `hcore`, `eri`, etc. in order to work. This is fine, because all drivers that we have at the moment provide these fields, except for the `FCIDumpDriver`. However, this driver will likely only be used with already reduced output. We have two options now:
1. Stick with this limitation (for now) and document it (I would add that depending on the outcome of this discussion).
2. We can also perform the AS reduction not from `AO->MO` (as done now) but within `MO->MO`. This requires a "dummy" `mo_coeff` object which has only `1`s as coefficients (in the correct places). These places can be inferred based on the AS configuration. Thus, I can add logic which checks whether the AO fields are available. If not, it will fall back to this method.

What do you think?

woodsp-ibm:
In the reduction that was done before freeze and the orbital specification were logically or'd together. In other words the freeze core was simple way without explicitly stating what the orbitals were. Since the freezing and reduction were all about removing it maybe easy to think of the logical or. But I think we can only apply one of these transformations right - we cannot have one with freeze core and another to change the remainder of the active space - or at least from the energies being saved they would be overwritten hence it appears not. But it seems fine to me to be able to say freeze core plus have some added definition of the other active space.

I agree that the prior active space was not very flexible - but in general to that point we were dealing only with small systems and it sufficed. It did have the ability to support both positive or negative indexes for convenience (where negative -1 is the highest entry, -2 next one etc. Maybe that is still the case here. Perhaps we should consider the specification for the orbitals to be a List of ether int or slice. That way a contiguous range is easier to define with a slice so each orbital does not need to be individually listed.

As to the limitation, is it much code to infer what is needed to get it to work with FCIDumpDriver? Even so we could go now with say raising an exception and later adding this code.

Does this completely do away with the reduction logic that we had before. I.e is there any point in having a similar transform based on that logic, since that did not have that requirement and worked across the board. The logic also permitted any single spin orbital to be removed, though that was never exposed in the Hamiltonian/Transformation as a capability rather as it was user facing it was defined in MOs and acted thus on both spin orbitals resulting from them.

And in regards of #64 it would be good to be more precise here. Also it might be nice for an example here showing orbitals before and after in some way.

Final thought. Since UCCSD had active space definition it would be nice if these were compatible so a user configures things the same way. Of course by the time UCCSD sees the set of orbitals the indexes it knows about would not be the same as the molecule started with so the user would have to account for that I think and give its active space in terms or what remained. Or is the plan to drop it from UCCSD? Having unoccupied orbital qubits that don't do anything is inefficient so maybe UCCSD should always act on the full space of what remains and drop active space from UCCSD. If its s more than one place though, maybe like excitation logic, there should be some active space logic that ends up in common place  for any modules that need it.

mrossinek:
> In the reduction that was done before freeze and the orbital specification were logically or'd together. In other words the freeze core was simple way without explicitly stating what the orbitals were. Since the freezing and reduction were all about removing it maybe easy to think of the logical or. But I think we can only apply one of these transformations right - we cannot have one with freeze core and another to change the remainder of the active space - or at least from the energies being saved they would be overwritten hence it appears not. But it seems fine to me to be able to say freeze core plus have some added definition of the other active space.

Yes, it is not possible to combine `freeze_core` and `active_space`. But the reasoning here is that it really is not necessary because the specification of the `active_space` is so easy.

> I agree that the prior active space was not very flexible - but in general to that point we were dealing only with small systems and it sufficed. It did have the ability to support both positive or negative indexes for convenience (where negative -1 is the highest entry, -2 next one etc. Maybe that is still the case here. Perhaps we should consider the specification for the orbitals to be a List of ether int or slice. That way a contiguous range is easier to define with a slice so each orbital does not need to be individually listed.

I am not sure whether this adds a lot of benefit. Especially because in python you could just use `range(n, m)` and give that list as the active orbital indices.

> As to the limitation, is it much code to infer what is needed to get it to work with FCIDumpDriver? Even so we could go now with say raising an exception and later adding this code.

It would require some more case handling and is not very nice to integrate in that sense. But in terms of code it is just the addition of one or a few functions. In total likely less than 100 lines of code.
I tend to leave it out for now because (given the time frame until the release) it is not really required.


> And in regards of #64 it would be good to be more precise here. Also it might be nice for an example here showing orbitals before and after in some way.

Yes, I will address that today :+1: 

> Final thought. Since UCCSD had active space definition it would be nice if these were compatible so a user configures things the same way. Of course by the time UCCSD sees the set of orbitals the indexes it knows about would not be the same as the molecule started with so the user would have to account for that I think and give its active space in terms or what remained. Or is the plan to drop it from UCCSD? Having unoccupied orbital qubits that don't do anything is inefficient so maybe UCCSD should always act on the full space of what remains and drop active space from UCCSD. If its s more than one place though, maybe like excitation logic, there should be some active space logic that ends up in common place for any modules that need it.

The new `UCC` implementation will not have the same concept of active space as it used to have. Either, you can use this transform directly, or you can provide a custom excitation generator which limits the excitations to a subset of orbitals (although the transformer use is much easier and achieves the same).

woodsp-ibm:
>Yes, it is not possible to combine freeze_core and active_space. But the reasoning here is that it really is not necessary because the specification of the active_space is so easy.

So if I want to freeze core, which is very common, and then additionally restrict the active space, I have to spell it all out in full. Before I could say freeze core and do not include these 2 orbitals (say -3,-2) for LiH the two unoccupied orbitals below the upper one (that is where the negative was convenient though of course the orbitals can be stated from 0 and made postive. Its just convience.

Will I be able to give two active space transformers, one to freeze core and then another to act on the remainder. Will I have to worry about order and the indexes etc.

> I am not sure whether this adds a lot of benefit. Especially because in python you could just use range(n, m) and give that list as the active orbital indices.

I should look more closely at the spec. I was assuming I would state orbitals that are occupied that I want in active space as well as those that are unoccupied. I guess this would be two ranges (assuming they are continuous blocks). Since doing slice(n, m) would have the same notion the list of orbitals, assuming it was 2 'ranges' would be [slice(n, m), slice(p, q)] I guess it depends on what sort of active spaces you foresee in terms of how useful that might be in being able to specify them.

> or you can provide a custom excitation generator which limits the excitations to a subset of orbitals (although the transformer use is much easier and achieves the same).
Well while it might achieve the same result since the operator contains the qubits for those orbitals its not as efficient when it comes to running it - or at least this is how things were. It was better to prune the orbitals out early if there would be no excitations to them rather than leaving them in. But having the active space in UCCSD does give flexibility.

mrossinek:
> So if I want to freeze core, which is very common, and then additionally restrict the active space, I have to spell it all out in full. Before I could say freeze core and do not include these 2 orbitals (say -3,-2) for LiH the two unoccupied orbitals below the upper one (that is where the negative was convenient though of course the orbitals can be stated from 0 and made positive. Its just convenience.

Yes, currently you would have to spell out the full active space in this scenario. But I am starting too see the benefit of being able to specify a cutoff for the unoccupied orbitals. Would a simply index which specifies the maximum index to consider suffice here?

> Will I be able to give two active space transformers, one to freeze core and then another to act on the remainder. Will I have to worry about order and the indexes etc.

I have not tested that but I do not see a problem with applying the transformer twice (although we would have to handle the `energy_shift` differently then, using `+=`). However, the user would have to account for the index shift itself.

> I should look more closely at the spec. I was assuming I would state orbitals that are occupied that I want in active space as well as those that are unoccupied. I guess this would be two ranges (assuming they are continuous blocks). Since doing slice(n, m) would have the same notion the list of orbitals, assuming it was 2 'ranges' would be [slice(n, m), slice(p, q)] I guess it depends on what sort of active spaces you foresee in terms of how useful that might be in being able to specify them.

It would be a single range if your occupied and unoccupied orbitals are in a contiguous block.

> Well while it might achieve the same result since the operator contains the qubits for those orbitals its not as efficient when it comes to running it - or at least this is how things were. It was better to prune the orbitals out early if there would be no excitations to them rather than leaving them in. But having the active space in UCCSD does give flexibility.

I see your point. In the UCCSD PR I was wondering whether we should passing a `filter` callable to the excitation generator such that this scenario becomes easier. How common is this use case though?

woodsp-ibm:
> > But having the active space in UCCSD does give flexibility.

> How common is this use case though?

I do not really know. The active space was added to UCCSD after we had the original orbital reduction; I do not recall any rationale/use case in regards of doing it there versus earlier in the stack.

mrossinek:
Alright, despite our recent conversations I decided to add a `remove_orbitals` argument now. It should be rather self-explanatory.

@pbark @woodsp-ibm This PR is now fully ready for review :slightly_smiling_face:

qiskit_nature/transformers/active_space_transformer.py
```diff
@@ -56,31 +56,66 @@ class ActiveSpaceTransformer(BaseTransformer):
...
+            num_alpha: The optional number of active alpha-spin electrons.
             active_orbitals: A list of indices specifying the molecular orbitals of the active
                              space. This argument must match with the remaining arguments and should
                              only be used to enforce an active space that is not chosen purely
                              around the Fermi level.
+            freeze_core: A convenience argument to quickly enable the inactivity of the
+                         `QMolecule.core_orbitals`. This keyword overwrites the use of all other
+                         keywords (except `remove_orbitals`) and, thus, cannot be used in
+                         combination with them.
+            remove_orbitals: A list of indices specifying molecular orbitals which are removed in
```
woodsp-ibm, line 99: More and more these feel like independent transformers. A subclass of this just having freeze_core and remove_orbitals that computes the active space, and internally delegates the actual transformation to an instance of this it creates, when it needs to do the transform, does that not work and make things simpler for an end user? I know having a separate class was discussed before but it was avoided due to complexity. Maybe I am missing something...

qiskit_nature/transformers/active_space_transformer.py
```diff
@@ -56,31 +56,66 @@ class ActiveSpaceTransformer(BaseTransformer):
...
     References:
         - *M. Rossmannek, P. Barkoutsos, P. Ollitrault, and I. Tavernelli, arXiv:2009.01872
           (2020).*
     """
 
-    def __init__(self, num_electrons: int, num_orbitals: int, num_alpha: Optional[int] = None,
-                 active_orbitals: Optional[List[int]] = None):
+    def __init__(self,
+                 num_electrons: Optional[int] = None,
+                 num_molecular_orbitals: Optional[int] = None,
```
woodsp-ibm, line 69: So if say I have a triplet where the active space comprises 6 electrons. I say num_electrons is 6 and num_alpha is 2 as 2 of these are unpaired with a beta? So two fully occupied and two half occupied orbitals.

qiskit_nature/transformers/active_space_transformer.py
```diff
@@ -94,112 +129,152 @@ def transform(self, q_molecule: QMolecule) -> QMolecule:
...
+        # compute number of inactive electrons
         nelec_inactive = nelec_total - self.num_electrons
         if self.num_alpha is not None:
-            if not beta:
+            if not self._beta:
                 warning = 'The provided instance of QMolecule does not provide any beta ' \
-                          + 'coefficients but you tried to specify a separate number of alpha ' \
-                          + 'electrons. Continuing as if it does not matter.'
+                          + 'coefficients but you tried to specify a separate number of alpha' \
+                          + ' electrons. Continuing as if it does not matter.'
```
woodsp-ibm, line 253: The reason for the comment above was that I was trying to understand this warning. Basically you are saying that if you gave alpha we were expecting an open shell molecule, but a closed shell computation has been done since there are not separate alpha and beta coeffs since they are the same (and hence no beta). Rather than saying `Continuing as if it does not matter` can we say something more like. 'The system has no orbitals that are single occupied by alpha electrons, the alpha electrons value supplied {self.num_alpha} will be be ignored`

qiskit_nature/transformers/active_space_transformer.py
```diff
@@ -56,31 +56,66 @@ class ActiveSpaceTransformer(BaseTransformer):
...
+                             _unoccupied_ orbitals, which can be removed without taking any energy
+                             shifts into account.
         """
         self.num_electrons = num_electrons
-        self.num_orbitals = num_orbitals
+        self.num_molecular_orbitals = num_molecular_orbitals
         self.num_alpha = num_alpha
         self.active_orbitals = active_orbitals
+        self.freeze_core = freeze_core
+        self.remove_orbitals = remove_orbitals
```
woodsp-ibm, line 110: Do we want these on public properties - private is fine. If they need to be made public then normally we have public getter/setter properties with docstrings.

qiskit_nature/transformers/active_space_transformer.py
mrossinek: Yes I was starting to think this as well... I will have a look tomorrow :+1:

qiskit_nature/transformers/active_space_transformer.py
mrossinek: No it's a little bit different than that. But I am wondering whether it would be easier (and also more consistent) if I just switched this to also take `Tuple[int, int]` for `num_electrons`. This is what everyone is used to in Qiskit anyways.
For simplicity, I can also make it a `Union[int, Tuple[int, int]]` which will assume singlet-spin.
I will also give this a thought tomorrow :+1:

qiskit_nature/transformers/freeze_core_transformer.py
```diff
@@ -0,0 +1,110 @@
...
+        Returns:
+            A new `QMolecule` instance.
+
+        Raises:
+            QiskitNatureError: If more electrons or orbitals are requested than are available, if an
+                               uneven number of inactive electrons remains, or if the number of
+                               selected active orbital indices does not match
+                               `num_molecular_orbitals`.
+        """
+        q_molecule_new = super().transform(q_molecule)
```
woodsp-ibm, line 73: Ah ok, I'll look more over these later. I had thought perhaps that you would create an ActiveSpaceTransformer at this point, given that the QMolecule is now known, and just figure the correct active space configuration and just call its transform. This would then just extend the base Transformer class and delegate to ActiveSpaceTransformer which would just be created/used internally. Maybe the interface did not permit the removal in the way that this does hence the sub-classing. Though the dictionary keys then for the energy extracted might not be as expected unless this was documented as a convenience that uses ActiveSpaceTransformer to simply set a frozen core config

mrossinek: That would be an option to but I found this more intuitive. Since this does not change anything user-facing we can change this later if we want to.

woodsp-ibm (approved):

PR #79, opened 2021-03-10
Vibrational problem

dlasecki:
### Summary
Vibrational problem implementation


qiskit_nature/problems/second_quantization/base_problem.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+
+        Args:
+            driver: A driver encoding the molecule information.
+            transformers: A list of transformations to be applied to the molecule.
+        """
+        if transformers is None:
+            transformers = []
+        self.driver = driver
+        self.transformers = transformers
```
mrossinek, line 45: We can also do the following but I am not opinionated about this:
```suggestion
        self.driver = driver
        self.transformers = transformers or []
```

qiskit_nature/problems/second_quantization/base_problem.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @abstractmethod
+    def second_q_ops(self):
+        """Returns a list of `SecondQuantizedOp` created based on a driver and transformations
+        provided.
+
+        Returns:
+            A list of `SecondQuantizedOp` in the following order: ... .
+        """
+        return
```
mrossinek, line 55:
```suggestion
        raise NotImplementedError
```

qiskit_nature/problems/second_quantization/base_problem.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @abstractmethod
+    def _transform(self, molecule):
+        return
```
mrossinek, line 59:
```suggestion
        raise NotImplementedError
```

qiskit_nature/problems/second_quantization/vibrational/spin_op_builder.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    Builds a spin operator based on a WatsonHamiltonian object.
+
+    Args:
+        watson_hamiltonian (WatsonHamiltonian): WatsonHamiltonian instance.
+        basis_size: size of a basis
+        truncation_order: order at which an n-body expansion is truncated
+
+    Returns:
+        SpinOp: SpinOp built from a QMolecule object.
```
mrossinek, line 29:
```suggestion
        SpinOp: SpinOp built from a `WatsonHamiltonian`.
```

qiskit_nature/problems/second_quantization/vibrational/spin_op_builder.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+
+    num_modes = watson_hamiltonian.num_modes
+    basis_size = [basis_size] * num_modes
+    boson_hamilt_harm_basis = HarmonicBasis(watson_hamiltonian,  # type: ignore
```
mrossinek, line 34: I think we should make `HarmonicBasis` configurable and instead take any `BosonicBasis` as an input somewhere.

qiskit_nature/problems/second_quantization/vibrational/vibrational_problem.py
```diff
@@ -0,0 +1,56 @@
...
+        """Returns a list of `SecondQuantizedOp` created based on a driver and transformations
+        provided.
+
+        Returns:
+            A list of `SecondQuantizedOp` in the following order: ... .
+        """
+        watson_hamiltonian = self.driver.run()
+        watson_hamiltonian_transformed = self._transform(watson_hamiltonian)
+        basis_size = 1  # TODO how to get it?
+        truncation_order = 3  # TODO how to get it?
```
mrossinek, line 46: Afaict from the old way of using `BosonicTransformation` (https://github.com/Qiskit/qiskit-aqua/blob/master/test/chemistry/test_bosonic_esc_calculation.py#L60-L64) this will need to be an input to the constructor. @paulineollitrault Please correct me, if these can be inferred in some other way.

qiskit_nature/algorithms/ground_state_solvers/orbital_optimization_vqe.py
```diff
@@ -208,8 +208,8 @@ def _initialize_additional_parameters(self, driver: BaseDriver):
             self._set_bounds(self._orbital_rotation.parameter_bound_value)
         if self._iterative_oo_iterations < 1:
             raise QiskitNatureError(
-                            'Please set iterative_oo_iterations parameter to a positive number,'
-                            ' got {} instead'.format(self._iterative_oo_iterations))
+                'Please set iterative_oo_iterations parameter to a positive number,'
```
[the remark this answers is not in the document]
dlasecki: Lint was complaining a lot.

qiskit_nature/problems/second_quantization/base_problem.py
dlasecki: Your way is much better. Thanks.

qiskit_nature/problems/second_quantization/vibrational/spin_op_builder.py
dlasecki: Sure. Let's deal with it a bit later. Currently we only have a `HarmonicBasis` implemented.

qiskit_nature/problems/second_quantization/vibrational/vibrational_problem.py
dlasecki: I have been inclined towards putting it to the `VibrationalProblem` constructor. Another option would be arguments of `second_q_ops` but it conflicts with a signature of that method in a parent class.

qiskit_nature/algorithms/ground_state_solvers/orbital_optimization_vqe.py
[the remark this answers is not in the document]
mrossinek: Even in the CI?

qiskit_nature/problems/second_quantization/vibrational/vibrational_problem.py
mrossinek: No `second_q_ops` should not have any arguments. Once you call that, everything should really be set.
The idea here is that it should feel like a property that you query

dlasecki:
Please take a look at the current code. Updated SpinOp and relevant tests were taken from the PR #68.
 
Tests are passing but there are still some mypy errors:

```
qiskit_nature/operators/second_quantization/spin_op.py:219: error: Incompatible types in assignment (expression has type "List[Tuple[str, int]]", variable has type "Union[str, List[Tuple[str, complex]], List[Tuple[str, float]], Tuple[Any, Any]]")
qiskit_nature/problems/second_quantization/vibrational/vibrational_problem.py:46: error: "BaseDriver" has no attribute "run"
qiskit_nature/problems/second_quantization/molecular/molecular_problem.py:50: error: "BaseDriver" has no attribute "run"
```

Last 2 errors come from the fact of me introducing the class `BaseProblem` that operates on `BaseDriver` that do not have a run method. `BaseDriver` problem has been discussed with @mrossinek and we decided to postpone it. Perhaps we should also postpone introducing `BaseProblem`, what do you think?

I am not sure about the first mypy error. @ikkoham ? I introduced an additional input type that uses float coefficients instead of `complex` coefficients because `VibrationalSpinOp` operates on `float` coefficient only as of now (it starts from the `HarmonicBasis` class). Somehow `float `is not considered a subtype of `complex` in type hints...

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -166,27 +170,23 @@ class SpinOp(SecondQuantizedOp):
...
-            str,
-            List[Tuple[str, complex]],
-            Tuple[np.ndarray, np.ndarray],
-        ],
-        spin: Union[float, Fraction] = Fraction(1, 2),
+            self,
+            data: Union[
+                str,
+                List[Tuple[str, complex]],
+                List[Tuple[str, float]],
```
ikkoham, line 178: This line is unnecessary because float is a virtual subclass of complex.

qiskit_nature/problems/second_quantization/vibrational/vibrational_labels_validator.py
```diff
@@ -0,0 +1,120 @@
...
+            if _is_order_incorrect(mode_index, last_mode_index, modal_index, last_modal_index, op,
+                                   last_op):
+                raise ValueError(
+                    f"Incorrect order of operators for label {label} and previous label "
+                    f"{str(last_op) + '_' + str(last_mode_index) + '*' + str(last_modal_index)}.")
+
+            last_op, last_mode_index, last_modal_index = op, mode_index, modal_index
+
+            increment = 1 if op == "+" else -1
+            check_list[int(mode_index)] += increment
```
mrossinek, line 84:
```suggestion
            check_list[int(mode_index)] += 1 if op == "+" else -1
```

qiskit_nature/problems/second_quantization/vibrational/vibrational_labels_validator.py
```diff
@@ -0,0 +1,120 @@
...
+                      not valid_vibr_label_pattern.match(label)]
+    if invalid_labels:
+        raise ValueError(f"Invalid labels: {invalid_labels}")
+
+
+def _validate_indices(vibrational_labels: List[Tuple[str, float]], num_modes: int,
+                      num_modals: List[int]):
+    for labels, _ in vibrational_labels:
+        coeff_labels_split = labels.split(" ")
+        check_list = [0] * num_modes
```
mrossinek, line 64: Either a better name or a comment explaining what this variable is used for would be great :+1:

qiskit_nature/problems/second_quantization/vibrational/vibrational_labels_validator.py
```diff
@@ -0,0 +1,120 @@
...
+                    f"Incorrect order of operators for label {label} and previous label "
+                    f"{str(last_op) + '_' + str(last_mode_index) + '*' + str(last_modal_index)}.")
+
+            last_op, last_mode_index, last_modal_index = op, mode_index, modal_index
+
+            increment = 1 if op == "+" else -1
+            check_list[int(mode_index)] += increment
+        if not all(v == 0 for v in check_list):
+            raise ValueError(
+                f"Modes of raising and lowering operators do not agree for labels {labels}.")
```
mrossinek, line 87: You could even use the `check_list` to only print the violating labels, right?

qiskit_nature/problems/second_quantization/vibrational/vibrational_labels_validator.py
```diff
@@ -0,0 +1,120 @@
...
+                         last_modal_index: int, op: str, last_op: str) -> bool:
+    return modal_index == last_modal_index and mode_index == last_mode_index and op == last_op
+
+
+def _is_order_incorrect(mode_index: int, last_mode_index: int, modal_index: int,
+                        last_modal_index: int, op: str, last_op: str) -> bool:
+    return _is_mode_order_incorrect(mode_index, last_mode_index) or \
+           _is_modal_order_incorrect(last_mode_index, mode_index, last_modal_index, modal_index) \
+           or _is_operator_order_incorrect(mode_index, last_mode_index, modal_index,
+                                           last_modal_index, op, last_op)
```
mrossinek, line 105: I think you can inline those methods with a comment explaining each rather than splitting out into extra functions. But here I don't mind.

qiskit_nature/problems/second_quantization/vibrational/vibrational_spin_op_builder.py
```diff
@@ -0,0 +1,47 @@
...
+        truncation_order: order at which an n-body expansion is truncated
+
+    Returns:
+        SpinOp: SpinOp built from a WatsonHamiltonian object.
+    """
+
+    num_modes = watson_hamiltonian.num_modes
+    if isinstance(basis_size, int):
+        basis_size = [basis_size] * num_modes
+    # TODO make HarmonicBasis an argument and support other bases when implemented
```
mrossinek, line 41: Do you want to address this before merging or in the future?

qiskit_nature/problems/second_quantization/vibrational/vibrational_spin_op_builder.py
```diff
@@ -0,0 +1,47 @@
...
+    boson_hamilt_harm_basis = HarmonicBasis(watson_hamiltonian,
+                                            # type: ignore
```
mrossinek, line 43: Why is this necessary?

qiskit_nature/transformers/base_transformer.py
```diff
@@ -18,14 +18,15 @@
...
 class BaseTransformer(ABC):
-    """The interface for implementing methods which map from one `QMolecule` to another. These
-    methods may or may not affect the size of the Hilbert space.
+    """The interface for implementing methods which map from one `QMolecule` or
+    'WatsonHamiltonian' to another. These methods may or may not affect the size of the Hilbert
+    space.
     """
 
     @abstractmethod
     def transform(self, q_molecule: QMolecule) -> QMolecule:
```
mrossinek, line 27: Since you want to add the `WatsonHamiltonian` to this interface, it should also go in the type hint here. Although that will likely lead to `mypy` issues, too..

test/problems/second_quantization/vibrational/test_vibr_to_spin_op_label_converter.py
```diff
@@ -0,0 +1,73 @@
...
+# TODO add more tests
+class TestVibrToSpinOpLabelConverter(QiskitNatureTestCase):
+    """Tests Vibrational Spin Op to Spin Op Label Converter."""
+
+    def test_convert_to_spin_op_labels(self):
+        """Tests that VibrationalSpinOp labels are converted to SpinOp labels correctly."""
+        expected_spin_op_labels = [('-_1 +_0', 1215.375), ('-_5 +_4 -_1 +_0', -6.385)]
+        vibrational_labels = [('-_0*1 +_0*0', 1215.375), ('-_2*1 +_2*0 -_0*1 +_0*0', -6.385)]
+        num_modes = 3
+        num_modals = 2
```
mrossinek, line 27: I think you can use `@data` here to reduce some of the code duplication

test/problems/second_quantization/vibrational/test_vibr_to_spin_op_label_converter.py
```diff
@@ -0,0 +1,73 @@
...
+    convert_to_spin_op_labels
+
+
+# TODO add more tests
+class TestVibrToSpinOpLabelConverter(QiskitNatureTestCase):
+    """Tests Vibrational Spin Op to Spin Op Label Converter."""
+
+    def test_convert_to_spin_op_labels(self):
+        """Tests that VibrationalSpinOp labels are converted to SpinOp labels correctly."""
+        expected_spin_op_labels = [('-_1 +_0', 1215.375), ('-_5 +_4 -_1 +_0', -6.385)]
```
mrossinek, line 24: Where do these random numbers come from?

test/problems/second_quantization/vibrational/test_vibrational_label_builder.py
```diff
@@ -0,0 +1,163 @@
...
+
+class TestVibrationalLabelBuilder(QiskitNatureTestCase):
+    """Tests Vibrational Label Builder."""
+
+    def test_create_labels(self):
+        """Tests that correct labels are built."""
+        logfile = self.get_resource_path('CO2_freq_B3LYP_ccpVDZ.log')
+        driver = GaussianForcesDriver(logfile=logfile)
+
+        # TODO extract somewhere
```
mrossinek, line 28: Please do :wink:

mrossinek (changes requested):
Nice work @dlasecki! :+1: 
As usual some nit-picking comments from my side :wink:

qiskit_nature/operators/second_quantization/vibrational_spin_op.py
```diff
@@ -0,0 +1,96 @@
...
+[1]: https://en.wikipedia.org/wiki/Spin_(physics)#Higher_spins
+"""
+from fractions import Fraction
+from typing import List, Tuple, Union
+
+from .. import SpinOp
+from ...problems.second_quantization.vibrational.vibr_to_spin_op_label_converter import \
+    convert_to_spin_op_labels
+from ...problems.second_quantization.vibrational.vibrational_labels_validator import \
+    validate_vibrational_labels
```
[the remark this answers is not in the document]
dlasecki: We could also move these classes to where the operator sits. I think I did not have a good reason for keeping them in the problem module.

qiskit_nature/problems/second_quantization/vibrational/vibr_to_spin_op_label_converter.py
```diff
@@ -0,0 +1,86 @@
...
+        for mode_len in num_modals:
+            summed += mode_len
+            partial_sum_modals.append(summed)
+        return partial_sum_modals
+    else:
+        raise ValueError(f"num_modals of incorrect type {type(num_modals)}.")
+
+
+def _build_coeff_spin_op_labels(labels: str, partial_sum_modals: List[int]) -> str:
+    coeff_labels_split = labels.split(" ")
```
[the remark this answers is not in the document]
dlasecki: Yep

qiskit_nature/problems/second_quantization/vibrational/vibrational_labels_validator.py
```diff
@@ -0,0 +1,120 @@
...
+        Args:
+            vibrational_labels: list of vibrational labels with coefficients.
+            num_modes: the number of modes.
+            num_modals: the number of modals.
+
+        Raises:
+            ValueError: if invalid vibrational labels provided.
+        """
+    if isinstance(num_modals, int):
+        num_modals = [num_modals] * num_modes
```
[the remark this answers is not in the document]
dlasecki: I assumed there might be several entry points.

qiskit_nature/problems/second_quantization/vibrational/vibrational_labels_validator.py
```diff
@@ -0,0 +1,120 @@
...
+    if invalid_labels:
+        raise ValueError(f"Invalid labels: {invalid_labels}")
+
+
+def _validate_indices(vibrational_labels: List[Tuple[str, float]], num_modes: int,
+                      num_modals: List[int]):
+    for labels, _ in vibrational_labels:
+        coeff_labels_split = labels.split(" ")
+        check_list = [0] * num_modes
+        last_op, last_mode_index, last_modal_index = "+", num_modes, max(num_modals)
```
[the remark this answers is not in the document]
dlasecki: Makes sense.

qiskit_nature/problems/second_quantization/vibrational/vibrational_spin_op_builder.py
dlasecki: Do you mean the ignore? Not sure, it is a copy-paste from a deprecated class.

qiskit_nature/transformers/base_transformer.py
dlasecki: That was my initial try but mypy issues, exactly.

qiskit_nature/problems/second_quantization/vibrational/vibrational_labels_validator.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _is_index_out_of_range(mode_index: int, num_modes: int, modal_index: int,
+                           num_modals: List[int]) -> bool:
+    return mode_index >= num_modes or modal_index >= num_modals[int(mode_index)]
+
+
+def _is_label_duplicated(mode_index: int, last_mode_index: int, modal_index: int,
+                         last_modal_index: int, op: str, last_op: str) -> bool:
+    return modal_index == last_modal_index and mode_index == last_mode_index and op == last_op
```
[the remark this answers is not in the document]
dlasecki: I think it is sufficient.
Case 1: several duplicated labels next to each other -> we get an error until we deduplicate to a single one,
Case 2: several duplicated labels not next to each other -> we get an ordering error.

qiskit_nature/problems/second_quantization/vibrational/vibrational_spin_op_builder.py
mrossinek: Yes. This comment disables `mypy` for the variable in question but I cannot see which variable it is referring to here. Please remove it and see what breaks. Then you can add it back in (if necessary) while making sure it is clear which variable it tells `mypy` to ignore.

qiskit_nature/problems/second_quantization/vibrational/vibr_to_spin_op_label_converter.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    if isinstance(num_modals, int):
+        num_modals = [num_modals] * num_modes
+
+    if len(num_modals) != num_modes:
+        raise ValueError("num_modes does not agree with the size of num_modals")
+
+    validate_vibrational_labels(vibrational_labels, num_modes, num_modals)
+
+    partial_sum_modals = _calc_partial_sum_modals(num_modals)
```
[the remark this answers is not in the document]
dlasecki: Good suggestion, thanks,

qiskit_nature/transformers/base_transformer.py
mrossinek: You could add a `# type: ignore` comment in these cases just to get `mypy` to shut up for now, until we refactor the drivers and `QMolecule`/`WatsonHamiltonian` classes.

qiskit_nature/problems/second_quantization/vibrational/vibrational_labels_validator.py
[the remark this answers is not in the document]
mrossinek: Please add a unittest which covers this case though :+1:

qiskit_nature/problems/second_quantization/vibrational/vibrational_spin_op_builder.py
dlasecki: In the future. Would be better to see how the non-existing alternative to HarmonicBasis looks like and works.

qiskit_nature/problems/second_quantization/vibrational/vibrational_spin_op_builder.py
dlasecki: No complaints from mypy or lint when I got rid of it. Seems unnecessary.

qiskit_nature/problems/second_quantization/vibrational/vibr_to_spin_op_label_converter.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def _build_spin_op_label(label: str, partial_sum_modals: List[int]) -> str:
+    op, mode_index, modal_index = re.split('[*_]', label)
+    index = _get_ind_from_mode_modal(partial_sum_modals, int(mode_index), int(modal_index))
+    new_label = "".join([op, "_", str(index)])
+    return new_label
+
+
+def _get_ind_from_mode_modal(partial_sum_modals: List[int], mode_index: int,
```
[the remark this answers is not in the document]
dlasecki: I think it improves readability.

test/problems/second_quantization/vibrational/test_vibr_to_spin_op_label_converter.py
dlasecki: I copied them from an output for a particle I loaded. Not sure how to explain it here... They don't play any role in these tests, could be anything.

qiskit_nature/problems/second_quantization/vibrational/vibrational_label_builder.py
```diff
@@ -0,0 +1,57 @@
...
+
+    Args:
+        boson_hamilt_harm_basis: A data structure returned by a convert() method in `HarmonicBasis`.
+        truncation_order: order at which an n-body expansion is truncated.
+    Returns:
+        A list of labels and corresponding coefficients that describe a vibrational problem.
+    """
+    all_labels = []
+    for num_body in range(truncation_order):
+        num_body_data = boson_hamilt_harm_basis[num_body]
```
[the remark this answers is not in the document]
dlasecki: I think you are right.

qiskit_nature/operators/second_quantization/vibrational_spin_op.py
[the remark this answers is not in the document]
dlasecki: I moved things around a little.

qiskit_nature/drivers/base_driver.py
```diff
@@ -50,6 +50,13 @@ def __init__(self,
         self._hf_method = hf_method
         self._supports_molecule = supports_molecule
 
+    @abstractmethod
```
woodsp-ibm, line 53: I imagine the reason we never put a run() method here before is that the type returned from the two subclasses is different and hence we left the run methods to be specific for the fermionic and bosonic drivers but did not try to have some common parent. If this is needed for some reason then maybe it should return 'Any' ?

qiskit_nature/operators/second_quantization/vibrational_spin_op.py
```diff
@@ -0,0 +1,102 @@
...
+# (C) Copyright IBM 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+"""A Vibrational Spin operator.
```
woodsp-ibm, line 12: I am not sure that this Note will appear in the Sphinx. I know the class docstring an init docstring get merged. Perhaps build the html and check. Maybe its safer to have this note in the class docstring below (one starting at line 30)

qiskit_nature/operators/second_quantization/vibrational_spin_op_utils/vibr_to_spin_op_label_converter.py
```diff
@@ -0,0 +1,77 @@
...
+import operator
+import re
+from typing import List, Tuple, Union
+
+from qiskit_nature.operators.second_quantization.vibrational_spin_op_utils\
+    .vibrational_labels_validator import \
+    validate_vibrational_labels
+
+
+def convert_to_spin_op_labels(vibrational_labels: List[Tuple[str, complex]], num_modes: int,
```
woodsp-ibm, line 25: This is some separate module, and public supported API, outside of the VibrationalSpinOp class as its considered useful in its own right. I could see this as function in the VibrationalSpinOp - just wondering why we need this and the others in this folder that seem more like private helper functions of that class

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
```diff
@@ -22,9 +22,10 @@
...
 from .integrals_calculators import calc_total_magnetization_ints
 from .integrals_calculators import calc_total_particle_num_ints
+from ..base_problem import BaseProblem
 
 
-class MolecularProblem:
+class MolecularProblem(BaseProblem):
     """Molecular Problem"""
 
     def __init__(self, fermionic_driver: FermionicDriver,
```
woodsp-ibm, line 31: In the same way there is FermionicDriver in here I wonder whether we ought to have FermionicTransformer and likewise BosonicTransformer (here it would be a List of Fermionic ones of course)

qiskit_nature/transformers/base_transformer.py
```diff
@@ -13,24 +13,27 @@
...
+    """The interface for implementing methods which map from one `QMolecule` or
+    'WatsonHamiltonian' to another. These methods may or may not affect the size of the Hilbert
+    space.
     """
 
     @abstractmethod
-    def transform(self, q_molecule: QMolecule) -> QMolecule:
-        """Transforms one `QMolecule` into another one. This may or may not affect the size of the
-        Hilbert space.
+    def transform(self, molecule_data: Union[QMolecule, WatsonHamiltonian]):
```
woodsp-ibm, line 29: Will we ever have transformers that would work on either type. Or more like my comment above would it be better to define A FermionicTransform taking QMolecule and a BosonicTransform taking the WatsonHamiltonian

qiskit_nature/operators/second_quantization/vibrational_spin_op_utils/vibr_to_spin_op_label_converter.py
woodsp-ibm: Update, I see there was a conversation above about this - sorry I have been trying to catch up with things and did not see that first!

qiskit_nature/drivers/base_driver.py
dlasecki: Yes, this is indeed the case. I introduced a BaseProblem class and it makes sense that it operates on a BaseDriver (it happens that it also runs a driver). I could try 'Any'. We also discussed with Max that these classes will likely be refactored later somehow and we will try to come up with a better solution.

qiskit_nature/transformers/base_transformer.py
dlasecki: Not sure about it... @mrossinek ? If not, I agree with you.

qiskit_nature/transformers/freeze_core_transformer.py
```diff
@@ -12,15 +12,14 @@
...
-
 from .base_transformer import BaseTransformer
 from ..drivers import QMolecule
 
 
 class FreezeCoreTransformer(BaseTransformer):
     """The Freeze-Core transformer."""
 
-    def transform(self, q_molecule: QMolecule) -> QMolecule:
+    def transform(self, q_molecule: QMolecule) -> QMolecule:  # type: ignore
```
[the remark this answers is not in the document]
dlasecki: Yes

qiskit_nature/operators/second_quantization/vibrational_spin_op_utils/vibr_to_spin_op_label_converter.py
dlasecki: I can see your point. My main motivation was to extract logic with a certain responsibility from the main class. As you can see these modules end up being pretty lengthy. Then, I have been wondering if they could be useful to a user in their own right in case they do some steps on their own - but I am not convinced about it. I would be fine with making all methods in these modules "private" such that they are not exposed as a public API. Let me know if there is a better way to accomplish it and avoid importing a private method to VibrationalSpinOp.

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
mrossinek: Why is this field even called `fermionic_driver` rather than just `driver` as in `BaseProblem`? Sub-typing it should be fine

qiskit_nature/transformers/base_transformer.py
mrossinek: I don't think we will ever have transformers which support both types. Maybe you can also just specify `Any` as the type here and then subclass to a more specific type in those subclasses.

I would even argue that, since we do not even have any transformer for `WatsonHamiltonian` we simply leave that out for this release..

qiskit_nature/operators/second_quantization/vibrational_spin_op.py
```diff
@@ -0,0 +1,96 @@
...
+          - Raising operator
+        * - `-`
+          - :math:`S_-`
+          - Lowering operator
+    1. Labels
+    :class:`VibrationalSpinOp` accepts the notation that encodes raising (+) and lowering (-)
+    operators together with indices of modes and modals that they act on, e.g. "+_{mode_index}*{
+    modal_index}". Each modal can be excited at most once.
+    **Initialization**
+    # TODO
```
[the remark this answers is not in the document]
dlasecki: Me neither, do you have anything specific in mind? It is mostly based on the SpinOp docstring.

qiskit_nature/operators/second_quantization/vibrational_spin_op_utils/vibr_to_spin_op_label_converter.py
woodsp-ibm: I guess personally I am less worried about the resultant total length of VibrationalSpinOp if these are added at the end of it. The methods themselves are reasonable and so reading them in the overall file should be fine. It would only end up being 2 ro 3 hundred lines long right - puny compared to say the 2.6K that is QuantumCircuit!

qiskit_nature/problems/second_quantization/vibrational/vibrational_labels_validator.py
```diff
@@ -0,0 +1,120 @@
...
+    if isinstance(num_modals, int):
+        num_modals = [num_modals] * num_modes
+
+    _validate_data_type(vibrational_labels)
+    _validate_regex(vibrational_labels)
+    _validate_indices(vibrational_labels, num_modes, num_modals)
+
+
+def _validate_data_type(vibrational_labels: List[Tuple[str, float]]):
+    if not isinstance(vibrational_labels, list):
```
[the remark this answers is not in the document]
dlasecki: Hmm, probably going deeper would be the role of SpinOp since we here we have the same data structure as there.

qiskit_nature/operators/second_quantization/vibrational_spin_op.py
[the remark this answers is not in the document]
woodsp-ibm: You can always `make html` locally which uses Sphinx to build things out then you can see how things turn out and make any adjustments

pbark (approved):
LGTM.! Thanks @dlasecki 👍

PR #81, opened 2021-03-12
Remove the mixed-operator: SecondQuantizedOp

### LINKED ISSUE #20: Finish the SecondQuantizedOperator implementation
Migrated from Enterprise Github.
Creator: @mrossinek

- [ ] unittest all methods

- ~unify the `SecondQuantizedOp` as much as possible with the other classes~
- ~make the `SecondQuantizedSumOp` implement the `SumOp` interface~

mrossinek:
### Summary

Based on an offline discussion, we have decided to remove the `SecondQuantizedOp` class as it stands now.
For current applications it is mainly a container which everything needs to be placed without adding useful features.
Once we need a mixed-particle kind of operator, we can implement a `MixedOperator` which will be a "composite"-like implementation of the current `ParticleOp` interface.
This will allow seemless usage of the mixed- as well as pure-particle operators within the stack.

### Details and comments

- removes the mixed-operator implementation previously called `SecondQuantizedOp`
- renames the `ParticleOp` interface to `SecondQuantizedOp`

> Note: a future version of the old `SecondQuantizedOp` could be called `MixedOperator`.

Closes #20  (makes it obsolete until we need to consider `MixedOperator` which I would argue will be tracked in a separate issue)


qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -10,84 +10,47 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 
-"""The Second-Quantized Operator."""
+"""The Sum Operator base interface."""
```
pbark, line 13: Why is this changed? Should this be the Second-Quantizer Operator interface?

pbark (approved):

qiskit_nature/operators/second_quantization/second_quantized_op.py
mrossinek: This is a limitation of how git diffes these files. Here, the old `SecondQuantizedOp` is being compared against `ParticleOp` (which is now called `SecondQuantizedOp`).
What actually happened is:
- old `operators/second_quantization/second_q_op.py` got removed
- `operators/second_quantization/particle_op.py` moved to `operators/second_quantization/second_q_op.py`
But git diff shows it as:
- `operators/second_quantization/particle_op.py` got removed
- and diffes the changes on the `second_q_op.py` file

In short: I didn't change it, but git shows it as such because of how the files were renamed.

woodsp-ibm (approved):

PR #85, opened 2021-03-14
Hartree-Fock Refactoring

### LINKED ISSUE #71: Refactor HarteeFock initial state
The interface needs changing since the way mappings are now specified is different. Maybe add some state mapping to the mappers in order to allow the HF state to be mapped accordingly. Maybe like the UCCSD proposal this can then take QubitConvertor since it needs also information related to any symmetry reduction of the main (Hamiltonian) operator of the problem.

mrossinek:
### Summary

Closes #71 

### Details and comments

We can exploit the modularity of the operators and mappers and construct the HF initial state out of a FermioncOp quite simply.
Documentation, unittests and code quality are still lacking. The CI *will* fail as of now.

You can test the code with the following snippet:
```python
from qiskit_nature.circuit.library.initial_states import HartreeFock
from qiskit_nature.mappers.second_quantization import (JordanWignerMapper,
                                                       ParityMapper,
                                                       BravyiKitaevMapper)
from qiskit_nature.operators.second_quantization.qubit_converter import QubitConverter

for mapper in [JordanWignerMapper(), ParityMapper(), BravyiKitaevMapper()]:
    converter = QubitConverter(mappers=mapper)

    hf = HartreeFock(num_orbitals=8, num_particles=(1, 1), qubit_converter=converter)

    print(hf.draw())
```


qiskit_nature/circuit/library/initial_states/hartree_fock.py
```diff
@@ -26,49 +29,47 @@ class HartreeFock(QuantumCircuit):
...
                 self.x(i)
 
 
-def hartree_fock_bitstring(num_orbitals: int,
-                           num_particles: Union[Tuple[int, int], int],
-                           qubit_mapping: str = 'parity',
-                           two_qubit_reduction: bool = True,
-                           sq_list: Optional[List[int]] = None) -> List[bool]:
+def hartree_fock_bitstring(num_spin_orbitals: int,
+                           num_particles: Union[Tuple[int, int], int]) -> List[bool]:
```
woodsp-ibm, line 66: When we first started chemistry, and only did closed shell, we just had the int, but when we added open shell, and needed different number for alpha and beta, where the tuple was added, it was always our intent to deprecate supporting int and require a tuple. I do not mind whether its kept or not, but the stack always emitted num particles as a tuple which you would pass onto these methods, and hence a user would need to have created an int outside of that.

qiskit_nature/circuit/library/initial_states/hartree_fock.py
```diff
@@ -26,49 +29,47 @@ class HartreeFock(QuantumCircuit):
...
             if bit:
                 self.x(i)
 
 
-def hartree_fock_bitstring(num_orbitals: int,
-                           num_particles: Union[Tuple[int, int], int],
-                           qubit_mapping: str = 'parity',
-                           two_qubit_reduction: bool = True,
-                           sq_list: Optional[List[int]] = None) -> List[bool]:
+def hartree_fock_bitstring(num_spin_orbitals: int,
```
woodsp-ibm, line 65: What is the thinking here tapering this bitstring accordingly? The qubit_convertor has the z2_symmetries getter that would pass back a Z2Symmetry object with the sq_list in it. I imagine we will add the two_qubit_reduction to the Qubit_Convertor too, as a flag, so it can do that when the mapper supports it (the mapper has already had the flag added to it). The reason this method was public was so that it could be use by the stack, maybe the algorithms now, as input to determine the desired sector for ground state, and as such, since two qubit reduction  is done ahead of symmetry reduction the sqlist in that case would be empty. Of course by the time HF would be used to create the state then the symettry would be done and the sq_list (from Z2Symmetry) available

qiskit_nature/circuit/library/initial_states/hartree_fock.py
mrossinek: Sure I will update that and deprecate the `int`. I was confused to see it there but forgot to check again later :+1:

qiskit_nature/circuit/library/initial_states/hartree_fock.py
mrossinek: I saw only later that this method was used elsewhere in the stack. I am not sure to what degree this will still be necessary though.
Originally, I had planned to make this method return the fully mapped and tapered `QubitOperator`. It would then also take the `QubitConverter` as an input. Is that what you mean here?
I am not sure from the top of my head, if we can then extract the correct sector from the final result. Is that stored in the `QubitOperator` somehow? Or will we make the `QubitConverter` stateful and store it in there?

woodsp-ibm: The sector with the ground state, that we be chosen, is determined using that HF string. I think there is a pick/select sector. At the time we need to do symmetry reduction the operator will be mapped, and have had any two qubit reduction done. So the string needed to reflect that. I think the sqlist coming from the symmetry is used later in conjunction with UCCSD since by that time its working on the completly mapped/symmeterized operator.

Since QubitConvertor needs to know which sector to select when tapering, I was imagining the algorithm/problem knowing it was the ground state being computed that it would use the pick/select sector, using the HF info, and pass that in when calling the to_qubit_op(). So as mentioned, to taper it, it would just have the mapping and two q reduction flag settings to determine the sector from the mapped/2 q reduced operator.

This is what the code that did auto sector determination looked like before https://github.com/Qiskit/qiskit-aqua/blob/39067b4f9d345ca0fb2eeab41b231e263f828209/qiskit/chemistry/core/hamiltonian.py#L305-L313 The pick sector is further down

Yes, there is TaperedPauliSumOp (as a sub-class of PauliSumOp) that can carry the Z2Symmetry; but I think any operations on it will lose that information, as its a very simple sub class that does not override any methods to preserve that where it made sense - such as simplification, multiplication by constant  etc

mrossinek: After our offline discussion, I have added a TODO comment documenting the currently open design questions.
However, as decided offline, let's postpone the resolution of this TODO until we finalize the `QubitConverter` implementation and know the details of its API and integration into the workflow.

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -84,8 +86,10 @@ def get_solver(self, transformation):
                 qubit_mapping = transformation.qubit_mapping
                 two_qubit_reduction = transformation.molecule_info['two_qubit_reduction']
                 z2_symmetries = transformation.molecule_info['z2_symmetries']
-                initial_state = HartreeFock(num_orbitals, num_particles, qubit_mapping,
-                                            two_qubit_reduction, z2_symmetries.sq_list)
+
+                mapper = ParityMapper()
+                converter = QubitConverter(mappers=mapper)
```
woodsp-ibm, line 91: Just a comment. Given the operators need to be the same type, and we have changed the 2ndQop so that its not a container of different types we may well refactor this to just be a single mapper that is passed. When we have a MixedOperator later that could be the MixedMapper that is a contained too for individual mappers, so we still need one mapper at this level. Anyway by using this as a named param of QubitConvertor, rather than positional, this will need to be refactored to accomodate that. I guess we have a mix of named and not named here, for instance HF is done by position.

qiskit_nature/circuit/library/initial_states/hartree_fock.py
```diff
@@ -12,63 +12,79 @@
...
```
woodsp-ibm, line 80: Maybe we should think about the name here too - I think of a bitstring as a string of 0's and 1's  '010101111' which perhaps this was at some point. Anyway when we figure the QubitCOnvertor TODO maybe we can consider a better name.

test/components/variational_forms/test_uccsd_advanced.py
```diff
@@ -188,11 +188,9 @@ def test_uccsd_hf_qUCCSD(self):
 
         # initial state
         init_state = HartreeFock(
-            num_orbitals=fermionic_transformation.molecule_info['num_orbitals'],
-            qubit_mapping=fermionic_transformation._qubit_mapping,
-            two_qubit_reduction=fermionic_transformation._two_qubit_reduction,
+            num_spin_orbitals=fermionic_transformation.molecule_info['num_orbitals'],
             num_particles=fermionic_transformation.molecule_info['num_particles'],
-            sq_list=fermionic_transformation.molecule_info['z2_symmetries'].sq_list)
+            qubit_converter=self.qubit_converter)
```
woodsp-ibm, line 193: This will need to change since the idea was to test under different conditions. Here it was with tapering - maybe, depending how 'auto' is setup, this will have to change. I do not know whether it would be helpful to have a TODO comment in these or not - clearly they will need fixing up once the stack is altered and hopefully the use of the self.qubit_convertor is self evidently not what would be wanted since we want to change up the configurations that are tested

woodsp-ibm (approved):

qiskit_nature/circuit/library/initial_states/hartree_fock.py
mrossinek: Right, I had kind of ignored that the name no longer fits so nicely.. But I agree, let's do that when we tackle the other TODO :+1:

test/components/variational_forms/test_uccsd_advanced.py
mrossinek: Yes. I mainly had to change the unittests for the linters to be happy. All of the UCCSD-related unittests are disabled as of now. But once we merge this PR, I will replace them with new tests in #66 :+1: Thanks for pointing it out though!

pbark (approved):

PR #86, opened 2021-03-14
[WIP] ParticleHoleTransformer

pbark:
### Summary
ParticleHoleTransformer

### Details and comments
This is the implementation of the particle hole transformer in the qmolecule. This is WIP. 

This PR still misses: 
- [x] calculation of `mo_eri_ints `
`mo_eri_ints` are not properly calculated since the ParticleHoleTransformer operates in the `two_electron_integrals` that we still need to transform back. 
- [x] unittest for second quantized ops
- [ ] fix also dipole integrals 

When we have algorithms ready we need to add a Ground state calculation in the tests


qiskit_nature/transformers/particle_hole_transformer.py
```diff
@@ -12,13 +12,30 @@
...
 
+from .base_transformer import BaseTransformer
+from ..drivers import QMolecule
 
 class ParticleHoleTransformer(BaseTransformer):
     """The Particle/Hole transformer."""
 
+    def __init__(self, num_electrons: int, num_orbitals: int, num_alpha: int):
+        """Initializes a transformer which can reduce a `QMolecule` to a configured active space.
+
```
woodsp-ibm, line 29: Given the transformer is acting directly on a QMolecule, that contains all this information, it should not be necessary to pass this information in since it should all be there in the qmolecule. Indeed expecting the user to figure this when it may be in a series of tranforms that are affecting num electrons etc seems less than desirable.

qiskit_nature/transformers/particle_hole_transformer.py
```diff
@@ -28,5 +45,2102 @@ def transform(self, q_molecule: QMolecule) -> QMolecule:
...
+        self._convert_to_interleaved_spins()
+
+        print(self._h1)
+
+        h1_old_matrix = self._h1
+        h2_old_matrix = self._h2
+
+        if q_molecule.mo_onee_ints_b is not None:
+            raise QiskitNatureError("Particle Hole Transformer does not work with UHF")
```
woodsp-ibm, line 64: The ParticleHole from before was changed to work with open shell right. This test https://github.com/Qiskit/qiskit-aqua/blob/master/test/chemistry/test_particle_hole.py included `ROHF` and well as `UHF` tests. I think that above test would exclude all open shell usage,.

qiskit_nature/transformers/particle_hole_transformer.py
```diff
@@ -28,5 +45,2102 @@ def transform(self, q_molecule: QMolecule) -> QMolecule:
         Returns:
             A new `QMolecule` instance.
         """
-        # TODO
-        raise NotImplementedError()
+
+        n_qubits = self.num_orbitals * 2
+
+        self._h1 = q_molecule.one_body_integrals
+        self._h2 = q_molecule.two_body_integrals
```
woodsp-ibm, line 52: Is this not doable then on the integrals that are stored in qmolecule, ie. those directly from the driver like is being done for the active space.

qiskit_nature/transformers/particle_hole_transformer.py
```diff
@@ -28,5 +45,2102 @@ def transform(self, q_molecule: QMolecule) -> QMolecule:
...
+
+        q_molecule_new = copy.deepcopy(q_molecule)
+
+        q_molecule_new.mo_onee_ints = self._h1[:int(self._h1.shape[0]/2),:int(self._h1.shape[1]/2)]
+
+        # TODO calculation of mo_eri is still wrong, because of the doubling of the space.
+        q_molecule_new.mo_eri_ints = self._h2[:int(self._h2.shape[0]/2),:int(self._h2.shape[1]/2),
+                                              :int(self._h2.shape[2]/2),:int(self._h2.shape[3]/2)]
+
+        q_molecule_new.energy_shift['ParticleHoleTransformer'] = identities_new_sum
```
woodsp-ibm, line 143: Just noting, since its not in todo list, that the dipole matrices need to be adjusted too I think.

qiskit_nature/transformers/particle_hole_transformer.py
```diff
@@ -28,5 +45,2102 @@ def transform(self, q_molecule: QMolecule) -> QMolecule:
...
+        swapped_indices[2] = array_ind_two_body_term[3]
+        swapped_indices[3] = array_ind_two_body_term[2]
+
+        return swapped_indices
+
+    def _normal_order_integrals(self, n_qubits, n_occupied, array_to_normal_order,
+                               array_mapping, h1_old, h2_old,
+                               h1_new, h2_new):
+        """
+        Given an operator and the rFs and rsgtu from Gaussian it produces new
```
woodsp-ibm, line 254: Can we amend this comment - it has been in the old code forever and never fixed. The logic may have been developed originally under that driver but now the qmolecule/integrals are driver independent

qiskit_nature/transformers/particle_hole_transformer.py
```diff
@@ -28,5 +45,2102 @@ def transform(self, q_molecule: QMolecule) -> QMolecule:
...
+                                  ind_ini_term[2], ind_ini_term[3]]
+                    ind_old_h2 = self.last_two_indices_swap(ind_old_h2)
+
+                    h2_new[ind_new_h2[0]][ind_new_h2[1]][
+                        ind_new_h2[2]][ind_new_h2[3]] \
+                        += 0.5 * temp_sign_h2 * \
+                           h2_old[ind_old_h2[0]][ind_old_h2[1]][ind_old_h2[2]][ind_old_h2[3]]
+
+                else:
+                    print('ERROR 1')
```
woodsp-ibm, line 443: And can we improve this to raise an error, if that is what is needed (or log a warning or whatever) not print.

qiskit_nature/drivers/qmolecule.py
```diff
@@ -116,15 +116,34 @@ def __init__(self, filename=None):
         self.nuclear_dipole_moment = None
         self.reverse_dipole_sign = False
 
+        # Custom one and two body integrals
+        self._one_body_integrals = None
+        self._two_body_integrals = None
```
woodsp-ibm, line 121: I have to say that there is nothing much about this I like. Unlike changing the basic integrals where the qmolecule data is self consistent this now sets these such it will return whatever is set in these. Its not saved in the HDF5, like the rest of the data, that includes the particle hole energy that is extracted, so that would be wrong of reconstituting it from an HDF5 as it would say particle hole energy was extracted but the data return from here would not reflect that.
And as I mentioned the dipoles needed a like treatment so I presume you have in mind the same for those

qiskit_nature/transformers/particle_hole_transformer.py
pbark: You are right. I am fixing this now.

qiskit_nature/transformers/particle_hole_transformer.py
pbark: This was removed by @mrossinek. This is fixed.

qiskit_nature/transformers/particle_hole_transformer.py
pbark: We need to use the full one and two body integrals and not the ones stored in the qmolecule. I made sure that the PH transformer is the last to be applied in the  molecular problem transformation

qiskit_nature/transformers/particle_hole_transformer.py
pbark: I do not think that we ever changed the dipole integrals accordingly in the PH picture. Do you have any reference in the old code for this? I cannot recall any implementation on dipole ints.

qiskit_nature/transformers/particle_hole_transformer.py
pbark: this is fixed in Max commits

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
```diff
@@ -73,8 +73,18 @@ def second_q_ops(self) -> List[SecondQuantizedOp]:
...
         for transformer in self.transformers:
-            q_molecule = transformer.transform(q_molecule)
+            # ensure Particle Hole Transformer is executed last
+            if transformer.__class__.__name__ == 'ParticleHoleTransformer':
+                ParticleHoleTransformer = transformer
+            else:
+                q_molecule = transformer.transform(q_molecule)
+
+        if ParticleHoleTransformer is not None:
+            q_molecule = ParticleHoleTransformer.transform(q_molecule)
```
woodsp-ibm, line 86: I thought the intent was to have some sort of priority information in the transformer itself that they could be ordered here according to that priority to execute the transformation. And then whatever was put in as the priority for ParticleHole it could be done such that it would be last. This does work for now, but at the very minimum a TODO could be put in the code to more have ordering info supplied by transformers, whatever that might be.

Also it would be better to just so instanceof rather than - variables there also should not be mixed case - it makes it look like a class name and is kinda confusing - but then I imagine lint will complain too...

qiskit_nature/transformers/particle_hole_transformer.py
woodsp-ibm: In the past PH acted on h1/h2 of FermionicOperator. We built out main operator (hamiltonian) and all dipoles as a set of FermionicOperators. They all transformed under the active space reduction, particle hole etc transformation capabilities of the FermionicOperator. You can see it here https://github.com/Qiskit/qiskit-nature/blob/f161ae83de1655aa55dde5dbf4fbbc32e4bb1348/qiskit_nature/transformations/fermionic_transformation.py#L301

qiskit_nature/problems/second_quantization/molecular/molecular_problem.py
pbark: I think this is fixed now.

pbark:
Tomorrow I will work on the dipole integrals fix. @mrossinek do we need to keep the legacy test?

mrossinek:
After some lengthy discussions and investigations offline, @pbark and I have decided to postpone the implementation of the `ParticleHoleTransformer` until a later release. I have tried to summarize our thought process in [this gist](https://gist.github.com/mrossinek/3cc65837d2a998856f202ef341e7f455) in the hope to save us the hassle of digging through the same problems when we tackle this in the future.

I will refrain from summarizing everything again here, because I don't think it will be more understandable than the gist. Although it is quite lengthy, please be sure to check out the conclusions at its end.

Issue #83, opened 2021-03-12
Integrate QubitConverter into the algorithm stack


mrossinek:
### What is the expected enhancement?

#60 added the `QubitConverter` which should be integrated into the algorithm stack.
The rationale is the following:

- In the past we had multiple keyword arguments on the `FermionicTransformation` (and elsewhere) which configured the `qubit_mapping` and `z2_symmetries`, etc.
- Now, we replaced these transformations with the `problems` module in which we only add problem-describing parameters.
- Since the aforementioned keywords deal with "how the algorithm solves the problem", these should go into the algorithms stack.
- Thus, we need to replace the legacy `transformation` argument in the algorithms with a `converter` argument.

PR #87, opened 2021-03-19
Minor cleanups

### LINKED ISSUE #64: Disambiguate `num_orbitals`

mrossinek:
### Summary

This PR is a collection of minor cleanups.
It also closes #64

### Details and comments

Grepping for `num_orbitals` still returns results in modules/classes which will be removed before the release. I did not bother changing things there, too. These include:
- `qiskit_nature.transformations`
- `qiskit_nature.components`

`qiskit_nature.algorithms` will be tackled during #83.


pbark (dismissed):

mrossinek:
Thanks for the approval. I think I will actually add some more things here, so I converted this to a draft for now.

qiskit_nature/mappers/second_quantization/__init__.py
```diff
@@ -13,25 +13,9 @@
...
-from .direct_mapper import DirectMapper
 from .fermionic_mapper import FermionicMapper
 from .jordan_wigner_mapper import JordanWignerMapper
 from .linear_mapper import LinearMapper
-from .logarithmic_mapper import LogarithmicMapper
 from .parity_mapper import ParityMapper
 from .qubit_mapper import QubitMapper
 from .spin_mapper import SpinMapper
-
-__all__ = [
```
woodsp-ibm, line 26: Why was this completely removed? This is normal to have in the init https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces

Also I would expect this file will have to contain sphinx documentation that is similar to what you see in other init files. That still needs doing for this and others I imagine.

mrossinek: This is not handled consistently within Nature at the moment. I saw it was not there elsewhere and thus removed it.
I will make an effort to make this consistent instead.

I will also go through all `__init__` files and a `TODO(sphinx): documentation` comment which is easy to grep for when we get to the final preparation stages of this release. This is not something I am going to tackle in this PR.

mrossinek:
I would also like to suggest replacing `indexes` with `indices` everywhere in the code base. What do you think?

qiskit_nature/mappers/second_quantization/__init__.py
woodsp-ibm: Taking care of the code is one thing. Given we often allow the user to import modules from its parent then not all init files need docs. We will have to look over the docs anyway and sort things out which yes would be a separate PR later once things are a bit more ready. As a first pass it needs at least the navigation structure in place - then that can be filled out. Anyway its for later....

mrossinek:
Besides my question above, I cannot think of any other pressing things which I would like to fix. Thus, this is now ready for review :+1:

woodsp-ibm:
> I would also like to suggest replacing indexes with indices everywhere in the code base. What do you think?

Either are accepted as plural of index. If things are consistent everywhere I guess that would be good. If its in comments I would think its very low priority to change - for a param name etc., i.e. something that would end up in users code, then it would be good to have things consistent - do we have that where its a parameter name or something?

mrossinek:
Both variants are being used in comments and as variable names (or parts thereof). `indices` appears to be the more widely used one.
[indexes.txt](https://github.com/Qiskit/qiskit-nature/files/6183818/indexes.txt)
[indices.txt](https://github.com/Qiskit/qiskit-nature/files/6183820/indices.txt)

woodsp-ibm:
For internal variable names we can change it any time. Its more for param names on public API where we could expect the name to appear in user code. It would be nice to be consistent. I am fine with `indices` as that I think is more used in general

mrossinek:
Although this is likely to not catch all cases a quick grep for `def.*indexes` and `def.*indices` returns only a single result for the former (a function of `UVCC`) and multiple for the latter as it is used in some args of private functions around the repo.

mrossinek:
I tried a `s/indexes/indices/g` on Qiskit Nature... Let's see if the CI is still happy.
I also found some `DeprecationWarning`s being raised by the mappers because they used legacy methods from the Pauli operators. I fixed those, too.

pbark (approved):

pbark:
Thanks @mrossinek for taking care of this!

mrossinek:
I force pushed this branch because a mixture of online merges of the `master` branch and local ones confused this branch into thinking it included some others PRs. This branch is fixed now and contains the correct changes.
@pbark Please re-approve because this process dismissed your approval

woodsp-ibm (approved):

PR #88, opened 2021-03-22
Add Normal Order to FermionicOp

ikkoham:
### Summary

Add `FermionicOp.to_normal_order()`.
It returns the normal ordered fermionic operator that is equivalent to self.


qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -358,6 +371,36 @@ def _to_sparse_label(label):
             filter(None, (label_transformation[c].format(i=i) for i, c in enumerate(label)))
         )
 
+    def to_normal_order(self) -> "FermionicOp":
+        """Convert to the normal order. The returned operator is a sparse label mode."""
```
woodsp-ibm, line 375: This does not convert self right? it returns a new sparse label FermionicOp that is converted to normal mode.

Why does it change self though - it sets `self.sparse_label = False` which to me looks wrong if its building out a complete new one based on the current, it should not affect/change the current one i.e. self though right?

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -210,9 +213,18 @@ def __repr__(self) -> str:
...
         if len(self) == 1:
             if data[0][1] == 1:
-                return f"FermionicOp('{data[0][0]}', register_length={self.register_length})"
-            return f"FermionicOp({data[0]}, register_length={self.register_length})"
-        return f"FermionicOp({data}, register_length={self.register_length})"  # TODO truncate
+                data_str = f"'{data[0][0]}'"
+            data_str = f"'{data[0]}'"
+        data_str = f"{data}"
+
+        # TODO truncate
```
woodsp-ibm, line 220: If this TODO still pending - is it intended to truncate the data string if the data is large - like numpy does when printing arrays?

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Thanks. This is the bug.This method should not change the `self.sparse_label`. This is only for the conversion.

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Yes, that's what I intended.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -397,9 +404,11 @@ def to_normal_order(self) -> "FermionicOp":
                     True,
                 )
 
+        self.sparse_label = temp_sparse_label
+
         if isinstance(ret, FermionicOp):
             return ret
-        return FermionicOp(("", 0), self.register_length, True)
+        return FermionicOp(("", 0), self.register_length, self.sparse_label)
```
woodsp-ibm, line 411: Is this assuming that the current FermionicOp is sparse. I assumed the setting its sparse labal to False we because logic was used that depended on that. Hence whether it was False or not that set it as wanted. The label is later restored as it was originally. Now the method says it returns a sparse op, yet this line was changed from setting it True to be whatever is in the current operator. Might that be False? Or is this method assuming its sparse - in which case is that ok if its called on an operator that is not sparse - or should that raise an error.

ikkoham: Current FermionicOp is dense in default. I personally had a hard time with the implementation here, and I thought it would be more convenient to return the same type as the original operator, so I did this. But the doc string of the method said "it returns sparse label", so this should be True as you said. Thanks.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -358,6 +371,45 @@ def _to_sparse_label(label):
...
         )
 
+    def to_normal_order(self) -> "FermionicOp":
+        """Convert to the normal order. This is not usual normal order.
+        It returns the operator with normal order but equivalent to the input operator.
+        For example, in usual :math:`\\colon c_0 c_0^\\dagger\\colon = -c_0^\\dagger c_0`
+        where :math:`c_0` is an annihilation operator,
+        but this method returns :math:`-c_0^\\dagger c_0 + 1`.
+        The returned operator is a sparse label mode.
+        """
```
mrossinek, line 381: I don't quite understand what you are trying to say here.. Shouldn't this function always perform the computation according to:
```
c_i c_j^\dagger = \delta_ij - c_j^\dagger c_i
```
where `\delta_ij` is the Kronecker delta (i.e. `delta = 1 iff i == j else 0`)

According to the following test this is indeed the case:
```python
>>> print(FermionicOp("E_0", register_length=4).to_normal_order())
   * (1+0j)    # <- notice this weird formatting edge case
+ +_0 -_0 * (-1+0j)
>>> print(FermionicOp("-_0 +_1", register_length=4).to_normal_order())
+_1 -_0 * (-1+0j)
```

test/operators/second_quantization/test_fermionic_op.py
```diff
@@ -300,6 +300,50 @@ def test_label_display_mode(self, label, pre_processing):
...
+            targ = FermionicOp([("", 1), ("+_0 -_0", -1)], sparse_label=True)
+            self.assertFermionEqual(fer_op, targ)
+            self.assertFalse(orig.sparse_label)
+
+        with self.subTest("Test for multiple operators 1"):
+            orig = FermionicOp("-+")
+            fer_op = orig.to_normal_order()
+            targ = FermionicOp([("+_1 -_0", -1)], sparse_label=True)
+            self.assertFermionEqual(fer_op, targ)
+            self.assertFalse(orig.sparse_label)
```
mrossinek, line 338: As per my previous comment you even ensure that this works as it should :+1:

mrossinek (changes requested):
This is basically good to go, but one docstring appears quite wrong to me..

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Yes. I've already implemented it for the next PR.

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: You are right. In the usual normal order, `:c_i c_j†: = - c_j† c_i`, my intention is to note that this method behaves differently than usual.
(And I'll notice the weird formatting.. Thanks. I'll fix it with truncation in the next PR.)

mrossinek: What I meant is that I don't understand how this method is different than the usual normal oder?

ikkoham: I see.
Usual: `: c_0 c_0^†: = -c_0^† c_0.`
This method: `: c_0 c_0^†: = 1 - c_0^† c_0.`

mrossinek: No but this is exactly what I am saying. Because these are *fermionic* operators, the definition is:
```
:c_i c_j^†: = \delta_ij - c_j^† c_i
```
This means that the usual case is:
* `:c_i c_j^†: = - c_j^† c_i` if and only if `i != j`
* `:c_i c_i^†: = 1 - c_i^† c_i` if and only if `i == j`

Thus, this method does exactly what is expected from it.

ikkoham: Your reference wikipedia said
image

ikkoham: https://physics.stackexchange.com/questions/149419/normal-ordering-for-a-two-fermion-case

mrossinek: That is the single fermions section. You need to consider the multiple fermions case, which is this:
screenshot_1627985202
The most important part being the very last line on this screenshot

ikkoham: The last line is just the commutation relation, not the definition of the normal order.

ikkoham: image

ikkoham: If we adopt your definition, 
```
:f_2 f_1^\dagger f_2^\dagger: = -f_1^\dagger f_2 f_2^\dagger = - f_1^\dagger (1 - f_2^\dagger f_2) = - f_1^\dagger + f_1^\dagger f_2^\dagger f_2.
```
However Wikipedia said `:f_2 f_1^\dagger f_2^\dagger: = f_1^\dagger f_2^\dagger f_2`.

ikkoham: And, the case for single fermion means we only have single register.

ikkoham: In the first place, the normal ordered product is used to remove divergence of vacuum expectation value in field theory, so it can't be an equivalent operator.

pbark: @ikkoham thanks for explaining this. Please add a reference to the docstring for future reference so that this is clear. I think that the misunderstanding here is that @ikkoham is talking about the definition of normal order (that is indeed right) and @mrossinek is taking about how to do an algebraic transformation of a random ordered operator to a normal ordered operator (that needs to take into account the commutation rules and the above definition is also right). The important is that the code does the transformation to normal order in the proper way. Thanks!

mrossinek: Thanks for clarifying @pbark! In that case the docstring should simply reflect that the operator gets transformed according to the anticommutation rules :+1:

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -358,6 +371,51 @@ def _to_sparse_label(label):
...
+        The returned operator is a sparse label mode.
+
+        .. note::
+
+            This method does not return usual normal ordering.
+            The order is calculated by commutation relation.
+            For example, in usual :math:`\\colon c_0 c_0^\\dagger\\colon = -c_0^\\dagger c_0`
+            where :math:`c_0` is an annihilation operator,
+            but this method returns :math:`1 - c_0^\\dagger c_0`.
+            See the reference: https://en.wikipedia.org/wiki/Normal_order#Multiple_fermions.
```
pbark, line 385:
```suggestion
            This method implements the transformation of an operator to the normal ordered operator. 
            The transformation is calculated by considering all commutation relations between the operators.
            For example, for the case :math:`\\colon c_0 c_0^\\dagger\\colon`
            where :math:`c_0` is an annihilation operator,
            this method returns :math:`1 - c_0^\\dagger c_0` due to commutation relations. 
            See the reference: https://en.wikipedia.org/wiki/Normal_order#Multiple_fermions.
```

pbark: @ikkoham thanks for the updates. I think the suggestion is more explanatory and helps understand better what we are doing. I still find the previous one a bit cryptic. @mrossinek what do you think?

ikkoham: Thank you so much. I was only worried about confusing the users who know the wick ordering. I think your improvement is nice. After Max's approval, I'll update it.

mrossinek: Thanks @pbark! I think your suggestion works very well :+1: 
@ikkoham if you can update this today, we can merge it right after :slightly_smiling_face:

mrossinek (approved):

pbark (approved):

PR #96, opened 2021-03-23
Add QubitConvertor symmetry reduction logic

### LINKED ISSUE #82: Finish the QubitConvertor implementation
A first pass was merged with Z2Symmetry support, this should be added to the function, along with the two qubit reduction based on the mappers.

If we provide a mechanism to map state from its fermionic to qubit equivalent, as we need for HartreeFock, then this should be added and exposed on the Qubit convertor as a capability too based on the function of state mapping being added to and provided by the mappers. So they would map state as well as operators to the qubit form.

woodsp-ibm:
Closes #82

- [X] Two qubit reduction support
- [X] Symmetry tapering support
- [X] Update for reuse by UCCSD to taper similarly etc
- [X] Unit testing newly added code as above
- [X]  Tidy up docstrings


qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -11,26 +11,92 @@
...
+    @property
+    def two_qubit_reduction(self) -> bool:
+        """Get two_qubit_reduction"""
+        return self._two_qubit_reduction
+
+    @two_qubit_reduction.setter
+    def two_qubit_reduction(self, value: bool) -> None:
+        """Set two_qubit_reduction"""
+        self._two_qubit_reduction = value
+        self._invalidate()
```
mrossinek, line 99: Do we always want to invalidate or only when this value actually changed?
The same can be asked for the other setters.

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -40,58 +106,235 @@ def z2symmetry_reduction(self) -> Optional[List[int]]:
...
         Args:
-            second_q_ops: The list of second quantized operators to be converted
+            second_q_ops: A second quantized operator, or list thereof to be converted.
+            num_particles: Needed for two qubit reduction to determine correct sector. If
+                not supplied, even if two_qubit_reduction is possible, it will not be done.
             z2symmetry_reduction: Optional z2symmetry reduction, the sector of the symmetry
 
         Returns:
-            A list of qubit operators or the same length as the second_q_ops list. The first
+            A qubit operator or list of thereof the same length as the second_q_ops list. The first
```
mrossinek, line 166:
```suggestion
            A qubit operator or list thereof of the same length as the second_q_ops list. The first
```

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -40,58 +106,235 @@ def z2symmetry_reduction(self) -> Optional[List[int]]:
...
+            qubit_ops.append(self._mapper.map(op))
 
         return qubit_ops
 
-    def _symmetry_reduce(self, qubit_ops: List[PauliSumOp]) -> List[Optional[PauliSumOp]]:
+    def _two_qubit_reduce(self, qubit_ops: List[PauliSumOp],
+                          num_particles: Optional[Tuple[int, int]])\
+            -> Tuple[List[Optional[PauliSumOp]], bool]:
+        if self._two_qubit_reduction and not self._mapper.allows_two_qubit_reduction:
+            logger.warning("Ignoring requested two qubit reduction as mapping does not support it")
```
mrossinek, line 248:
```suggestion
            logger.warning("Ignoring requested two qubit reduction as mapper does not support it")
```

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -40,58 +106,235 @@ def z2symmetry_reduction(self) -> Optional[List[int]]:
...
+        # now that a conversion was successful
+        self._set_valid(did_2_q, num_particles, z2symmetries)
 
-        return qubit_ops_reduced
+        if wrapped:
+            qubit_ops_tapered = qubit_ops_tapered[0]
+
+        return qubit_ops_tapered
+
+    def convert_more(self, second_q_ops: Union[SecondQuantizedOp, List[SecondQuantizedOp]]
```
mrossinek, line 196: I am wondering whether it may be nice to merge this and the previous method (at least for the public API) into a single `convert` method. `num_particles` and `z2symmetry_reductions` are `Optional` already and we can leverage that to our advantage here.

For example:
```python
converter = QubitConverter(mapper=ParityMapper(), two_qubit_reduction=True)

initial_qubit_op = converter.convert(initial_second_q_op, num_particles)

# At this point, `_check_valid` will return True and we already have all the information necessary. Omitting the optional args can thus trigger the current `convert_more` function.
more_qubit_ops = converter.convert(more_second_q_ops)
```

If a user would want to re-use a converter object without using the state (for which I cannot think of a reason but who knows...) he can simply invalidate the state before calling `convert` again.

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -40,58 +106,235 @@ def z2symmetry_reduction(self) -> Optional[List[int]]:
...
-            self._z2symmetries = Z2Symmetries([], [], [], None)
+
+        return reduced_qubit_ops, did_2_q
+
+    def _find_symmetry_reduce(self, qubit_ops: List[PauliSumOp])\
+            -> Tuple[List[Optional[PauliSumOp]], Z2Symmetries]:
+
+        if self.z2symmetry_reduction is None:
+            tapered_qubit_ops = qubit_ops
+            z2_symmetries = self._no_symmetries
```
mrossinek, line 267: If you add a return statement here, you can get red if of `else` and dedent the rest of this function.

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -40,58 +106,235 @@ def z2symmetry_reduction(self) -> Optional[List[int]]:
...
+        else:
+            z2_symmetries = Z2Symmetries.find_Z2_symmetries(qubit_ops[0])
+            if z2_symmetries.is_empty():
+                logger.debug('No Z2 symmetries found')
+                tapered_qubit_ops = qubit_ops
```
mrossinek, line 272: Same here

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -40,58 +106,235 @@ def z2symmetry_reduction(self) -> Optional[List[int]]:
...
+                    if i == 0 and not commutes:
+                        raise QiskitNatureError('Z2 symmetry failure. Main operator must commute '
+                                                'with symmetries found from it!')
+                    commuted.append(commutes)
+                    logger.debug("Qubit operators commuted with symmetry %s", commuted)
+
+                # set tapering for the given sector and taper operators
+                z2_symmetries.tapering_values = self._z2symmetry_reduction
+                tapered_qubit_ops = []
+                for i, commutes in enumerate(commuted):
```
mrossinek, line 297: Do we need to have this extra `for`-loop? I think it would be better/faster to integrate it with the one above

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -40,58 +106,235 @@ def z2symmetry_reduction(self) -> Optional[List[int]]:
...
+                    op = z2_symmetries.taper(qubit_ops[i]) if commutes else None
+                    tapered_qubit_ops.append(op)
+
+        return tapered_qubit_ops, z2_symmetries
+
+    def _symmetry_reduce(self, qubit_ops: List[PauliSumOp])\
```
mrossinek, line 303: This entire function is almost an exact duplicate of the previous one.
If my comment about merging `convert` and `convert_more` will not be pursued these two functions can probably be split more conveniently in any case in order to minimize duplication.

mrossinek (commented):
This already looks quite good to me! I left some comments for your consideration :slightly_smiling_face:

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: I did not make any value check - it seemed to me an unnecessary complication. If you were setting it you were likely going to reuse it for a different conversion rather than do more of the same conversions. Yes you might set it exactly the same but as I said I did not believe it was worth the extra value checks

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: Yes, it maybe the comments need updating. When I am done with the testing and reviewing that I will fix things up. As it is the grammar there is messed up as I already made one change

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: Since a mapper performs a mapping either seemed fine to me - it was the mapping itself that did not support it where the mapper implemented the mapping.

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: I only just added the convert_more. They are different but similar. I am more trying to get the code together and tested at present. If you want two q reduction and auto symmetry reduction then the user has to supply num particles. When converting again it uses the stored num particles from before.

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: I prefer a single return at the end of a method unless its an exception out of it.

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: I am trying to get the code operational at present and have not worried too much about optimizing it yet. I would have done it as two since it was more convenient with the separation of the steps. Since we have few operators I doubt there is any performance impact

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: I do not know. They are different but similar - one needs to find symmetries the other not, etc. Yes one could do it with set of if elses. It was easier to do it like this for now since they are separate

qiskit_nature/operators/second_quantization/qubit_converter.py
mrossinek: Ah because the warning could eitherwise suggest that the current implementation of the `mapper` is not capable of doing so, but the problem is that it is incompatible with the `mapping` itself. I see.

qiskit_nature/operators/second_quantization/qubit_converter.py
mrossinek: Yes, that's what I am saying. I don't think it's necessary to have this in two methods because the `convert` method itself could reuse the stored values (if they are available).

qiskit_nature/operators/second_quantization/qubit_converter.py
mrossinek: We want the code to be scalable for the future no matter what. Hence the comment.
And its fine that you are just getting things together right now. I just wanted to give you this input for consideration.

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: My feeling at present is that the convertor should just convert, given the set of parameters it has just been passed, without relying on magic values in there. Sector can be passed on constructor so user can set things manually, though given something like auto sector that can be overridden. So already it has value tests. My preference is that if you want to convert the same was as a prior one, and supply just 2nd q ops, that you do it explicitly. This way there should be less issues of having values that unexpectedly cause the convert to use any stored state it has.

mrossinek: Hm I see. But that would require the code to be changed such that you can pass in `Z2Symmetries` rather than a sector right? Where would these `Z2Symmetries` be stored? Would the initial `convert` have to return them too?

qiskit_nature/operators/second_quantization/qubit_converter.py
mrossinek: Yes, let's postpone internal refactorings until after the upcoming release

dlasecki:
One minor modification might be necessary - in QEOM we need to have an access to an `untapered_qubit`, i.e. after a mapping but before reductions. We discussed with Max that it could be achieved my making `_map_to_qubits` and `_symmetry_reduce` public so that one could save an `untapered_qubit` for methods that require it.

woodsp-ibm:
>  in QEOM we need to have an access to an untapered_qubit

Arguably you can get the mapper from it and just use the map method on whatever operator you have

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: No it needs a sector - the sector can be put as tapering_values in Z2Symmetry once symmetries are found, to tell the taper method to just give back the operator for that sector not all of them,

woodsp-ibm:
I updated the API.

`convert`: converts a single operator, the main one only. Does all the mapping, two q reduce and symmetry

`convert_match`: Will convert aux_ops or any other operators to match saved state from convert. Whereas before this was policed    that a prior convert had to be done, now it does not. It will use whatever state it has. If the problem has a list where one is the main operator and the other aux ops, it will need to split the list to pass things separately now.

If for some reason, mainly for tests, or perhaps some advance use, there is a `force_match` method that allows you to set the state, ie num particles and a Z2Symmetry object with valid tapering values (same state that convert could generate). This allows one to setup a qubit convertor without requiring a convert that would normally find the symmetry of main operator. This allows more flexibility but more risk since it is not policed as it was before, which caused UCC tests etc to fail as it was tested with an unused QubitConvertor whereas in practice we expected one already used in the stack to be passed in so excitation operators could be conformed to match. This now allows the testing (or any other advanced use) to use different QubitConvertors, though in practice the normal usage will be to re-use the one used for the main operator conversion.

`map`: Even though one can get the mapper and call map on it, assuming all you want to do is the mapping, I added a convenience method to do the mapping given a 2nd q operator or list thereof.

qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -240,7 +240,14 @@ def _build(self) -> None:
             excitation_ops = self.excitation_ops()
 
             logger.debug('Converting SecondQuantizedOps into PauliSumOps...')
-            self.operators = self.qubit_converter.to_qubit_ops(excitation_ops)
+            # Convert operators according to saved state in converter from the conversion of the
+            # main operator since these need to be compatible. If Z2 Symmetry tapering was done
+            # the it may be that one or more excitation operators do not commute with the
```
mrossinek, line 245:
```suggestion
            # it may be that one or more excitation operators do not commute with the
```

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -11,26 +11,93 @@
...
+    the :class:`~qiskit_nature.mappers.second_quantization.ParityMapper` that introduces known
+    symmetries, by virtue of the mapping, that can be exploited to reduce the size of the
+    problem, i.e the qubit operator, by two qubits. The two qubit reduction setting indicates
+    to do this where possible - i.e. mapper supports it and the number of particles in the
+    Fermionic system is provided for the conversion. (The number of particles is used to
+    determine the symmetry)
+
+    Also this converter supports Z2 Symmetry reduction to reduce a problem (operator) size
+    based on mathematical symmetries that can be detected in the operator. A knowledgeable user
+    can define which sector the problem solution lies in. This sector information can be also
```
mrossinek, line 48:
```suggestion
    can define which sector the problem solution lies in. This sector information can also
```

mrossinek (dismissed):
Thanks a lot @woodsp-ibm! This looks very good to me now! I still found two nit-picky comments, but we can also fix them another time.

pbark (dismissed):

pbark (approved):

mrossinek (approved):

PR #92, opened 2021-03-22
Adding the linear mapper

pbark:
### Summary
Linear Mapper implementation 

### Details and comments

I still need to 

- [x] implement the mode_based_mapping for the spin operator with transformed xyzi
- [x] finish the unittest

PR #98, opened 2021-03-24
VibrationalOp and DirectMapper

mrossinek:
### Summary

This PR refactors the `VibrationalSpinOp` into a `VibrationalOp`.
It also adds the `DirectMapper` which is a straight forward migration of the legacy `BosonicOperator.mapping(mapping='direct')` implementation.

### Details and comments

The reason for this drastic (and short-notice) refactoring of the `VibrationalSpinOp` is a number of issues which we ran into with the planned compatibility of the `VibrationalSpinOp` and the `LinearMapper` (#92). The biggest problem was that each `modal` of the `VibrationalSpinOp` needs to be able to be mapped to a single qubit because it should represent a 2-state system. However, with the `LinearMapper` this would not have been possible and would have required (e.g.) a _linear_ encoding of the `modes` but a _logarithmic_ encoding of each `modal` within them...

As this is rather convoluted we have decided to discard the idea and introduce a new **standalone** `VibrationOp` which does not attempt to be related to the `SpinOp`. Since the `VibrationOp` has the requirement of allowing merely raising and lowering operators (i.e. an alphabet of `+` and `-`) it is actually closer to the `FermionicOp`. In fact, working on it just now, I believe it could even be fully compatible. But I am not sure about the physical logic of this and, thus, decided to leave it a standalone object for now.

### Open Tasks
- [x] Improve documentation
- [x] Unify naming of `num_modals` and `basis_size` throughout the stack


qiskit_nature/operators/second_quantization/vibrational_op.py
```diff
@@ -0,0 +1,356 @@
...
+            if not new_data:
+                return VibrationalOp(("I_0*0", 0), self._num_modes, self._num_modals)
+
+            return VibrationalOp(new_data, self._num_modes, self._num_modals)
+
+        raise TypeError(
+            f"Unsupported operand type(s) for *: 'VibrationalOp' and '{type(other).__name__}'"
+        )
+
+    # Map the products of two operators on a single fermionic mode to their result.
```
ikkoham, line 294: Just question. VibrationOp has a algebra same with fermion? or boson?

mrossinek: Thanks for spotting this! I did copy a lot of code from `FermionicOp` but the modals are `bosonic` in nature! (confirmed with @paulineollitrault)

I just updated the code to not do the anti-commutation (i.e. take the sign of permutations into account) as we do in the `FermionicOp`. I hope this answers your question :slightly_smiling_face:

mrossinek:
@woodsp-ibm @pbark Please review this PR when you have time. We need this in order to proceed with UVCC etc. and finally with the vibrational stack in the algorithms.

qiskit_nature/operators/second_quantization/vibrational_op.py
```diff
@@ -0,0 +1,338 @@
...
+    `+_{mode_index}*{modal_index}`. Each modal can be excited at most once.
+
+    For more information on how the sparse label notation works, please refer to the more extensive
+    examples in the :class:`SpinOp` class.
+
+    2. Dense Labels:
+
+    Internally, a :class:`VibrationalOp` is stored as a dense label in an identical notation to that
+    of the :class:`FermionicOp`. Thus, each modal is mapped onto a position of the string and can be
+    one of the following characters: `I`, `+`, `-`, `N` and `E`. Here, `N` and `E` are shorthands
```
woodsp-ibm, line 64:
```suggestion
    one of the following characters: `I`, `+`, `-`, `N` and `E`. Here, `N` and `E` are shorthand
```
Very minor comment. To me the plural form reads oddly here - saying they are shorthand for ie short for in seems fine. It also means its not needed in the custom dictionary so I guess its not common in the plural

qiskit_nature/operators/second_quantization/vibrational_op.py
```diff
@@ -0,0 +1,338 @@
...
+
+from .second_quantized_op import SecondQuantizedOp
+
+
+class VibrationalOp(SecondQuantizedOp):
+    """Vibration type operators.
+
+    **Label**
+
+    :class:`VibrationalOp` supports two kinds of labels: sparse and dense ones.
```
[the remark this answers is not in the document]
mrossinek: Yes I agree that makes sense. But yes, let's address it that when we go over the docs in a separate PR

qiskit_nature/operators/second_quantization/vibrational_op.py
```diff
@@ -0,0 +1,338 @@
...
+
+    Internally, a :class:`VibrationalOp` is stored as a dense label in an identical notation to that
+    of the :class:`FermionicOp`. Thus, each modal is mapped onto a position of the string and can be
+    one of the following characters: `I`, `+`, `-`, `N` and `E`. Here, `N` and `E` are shorthands
+    for `+-` and `-+` applied on the same modal, respectively.
+    The length of a dense label is given by `register_length` and can be computed as the sum of all
+    modals over all modes.
+
+    For more information on how the dense label notation works, please refer to the more extensive
+    examples in the :class:`FermionicOp` class.
```
woodsp-ibm, line 70: This is a good example of use of class since it refers to another one and the hyperlink will allow a reader to go straight to it and see

qiskit_nature/operators/second_quantization/vibrational_op.py
```diff
@@ -0,0 +1,338 @@
...
+            op = reduce(lambda a, b: a + b, ops)
+
+            self._labels = op._labels.copy()
+
+    def __repr__(self) -> str:
+        if len(self) == 1:
+            if self._coeffs[0] == 1:
+                return f"VibrationalOp('{self._labels[0]}')"
+            return f"VibrationalOp({self.to_list()[0]})"
+        return f"VibrationalOp({self.to_list()})"  # TODO truncate
```
woodsp-ibm, line 164: Looking at TODO here

qiskit_nature/operators/second_quantization/vibrational_op.py
```diff
@@ -0,0 +1,338 @@
...
+    def num_modals(self) -> List[int]:
+        """The number of modals.
+        Returns:
+            The number of modals
+        """
+        return self._num_modals
+
+    @property
+    def register_length(self) -> int:
+        """Getter for the length of the register that the VibrationalOp `self` acts on."""
```
woodsp-ibm, line 194: Just saying Gets the register length would suffice - the stuff about VibrationalOp self should be implicit right - these are all properties of the class.

qiskit_nature/operators/second_quantization/vibrational_op.py
```diff
@@ -0,0 +1,338 @@
...
+        return VibrationalOp(
+            list(
+                zip(self._labels + other._labels, np.hstack((self._coeffs, other._coeffs)).tolist())
+            ),
+            self._num_modes,
+            self._num_modals,
+        )
+
+    def to_list(self) -> List[Tuple[str, complex]]:
+        """Getter for the operator_list of `self`"""
```
woodsp-ibm, line 224: This is not a getter like the properties above. More a method to return the operator as a List of (label, coeff) tuples

qiskit_nature/operators/second_quantization/vibrational_op.py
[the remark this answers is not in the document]
woodsp-ibm: In general I had imagined going over the docs was more around formatting, ensuring things were in tables of contents etc. While these are links and hence might fall into that category they are also more about individual file content which I had not really intended to really spend reading. If you are also going to look over things at that latter level then its fine.

qiskit_nature/operators/second_quantization/vibrational_op.py
```diff
@@ -0,0 +1,338 @@
...
+
+    @classmethod
+    def _single_mul(cls, label1: str, label2: str) -> Tuple[str, bool]:
+        if len(label1) != len(label2):
+            raise QiskitNatureError("Operators act on Fermion Registers of different length")
+
+        new_label = []
+
+        for pair in zip(label1, label2):
+            # Check what happens to the symbol
```
woodsp-ibm, line 330: What does this mean? If its stating what the code is doing or should be doing can it not be more explicit? Like check if symbol becomes 0 which is discarded. Check what happens is just so abstract

qiskit_nature/operators/second_quantization/vibrational_op_utils/vibrational_dense_label_converter.py
```diff
@@ -10,20 +10,19 @@
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
 """
-This module converts labels from the `VibrationalSpinOp` to 'SpinOp` notation.
+This module converts labels from the `VibrationalOp` to 'SpinOp` notation.
```
woodsp-ibm, line 13: These would be good for :class: labels. (Last comment on the matter...)

qiskit_nature/operators/second_quantization/vibrational_op_utils/vibrational_labels_validator.py
```diff
@@ -29,7 +32,8 @@ def _validate_vibrational_labels(vibrational_labels: List[Tuple[str, complex]],
             * operators in each label are sorted in the decreasing order of modes and modals,
             if both are equal then '+' comes before '-' (i.e. they are normal ordered),
             * for each `+` operator in each label, there is a corresponding '-' operator
-            acting on the same mode (i.e. the number of particles is preserved per mode).
+            acting on the same mode (i.e. the number of particles is preserved per mode) [this last
+            case will only log a warning]
```
woodsp-ibm, line 36: What does this mean - A warning is logged is particles are preserved (or is the operators are such that this is not the case that they are preserved). I was trying to read that in conjunction with the warning below but that just says the number mismatches which leads me to believe that the warning is emitted if thats not the last case and particles are not preserveed

qiskit_nature/operators/second_quantization/vibrational_op_utils/vibrational_dense_label_converter.py
```diff
@@ -10,20 +10,19 @@
...
 
-from qiskit_nature.operators.second_quantization.vibrational_spin_op_utils \
-    .vibrational_labels_validator import _validate_vibrational_labels
+from .vibrational_labels_validator import _validate_vibrational_labels
 
 
-def _convert_to_spin_op_labels(vibrational_labels: List[Tuple[str, complex]], num_modes: int,
-                               num_modals: Union[int, List[int]]) -> List[Tuple[str, complex]]:
-    """Converts `VibrationalSpinOp` labels to `SpinOp` labels.
+def _convert_to_dense_labels(vibrational_labels: List[Tuple[str, complex]], num_modes: int,
```
woodsp-ibm, line 23: These are all private methods yet used outside this package. For me this whole deal with these util modules that are in a separate pkg from the sole code that uses them is not very good. If an operator really needed splitting into modules, and to me combining these into the main one would not make the file too long, then better in my mind to move the VibrationalOp into its own pkg along with these. The VibrationalOp can be exported from the package (ie import in init and none of these util methods that can still be public). And the VibrationalOp can for convenience be in the operator init too at the level above so users would import it from there even if in its own pkg since it comprises a set of modules, not just one.

qiskit_nature/problems/second_quantization/vibrational/vibrational_op_builder.py
```diff
@@ -9,38 +9,40 @@
...
-
     num_modes = watson_hamiltonian.num_modes
-    if isinstance(basis_size, int):
-        basis_size = [basis_size] * num_modes
-    # TODO make HarmonicBasis an argument and support other bases when implemented
+
+    if isinstance(num_modals, int):
+        num_modals = [num_modals] * num_modes
+
+    # TODO: make HarmonicBasis an argument and support other bases when implemented
```
woodsp-ibm, line 42: I guess if this argument is added on the end in the future and defaults to HarmonicBasis internally then things would stay compatible. Just a comment against the TODO which I am assuming the intent is that this would not be done this release

qiskit_nature/operators/second_quantization/vibrational_op.py
mrossinek: Copied over from `FermionicOp`. Thus it was placed by @ikkoham.
I assume this is something we will address in the future

qiskit_nature/operators/second_quantization/vibrational_op.py
mrossinek: These comments were copied over from `FermionicOp`. I will update them here and in that class, too.

qiskit_nature/operators/second_quantization/vibrational_op_utils/vibrational_labels_validator.py
mrossinek: I clarified this docstring.

qiskit_nature/operators/second_quantization/vibrational_op_utils/vibrational_dense_label_converter.py
mrossinek: I will get rid of the `utils` module and migrate everything into the `VibrationalOp` then. I really just followed what we had for the `VibrationalSpinOp` before.

woodsp-ibm (approved):

PR #104, opened 2021-03-31
Auto symmetry reduction

pbark:
### Summary
Added the auto symmetry detection function in the Molecular problem to be used as callbacks for symmetry autodetection in the qubit converter


mrossinek:
Once 101 is merged this PR should also take care of integrating this auto-detection into the problem classes

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -52,12 +52,16 @@ class QubitConverter:
...
 
         Args:
             mapper: A mapper instance used to convert second quantized to qubit operators
             two_qubit_reduction: Whether to carry out two qubit reduction when possible
+            auto_symmetry_reduction: Whether to perform automatic symmetry reduction. This requires
+                the `sector_locator` argument of `convert` to be specified. This option will also be
+                overwritten if `z2symmetry_reduction` is not `None`.
             z2symmetry_reduction: An optional sector definition so the desired operator, from
                the tapered set, containing the problem solution is returned. This is a list of
                -1 and 1's to define the sector, where the list size is the number of symmetries
```
woodsp-ibm, line 67: If we are specifying the auto reduction here maybe we should just have it like before where z2_symmetry reduction was either None, a manually given sector (List[int]) or the 'auto' string. Evidently auto will only work with problems/algorithms that supply a the sector locator to convert. By having auto, manual or none in the same parameter I think its clearer to an end user.

As it stands now, the one from sector locator would override any manual one (as long as it returns a non-None value back which is what I expected). Had it returned None. With both on the one param, as they were before, we could never get into that situation since it would no symmetry, manual or auto.

pbark:
LGTM!

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -52,21 +52,31 @@ class QubitConverter:
...
         Args:
             mapper: A mapper instance used to convert second quantized to qubit operators
             two_qubit_reduction: Whether to carry out two qubit reduction when possible
-            z2symmetry_reduction: An optional sector definition so the desired operator, from
-               the tapered set, containing the problem solution is returned. This is a list of
-               -1 and 1's to define the sector, where the list size is the number of symmetries
-               of the main operator.
+            z2symmetry_reduction: If z2 symmetry reduction should be applied to resulting
+                qubit operators that are computed. For each symmetry detected the operator will be
+                split in two where each requires one qubit less for computation. So for example
```
dlasecki, line 63: #
```suggestion
                split into two where each requires one qubit less for computation. So for example
```

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -52,21 +52,31 @@ class QubitConverter:
...
         """
 
         Args:
             mapper: A mapper instance used to convert second quantized to qubit operators
             two_qubit_reduction: Whether to carry out two qubit reduction when possible
-            z2symmetry_reduction: An optional sector definition so the desired operator, from
-               the tapered set, containing the problem solution is returned. This is a list of
-               -1 and 1's to define the sector, where the list size is the number of symmetries
-               of the main operator.
+            z2symmetry_reduction: If z2 symmetry reduction should be applied to resulting
```
dlasecki, line 61:
```suggestion
            z2symmetry_reduction: Indicates whether a z2 symmetry reduction should be applied to resulting
```

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -52,21 +52,31 @@ class QubitConverter:
...
+                qubit operators that are computed. For each symmetry detected the operator will be
+                split in two where each requires one qubit less for computation. So for example
+                3 symmetries will split in the original operator into 8 new operators each
```
dlasecki, line 64: `
```suggestion
                3 symmetries will split the original operator into 8 new operators each
```

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -52,21 +52,31 @@ class QubitConverter:
...
+                requiring 3 less qubits. Now only one of these operators will have the ground state
+                and be the correct symmetry sector needed for the ground state. Setting 'auto' will
+                use an automatic computation of the correct sector. If from other experiments, with
```
dlasecki, line 67:
```suggestion
                use an automatic computation of the correct sector. If the sector is known
```

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -52,21 +52,31 @@ class QubitConverter:
...
+                the z2symmetry logic, the sector is known, then the tapering values of that sector
```
dlasecki, line 68:
```suggestion
              from other experiments with the z2symmetry logic, then the tapering values of that sector
```

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -52,21 +52,31 @@ class QubitConverter:
...
+                can be provided (a list of int of values -1, and 1). The default is None
+                meaning no symmetry reduction is done. Note that dipole and other operators
+                such as spin, num particles etc are also symmetry reduced according to the
```
dlasecki, line 71:
```suggestion
                such as spin, num particles etc. are also symmetry reduced according to the
```

qiskit_nature/problems/second_quantization/base_problem.py
```diff
@@ -73,6 +73,19 @@ def _transform(self, data):
             data = transformer.transform(data)
         return data
 
+    def symmetry_sector_locator(self, z2_symmetries: Z2Symmetries) -> Optional[List[int]]:
+        # pylint: disable=unused-argument
+        """Given the detected Z2Symmetries can determine the correct sector of the tapered
```
dlasecki, line 78:
```suggestion
        """Given the detected Z2Symmetries, it can determine the correct sector of the tapered
```

qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
```diff
@@ -136,3 +138,32 @@ def filter_criterion(self, eigenstate, eigenvalue, aux_values):
...
+
+    def symmetry_sector_locator(self, z2_symmetries: Z2Symmetries) -> Optional[List[int]]:
+        """Given the detected Z2Symmetries can determine the correct sector of the tapered
+        operators so the correct one can be returned
+
+        Args:
+            z2_symmetries: the z2 symmetries object.
+
+        Returns:
+            the sector of the tapered operators with the problem solution
```
dlasecki, line 150: Some docs have dots at the end and some do not. Would be good to make it consistent.

dlasecki (changes requested):
Just some typos in docs.

qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
mrossinek, line 150:
```suggestion
            The sector of the tapered operators with the problem solution.
```

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -52,26 +52,31 @@ class QubitConverter:
...
+                requiring 3 less qubits. Now only one of these operators will have the ground state
+                and be the correct symmetry sector needed for the ground state. Setting 'auto' will
+                use an automatic computation of the correct sector. If the sector is known
+              from other experiments with the z2symmetry logic, then the tapering values of that
+                sector can be provided (a list of int of values -1, and 1). The default is None
+                meaning no symmetry reduction is done. Note that dipole and other operators such as
+                spin, num particles etc. are also symmetry reduced according to the symmetries found
+                in the main operator if this operator commutes with the main operator symmetry. If
+                it does not then the operator will be discarded since no meaningful measurement can
+                take place.
```
woodsp-ibm, line 74: This is really around the convert_match from the perspective here that if you do symmetry reduction on `convert` then `convert_match` will do the same symmetry tapering/reduction as long as the operator commutes with the symmetry. If not then qubit operator that would normally be output (so at that position in the List when a List is supplied) will be set to None. Of course this applies to any operators such as dipoles etc that are passed by the algorithm into this convertor.

If we want to avoid updating this at present, due to another build cycle we can, but I think it would be better to describe it as what it does from this components perspective relating it to the overall behavior in the stack like I outlined above.

mrossinek: I don't fully understand what you want to say here. I admit just copying over this docstring from Aqua..
Are you just suggesting to clarify that additional operators are dealt with by `convert_match`?

woodsp-ibm: In Aqua or course this was all done as part of Transformation (formerly core.Hamiltonian) where that itself built out the dipoles etc in that same module. Here the QubitConvertor does nothing aside from what it is asked to do. It has no notion of dipoles or anything. Just some operator that it is asked to `convert`, an possibly taper (do symmetry reduction) and then other operators as passed later to `convert_match` where it will convert them the same way, but instead of finding symmetry it just matches the symmetry of the main operator, if you will, ie. it checks the operator commutes with symmetry. I think I am saying the text, and no you say the source, is more about the stack behavior when combined say with an ElectronicStructureProblem than about the QubitConvertor behavior. Thats what I tried to express above about convert and convert_match that I think should be the way things are described wrt putting None for non-commuting operators - now one give an example saying these operator could be dipoles etc if that seems helpful for a user.

woodsp-ibm (approved):

Issue #52, opened 2020-11-30
QEOM energies differ for Statevector and Qasm simulator


MariaSapova:
### Information

- **Qiskit Aqua version**: 0.8.1
- **Python version**: 3.6.8
- **Operating system**: Ubuntu

### What is the current behavior?

For H2 molecule QEOM excited state energies differ running with Statevector and Qasm simulators.

### Steps to reproduce the problem

Please, run the [gist](https://gist.github.com/MariaSapova/0912242cfb0b6e344d86367330f5e94d)

### What is the expected behavior?

Close excited state energies 

### Suggested solutions

Seems like the problem is somewhere in eval() method, but I'm not sure about that.

mrossinek:
I investigated this bug together with @Cryoris again today but we have not been able to track down the exact cause of it yet.. We opened a PR to fix a sub-issue of this but I am afraid a fix for this issue will not make it into the upcoming release.
I will revisit this again in more detail later.

Anthony-Gandon:
This issue seems to have been solved as a side effect of the new Qiskit release.
The problematic methods for this issue (`eval_op()` and `evaluate_operators()`) have been deprecated and replaced.

The new behaviour of the QEOM calculation does not show this difference between the noisy and state vector "backend" (this has also been replaced): see this [gist](https://gist.github.com/Anthony-Gandon/fe4c4c13acaafb2ad7df351c0ed8ac01)

mrossinek:
Thank you, Anthony! :+1:

PR #105, opened 2021-03-31
Fix #53

### LINKED ISSUE #53: GroundStateEigensolver evaluate_operators() works incorrectly with qasm
### Information

- **Qiskit Aqua version**: qiskit-master
- **Python version**: 0.3.7

I investigated my issue Qiskit/qiskit-nature#52, where matrix elements of Q matrix in qEOM were completely incorrect. It turns out that using evaluate_operators() from GroundStateEigensolver produces incorrect expectation values when running with qasm simulator.

See Qiskit/qiskit-nature#52.
Also there is a more simple case: after running VQE-UCCSD calculation I get optimal parameters for the circuit. If I use evaluate_operators() to get the energy on the optimized circuit, the value is different from the one which I got by VQE procedure.

Expectation value of the operator should not strongly depend on the simulator.

### Suggested solutions

I copied the code of construct_expectation() from VQE and it helped me to get results equivalent to the statevector ones.

mrossinek:
### Summary

Closes #53 

### Details and comments

When working with QuantumInstance-based solvers, we also need to account
for the Expectation-conversion of the expectation operator. Otherwise,
QASM-based calculations will not return the expected results when
evaluating additional operators.

@woodsp-ibm Is this an issue you would like to see fixed in Aqua before it gets deprecated?

### DIFF
--- a/qiskit_nature/algorithms/ground_state_solvers/ground_state_eigensolver.py
+++ b/qiskit_nature/algorithms/ground_state_solvers/ground_state_eigensolver.py
@@ -128,6 +128,8 @@ class GroundStateEigensolver(GroundStateSolver):
         """
         # try to get a QuantumInstance from the solver
         quantum_instance = getattr(self._solver, 'quantum_instance', None)
+        # and try to get an Expectation from the solver
+        expectation = getattr(self._solver, 'expectation', None)
 
         if not isinstance(state, StateFn):
             state = StateFn(state)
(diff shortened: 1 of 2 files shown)


woodsp-ibm:
Is this additional operators supplied in by an end user as different from the aux_ops we normally build out - I thought they worked ok as they are the dipoles etc and I had not seen any issues in regards of those and incorrect values - at least that I recall. But I do have some vague recollection of an issue with expectation and qasm, but I recall that being around QEOM?

mrossinek:
Yes this issue showed up first in #52. But the exact source of that error I have not been able to track down yet.
This, however, is a fix to a standalone problem which is easy enough to fix now.

dlasecki (approved):

pbark (approved):

PR #116, opened 2021-04-05
Fix documentation and init files

### LINKED ISSUE #29: Drivers should explicitly mention what values they allow for basis kwarg

There is no way to tell from the documentation what values are supported for the `basis` kwarg in the drivers.  Only `PyQuanteDriver` gives the values, the rest just say `basis (str) – Basis set`.

For reference see: https://qiskit.org/documentation/apidoc/qiskit.chemistry.drivers.html

woodsp-ibm:
Updating the documentation

- [X] Remove imports promoted up to parent of second quantization folders
- [X] Ensure all folders are included, that are necessary, in navigation
- [X] Ensure all classes are included in docs & init files
- [X] Update notebooks for QubitConverter import
- [X] Update text

Fixes #29

As always more could be done to improve docs and the class docstrings to better describe etc. But for now I think its  decent update that includes all the classes etc in the navigation structure with docs and init files reflect the same structure.


qiskit_nature/algorithms/excited_states_solvers/excited_states_eigensolver.py
```diff
@@ -18,10 +18,10 @@
 from qiskit.algorithms import Eigensolver
 from qiskit.opflow import PauliSumOp
 
+from qiskit_nature.converters.second_quantization import QubitConverter
 from qiskit_nature.operators.second_quantization import SecondQuantizedOp
-from qiskit_nature.operators.second_quantization.qubit_converter import QubitConverter
-from qiskit_nature.problems.second_quantization.base_problem import BaseProblem
-from qiskit_nature.results.eigenstate_result import EigenstateResult
+from qiskit_nature.problems.second_quantization import BaseProblem
```
mrossinek, line 23: Just noting this once: I see that you unified the import statements to always import clases from the module they are first exposed in. I do agree with this and am wondering whether we should make this an explicit convention. If we do want to do that, can we enforce this with some linter? However, I cannot think of a solution from the top of my head..

qiskit_nature/converters/second_quantization/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+SecondQuantizedOp Mappers (:mod:`qiskit_nature.converters.second_quantization`)
```
mrossinek, line 14:
```suggestion
SecondQuantizedOp Converters (:mod:`qiskit_nature.converters.second_quantization`)
```

qiskit_nature/converters/second_quantization/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+===============================================================================
+
+.. currentmodule:: qiskit_nature.converters.second_quantization
+
+The classes here are used to convert fermionic, vibrational and spin operators to qubit operators,
```
mrossinek, line 19: Why not simply `used to convert second quantized operators`?
You can add an example like `(e.g. fermionic operators)` but making this a list which appears to be exhaustive is not really a future-proof idea.

qiskit_nature/converters/second_quantization/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+using mappers and other techniques that can also reduce the problem such as leveraging
+Z2 Symmetries.
```
mrossinek, line 21:
```suggestion
using mappers and other techniques that can also reduce the problem size, for example by leveraging Z2 Symmetries.
```

qiskit_nature/mappers/second_quantization/__init__.py
```diff
@@ -16,6 +16,13 @@
 
 .. currentmodule:: qiskit_nature.mappers.second_quantization
 
+The classes here are used to convert fermionic, vibrational and spin operators to qubit operators.
```
mrossinek, line 19: Same as before: we should not hard-code this list of operator types.

qiskit_nature/operators/__init__.py
```diff
@@ -11,31 +11,16 @@
...
+==========================================
 
 .. currentmodule:: qiskit_nature.operators
 
-Operators
-==========
-
-Second-Quantization Operators
-+++++++++++++++++++++++++++++
+Operators for different systems such as fermionic, vibrational and spin.
```
mrossinek, line 19: Especially here I find this list a little bit confusing. If you do want to mention examples here, you should also make clear how they differ. But in fact I think it is not needed here at all.

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -24,13 +26,15 @@
...
+from qiskit_nature.operators.second_quantization import SecondQuantizedOp
 
 logger = logging.getLogger(__name__)
 
 
 class QubitConverter:
-    """A converter from Second-Quantized to Qubit Operators.
+    """ **DEPRECATED** A converter from Second-Quantized to Qubit Operators.
+
+    The QubitConventer was moved to `converters.second_quantization`
```
mrossinek, line 37: Do we need to say `qiskit_nature.converter.second_quantization`?

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -69,6 +73,11 @@ def __init__(self,
                 sector can be provided (a list of int of values -1, and 1). The default is None
                 meaning no symmetry reduction is done.
         """
+        warnings.warn('This QubitConverter is deprecated as of 0.1.0, '
```
mrossinek, line 76: Aren't we already at `0.2.0` right now? I think this should be either `0.2.1` or `0.3.0`, probably the latter.

qiskit_nature/operators/second_quantization/qubit_converter.py
```diff
@@ -69,6 +73,11 @@ def __init__(self,
...
+                      'and will be removed no earlier than 3 months after the release. '
```
mrossinek, line 77: Also, since this deprecation string always has the same general structure maybe we should think about extracting that into a wrapper method which takes only a few arguments (class name, version number, new location, etc.)

Just an FYI...

qiskit_nature/transformers/__init__.py
```diff
@@ -16,10 +16,16 @@
 
 .. currentmodule:: qiskit_nature.transformers
 
+Transformers act on a :class:`~qiskit_nature.drivers.QMolecule` to produce an altered copy of it
+as per the specific transformer. So for instance the :class:`FreezeCoreTransformer` will alter the
+integrals and number of particles in a way that freezes the core orbitals out, storing an extracted
```
mrossinek, line 21:
```suggestion
integrals and number of particles in a way that freezes the core orbitals, storing an extracted
```

Also do we want to assume that the reader knows what `core orbitals` are? Otherwise we can explicitly refer to `QMolecule.core_orbitals`.

mrossinek (commented):
This already looks quite good to me! Just left some minor comments :+1:

docs/tutorials/01_electronic_structure.ipynb
```diff
@@ -323,7 +306,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.10"
+   "version": "3.6.1"
```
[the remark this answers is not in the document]
woodsp-ibm: Well we still support 3.6 - I value diversity so we are not all running on 3.9!

qiskit_nature/algorithms/__init__.py
```diff
@@ -11,22 +11,17 @@
...
+============================================
 
 .. currentmodule:: qiskit_nature.algorithms
 
-Nature Algorithms
-=================
-These are algorithms configured and/or functioning using chemistry specific knowledge. See also
-:mod:`~qiskit.algorithms` for other algorithms in these categories which may also
-be used for chemistry problems such as :class:`~qiskit.algorithms.VQE`.
+These are nature science algorithms to solve specific problems such as find the ground state
```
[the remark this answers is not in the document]
woodsp-ibm: I'll fix it if it breaks!

qiskit_nature/__init__.py
```diff
@@ -16,111 +16,11 @@
 
 .. currentmodule:: qiskit_nature
 
-This is Qiskit Nature module that provides for experimentation with
-domain problems such as ground state energy and excited state energies of molecules.
+The Qiskit Nature module provides function to experiment with quantum computing for natural
+science problems, such as in chemistry and physics. For example computing the ground state energy
+or excited state energies of molecules.
```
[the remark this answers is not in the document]
woodsp-ibm: As I mentioned text could be improved overall. What was there was way out of date; I tried to come up with some brief, quick overview. We can add/improve things over time etc

qiskit_nature/algorithms/excited_states_solvers/excited_states_eigensolver.py
woodsp-ibm: Normally this is how we are importing elsewhere where we expose the class/method via the package init file. For sure this is how any tests should be done, since this is how we would normally expect users to do it - it corresponds to the API reference and how we document these.

qiskit_nature/operators/__init__.py
woodsp-ibm: I think we should have some text in these sections that gives a bit of an overview; not everyone that looks over the API ref is as knowledgeable on the topic. I put a single sentence just to describe what this has and I agree things could be improved. Maybe differences are not needed since they should say individually in the class what they are for etc and differences would be seen. But some aspects might be useful like, since fermions have spin why is that not covered by the spin op that it is special case, as in things maybe not so clear to some people.

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: This has never been included in the documentation so any text there would just be read when you view the source. I was hoping that since we only just released not too many people yet would be using it from where it was. We did in test cases and notebooks, but I changed all that to the new location. This was kept just in case any body started using it.

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: Master is at 0.2.0 but stable is at 0.1.0. The intent is that this is fixed asap (the QubitConvertor location to avoid cycles) in 0.1.1. As I mentioned in another comment above I have changed every where we used it so its just a question of did anyone start using this in the short time since 0.1.0 came out. If they did they will now find its deprecated. I could make it 0.1.1 but we would not normally deprecate on bug fix releases - of course having something deprecated from the initial release is weird too, but I am hoping its not noticed too much given the short timeframe.

qiskit_nature/operators/second_quantization/qubit_converter.py
woodsp-ibm: @manoelmarques I know Manoel had done something like that for Aqua before. Maybe it would be useful to have in the repos in general to facilitate things. But this can be added to master at any time.

qiskit_nature/transformers/__init__.py
woodsp-ibm: Some new to chemistry may not know - in general in this text its a question of what we need to explain to the set of users that would be using this. Hence in general improving the text to explain things, or at least link to (external) reference material that does, is useful.

pbark (approved):

mrossinek:
@woodsp-ibm There are still some comments left. Would you mind taking a look at them?

woodsp-ibm:
> There are still some comments left. Would you mind taking a look at them?

I saw there were some minor comments/changes to docstrings but I saw nothing incorrect that needed correction. Since there are the CI failures and other things I had been looking at these since the text/docstrings in general could be improved everywhere. If I missed something you feel is vital to change now let me know.

mrossinek:
> I saw there were some minor comments/changes to docstrings but I saw nothing incorrect that needed correction. Since there are the CI failures and other things I had been looking at those, as the text/docstrings in general could be improved everywhere. If I missed something you feel is vital to change now let me know.

No worries, I think it is fine! I just wanted to ping you cause you did not comment on my comments so I was not sure whether you saw them

mrossinek (approved):

PR #135, opened 2021-04-14
Code refactoring

dlasecki:
### Summary

Fixed certain things that SonarLint complained about. They are pretty minor improvements.

### Details and comments

Not sure if pulling out certain strings as constants is a good step... Perhaps we could name them better?

### DIFF
--- a/qiskit_nature/mappers/second_quantization/linear_mapper.py
+++ b/qiskit_nature/mappers/second_quantization/linear_mapper.py
@@ -46,17 +46,17 @@ class LinearMapper(SpinMapper):
 
                 if n_x > 0:
                     operator_on_spin_i.append(
-                        reduce(operator.matmul, [spinx for _ in range(int(n_x))])
+                        reduce(operator.matmul, [spinx] * int(n_x))
                     )
 
                 if n_y > 0:
                     operator_on_spin_i.append(
-                        reduce(operator.matmul, [spiny for _ in range(int(n_y))])
+                        reduce(operator.matmul, [spiny] * int(n_y))
                     )
 
                 if n_z > 0:
                     operator_on_spin_i.append(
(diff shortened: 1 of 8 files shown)


qiskit_nature/mappers/second_quantization/linear_mapper.py
```diff
@@ -46,17 +46,17 @@ def map(self, second_q_op: SpinOp) -> PauliSumOp:
 
                 if n_x > 0:
                     operator_on_spin_i.append(
-                        reduce(operator.matmul, [spinx for i in range(int(n_x))])
+                        reduce(operator.matmul, [spinx for _ in range(int(n_x))])
```
mrossinek, line 49: Isn't this just `[spinx] * int(n_x)`?
Same for the others..

mrossinek (changes requested):
As far as I know we may *not* change anything in the `gauopen` module. That is also the reason why it get's excluded from all linting jobs. Please undo those changes.

Other than that just a minor comment, thanks @dlasecki :+1:

qiskit_nature/mappers/second_quantization/linear_mapper.py
dlasecki: Yes, good catch.

manoelmarques:
Yes, the folder "gauopen" has source code from Gaussian that we don't touch. I even have it not checked by style, pylint, mypy etc.

mrossinek (approved):

manoelmarques:
I added the "backport" status because even though it is not a bug fix, having those changes in stable will reduce the possibility of merge conflicts in future backports.

pbark (approved):

woodsp-ibm:
This should NOT have been labelled backport. Backporting to stable branch is just for bug fixes deemed critical enough to be part of the next point release.

PR #142, opened 2021-04-15
Code refactoring

dlasecki:
### Summary
Introduces a little bit of structure to second quantized operators generated by problems. 

### Details and comments
It deals with a bad practice of addressing the main operator via an index. In the future we might want to think about introducing even more structure by subclassing for example, depending on the problem at hand.
Let me know if you have better ideas for naming.

### DIFF
--- a/qiskit_nature/problems/second_quantization/problem_second_quant_ops.py
+++ b/qiskit_nature/problems/second_quantization/problem_second_quant_ops.py
@@ -1,0 +1,53 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+"""Problem Second Quantized Operators"""
+from typing import List
+
+from qiskit_nature.operators.second_quantization import SecondQuantizedOp
+
+
+class ProblemSecondQuantOps():
+    """Problem Second Quantized Operators"""
+
+    def __init__(self, main_operator: SecondQuantizedOp,
+                 aux_operators_list: List[SecondQuantizedOp] = None):
+        """
+
+        Args:
+            main_operator: main problem-related second quantized operator.
+            aux_operators_list: a list of auxiliary problem-related second quantized operators.
+        """
+
+        self._main_operator = main_operator
+        self._aux_operators_list = aux_operators_list
+
+    def __len__(self):
+        length = 0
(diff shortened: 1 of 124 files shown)


woodsp-ibm (changes requested):
Since the code is now released changes to public interfaces need to be deprecated. So for example it seems to me for instance a new method would need to be added to replace the old one that returns a List whereby that one is deprecated and the user is referred to the new.

Since there is also discussion in changing VQE (ie min eigen solvers) - to say allow a dictionary instead of List - do we want to delay changing things until that is done since this seems to be related since ultimately these need to get passed to VQE?

qiskit_nature/problems/second_quantization/problem_second_quant_ops.py
```diff
@@ -0,0 +1,53 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+"""Problem Second Quantized Operators"""
+from typing import List
+
+from qiskit_nature.operators.second_quantization import SecondQuantizedOp
+
+
+class ProblemSecondQuantOps():
```
Cryoris, line 18: I think it would be nice to stay consistent with the other appearances of "quantized", i.e. since we have `SecondQuantizedOp` rather have `ProblemSecondQuantizedOps`?

qiskit_nature/problems/second_quantization/problem_second_quant_ops.py
```diff
@@ -0,0 +1,53 @@
...
+    """Problem Second Quantized Operators"""
+
+    def __init__(self, main_operator: SecondQuantizedOp,
+                 aux_operators_list: List[SecondQuantizedOp] = None):
```
Cryoris, line 22: This misses the typhint for `None` 🙂 
```suggestion
                 aux_operators_list: Optional[List[SecondQuantizedOp]] = None):
```

dlasecki:
> Since the code is now released changes to public interfaces need to be deprecated. So for example it seems to me for instance a new method would need to be added to replace the old one that returns a List whereby that one is deprecated and the user is referred to the new.
> 
> Since there is also discussion in changing VQE (ie min eigen solvers) - to say allow a dictionary instead of List - do we want to delay changing things until that is done since this seems to be related since ultimately these need to get passed to VQE?

Sure, let's wait and see how the situation develops. Good point about deprecations - I will rely on your suggestions there.

mrossinek:
The idea of the `ProblemSecondQuantizedOps` class will largely be implemented by the `Property` objects introduced in #220. In combination with the planned step of #243 this should mostly alleviate the need for this PR, right? @dlasecki

dlasecki:
> The idea of the `ProblemSecondQuantizedOps` class will largely be implemented by the `Property` objects introduced in #220. In combination with the planned step of #243 this should mostly alleviate the need for this PR, right? @dlasecki

Yes, I agree. Let's close this.

PR #145, opened 2021-04-20
Fix CI slow tests

### LINKED ISSUE #122: Fix the scheduled CI runs

The scheduled CI runs which execute the `slow_test`s has been failing for the past 10 days.
The failing tests are the following:
```
test.algorithms.ground_state_solvers.test_groundstate_eigensolver.TestGroundStateEigensolver.test_uccsd_hf_qasm
test.algorithms.ground_state_solvers.test_groundstate_eigensolver.TestGroundStateEigensolver.test_uccsd_hf_aer_qasm
```
Although the Aer test passes in some CI runs.

Nothing special to do. Just run the `slow_test`:
```
% export QISKIT_TESTS="run_slow"; python -m unittest test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
```

The tests should pass.

### Suggested solutions

We need to investigate where exactly the fault lies. It could just be that the seeds changed and the values need to be updated. But I think this may be unlikely.

woodsp-ibm:
Fixes #122 

This was simple change to seed to allow the slow CI tests too achieve the expected test values and pass.

Also, created a run_slow label that, if applied to PR, will cause all slow tests to run, besides running on nightly CI.

### DIFF
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -16,6 +16,9 @@ inputs:
   event-name:
     description: 'Actions event'
     required: true
+  run-slow:
+    description: 'Run slow tests or not'
+    required: true
   python-version:
     description: 'Python version'
     required: true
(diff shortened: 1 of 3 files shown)


manoelmarques (approved):

mrossinek (approved):
Thanks @woodsp-ibm for fixing the seed!
And thanks @manoelmarques for the CI label! That is very useful :+1:

mrossinek:
I overlooked that there were two slow tests which were failing on the scheduled CI runs.. I am not sure as to why it passed on the CI here though.. @woodsp-ibm could you please take another look? (https://github.com/Qiskit/qiskit-nature/runs/2405903539?check_suite_focus=true)

PR #146, opened 2021-04-21
Use simpler, more performant construction of array in qubit mapper

jlapeyre:
This constructs a numpy array directly instead of constructing and converting a list.
The construction is both simpler and more performant, although it is unlikely to make the enclosing function faster
in this case.

### DIFF
--- a/qiskit_nature/mappers/second_quantization/qubit_mapper.py
+++ b/qiskit_nature/mappers/second_quantization/qubit_mapper.py
@@ -96,7 +96,7 @@ class QubitMapper(ABC):
             return real_part + imag_part
 
         # 1. Initialize an operator list with the identity scaled by the `self.coeff`
-        all_false = np.asarray([False] * nmodes, dtype=bool)
+        all_false = np.zeros(nmodes, dtype=bool)
 
         ret_op_list = []
 
--- a/qiskit_nature/mappers/second_quantization/qubit_mapper.py
+++ b/qiskit_nature/mappers/second_quantization/qubit_mapper.py
@@ -96,7 +96,7 @@ class QubitMapper(ABC):
             return real_part + imag_part
 
         # 1. Initialize an operator list with the identity scaled by the `self.coeff`
-        all_false = np.zeros(nmodes, dtype=bool)
+        all_false = np.full(nmodes, False)
 
         ret_op_list = []
 


mrossinek (dismissed):
Thanks @jlapeyre!

woodsp-ibm:
One could argue that the original construct is a little easier to see that it's an array of False length of num_modes that is wanted. The latter construct does of course give the same but I'd argue it not quite as evident - having code readability is important too and if the construct is not really all that much better, for the size of systems we would be dealing with, where its not really any performance impact, we need to consider that too; Maybe in this case just a comment that went `# [False] * num_nodes` would be helpful

jlapeyre:
But, it's not an array of builtin `False`, it's an array `numpy._bool`. I'm not suggesting premature optimization at the expense of readability. I think it's important to adopt a habit of using a clean, efficient, interface provided by a library, unless there is a good reason to do otherwise.

woodsp-ibm:
I like the np.full better I think. Seeing np.zeros makes me think more numeric, then I have to take notice of the bool type which is why I thought some comment there might be helpful.

jlapeyre:
`full` is slower than `zeros`, but after investigating a bit, I am thinking it might be more readable/coherent. And it's ok since this is not a hot spot.

In many languages booleans are integers. If this were the case in Python, I might argue that `zeros` is correct. But, the Python type ~~system~~ situation is not coherent on this point.
The following all evaluate to `True`:

```python
import numpy as np
from numbers import Integral

isinstance(False, Integral)
not isinstance(np.bool_(False), Integral)
not isinstance(np.bool_(False), np.integer)
not isinstance(np.bool_(False), np.number)
isinstance(int64(1), np.integer)
```

So a Python `bool` is in some sense an integer. And numpy doesn't think `np.bool_` is any kind of number. But, `np.bool_` is still supported by a function that returns an array of "zeros", which is a bit confusing to me. I think `full` avoids this issue.

jlapeyre:
It should also be pointed out that `np.zeros` is used frequently in this context elsewhere in qiskit_nature. Eg, `np.zeros(nmodes, dtype=bool)`

mrossinek (approved):

woodsp-ibm (approved):
> It should also be pointed out that np.zeros is used frequently in this context

Oh well, one small step on the road to improving code readability in any case. Thanks John.

PR #160, opened 2021-04-29
Fix test case for gaussian driver

woodsp-ibm:
There are driver tests for other methods such as uhf, rohf which are on larger molecules. To keep tests small/fast these have been done on frozen core but the logic around that was changed. The base test case for all the driver method tests, used to set frozen core to True as default. It seems this logic was changed so that False is now in effect the default, since transformer(s) need to be passed to alter that. While the pyquante, pyscf and psi4 tests were all changed, presumably as these are run in CI, the gaussian one had not been and fails if run. This changes that test case to make it logically the same as before by passing a FreezeCoreTransformer, as indeed the other driver tests were changed to do.


mrossinek (approved):
Thanks for looking into this @woodsp-ibm! :+1:

Issue #129, opened 2021-04-13
General methods for second quantized operators


LaurinFischer:
### What is the expected enhancement?

There is some general functionality which would be nice to have for all types of second-quantized ops:

1. Checks for hermiticity of an operator (Is the operator equivalent to its adjoint)?
2.  A direct way to compute commutators, and checks for commutativity.
3. I think it would be nice to also have a `.to_matrix` method for the `FermionicOp` as is already in place for the `SpinOp`. 
4. For the `.to_matrix()` methods it would be nice to return a sparse datatype such as one of the `scipy.sparse` formats like `csc`. Maybe this could be its own method `.to_sparse_matrix()`. 

Points 1. and 2. should be straightforward to implement with the current arithmetics already in place.  

For 3. one needs to make a choice whether to use the full Fock space (with varying particle number) as a basis or only a certain particle-number (and fixed spin?) subspace. It might be nice to have this be specified by the user.

mrossinek:
@ikkoham Would you also be willing to look into these suggestions?

The sparse-matrix suggestions could be tied into a separate tasks of restructuring the operators to an internal sparse representation.

ikkoham:
Thank you for nice suggestion.

> 1. Checks for hermiticity of an operator (Is the operator equivalent to its adjoint)?

`SecondQuantizedOp.is_hermite() -> bool` is nice!

> 2. A direct way to compute commutators, and checks for commutativity.

We have two choices.
1. `SecondQuantizedOp.is_commutative(op: SecondQuantizedOp) -> bool`
2. `commutator(op1: SecondQuantizedOp, op2: SecondQuantizedOp).is_zero()`. I've made this in opflow.

> 3. I think it would be nice to also have a .to_matrix method for the FermionicOp as is already in place for the SpinOp.

The basis of`FermionicOp.to_matrix()` isn't trivial as you said. Which basis is convenient for chemistry users?

> 4. For the .to_matrix() methods it would be nice to return a sparse datatype such as one of the scipy.sparse formats like csc. Maybe this could be its own method .to_sparse_matrix().

Nice. See https://github.com/Qiskit/qiskit-nature/blob/master/qiskit_nature/operators/second_quantization/spin_op.py#L439. This parameter is same with quantum_info.Pauli.

LaurinFischer:
Thanks for the reply!

> The basis of `FermionicOp.to_matrix()` isn't trivial as you said. Which basis is convenient for chemistry users?

[...]

PR #161, opened 2021-04-30
Hermiticity checks for SecondQuantizedOps

LaurinFischer:
### Summary
Added method `is_hermitian() -> bool` for `SecondQuantizedOp`

### Details and comments
Also made `to_list()` an `abstractmethod` of `SecondQuantizedOp`.

Related to first point in #129


qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -54,3 +54,12 @@ def reduce(self, atol: Optional[float] = None, rtol: Optional[float] = None):
...
         raise NotImplementedError
+
+    @abstractmethod
+    def to_list(self) -> List[Tuple[str, complex]]:
+        """Returns the operators internal contents in list-format."""
+        raise NotImplementedError
+
+    def is_hermitian(self) -> bool:
+        """Checks whether the operator is hermitian"""
+        return frozenset(self.to_list()) == frozenset(self.adjoint().to_list())
```
woodsp-ibm, line 65: In creating a set, from a list, any duplicates in the list would get dropped of course. So as long as the adjoint operator has the exact same number terms, and nothing is simplified, the corresponding duplicates, would get dropped there too. Hence given the current code, I think this technique works. Maybe tests should have some more complex examples that would cover this aspect

LaurinFischer: Yes that is the idea behind this method as a workaround of not having an exact `__eq__` method for the `SecondQuantizedOps`. When there is a `to_matrix`-method in place for all operator types this could be used as a more precise (albeit expensive) method of comparison.

ikkoham: Applying `reduce()` before the equality check may be nice to remove duplicated terms.But as @LaurinFischer said, this is workaround until `to_matrix()`. It might be nice to merge this PR after other PRs including `to_matrix()`.

ikkoham:
Thanks @LaurinFischer. I think the direction is very nice. I have a minor comment. This PR changes the API, so the release note is required. https://github.com/Qiskit/qiskit-nature/blob/main/CONTRIBUTING.md#adding-a-new-release-note

qiskit_nature/operators/second_quantization/second_quantized_op.py
woodsp-ibm: to_matrix() though seems like an expensive solution to this. My comment was also to point out that if anyone change adjoint() such that the resultant operator was simplified then the list based test would fail I think - as such in the unit tests it would be good to have one that has duplicates that we might catch such a change (We have had problems with list/set based comparison in the past hence my commenting etc)

LaurinFischer: Thanks for the feedback. I agree with Steve and have made sure that the tests are performed on operators that do have duplicate terms in their `to_list()` such that if the implementations of `.adjoint()` were to change this should be caught.

woodsp-ibm: Nice, thanks. But since each subclass has their own adjoint implementation might it be prudent to do a unit test for each, i.e. include SpinOp and VibrationalOp to assure things work as expected.

LaurinFischer: As Ikko suggested above, I added a `reduce()` before the comparison to further proof against any duplicate terms being wrongfully discarded. Also tests now include all operator types, i.e. FermionicOp, SpinOp and VibrationalOp.

woodsp-ibm (approved):

mrossinek (approved):
Thanks @LaurinFischer!

pbark (approved):
This looks good! Thanks a lot @LaurinFischer !

PR #179, opened 2021-05-10
Move ``VQEProgram`` to Nature

Cryoris:
### Summary

Move the ``VQEProgram`` from the qiskit-runtime repository to qiskit-nature.

### Details and comments

We want to avoid another repository that needs to be pip-installed to be able to use the runtime programs. Thus, we're moving the VQE program to qiskit-nature.

The ``VQEProgram`` is a wrapper to the server-side VQE script and embeds it in our `MinimumEigensolver` interface.

### DIFF
--- a/qiskit_nature/runtime/vqe_program.py
+++ b/qiskit_nature/runtime/vqe_program.py
@@ -1,0 +1,367 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""The Qiskit Nature VQE Quantum Program."""
+
+
+from typing import List, Union, Callable, Optional, Any, Dict
+import numpy as np
+
+from qiskit import QuantumCircuit
+from qiskit.exceptions import QiskitError
+from qiskit.providers import Provider
+from qiskit.providers.backend import Backend
+from qiskit.algorithms import MinimumEigensolver, MinimumEigensolverResult, VQEResult
+from qiskit.opflow import OperatorBase, PauliSumOp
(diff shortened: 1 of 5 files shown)


test/runtime/fake_vqeruntime.py
```diff
@@ -0,0 +1,99 @@
...
+
+from typing import Dict, Any
+import numpy as np
+from qiskit.circuit import QuantumCircuit
+from qiskit.opflow import PauliSumOp
+from qiskit.providers import Provider
+from qiskit_nature.runtime import VQEProgramResult
+
+
+class FakeVQEJob:
```
pbark, line 23: @Cryoris maybe instead of `Fake` we can use `Test`?  It just seems a bit strange to me. But this is only minor.

pbark: ok I see that by Fake you mean something that mimics the VQEJob. SHould be fine then.

Cryoris: I was stealing the name from `FakeBackend` 😄 We could call it `Mock` if you prefer, but I think `Test` is only for classes that test something, whereas this is a "mock" or "fake" class that pretends to be a job.

qiskit_nature/runtime/vqe_program.py
```diff
@@ -0,0 +1,372 @@
...
+
+    def __init__(
+        self,
+        ansatz: QuantumCircuit,
+        optimizer: Optional[Dict[str, Any]] = None,
+        initial_point: Optional[np.ndarray] = None,
+        provider: Optional[Provider] = None,
+        backend: Optional[Backend] = None,
+        shots: int = 1024,
+        measurement_error_mitigation: bool = True,
```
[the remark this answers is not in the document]
Cryoris: Good point, I'll change it 👍🏻

test/runtime/fake_vqeruntime.py
pbark: I think is fine. I misinterpret it in the first pass, that's why I marked it as resolved ;)

pbark (dismissed):

manoelmarques (approved):

woodsp-ibm (approved):

manoelmarques (approved):

manoelmarques (approved):

woodsp-ibm (approved):

Issue #166, opened 2021-05-04
General Driver Updates


mrossinek:
### What is the expected enhancement?

#147 also suggest some general improvements to the drivers. The implementation of these is tracked by this issue:

- [x] relocate `qiskit_nature.drivers.*` into `qiskit_nature.drivers.second_quantization.*`
- [x] relocate `qiskit_nature.transformers.*` into `qiskit_nature.transformers.second_quantization.*`
- [x] check the "file"-based drivers for general improvements:
  - [x] `fcidumpd`
  - [x] `hdf5d`
- [x] check the "program"-based drivers for general improvements:
  - [x] `pyscfd`
  - [x] `pyquanted`
  - [x] `psi4d`
  - [x] `gaussiand`

While performing the checks for general improvments we should also check for the ability of the driver to perform more input validation and update the documentation of all methods.

mrossinek:
With #173 nearing its merging and #180 following suite shortly after, two major (and the most important) steps of this issue will be addressed soon. None of the other drivers are technically high priority right now because none of them immediately block other implementations planned in the near future.
Nonetheless, all driver codes should be checked in order to determine what kind of updates are possible and to estimate the time such updates would take on a case by case basis. Most drivers are likely to undergo very little changes because besides PySCF all but one other driver are not interfacing another Python-based code, limiting the code complexity which could be updated in a similar style to #180.

mrossinek:
As part of migrating the new drivers to the property framework (#264) I have checked each open driver for an estimated amount of time required to refactor it. I will summarize below:

| Driver | Estimated Time required |
|-|-|
| pyscf | Done and fully migrated after #264 |
| pyquante | very similar to PySCF, portable in 0.5-1 days |
| psi4 | relies on `hdf5` framework (see below) |
| gaussian (electronic) | relies on `gauopen` module, mostly Property integration required, (maybe some function reorganization), portable in 0.5-1 days |
| gaussian (vibrational + log) | relies on `GaussianLogResult`, portable in 0.5-1 days |
| fcidump | basically done after #264 |
| hdf5 | requires more discussion (#302) |

The migration of the remaining drivers will likely not be part of the `0.2.0` release.

mrossinek:
A quick update on this epic:
- [ ] `pyquante`: handled by #525 
- [ ] `psi4`: handled by #521 (blocked by #461)
- [ ] `gaussian`: prepared offline (blocked by #461 and #521 [latter due to minor general refactoring in that PR])
[...]

PR #180, opened 2021-05-10
PySCF Driver Updates

### LINKED ISSUE #28: pyscf driver does no checking on input arguments at instantiation
### Information

- **Qiskit Aqua version**: lastest
- **Python version**:
- **Operating system**:

for example:
```python
driver = PySCFDriver(molecule = 'happy holidays', unit=UnitsType.ANGSTROM, basis='happy holidays')
```
does not generate an error when the object is built.  The `unit` kwarg does look for an enum, but that is it.  This leads to hard to read errors down the line:

e.g. the above gives:

```
AttributeError: 'str' object has no attribute 'geometry'
```

or

```python
driver = PySCFDriver(molecule = molecule, unit=UnitsType.ANGSTROM, basis='happy holidays')
```

gives:

```
QiskitChemistryError: 'Failed electronic structure computation'
```

mrossinek:
### Summary

This PR continues the work #166 by adding general improvements to the `PySCFDriver`. The others drivers will need to be tackled one-by-one in a similar fashion.

### Details and comments

- [x] integration of `integrals.py` as class methods
- [x] improved splitting of class methods into more logical subunits
- [x] updated docstrings
- [x] closes #28:
  - explicit checking of all input arguments is _not_ feasible
  - unfortunately, PySCF also does not provide simple means to validate the arguments through some method call
  - BUT, through more logical method splitting, a more meaningful error will now be raised (e.g. [this one](https://github.com/mrossinek/qiskit-nature/blob/41b2945ea572e355aace5e877a135c4a538cf904/qiskit_nature/drivers/pyscfd/pyscfdriver.py#L371))
- [x] also re-evaluates #29 for this driver by updating all links to the new PySCF website (https://pyscf.org)
- [x] update unittests 

:warning: This PR as significant conflicts with #173!


mrossinek:
@woodsp-ibm While this is still a WIP, it would be great if you could already take a look at this and provide me some early feedback :+1:

test/drivers/pyscfd/test_driver_problem_pyscf.py
```diff
@@ -122,6 +123,7 @@ def test_lih_rhf_bk(self):
         )
         self._assert_energy_and_dipole(result, "lih")
 
+    @slow_test
```
woodsp-ibm, line 126: I am not sure these are all that slow. I would be reluctant not to have these in there running normally under CI.

qiskit_nature/drivers/pyscfd/pyscfdriver.py
```diff
@@ -53,47 +68,79 @@ def __init__(
...
+            charge: The charge of the molecule.
+            spin: The spin of the molecule. In accordance with PySCF's definition, the spin equals
+                :math:`2*S`, where :math:`S` is the total spin number of the molecule.
+            basis: A basis set name as recognized by PySCF (3_), e.g. `sto3g` (the default), `321g`,
+                etc. Note, that more advanced configuration options like a Dictionary or custom
+                basis sets are not allowed for the moment. Refer to 4_ for an extensive list of
+                PySCF's valid basis set names.
+            hf_method: The SCF method type to be used for the PySCF calculation. While the name
+                refers to HF methods, the PySCFDriver also supports KS methods. Refer to the
+                ``HFMethodType`` for a list of the supported methods.
```
woodsp-ibm, line 100: If the name needs this further explanation around hf would it be better to have this as a different name? Maybe simply just `method` that accommodates either and deprecate the `hf_method`; same for HFMethodType maybe it needs an enum specific to PySCF - that one supported  all the drivers and was used in PyQuante where these additional methods would be invalid.

qiskit_nature/drivers/pyscfd/pyscfdriver.py
```diff
@@ -125,49 +302,386 @@ def _check_valid():
...
+        """
+        try:
+            hf_method_name = self._hf_method.upper()
+            hf_method_cls = getattr(scf, hf_method_name)
+        except AttributeError as exc:
+            raise QiskitNatureError("Failed to load {} HF object.".format(hf_method_name)) from exc
+
+        self._calc = hf_method_cls(self._mol)
+
+        if "KS" in hf_method_name:
```
woodsp-ibm, line 425: Since we have enum here we should just be able to ask if the enum is one of the set of them that should be acted upon - that way we don't have some string "KS" as comparison that is we changed enum names, with set the refactoring would tale care of things.

qiskit_nature/drivers/pyscfd/pyscfdriver.py
```diff
@@ -125,49 +302,386 @@ def _check_valid():
...
+            "basis={}".format(self._basis),
+            "hf_method={}".format(self._hf_method),
             "conv_tol={}".format(self._conv_tol),
             "max_cycle={}".format(self._max_cycle),
             "init_guess={}".format(self._init_guess),
             "max_memory={}".format(self._max_memory),
-            "",
         ]
-        q_mol.origin_driver_config = "\n".join(cfg)
+        if "ks" in self._hf_method.lower():
```
woodsp-ibm, line 483: I guess this "ks" string detection is lower case versus the above I commented on where it was uppercase; again with an enum I think we should be able to leverage that as described before

qiskit_nature/drivers/pyscfd/pyscfdriver.py
```diff
@@ -125,49 +302,386 @@ def _check_valid():
...
+
+        This method is part of the public interface to allow the user to easily overwrite it in a
+        subclass to further tailor the behavior to some specific use case.
+
+        Raises:
+            QiskitNatureError: If an invalid HF method type was supplied.
+        """
+        try:
+            hf_method_name = self._hf_method.upper()
+            hf_method_cls = getattr(scf, hf_method_name)
```
woodsp-ibm, line 419: I guess this works - but it ends up more obscure for me and could do with explanation for any future readers if you choose to want to keep this. Certainly the switch/case (ifs from before) would grow larger here but to me its clearer/cleaner - now the enum names I guess have to match the method names.

test/drivers/pyscfd/test_driver_problem_pyscf.py
mrossinek: I do not see the benefit of the `OH` tests compared to the existing `H2` *and* `LiH` tests. The `OH` tests are, relative to the other ones, *very* slow. Hence, I disabled them. If you disagree strongly we can re-enable them. However, for the record, I don't think they are necessary.

qiskit_nature/drivers/pyscfd/pyscfdriver.py
mrossinek: I agree that the naming is not ideal. Given our plans for the `MoleculeDriver` I think we can tackle that then, because we need to re-evaluate how this Enum ties into the general case there, too.

qiskit_nature/drivers/pyscfd/pyscfdriver.py
mrossinek: The reason why I went with this comparison style is that I did not to duplicate half of the Enum as a set here. And this comparison properly reflects the nature of the check which we care about.

qiskit_nature/drivers/pyscfd/pyscfdriver.py
mrossinek: I can add some explanatory comment. I will not revert it to the `if-elif-else` switches.
I will wait with this update until we have resolved the other discussion points above.

test/drivers/pyscfd/test_driver_problem_pyscf.py
woodsp-ibm: The doublet OH was included since its not a closed shell molecule to start with unlike LiH. Its a bit bigger and hence the bit slower. It was to make sure we were doing things right with either type of molecule across the different methods - it was originally called test driver methods. How long does it take to run that you consider it slow? If anything I could support perhaps putting some of them oh_uhf/oh_rohf (same for lih_uhf/rohf/rhf) as slow where they test the same thing but with a different mapping to thin things down. Arguably if it works for one mapping it should be good for the others - if it does not its more likely the mapper is broke than the driver. Having said that I am not sure we have mapping tests with such coverage of open shell molecules elsewhere.

qiskit_nature/drivers/pyscfd/pyscfdriver.py
woodsp-ibm: I guess in my mind it makes the code 'brittle' in that refactoring an enum, should we ever want to change any name, this detection here would break. Listing the set and simply saying `if x in [A, B, C]` works and if we refactor C to D that code should get changed by the refactor whereas some check on "C" would not.

qiskit_nature/drivers/pyscfd/pyscfdriver.py
woodsp-ibm: Thats fine - if its documented - though personally I am not so keen on the direct couping of names in the enum - that currently has wider use in Nature, to names that match PySCF driver methods, though I understand RHF/ROHF and UHF are universal enough - maybe that's true of these KS methods too.

mrossinek:
I will try to tackle @woodsp-ibm comments later today or early tomorrow.
However, I strongly believe we should merge #173 first. I will ping that one again now

mrossinek:
I just rebased this branch after the various updates to the `PySCFDriver` and its relocation. I also addressed @woodsp-ibm's comments so hopefully this is now entirely ready for review :+1:

qiskit_nature/drivers/second_quantization/pyscfd/pyscfdriver.py
```diff
@@ -10,25 +10,39 @@
...
+    from pyscf import __version__ as pyscf_version
+    from pyscf import ao2mo, dft, gto, scf
+    from pyscf.lib import chkfile as lib_chkfile
+    from pyscf.lib import logger as pylogger
+    from pyscf.lib import param
+    from pyscf.tools import dump_mat
+
+    warnings.filterwarnings("ignore", category=DeprecationWarning, module="pyscf")
+except ImportError:
+    logger.info("PySCF is not installed. See https://sunqm.github.io/pyscf/install.html")
```
woodsp-ibm, line 44: The former code was fixed with an updated message for the changed install location as follows
```suggestion
    logger.info("PySCF is not installed. See https://pyscf.org/install.html")
```

test/drivers/second_quantization/gaussiand/test_driver_methods_gaussian.py
```diff
@@ -10,18 +10,18 @@
...
-from test.drivers.second_quantization.test_driver_methods_gsc import TestDriverMethods
+from test.drivers.second_quantization.test_driver_problem import TestDriverProblem
 from qiskit_nature.drivers.second_quantization import GaussianDriver
 from qiskit_nature import QiskitNatureError
 from qiskit_nature.transformers.second_quantization import FreezeCoreTransformer
 
 
-class TestDriverMethodsGaussian(TestDriverMethods):
-    """Driver Methods Gaussian tests"""
+class TestDriverProblemGaussian(TestDriverProblem):
```
woodsp-ibm, line 23: The original rationale why this was called TestDriverMethods was that it was really around tests for the different methods - rhf, rohf, uhf

qiskit_nature/drivers/second_quantization/pyscfd/pyscfdriver.py
mrossinek: Thanks! I must have missed this one. After fixing it grepping for `sunqm` no longer returns any results in the codebase :+1:

test/drivers/second_quantization/gaussiand/test_driver_methods_gaussian.py
mrossinek: I see. I renamed it because these tests are now all based on the `Problem` classes whereas the other drivers tests are more basic. I don't think it matters too much though.

woodsp-ibm: They were named based on _what_ they were testing not _how_ - i.e. it's not about the problem, that's just a means to an end to verify the function/result. If we could exact diagonalize a FermionicOp to get a min eig value it could be done that way given what is tested perhaps - before we used to use core Hamiltonian and Numpy - using Problem and GroundStateSolver is just the latest way.

mrossinek:
@pbark @woodsp-ibm A fresh review on this PR would also be great :upside_down_face:

test/drivers/second_quantization/gaussiand/test_driver_methods_gaussian.py
mrossinek: After an offline-sync and re-evaluating my view on these tests I have decided to revert this naming change. Initially I was thinking differently about this but I agree that this is a change which does not add any value. I will revert this later today.

mrossinek:
All comments (here and offline) should be resolved now, @woodsp-ibm @pbark

woodsp-ibm (approved):

pbark (approved):
Thanks @mrossinek! LGTM!

PR #187, opened 2021-05-12
Pin Sphinx version below 4.0.0

mrossinek:
### Summary

Sphinx 4 introduced some new warnings which cause our CI to fail.
For now, this PR pins the version below 4.

I will open an issue to track the progress of resolving these warnings shortly.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -35,7 +35,7 @@ jobs:
       - uses: ./.github/actions/install-nature
       - name: Install Dependencies
         run: |
-          pip install jupyter 'sphinx<4' nbsphinx sphinx_rtd_theme 'matplotlib<3.3.0' qiskit-terra[visualization]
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-docs.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - uses: ./.github/actions/install-nature
       - name: Install Dependencies
         run: |
-          pip install jupyter sphinx nbsphinx sphinx_rtd_theme 'matplotlib<3.3.0' qiskit-terra[visualization]
+          pip install jupyter 'sphinx<4' nbsphinx sphinx_rtd_theme 'matplotlib<3.3.0' qiskit-terra[visualization]
```
mrossinek, line 38: @manoelmarques I wanted to get this fix out quickly but maybe we should think about if we can remove these hardcoded packages in favor of using `requirements-dev.txt` or similar?

manoelmarques: Actually in the CI we don't even need to install sphinx in here because at that point it has already been installed from requirements-dev.txt.

manoelmarques (approved):

PR #199, opened 2021-05-20
Fix print of ElectronicStructureResult

### LINKED ISSUE #198: `ElectronicStructureResult` throws error when printed and dipole moments not summed properly.
### Information

- **Qiskit Nature version**: `main`
- **Python version**: 3.7.10
- **Operating system**: MacOS

The `ElectronicStructureResult` throws an error when attempting to print it.

Problems:
1) Division of a `None` by a number.
2) Total dipole moment displays `None` if any elements of dipole moments are `None`. 

Run the following snippet:

```python
from qiskit_nature.drivers import PySCFDriver
from qiskit_nature.transformers.freeze_core_transformer import FreezeCoreTransformer
from qiskit_nature.algorithms.ground_state_solvers.minimum_eigensolver_factories import NumPyMinimumEigensolverFactory
from qiskit_nature.algorithms.ground_state_solvers.ground_state_eigensolver import GroundStateEigensolver

driver = PySCFDriver(atom="Li 0 0 0; H 0 0 1.6", basis="sto3g")
problem = ElectronicStructureProblem(driver, [FreezeCoreTransformer()])
qubit_converter = QubitConverter(
    ParityMapper(), two_qubit_reduction=True, z2symmetry_reduction="auto",
)

solver = NumPyMinimumEigensolverFactory()
gsc = GroundStateEigensolver(qubit_converter, solver)

result = gsc.solve(problem)

print(result)
```
You will observe the following error:

```TypeError: unsupported operand type(s) for /: 'NoneType' and 'float'```

The `ElectronicStructureResult`  should be printed and the total dipole moments should be summed correctly. 
For instance:
```
=== GROUND STATE ENERGY ===
 
* Electronic ground state energy (Hartree): -8.90847269193
  - computed part:      -1.089782396349
  - FreezeCoreTransformer extracted energy part: -7.818690295581
~ Nuclear repulsion energy (Hartree): 1.025934879643
> Total ground state energy (Hartree): -7.882537812287
 
=== MEASURED OBSERVABLES ===
 
  0:  # Particles: 2.000 S: 0.000 S^2: 0.000 M: 0.000
 
=== DIPOLE MOMENTS ===
 
~ Nuclear dipole moment (a.u.): [0.0  0.0  2.92416221]
 
  0: 
  * Electronic dipole moment (a.u.): [None  None  4.74515931]
    - computed part:      [None  None  4.74910617]
    - FreezeCoreTransformer extracted energy part: [0.0  0.0  -0.00394686]
  > Dipole moment (a.u.): [None  None  -1.8209971]  Total: 1.8209971
                 (debye): [None  None  -4.62851252]  Total: 4.62851252
```

### Suggested solutions

Replace the respective methods  in `qiskit_nature.results.electronic_structure.py`by these and you will obtain the result above:

``` @property
    def total_dipole_moment(self) -> Optional[List[float]]:
        """ Returns total dipole of moment """
        if self.dipole_moment is None:
            return None  # No dipole at all
        dms = [dip for dip in self.dipole_moment[0] if dip is not None]
        return [np.sqrt(np.sum(np.power(dms, 2)))] 

    @property
    def total_dipole_moment_in_debye(self) -> Optional[List[float]]:
        """ Returns total dipole of moment in Debye """
        tdm = self.total_dipole_moment
        return [dip / QMolecule.DEBYE if dip is not None else None for dip in tdm]```

### PR DESCRIPTION
**Pull request by Brogis1**
### Summary

This PR fixes #198.

### Details and comments
There were problems printing the `ElectronicStructureResult `and the total dipole moment was not right.
I am not sure if we have tests for the results classes. 

@dlasecki  @pbark What do you think?

### CODE REVIEW

File: qiskit_nature/results/electronic_structure_result.py [hunk 1]
```diff
@@ -135,24 +135,17 @@ def reverse_dipole_sign(self, value: bool) -> None:
...
         if self.dipole_moment is None:
             return None  # No dipole at all
-        tdm: List[float] = []
-        for dip in self.dipole_moment:
-            if np.any(np.equal(list(dip), None)):
-                tdm.append(None)  # One or more components in the dipole is None
-            else:
-                tdm.append(np.sqrt(np.sum(np.power(list(dip), 2))))
-        return tdm
+        dms = [dip for dip in self.dipole_moment[0] if dip is not None]
```
mrossinek, line 141: What will happen here if `dip is None`? I think in that case the following line may be a problem

qiskit_nature/results/electronic_structure_result.py
```diff
@@ -135,24 +135,17 @@ def reverse_dipole_sign(self, value: bool) -> None:
 
     @property
     def total_dipole_moment(self) -> Optional[List[float]]:
-        """Returns total dipole of moment"""
+        """ Returns total dipole of moment """
```
mrossinek, line 138: Please run `make black` locally to fix the formatting of this file. This should also fix the CI

mrossinek (changes requested):
Just a minor question.
However, please also add a unittest which checks that this bug is fixed and a release note (for that, you need to run `reno new <some descriptive name without spaces>` and then edit the file whose path will be printed by that command).

qiskit_nature/results/electronic_structure_result.py
Brogis1: Thanks Max!

qiskit_nature/results/electronic_structure_result.py
Brogis1: Sure, I think it is solved now.

qiskit_nature/results/electronic_structure_result.py
```diff
@@ -136,23 +136,17 @@ def reverse_dipole_sign(self, value: bool) -> None:
...
-        if self.dipole_moment is None:
-            return None  # No dipole at all
-        tdm: List[float] = []
-        for dip in self.dipole_moment:
-            if np.any(np.equal(list(dip), None)):
-                tdm.append(None)  # One or more components in the dipole is None
-            else:
-                tdm.append(np.sqrt(np.sum(np.power(list(dip), 2))))
-        return tdm
+        if np.all(np.equal(self.dipole_moment[0], None)):
```
woodsp-ibm, line 139: There are 3 different conditions to take into account here.
1. No dipole moments for X,Y,Z were computed at all
2. Dipole was computed and we have X,Y and Z values returned which can be summed for a total
3. Z2 Symmetry reduction was done, with one or more of the X,Y or Z dipole operators failing to commute with symmetry hence None would be returned for that moment(s), i.e letting caller know its unknown since it could not be validly measured. In this case we have some dipole moments but not enough to compute a total. That logic is reflected in the original code as far as I can see. I do not believe we should report a total when we cannot measure some of them since it can be incorrect which is why we would in the past we would not have shown it

test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
```diff
@@ -417,6 +417,16 @@ def test_freeze_core_z2_symmetry_compatibility(self):
...
 
+    def test_total_dipole(self):
+        """Regression test against #198.
+
+        An issue with calculating the dipole moment that had division None/float.
+        """
+        solver = NumPyMinimumEigensolverFactory()
+        calc = GroundStateEigensolver(self.qubit_converter, solver)
+        res = calc.solve(self.electronic_structure_problem)
+        self.assertAlmostEqual(res.total_dipole_moment_in_debye[0], 0.0, places=1)
```
mrossinek, line 428: I was rather thinking of a test for `print(result)`. While this test is also good to have, explicitly calling `print(result)` should also be tested as this is how you found this bug.
Simply calling `print(result)` should never break anything so calling that explicitly will ensure that we never break that in any way (whether it is dipole-related or not).

qiskit_nature/results/electronic_structure_result.py
Brogis1: Thanks, ok so replacing the np.all() by to np.any() should do the trick. 
Sorry, it is confusing having a None for total dipole, when there is one of its dimensions computed.
Assumed it was another problem.
Perhaps a text-like warning for the users would be good.

woodsp-ibm: I would need to look more carefully - I am not sure that caters to when there is no dipole computed at all, maybe that is done outside the method since we should not be trying to compute a total at all in that case - e.g when using the PyQuante driver we have no dipole moment ints and hence we never attempt any dipole computation since we cannot create any dipole operators ( the other cases are when we do that we need to take care of Z2Symmetry, that is done on the main operator, that we can still validly do it on the dipole moments - if not None is the result to indicate the failure). 
How to represent things then in text format that is readable for the end user is a choice. None means you need to understand what that means - but then that block was all formatted so anything longer might be harder to do - if None is acceptable as an individual moment maybe ?, or Unknown, Cannot compute, or something else can be written for a total. Maybe as you suggest we could instead leave None in all places and if None shows up then simply add some text to explain the meaning in the form of say a footnote under the dipole output.

mrossinek:
@Brogis1 I finally had a moment of time to add the print-unittest I suggested. I also reverted a part which @woodsp-ibm was not too sure about.

Brogis1:
Excellent, thanks for this.

qiskit_nature/results/electronic_structure_result.py
mrossinek: @woodsp-ibm I am not sure if you noticed but I reverted this questionable change.

woodsp-ibm (approved):

mrossinek (approved):

PR #201, opened 2021-05-21
Replace internal data in FermionicOp

ikkoham:
### Summary

This is another idea for #197.


mrossinek (commented):
I think this looks to be a promising direction! Thanks @ikkoham 

Some general thoughts:
- we tend to be verbose with names. So I would replace `_sub_label` with `_substituted_label` (because otherwise I tend to think of a *partial* label). Also maybe `_to_sp_label -> _to_sparse_label`
- what is the plan for `_LegacyFermionicOp`? Is this something you want to keep around for testing purposes?
- How does this PR affect (or how is it affected by) the ideas outlined in #129?

ikkoham:
@mrossinek Thank you for your quick response.

> we tend to be verbose with names. So I would replace _sub_label with _substituted_label (because otherwise I tend to think of a partial label). Also maybe _to_sp_label -> _to_sparse_label

OK

> what is the plan for _LegacyFermionicOp? Is this something you want to keep around for testing purposes?

I will remove it in the future. (I think this blocks other issues, so this should be merged asap.)

> How does this PR affect (or how is it affected by) the ideas outlined in #129?

As long as #129 uses the public API, I don't think this PR will affect it.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -249,157 +231,112 @@ def mul(self, other: complex) -> "FermionicOp":
...
 
+        Args:
+            sparse: The labels in the return list are sparse or not. (Default: False)
+
         Returns:
             A list of tuples consisting of the dense label and corresponding coefficient.
         """
-        return list(zip(self._labels, self._coeffs.tolist()))
+        if sparse:
+            return self._data[:]
```
mrossinek, line 282:
```suggestion
            return self._data
```

mrossinek (commented):
Alright, except for this minor change I think this looks good now. Thanks @ikkoham!
Next, we need to check how/if the rest of the stack is affected by the new output format of `op.to_list(sparse=True)`.

I was also thinking, that once we remove the `_LegacyFermionicOp`, it may make sense to provide a "pretty-printing" method for the user. A normal printing of the operator would return the sparse label representation, but we could provide another method to `reduce()` things first and print a dense label representation. While this does loose some information, it can arguably be easier to read. But this is an idea for another PR..

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: This `[:]` means copy of list in Python. I think this copy is necessary because of immutability.

mrossinek: Ah I did not know that. A comment would be useful there in order to avoid future confusion.
Alternatively, why not call `.copy()` directly?

ikkoham:
@mrossinek Thank you for your review.

Currently, the methods of private classes are not available from outside (user side). All the outputs except `to_list(sprse=True)` are in dense. There may be room to change this part.
Since there is no way to pass options to `__str__`, it might be a good idea to use other methods like `pretty_print` instead. Alternatively, we could create a property `label_mode` to control the output.

mrossinek:
> Since there is no way to pass options to __str__, it might be a good idea to use other methods like pretty_print instead.
Yes, that is what I meant above as well. But we can leave that for a future PR for the sake of getting this one done.

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Both are often used, and there is no deep meaning to my choice. `.copy()` is equivalent to `[:]`. https://docs.python.org/3/tutorial/datastructures.html#more-on-lists
 I will change `.copy()` here.

ikkoham, line 282:
```suggestion
            return self._data.copy()
```

ikkoham:
I have deprecated to use `to_list()`'s option sparse. Alternatively I have made setter for the label mode. Code example can be seen from the test code.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -364,42 +275,83 @@ def to_list(self) -> List[Tuple[str, complex]]:
...
+        Raises:
+            ValueError: invalid mode is given
+        """
+        mode_lower = mode.lower()
+        if mode_lower == "dense":
+            cls._label_display_mode_is_dense = True
+        elif mode_lower == "sparse":
+            cls._label_display_mode_is_dense = False
+        else:
+            raise ValueError(f"Invalid `mode` {mode} is given. `mode` must be dense or sparse.")
```
mrossinek, line 318:
```suggestion
            raise ValueError(f"Invalid `mode` {mode} is given. `mode` must be 'dense' or 'sparse'.")
```
Just to make it that tiny bit more explicit what is meant here.

mrossinek:
Other then the minor comment above this looks good to me! Thanks, @ikkoham :slightly_smiling_face:

mrossinek (dismissed):
LGTM! Thanks, @ikkoham 
Let's wait for @pbark and/or @woodsp-ibm to also take a look..

mrossinek (approved):

pbark (approved):
Thanks @ikkoham ! LGTM!

pbark:
@woodsp-ibm Do you have any comment here or should we go ahead and merge it?

woodsp-ibm (approved):
When the discussing removing the Legacy operator in the future, presumably we are talking about ahead of the next release. The version of FermionicOp as it was, is of course in the history here but we want to merge a copy of it for more direct access for now for it seems - I guess that's fine if its really needed

PR #203, opened 2021-05-24
Fixing the 404 Error on the page about qiskit_nature.drivers.pyscfd.html

WiFisunset:
Fixed a link error on the following page (Note: Line 220's link on the original page leads to a 404 Error): https://qiskit.org/documentation/nature/apidocs/qiskit_nature.drivers.pyscfd.html#pyscf-installation

### Summary
Fixed a link error on the following page (Note: Line 220's link on the original page leads to a 404 Error): https://qiskit.org/documentation/nature/apidocs/qiskit_nature.drivers.pyscfd.html#pyscf-installation

404-error

### Details and comments
The file in the github is an rst file. Unsure where to PR the html file, which is why I'm linking one to this rst file

### DIFF
--- a/docs/apidocs/qiskit_nature.drivers.pyscfd.html
+++ b/docs/apidocs/qiskit_nature.drivers.pyscfd.html
@@ -1,0 +1,376 @@
+
+
+
+<!DOCTYPE html>
+<!--[if IE 8]><html class="no-js lt-ie9" lang="en" > <![endif]-->
+<!--[if gt IE 8]><!--> <html class="no-js" lang="en" > <!--<![endif]-->
+<head>
+  <meta charset="utf-8">
+  
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  
+  <title>PYSCF Installation &mdash; Qiskit Nature 0.1.3 documentation</title>
+  
+
+  
+  
+  
+  
+
+  
+
+  
+  
+    
+
+  
+
+  <link rel="stylesheet" href="../_static/css/theme.css" type="text/css" />
+  <!-- <link rel="stylesheet" href="../_static/pygments.css" type="text/css" /> -->
+  <link rel="stylesheet" href="../_static/jupyter-sphinx.css" type="text/css" />
+  <link rel="stylesheet" href="../_static/thebelab.css" type="text/css" />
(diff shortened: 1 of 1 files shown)


WiFisunset:
It also would be good to note that for PySCF, it can't be installed on Windows (it can on an ubuntu subsystem on Windows 10 though).

Native Windows is not Supported: https://github.com/pyscf/pyscf/issues/750
pyscf-not-native-to-windows-10


I'll expand the page to mention this, and provide links as to how to set up WSL: 
https://docs.microsoft.com/en-us/windows/wsl/install-win10

woodsp-ibm:
Hi, for for noticing and bringing this to our attention. It seems that PySCF has a new documentation location and the former one, that we had linked to, is no longer valid.

Now the API documentation you see - the html files - is built from the source code docstsings (and a few .rst files) so fixing this requires fixing source code. The location that would need to be fixed is here, if you would like to do so, https://github.com/Qiskit/qiskit-nature/blob/472cd73f2e424402af87c2ab80f14dc39bf27188/qiskit_nature/drivers/pyscfd/__init__.py#L20-L23

and from what I can see the new location is https://pyscf.org/install.html so the line should look more like 

> According to the `PySCF installation instructions <https://pyscf.org/install.html>`__,

To see the html that is built from the source you would follow step 2 here https://github.com/Qiskit/qiskit-nature/blob/main/CONTRIBUTING.md#pull-request-checklist (basically run `make html`) and the output will be in docs/_build/html folder where you will see an index.html file that is built. Now to build the docs you would need some of the dev tools we use installed, such as sphinx. The simplest way is to cd into the qiskit-nature project folder you created and in the root you will see a requirements-dev.txt file, that lists the packages we use for development such as building html, doing lint etc. You can install from that file directly using `pip install -r requirements-dev.txt`.

In regards of statements about which platforms were supported that is covered by the PySCF installation page and we did not want to replicate information which could change in terms of what was supported or not, since its detailed there and our docs could easily get out of date if things were changed.

woodsp-ibm (changes requested):
The html file needs to be removed and the file fixed from which the html is built - see above detail

WiFisunset:
> The html file needs to be removed and the file fixed from which the html is built - see above detail

Thanks so much for your reply. I was able to get things working on WSL Ubuntu, ran pip install -r requirements-dev.txt, and used make html to see the output. 

I'll close out this pull request, since I've opened resolve #204.

PR #208, opened 2021-05-26
particle numbers and magnetization, wrong order in the doc

Jian2017:

### DIFF
--- a/qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
+++ b/qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
@@ -62,9 +62,9 @@ class ElectronicStructureProblem(BaseProblem):
         provided.
 
         Returns:
-            A list of `SecondQuantizedOp` in the following order: electronic operator,
-            total magnetization operator, total angular momentum operator, total particle number
-            operator, and (if available) x, y, z dipole operators.
+            A list of `SecondQuantizedOp` in the following order: Hamiltonian operator,
+            total particle number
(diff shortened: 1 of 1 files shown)


qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
```diff
@@ -62,9 +62,8 @@ def second_q_ops(self) -> List[SecondQuantizedOp]:
         provided.
 
         Returns:
-            A list of `SecondQuantizedOp` in the following order: electronic operator,
-            total magnetization operator, total angular momentum operator, total particle number
-            operator, and (if available) x, y, z dipole operators.
+            A list of `SecondQuantizedOp` in the following order: Hamiltonian operator,
+            total particle number operator, total angular momentum operator, total magnetization operator, and (if available) x, y, z dipole operators.
```
woodsp-ibm, line 66: Thanks for spotting this and making the correction. Since we have a line length limit on source code this line seems too long and I think it will fail CI testing and will need to be split so we end up with 3 lines more like the original.

woodsp-ibm, line 66:
```suggestion
            total particle number operator, total angular momentum operator, total magnetization
            operator, and (if available) x, y, z dipole operators.
```

woodsp-ibm: Since it failed I made a suggestion below and committed it - hopefully it will pass CI now.

Jian2017:
LGTM

woodsp-ibm (approved):
Thanks again for correcting this

Issue #117, opened 2021-04-05
Migrate `EvolvedOperatorAnsatz` and `StarAlgebraMixin` to Terra


woodsp-ibm:
The evolved_operator_ansatz  is a generic ansatz based on evolution of opflow operators. As such it would make more sense to have this in Terra, as part of the core circuit library, as it has potential for broader use

https://github.com/Qiskit/qiskit-nature/blob/master/qiskit_nature/circuit/library/ansatzes/evolved_operator_ansatz.py

Update: I noticed now the StarAlgebraMixin which I think is also intended to be moved across so I added this here.

https://github.com/Qiskit/qiskit-nature/blob/master/qiskit_nature/operators/second_quantization/star_algebra.py

mrossinek:
I agree that this sounds like a good idea!

@ikkoham I remember you had brought this up for `StarAlgebraMixin` back when you started working on this. Could you take care of this?

@Cryoris Shall we do this in a single PR or would you like to migrate `EvolvedOperatorAnsatz` separately?

Cryoris:
I think 2 might be better since these aren't really related and it keeps the history cleaner. I can take care of the `EvolvedOperatorAnsatz`.

ikkoham:
For `StarAlgebraMixin`, we need to dicsuss whether we should adopt [qargs](https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/quantum_info/operators/mixins/group.py#L116) or [permutation](https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/opflow/operator_base.py#L580).

mrossinek:
Based on what did we go with `permutation` before?

If the reason for the above is not (or no longer) valid/applicable, I agree that we should adopt the current way of how things are done in Terra :+1:

mrossinek:
Thanks @Cryoris for getting started on this!

@ikkoham could you tackle the `StarAlgebraMixin` within the next week, too?

ikkoham:
got it.

mrossinek:
Migration complete :tada: 

Now we need to deprecate the migrated classes here in Terra :+1:

mrossinek:
Since @ikkoham's part is merged now, I leave @Cryoris assigned to this issue. We had some pointers in #210 with regards to the `EvolvedOpAnsatz` but it appears as though it needs some extra care.
Can you take a look at this before the end of the month?

PR #210, opened 2021-05-26
Deprecate StarAlgebraMixin and dagger

ikkoham:
### Summary

Update internal API.

Related to: https://github.com/Qiskit/qiskit-nature/issues/117

I also deprecated the dagger:  https://github.com/Qiskit/qiskit-terra/pull/6423#discussion_r638887961


mrossinek (commented):
Thanks @ikkoham! Can you also include `EvolvedOpAnsatz` or do you want to handle that in a separate PR?

qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -63,3 +63,14 @@ def to_list(self) -> List[Tuple[str, complex]]:
     def is_hermitian(self) -> bool:
         """Checks whether the operator is hermitian"""
         return frozenset(self.reduce().to_list()) == frozenset(self.adjoint().reduce().to_list())
+
+    @property  # type: ignore
+    # pylint: disable=bad-docstring-quotes
+    @deprecate_function(
+        "Using the `dagger` property is deprecated as of version 0.2.0 and will be removed no "
+        "eariler than 3 months after the release date. As an alternative, use the `adjoint()` "
```
woodsp-ibm, line 71:
```suggestion
        "earlier than 3 months after the release date. As an alternative, use the `adjoint()` "
```

test/operators/second_quantization/test_fermionic_op.py
```diff
@@ -234,14 +234,14 @@ def test_pow(self):
...
-        """Test adjoint method and dagger property"""
+        """Test adjoint method"""
         with self.subTest("adjoint"):
             fer_op = ~FermionicOp([("+N", 3), ("N-", 1), ("--", 2 + 4j)])
             targ = FermionicOp([("-N", 3), ("N+", 1), ("++", (-2 + 4j))])
             self.assertFermionEqual(fer_op, targ)
 
-        with self.subTest("dagger"):
-            fer_op = FermionicOp([("+-", 1), ("II", 2j)]).dagger
+        with self.subTest("adojoint2"):
```
woodsp-ibm, line 243: Its just a test name of course  maybe you meant `adjoint2` without the o after ad?

woodsp-ibm:
Since this change requires what is in Terra main, i,e, version 0.18.0 when its released, I thin we should update the requirements.txt file to change this `qiskit-terra>=0.17.0` to `qiskit-terra>=0.18.0`

qiskit_nature/operators/second_quantization/star_algebra.py
```diff
@@ -1,115 +0,0 @@
-# This code is part of Qiskit.
```
woodsp-ibm, line 1: I imagine we delete this since we consider it an internal file even though it can be used by an end user if they build their own operator or whatever. The alternative would be to leave it and issue a deprecation message telling anyone that they should instead use the one in Terra.

qiskit_nature/operators/second_quantization/second_quantized_op.py
ikkoham: Thanks.

test/operators/second_quantization/test_fermionic_op.py
ikkoham: Sorry, my typo. I'll fix it.

ikkoham, line 243:
```suggestion
        with self.subTest("adjoint 2"):
```

ikkoham:
@mrossinek Thanks.
> Can you also include EvolvedOpAnsatz or do you want to handle that in a separate PR?

I think it is impossible in the current stage.
Simple replacement of import path does not work.
This is the issue https://github.com/Qiskit/qiskit-terra/issues/6425.

@woodsp-ibm Thank you for your review. I've updated.

mrossinek:
> I think it is impossible in the current stage.

Hm that is odd. @Cryoris can you check this, then? Probably in another PR in this case..

qiskit_nature/operators/second_quantization/star_algebra.py
```diff
@@ -43,9 +44,11 @@ def mul(self, other: complex):
         return NotImplementedError
 
     def __mul__(self, other: complex):
+        _warn()
```
mrossinek, line 47: Shouldn't it be enough to do this once in the `__init__()` method?

ikkoham: `StarAlgebra` does not have `__init__()`. First, I consider to use `__init_subclass__`, but this is dirty hack, so I decided to make as this.

ikkoham:
Note about EvolvedOpAnsatz: I said it doesn't work, but to be precise, it doesn't pass the test. I think the problem is that the order of the parameters has changed, or the bounds have changed, so the convergence value has changed.

qiskit_nature/operators/second_quantization/star_algebra.py
mrossinek: Well you could add an `__init__()` which throws the warning though, right?

ikkoham: No, it doesn't make sense because the subclass doesn't always call it.

mrossinek (approved):

pbark (approved):

Issue #167, opened 2021-05-04
Implement modular `Property` classes (a containerized `QMolecule`)


mrossinek:
### What is the expected behavior?

As proposed in #148 I suggest an implementation of a `Property` object which "modularizes" the contents of the `QMolecule` class. As such, these properties would centralize a number of data-specific logic currently spreading into the stack (most notably the builders, integral calculators and result interpreters present in the `problems` module).

The initial implementation of such containers can be mostly decoupled from the rest of the stack and the remaining ongoing work if we first implement them as an additional layer between the `tranformers` and the `problem`.

For more information, please refer to the Epic #148.

mrossinek:
Closed by #220.

Issue #148, opened 2021-04-21
QMolecule Design 


pbark:
Aligned with the drivers refactoring we need to converge to a new design for the QMolecule

mrossinek:
With the updated description of this issue above, I propose to convert this issue into an Epic after everyone had time to review it. I will create separate implementation-oriented issues for the three mentioned subtasks which can then be assigned to sprints in order to gradually close this (then) Epic.

I would like to stress the benefit of the proposed step-by-step implementation again, because it will allow us to start working on a prototype for the containers while working on #147 in parallel. If we get towards the end of #147 before reaching the third proposed subtask of this issue we will also have gained a better overview of the drivers themselves which will help in the discussion of step 3 once we get to it.

mrossinek:
As part of #173 we discussed the possibility to further split `qiskit_nature.drivers.second_quantization` into:
```
qiskit_nature/
    drivers/
        second_quantization/
            electronic/
                <...>
            vibrational/
                <...>
```
As such, renaming `QMolecule -> ElectronicDriverResult` and `WatsonHamiltonian -> VibrationalDriverResult` would come very naturally. However, we decided to postpone this until we have a clearer picture of how the implementation of this issue will affect the drivers. Thus, this should be revisited as part of **Step 3.** in the original issue description.

PR #220, opened 2021-06-01
Introduction of Property objects

mrossinek:
### Summary

This is a very early WIP for #167.

### Details and comments

As proposed in the aforementioned issue as well as the Epic #148, we want to introduce some modular `Property` objects, in order to containerize the `QMolecule` and make it fit better into the rest of the updated stack. This PR will make the first step into this direction. Below is a list of the changes:

- [x] Add `Property` objects for:
  - [x] `ElectronicEnergy`
  - [x] `ParticleNumber`
  - [x] `AngularMomentum`
  - [x] `Magnetization`
  - [x] `DipoleMoment`
  - [x] whatever comes out of the `WatsonHamiltonian`
- [x] generalize the electronic integrals to n-body terms
- [x] provide the means for arbitrary (_almost_) orbital basis transformation (at least the addition is of new bases is straight forward via the `Basis(Enum)`)
- ~leverage @jlapeyre's 2-body integral symmetry utilities as part of `_2BodyElectronicIntegrals`~ (EDIT: postponed)
- [x] tie the properties into the `StructureProblem`s:
  - [x] electronic (partially done, not very nice atm)
  - [x] vibrational
- [x] add unittests
- [x] add documentation

The tying into the `Problem` classes will be improved in the future. The details of that implementation depend on steps outlined in #148. The same goes for the `interpret()` method stubs. I have them there right now because I think they could fit well, but they will only become useful if/once we reach the last implementation step.

:warning: The API of this PR is by no means final. Public/private methods are bound to change. So are the names of all classes and methods.


mrossinek:
I started working on the `WatsonHamiltonian` and `Bosonic/HarmonicBasis` equivalents for this framework. As its not really fitting together yet I have not pushed these yet but I will update it here tomorrow. In the meantime hopefully the CI passes now.

mrossinek:
More WIP commits. Things I am not yet satisfied with/sure about which need to be addressed as part of this PR:
- [ ] the `Basis` handling
- [ ] the fact that `DipoleMoment` contains all three axes without subcomponents

Things that will likely be addressed in the future:
- [ ] some properties are mostly copied from the previously existing integral calculators. However, in some cases the label could be constructed directly in a simpler fashion
- [ ] the `interpret` method stubs will only be useful if we proceed with steps 2 and 3 of #148 
- [ ] if we do decide to go there, we probably need an overall container of these properties. This is what `ElectronicDriverResult` and `VibrationalDriverResult` may become

mrossinek:
Okay I think this is almost done. I am still missing some more unittests for the property objects themselves but I am too exhausted to finish these today. I will mark this as ready for review and it would be great if @woodsp-ibm, @pbark and @paulineollitrault can have a look at this while I am out-of-office next week :slightly_smiling_face:

Note, that I left some TODO markers for the followup steps of this PR as outlined in #148

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -355,3 +355,27 @@ def _to_sparse_label(label):
         return " ".join(
             filter(None, (label_transformation[c].format(i=i) for i, c in enumerate(label)))
         )
+
+    @classmethod
```
ikkoham, line 359: Why do we need these class methods? (same comment for `VibrationalSpinOp`)

dlasecki: It looks like they could be static but `@classmethod `forces them to still be part of the class and helps to avoid passing unnecessary `self`.

test/properties/electronic/integrals/test_two_body_electronic_integrals.py
```diff
@@ -0,0 +1,455 @@
...
+                            [0.0, 0.0, 0.0, 0.0],
+                            [0.0, 0.0, 0.0, 0.0],
+                        ],
+                        [
+                            [-2.0, -3.0, 0.0, 0.0],
+                            [-6.0, -7.0, 0.0, 0.0],
+                            [0.0, 0.0, 0.0, 0.0],
+                            [0.0, 0.0, 0.0, 0.0],
+                        ],
+                        [
```
dlasecki, line 139: Shall we store these big matrices in separate files?

test/properties/electronic/integrals/test_two_body_electronic_integrals.py
```diff
@@ -0,0 +1,455 @@
...
+                    ("-+IN", -6.5),
+                    ("ININ", 7.5),
+                    ("NI+-", 1),
+                    ("-+NI", -4),
+                    ("-++-", -5),
+                    ("+-NI", 2),
+                    ("+-+-", 3),
+                    ("INNI", 6),
+                    ("IN+-", 7),
+                    ("IINN", -5),
```
dlasecki, line 387: Separate file?

test/properties/electronic/integrals/test_two_body_electronic_integrals.py
```diff
@@ -0,0 +1,455 @@
...
+                    ("NNII", 1.5),
+                    ("NNII", -2.5),
+                    ("-+NI", -8.5),
+                    ("INNI", 9.5),
+                    ("-+-+", 12.5),
+                    ("IN-+", -13.5),
+                    ("-++-", -10.5),
+                    ("IN+-", 11.5),
+                    ("-+IN", -14.5),
+                    ("ININ", 15.5),
```
dlasecki, line 431: Separate file?

test/properties/vibrational/bases/test_harmonic_basis.py
```diff
@@ -0,0 +1,500 @@
...
+
+from qiskit_nature.properties.vibrational.bases import HarmonicBasis
+from qiskit_nature.properties.vibrational.integrals import VibrationalIntegrals
+
+
+@ddt
+class TestHarmonicBasis(QiskitNatureTestCase):
+    """Test HarmonicBasis."""
+
+    # TODO: extract data
```
dlasecki, line 28: Separate file?

test/properties/electronic/integrals/test_two_body_electronic_integrals.py
```diff
@@ -0,0 +1,455 @@
...
+                        [
+                            [0.0, -1.0, 0.0, 0.0],
+                            [-4.0, -5.0, 0.0, 0.0],
+                            [0.0, 0.0, 0.0, 0.0],
+                            [0.0, 0.0, 0.0, 0.0],
+                        ],
+                        [
+                            [-2.0, -3.0, 0.0, 0.0],
+                            [-6.0, -7.0, 0.0, 0.0],
+                            [0.0, 0.0, 0.0, 0.0],
```
dlasecki, line 251: Separate file?

dlasecki (changes requested):

qiskit_nature/operators/second_quantization/fermionic_op.py
mrossinek: The addition of these was motivated by https://github.com/Qiskit/qiskit-nature/pull/220/commits/850267550d686cd49813aa432d491609dda08c7c
That commit showed that allowing an empty operator initialization is not straight forward. Having a `zero` and `one` operator available on the class itself is pure convenience (which in the future we may use in other places, too)

test/properties/electronic/integrals/test_two_body_electronic_integrals.py
mrossinek: Yes, hence the `TODO` label in these places. I did not have the time to do this before leaving.

mrossinek:
Thanks, @dlasecki, for the review! I will address your comments later this week.
@woodsp-ibm @pbark did either of you have time to take a look at this PR while I was out of office?

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Personally, I don't like having multiple ways to do one thing, but I'll leave it to you. And, opflow is not planning to support these methods, so it will be inconsistent, but it might be a convenient method for second quantized op. 
Finally, this initialization can be applied as follows:
```python
FermionicOp(("", 0), register_length=register_length)
FermionicOp((f"I_{register_length - 1}", 0))
0 * FermionicOp(f"I_{register_length - 1}")
```

jlapeyre: I support having the concept and methods for zero and one. For creating and also predicates. I think initializing a zero and one is not straightforward; neither programmatically, nor conceptually. What exactly should count as zero?  I was struggling today with how to construct a zero `SparsePauliOp`. I sort of gave up and thought about opening an issue asking about it, and suggesting methods for zero and one.

qiskit_nature/properties/electronic/dipole_moment.py
```diff
@@ -0,0 +1,187 @@
+# This code is part of Qiskit.
```
pbark, line 1: I think we should use `electronic_structure` instead of `electronic` properties? (also in the directories)

qiskit_nature/properties/electronic/electronic_energy.py
```diff
@@ -0,0 +1,120 @@
...
+from .bases import ElectronicBasis
+from .integrals import (
+    ElectronicIntegrals,
+    OneBodyElectronicIntegrals,
+    TwoBodyElectronicIntegrals,
+)
+from ..property import Property
+
+
+class ElectronicEnergy(Property):
```
pbark, line 31: I proposed before the change of electronic to electronic_structure, but for this property I would keep it as ElectronicEnergy.

qiskit_nature/properties/electronic/particle_number.py
```diff
@@ -0,0 +1,110 @@
...
+
+        Returns:
+            An instance of this property.
+
+        Raises:
+            QiskitNatureError: if a WatsonHamiltonian is provided.
+        """
+        if isinstance(result, WatsonHamiltonian):
+            raise QiskitNatureError(
+                "You cannot construct an AngularMomentum from a WatsonHamiltonian. Please provide "
```
pbark, line 81: `AngularMomentum` should be `ParticleNumber` here

qiskit_nature/properties/electronic/magnetization.py
```diff
@@ -0,0 +1,81 @@
...
+
+        Returns:
+            An instance of this property.
+
+        Raises:
+            QiskitNatureError: if a WatsonHamiltonian is provided.
+        """
+        if isinstance(result, WatsonHamiltonian):
+            raise QiskitNatureError(
+                "You cannot construct an AngularMomentum from a WatsonHamiltonian. Please provide "
```
pbark, line 52: `AngularMomentum` -> `Magnetization`

qiskit_nature/properties/vibrational/__init__.py
```diff
@@ -0,0 +1,56 @@
+# This code is part of Qiskit.
```
pbark, line 1: `vibrational` properties -> `vibrational_structure` properties (same as in the electronic structure case)

pbark (changes requested):

test/properties/electronic/integrals/test_two_body_electronic_integrals.py
dlasecki: Ok, I only noticed the label in the last file.

qiskit_nature/properties/electronic/electronic_energy.py
mrossinek: Leaving this open for information purposes. Nothing to be resolved here, unless someone disagrees.

qiskit_nature/properties/property.py
```diff
@@ -0,0 +1,99 @@
...
+    def _validate_input_type(
+        cls,
+        result: Union[QMolecule, WatsonHamiltonian],
+        valid_type: Type[Union[QMolecule, WatsonHamiltonian]],
+    ) -> None:
+        if not isinstance(result, valid_type):
+            raise QiskitNatureError(
+                f"You cannot construct an {cls.__name__} from a {result.__class__.__name__}. "
+                f"Please provide a {valid_type.__name__} object instead."
+            )
```
mrossinek, line 81: Okay so the errors of the CI are actually related to this method. I tried to extract the `isinstance` checks into the `Property` baseclass by doing the above. Essentially, each property can now call something like the following: `cls._validate_input_type(result, QMolecule)` at the beginning of their `from_driver_result` method. This method will check the type of `result` against the provided `valid_type` and raise an error if these don't match.

This is now a problem due to our current duplicate `QMolecule` situation in `qiskit_nature.drivers` (deprecated) and `qiskit_nature.drivers.second_quantization`. Since the properties all use the new location `QMolecule` object, all `isinstance` checks will fail when given an instance of the deprecated object. This happens in the unittests for `tests/drivers/pyscfd` (which test the deprecated code). This also happens for the _stable_ tutorials for the same reason.

Now, I am not entirely sure what the best approach is to resolve this issue. @manoelmarques @woodsp-ibm do you have some suggestions?

mrossinek: I guess one approach would be to also import the deprecated `QMolecule` and hard-code a relation between it and it's new version into the `isinstance` check. Something like:
```
elif valid_type == QMolecule and not isinstance(result, LegacyQMolecule):
    # also raise an error
```
The same would have to be done for `WatsonHamiltonian`.

The problem with this approach is that it will only work as long as both `QMolecule` objects are actually still compatible. If this is subject to change before the release, the `from_driver_result` implementations would have to deal with that (maybe we would need a `from_legacy_driver_result`). This can happen (e.g.) if we replace the new `QMolecule` with an `ElectronicDriverResult` (which may not need such a method if it already consists of `Property` objects). Then, the `from_driver_result` method would essentially become a wrapper for legacy objects anyways.

mrossinek:
I've added some basic unittests and resolved the previous type validation issues with a temporary workaround which I will be able to resolve when starting to work on the second step of #148 for which I am opening a separate issue in a moment.

I consider this PR now ready for the final review stage.

qiskit_nature/properties/electronic_structure/dipole_moment.py
```diff
@@ -0,0 +1,160 @@
...
+                    "x", (qmol.x_dip_mo_ints, qmol.x_dip_mo_ints_b), qmol.x_dip_energy_shift
+                ),
+                "y": dipole_along_axis(
+                    "y", (qmol.y_dip_mo_ints, qmol.y_dip_mo_ints_b), qmol.y_dip_energy_shift
+                ),
+                "z": dipole_along_axis(
+                    "z", (qmol.z_dip_mo_ints, qmol.z_dip_mo_ints_b), qmol.z_dip_energy_shift
+                ),
+            },
+            dipole_shift={
```
woodsp-ibm, line 141: I guess I was not expecting nuclear part in here as a 'shift'. If this is total dipole moment then given we compute the electronic part it needs this to make the total. 
So now I am going to comment on the name ElectronicEnergy and not it has similar with nuclear as a shift. But why then is this called ElectronicEnergy and not say just Energy - i.e although we only compute the electronic part the dipole that does not appear in its name.

qiskit_nature/properties/electronic_structure/integrals/electronic_integrals.py
```diff
@@ -0,0 +1,141 @@
...
+            basis: the basis which these integrals are stored in. If this is initialized with
+                ``ElectronicBasis.SO``, these integrals will be used *ad verbatim* during the
+                mapping to a ``SecondQuantizedOp``.
+            matrices: the matrices (one or many) storing the actual electronic integrals. If this is
+                a single matrix, ``basis`` must be set to ``ElectronicBasis.SO``. Refer to the
+                documentation of the specific ``n-body`` integral types for the requirements in case
+                of multiple matrices.
+        """
+        self._basis = basis
+        assert num_body_terms >= 1
```
woodsp-ibm, line 56: I think we should be raising ValueError here or TypeError here as these are public APIs and users can pass whatever. Asserts are more for internal state checks to assure what we expect there meets whatever the condition is for debugging etc

qiskit_nature/properties/electronic_structure/integrals/two_body_electronic_integrals.py
```diff
@@ -0,0 +1,131 @@
...
+from ..bases import ElectronicBasis, ElectronicBasisTransform
+
+
+class TwoBodyElectronicIntegrals(ElectronicIntegrals):
+    """The 2-body electronic integrals."""
+
+    EINSUM_AO_TO_MO = "pqrs,pi,qj,rk,sl->ijkl"
+    EINSUM_CHEM_TO_PHYS = "ijkl->ljik"
+
+    ERI_TRUNCATION_LEVEL = 1e-12
```
woodsp-ibm, line 31: It might be useful to have quantity as a param on the constructor and default it to this value. That way if anyone wants they can build this out with a different threshold for truncation.

qiskit_nature/properties/__init__.py
```diff
@@ -0,0 +1,43 @@
+# This code is part of Qiskit.
```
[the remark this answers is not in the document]
mrossinek: The idea which I was following was that the properties are "quantization-agnostic" if you will. I.e. the same kind of properties should be usable in second- and first-quantization problems (once we have those). This will obviously require a `first_q_ops` addition to the interface (and possibly require different raw storage format) but I think in terms of usability this will simplify things for the user.

qiskit_nature/properties/electronic_structure/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+being conserved.
+
+Finally, if the driver which you used provided dipole moment integrals, this property will also be
+evaluated:
+
+.. autosummary::
+   :toctree: ../stubs/
+   :nosignatures:
+
+   TotalDipoleMoment
```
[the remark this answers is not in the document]
mrossinek: Is this just an FYI which we need to look out for in the future or do you actually suggest a change here?

qiskit_nature/properties/electronic_structure/angular_momentum.py
```diff
@@ -10,32 +10,80 @@
...
+        y_h1, y_h2 = _calc_s_y_squared_ints(self._num_spin_orbitals)
+        z_h1, z_h2 = _calc_s_z_squared_ints(self._num_spin_orbitals)
+        h_1 = x_h1 + y_h1 + z_h1
+        h_2 = x_h2 + y_h2 + z_h2
+
+        h1_ints = OneBodyElectronicIntegrals(ElectronicBasis.SO, h_1)
+        h2_ints = TwoBodyElectronicIntegrals(ElectronicBasis.SO, h_2)
+        return [(h1_ints.to_second_q_op() + h2_ints.to_second_q_op()).reduce()]
+
+    def interpret(self, result: EigenstateResult) -> None:
```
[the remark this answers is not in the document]
mrossinek: Since the implementation in the future will likely look similar to this:
```python
class ElectronicStructureProblem:
    def interpret(self, result: EigenstateResult) -> ElectronicStructureResult:
        new_result = ElectronicStructureResult(result)
        for prop in self.properties:
            prop.interpret(new_result)
```
I was thinking that these `interpret` methods add meaning to the `result`-object **in-place**. If this is unwanted we can change this when addressing #243.

qiskit_nature/properties/electronic_structure/bases/electronic_basis.py
```diff
@@ -0,0 +1,29 @@
...
+
+    This ``Enum`` simply names the available electronic bases. The ``SO`` basis is the _special_
+    basis into which an ``ElectronicEnergy`` must map its integrals before being able to perform the
+    mapping to a ``qiskit_nature.operators.second_quantization.SecondQuantizedOp``.
+    """
+
+    # pylint: disable=invalid-name
+    AO = "atomic"
+    MO = "molecular"
+    SO = "spin"
```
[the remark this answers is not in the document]
mrossinek: Yes, so far I did not see the need for interleaved matrices. This plays only a role in the `SO` basis so if we need it we can add a method to switch between the two formats rather easily.

qiskit_nature/properties/electronic_structure/dipole_moment.py
mrossinek: The `nuclear` shifts were added in view of the future change to include the `interpret` methods. This field is not being used now and we can discuss how to handle this when dealing with #243

qiskit_nature/properties/__init__.py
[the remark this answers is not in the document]
woodsp-ibm: I don;t know - I guess it depends on how these are exposed to the stack the user is dealing with and how much of them they ever use. Given drivers before are quantization dependent and the problem they will be used to solve - its hard for me to see what the simplification truly becomes for the user.

qiskit_nature/properties/electronic_structure/__init__.py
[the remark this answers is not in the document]
woodsp-ibm: I think I am saying that TotalDiploleMoment may not be calculable in those circumstances. Should this be just DipoleMoment with x,y and z components and a total field; where total would not exist if one (or more) did not commute. Or maybe even ElectronicDipoleMoment given my comment about the ElectronicEnergy property name.

qiskit_nature/properties/electronic_structure/angular_momentum.py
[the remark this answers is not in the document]
mrossinek: I removed the `interpret` method stubs from this PR

qiskit_nature/properties/__init__.py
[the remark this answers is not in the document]
mrossinek: I think we have to options here:

#### qiskit_nature.properties
In this case we do **not** add submodules for different problem types and instead expose different operator converters on each property:
```
Property.second_q_ops()
Property.first_q_ops()
Property.sampling_q_ops()
etc...
```
If a property does not support a specific kind of problem, we can raise an error which should gracefully be handled by the calling frame.

The benefit would be a single property implementation with the potential to avoid code duplication and the ability to reuse the same internal raw data for different kind of problems.
However, the latter might also be a disadvantage if the raw formats for different problems for the same property are too diverse. In that case, we may be better off with the following option.

#### qiskit_nature.properties.{second_quantization,first_quantization,sampling,...}
This mimics the rest of stack layout but may lead to code duplication. Arguably, we should then have something like:
```
SecondQuantizedProperty.operators()
FirstQuantizedProperty.operators()
...
```

For now, I would prefer to go with option one. I think only time will tell whether we really need the second case.

qiskit_nature/properties/electronic_structure/__init__.py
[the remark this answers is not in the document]
mrossinek: I see :thinking: Could we postpone this discussion until the tackling of #243? This will happen before the release anyways and seeing how this pans out in combination with `interpret` will help, I think..

qiskit_nature/properties/__init__.py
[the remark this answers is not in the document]
woodsp-ibm: I just wonder as this ends up giving two different patterns in regards of quantization. A property behaving as a Union with different methods for different quantizations, but other aspects being split into different types e.g. ElectronicStructureProblem has a `second_q_ops` - on the same basis one might expect it could have been similar and have a `first_q_ops` as well.

Consistency is also an aspect that helps people with the overall code - properties this way will be different than all aspects of the stack around it. Maybe a Union is simpler in that I only have one class to go to - but now offset that against it being different in the way quantization is dealt with.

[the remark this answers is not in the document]
mrossinek: I have relocated the code as per your suggestion. I agree that the consistency argument outweighs the others :+1:

woodsp-ibm (dismissed):

qiskit_nature/properties/second_quantization/vibrational/occupied_modals.py
```diff
@@ -0,0 +1,98 @@
...
+        return cls()
+
+    def second_q_ops(self) -> List[VibrationalOp]:
+        """Returns a list of operators each evaluating the occupied modal on a mode."""
+        num_modals_per_mode = self.basis._num_modals_per_mode
+        num_modes = len(num_modals_per_mode)
+
+        ops = []
+        for mode in range(num_modes):
+            ops.append(self._get_mode_op(mode))
```
dlasecki, line 78:
```suggestion
        ops = [self._get_mode_op(mode) for mode in range(num_modes)]
```

test/properties/second_quantization/electronic/integrals/test_two_body_electronic_integrals.py
```diff
@@ -0,0 +1,187 @@
...
+            ints_mo = ints_ao.transform_basis(transform_beta)
+            assert np.allclose(ints_mo._matrices[0], 16 * mat_aa)
+            assert np.allclose(ints_mo._matrices[1], 36 * mat_ba)
+            assert np.allclose(ints_mo._matrices[2], 81 * mat_bb)
+            assert np.allclose(ints_mo._matrices[3], 36 * mat_ba.T)
+
+        with self.subTest("Final basis match"):
+            ints_ao = TwoBodyElectronicIntegrals(ElectronicBasis.MO, (mat_aa, None, None, None))
+            ints_mo = ints_ao.transform_basis(transform)
+            assert ints_ao == ints_mo
```
dlasecki, line 114:
```suggestion
            self.assertEqual(ints_ao, ints_mo)
```
Applies to other places.

dlasecki (changes requested):

pbark (dismissed):

qiskit_nature/properties/second_quantization/electronic/electronic_energy.py
```diff
@@ -0,0 +1,141 @@
...
+from ..second_quantized_property import (
+    DriverResult,
+    ElectronicDriverResult,
+    SecondQuantizedProperty,
+)
+
+
+# TODO: extract into separate file?
+class _IntegralProperty(SecondQuantizedProperty):
+    """TODO."""
```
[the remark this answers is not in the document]
mrossinek: Thanks, I overlooked this one

qiskit_nature/properties/second_quantization/electronic/integrals/one_body_electronic_integrals.py
```diff
@@ -0,0 +1,99 @@
...
+        In this case of the 1-body integrals, the returned matrix is a block matrix of the form:
+        ``[[alpha_spin, zeros], [zeros, beta_spin]]``.
+
+        Returns:
+            A single matrix containing the ``n-body`` integrals in the spin orbital basis.
+        """
+        if self._basis == ElectronicBasis.SO:
+            return self._matrices  # type: ignore
+
+        matrix_a = self._matrices[0]
```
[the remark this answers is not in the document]
mrossinek: Not really. This tuple of matrices gets stored in the base class `ElectronicIntegrals`. It's length depends on which subclass we are dealing with (length is 2 in `OneBodyElectronicIntegrals` but 4 in `TwoBodyElectronicIntegrals`). Introducing a special type (e.g. a `NamedTuple`) would require registering this type in the base class which would need to be updated when we decided to add e.g. `ThreeBodyElectronicIntegrals`.

Given that the length and type of the `matrices` is asserted at construction time (I extracted the validation methods as you suggested above) I think this is okay as is.

dlasecki (approved):

pbark (approved):

PR #222, opened 2021-06-03
Update AdaptVQE to new, stateless VQE

Cryoris:
### Summary

Fix the Adapt VQE after Qiskit/qiskit-terra#6418 is merged.

**Note:** This can only be merged after the above PR in Terra is merged.

### Details and comments

AdaptVQE previously relied on an internal private method which has been deleted without deprecation (since it was private).


woodsp-ibm:
Since this change requires what is in Terra main, i,e, version 0.18.0 when its released, I think we should update the requirements.txt file to change this qiskit-terra>=0.17.0 to qiskit-terra>=0.18.0. This way when we backport it, it will have the right requirements on Terra.

mrossinek:
The CI will continue to fail until this is merged because 5 hours ago the related PR in Terra got merged :+1:

woodsp-ibm:
@Cryoris This is failing with some issue in vqe from the BopesSampler tutorial. Can you please take a look.

manoelmarques:
The error in the tutorial is due to a Terra decorator deprecate_function that is not wrapping properties correctly and returns a decorator wrapper instead of the property. If I remove the deprecate_decorator from:
```
@deprecate_function(
        """
The VQE.optimal_params property is deprecated as of Qiskit Terra 0.18.0
and will be removed no sooner than 3 months after the releasedate.
This information is part of the returned result object and can be
queried as VQEResult.optimal_point."""
    )
    @property
    def optimal_params(self) -> np.ndarray:
```

It works. If I keep the decorator, instead of returning an array, the property returns: 
```
<bound method deprecate_function.<locals>.decorator.<locals>.wrapper of <qiskit.algorithms.minimum_eigen_solvers.vqe.VQE object at 0x7ff8a2980b50>>
```
The Terra properties that use this decorator should revert to issue the warnings themselves.

woodsp-ibm:
I guess the BopesSampler should be fixed to avoid the deprecated method/property that it tripped over as described above. Since this is holding up CI maybe thats best done via another issue/PR.

mrossinek (approved):
@woodsp-ibm what issue did you see? The CI is passing, right?

mrossinek:
> @woodsp-ibm what issue did you see? The CI is passing, right?

Ah I see. Because this fix is merged now: https://github.com/Qiskit/qiskit-terra/pull/6599

pbark (approved):
LGTM

Cryoris:
I think we might still need to add the requirement Terra >= 0.18 before merging 🙂

mrossinek:
> I think we might still need to add the requirement Terra >= 0.18 before merging

This already happened or do we need another requirement? See: https://github.com/Qiskit/qiskit-nature/blob/main/requirements.txt

manoelmarques (approved):

woodsp-ibm:
I had also asked for the requirement to be updated in another PR which had a dependence on newer Terra see above comment https://github.com/Qiskit/qiskit-nature/pull/222#issuecomment-854017382.  Since we need to backport this as a bug fix release for the upcoming Terra to assure nature still works with the latest Terra we will have to remember to change the requirements.txt file on the stable backport for the release - @manoelmarques fyi

PR #229, opened 2021-06-11
Docs - Remove non overidden base methods/attributes from classes

manoelmarques:
### Summary
It removes overriden methods/attributes from Nature classes that were unnecessarily repeating all methods from QuantumCircuit and other Terra base classes and running their sample doc.code.  

Also, fixed the title levels of one tutorial that was showing multiple times in the docs index.


woodsp-ibm (approved):
Makes things quite a bit cleaner - thanks

PR #230, opened 2021-06-12
Fixed minor bug in README sample

### LINKED ISSUE #225: Possiblly a bug in the README sample
### Information

- **Qiskit Nature version 0.1.3**:
- **Python version 3.7**:
- **Operating system Linux**:

```
ansatz.compose(init_state, front=True)
```
This line probably does nothing because by default `compose` is not in place as of qiskit version 0.17.4

### Suggested solutions
Change the line to
```
ansatz.compose(init_state, front=True, inplace=True)
```

shil-m:
### Summary
Updated the code for initial state as described by issue #225


woodsp-ibm (approved):
Thanks for fixing this.

PR #232, opened 2021-06-15
Implement BKSF for `FermionicOp` input

jlapeyre:
This PR implements the Bravyi-Kitaev Super-Fast fermionic to qubit operator transform taking a `FermionicOp` as input.  When complete, this PR will satisfy #9 

#### Summary

* There already exists an implementation in qiskit-aqua, which takes `FermionicOperator` as input. The latter is in the form of a rank-two and a rank-four tensor representing interaction integrals (and a constant). This PR reuses much of that code.

* DONE ~~This operators returned by this PR and the aqua code do not agree in a few terms (plus there is a global minus sign) difference.~~

* ~~The class in this PR `BravyiKitaevSFMapper` accepts either `FermionicOperator` or `FermionicOp`. For the former the output is (it should be rechecked) the same as the original code in aqua.~~

* EDIT: *This timing is for the old `FermionicOp` structure, before the performance regression.* For some input, this mapper takes about the same amount of time as the sum of the times for computing the molecule with PySCF and for creating the corresponding `FermionicOp`. For `H_2` and the `631g` basis set, each takes `100-200`ms.  There are some obvious inefficiencies in the mapper that could be improved.

#### Tasks

- [x] Correct and verify that the output. This involves tracking down the discrepancy. I have not spent much time on this; in fact it may be fairly easy.
- [x] Refactoring, some of which is noted in comments.
- [x] Tests. EDIT: all tests from aqua are included. But, more are needed.
- [x] Release note.

~~EDIT: All the tests that were in bksf in aqua.chemistry have been ported here, and they pass. However, they are not enough to test the essential behavior of this module.~~
EDIT: Additional tests have been implemented.


dlasecki:
Hello, a general comment. Basically all variable names are short and enigmatic. Would it be possible to refactor them such that they are more descriptive?

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+    for edge_index in range(edge_list.shape[1]):
+        if set((i, j)) == set(edge_list[:, edge_index]):
+            position_ij = edge_index
+            break
+
+    w[position_ij] = 1
+
+    for edge_index in range(qubit_position_i.shape[1]):
+        i_i, j_j = qubit_position_i[:, edge_index]
+        i_i = 1 if i_i == 0 else 0  # int(not(i_i))
```
dlasecki, line 359: Shall we use the commented alternative?

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+        if edge_list[i_i][j_j] < j:
+            v[j_j] = 1
+
+    qubit_position_j = np.asarray(np.where(edge_list == j))
+    for edge_index in range(qubit_position_j.shape[1]):
+        i_i, j_j = qubit_position_j[:, edge_index]
+        i_i = 1 if i_i == 0 else 0  # int(not(i_i))
+        if edge_list[i_i][j_j] < i:
+            v[j_j] = 1
```
dlasecki, line 368: We might extract the for loop to a separate function and avoid code duplication.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+        phase = 1
+    else:
+        raise ValueError("unexpected sequence of operators", facs)
+    return facs_out, phase
+
+
+class BravyiKitaevSFMapper(FermionicMapper):
+    """The Bravyi-Kitaev super fast fermion-to-qubit mapping.
+
+    Reference arXiv:1712.00446
```
dlasecki, line 446: Can we also put the full citation for people who don't want to google it?

test/mappers/second_quantization/test_bksf_mapper.py
```diff
@@ -0,0 +1,157 @@
...
+        qterm_b1 = edge_operator_bi(edge_list, 1)
+        qterm_b2 = edge_operator_bi(edge_list, 2)
+        qterm_b3 = edge_operator_bi(edge_list, 3)
+
+        ref_qterm_b0 = SparsePauliOp("IIIZZZ")
+        ref_qterm_b1 = SparsePauliOp("IZZIIZ")
+        ref_qterm_b2 = SparsePauliOp("ZIZIZI")
+        ref_qterm_b3 = SparsePauliOp("ZZIZII")
+
+        with self.subTest("Test 1"):
```
dlasecki, line 53: Can we give a more descriptive name than Test 1? What does it test exactly? Applies to other places too.

test/mappers/second_quantization/test_bksf_mapper.py
```diff
@@ -0,0 +1,157 @@
...
+                ("IIIN", (-0.4759487152209648 + 0j)),
+                ("IINI", (-1.2524635735648986 + 0j)),
+                ("IINN", (0.48217928821207245 + 0j)),
+                ("INII", (-0.4759487152209648 + 0j)),
+                ("ININ", (0.697393767423027 + 0j)),
+                ("INNI", (0.6634680964235684 + 0j)),
+                ("NIII", (-1.2524635735648986 + 0j)),
+                ("NIIN", (0.6634680964235684 + 0j)),
+                ("NINI", (0.6744887663568382 + 0j)),
+                ("NNII", (0.48217928821207245 + 0j)),
```
dlasecki, line 121: Potential candidate for storing in a separate file.

test/mappers/second_quantization/test_bksf_mapper.py
```diff
@@ -0,0 +1,157 @@
...
+                ("IZZI", (0.34297063344496626 + 0j)),
+                ("XIIX", (0.04532220205287402 + 0j)),
+                ("YIIY", (0.04532220205287402 + 0j)),
+                ("YZZY", (0.04532220205287402 + 0j)),
+                ("ZIIZ", (0.3317340482117842 + 0j)),
+                ("ZIZI", (-0.22278593040418454 + 0j)),
+                ("ZXXZ", (0.04532220205287402 + 0j)),
+                ("ZYYZ", (0.04532220205287402 + 0j)),
+                ("ZZII", (-0.22278593040418454 + 0j)),
+                ("ZZZZ", (0.24108964410603623 + 0j)),
```
dlasecki, line 140: Potential candidate for storing in a separate file.

test/mappers/second_quantization/test_bksf_mapper.py
```diff
@@ -0,0 +1,157 @@
...
+        pauli_sum_op = BravyiKitaevSFMapper().map(h2_fop)
+
+        op1 = _sort_simplify(expected_pauli_op)
+        op2 = _sort_simplify(pauli_sum_op.primitive)
+
+        with self.subTest("Map H2 frome sto3g basis, number of terms"):
+            self.assertEqual(len(op1), len(op2))
+
+        with self.subTest("Map H2 frome sto3g basis result"):
+            assert op1 == op2
```
dlasecki, line 153:
```suggestion
            self.assertEqual(op1, op2)
```

dlasecki (changes requested):

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+    qubit_op = (id_op - b_p) * (id_op - b_q)
+    if p == s:  # two commutations
+        final_coeff = 0.25
+    else:  # one commutation
+        final_coeff = -0.25
+    qubit_op = (final_coeff * h2_pqrs) * qubit_op
+    return qubit_op
+
+
+## SW2018 eq 35
```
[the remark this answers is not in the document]
jlapeyre: The idea here is to put abbreviated references to equation numbers. And an expanded citation in one place.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+
+from qiskit.opflow import PauliSumOp
+from qiskit.quantum_info.operators import Pauli, SparsePauliOp
+from qiskit_nature.operators.second_quantization import FermionicOp
+from .fermionic_mapper import FermionicMapper
+
+
+def _pauli_id(n_qubits: int):
+    """
+    Return an `n_qubits`-identity `SparsePauliOp`.
```
pbark, line 26:
```suggestion
    Return an `n_qubit`-identity `SparsePauliOp`.
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+        a_ps = edge_operator_aij(edge_list, p, s)
+        a_ps = -a_ps if s < p else a_ps
+        qubit_op = (a_ps * b_s + b_p * a_ps) * (id_op - b_q)
+        final_coeff = 1j * -0.25
+    elif q == s:
+        b_r = edge_operator_bi(edge_list, r)
+        a_pr = edge_operator_aij(edge_list, p, r)
+        a_pr = -a_pr if r < p else a_pr
+        qubit_op = (a_pr * b_r + b_p * a_pr) * (id_op - b_q)
+        final_coeff = 1j * 0.25
```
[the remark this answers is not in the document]
jlapeyre: Yes. I was thinking the same thing. Many of your comments are on code that I copied in from the old version. But, I did not get around to refactoring/commenting/clarifying everything yet.... oh btw. Everything like `a_pr * b_r` is deprecated in favor of `b_r & a_pr`. This should probably be fixed before release.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+    b_p = edge_operator_bi(edge_list, p)
+    id_op = _pauli_id(edge_list.shape[1])
+    qubit_op = (0.5 * h1_pq) * (id_op - b_p)  # SW2018 eq 33
+    return qubit_op
+
+
+## SW2018 eq 34
+def _coulomb_exchange(
+    edge_list: List, p: int, q: int, s: int, h2_pqrs: float
+):  # pylint: disable=invalid-name
```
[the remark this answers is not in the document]
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+
+
+def _pauli_id(n_qubits: int):
+    """
+    Return an `n_qubits`-identity `SparsePauliOp`.
+    """
+    return SparsePauliOp(Pauli((np.zeros(n_qubits, dtype=bool), np.zeros(n_qubits, dtype=bool))))
+
+
+def _number_operator(edge_list: List, p: int, h1_pq: float):  # pylint: disable=invalid-name
```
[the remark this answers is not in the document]
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+        qubit_op = (a_ps * b_s + b_p * a_ps) * (id_op - b_q)
+        final_coeff = 1j * -0.25
+    elif q == s:
+        b_r = edge_operator_bi(edge_list, r)
+        a_pr = edge_operator_aij(edge_list, p, r)
+        a_pr = -a_pr if r < p else a_pr
+        qubit_op = (a_pr * b_r + b_p * a_pr) * (id_op - b_q)
+        final_coeff = 1j * 0.25
+    else:
+        raise ValueError("unexpected sequence of indices")
```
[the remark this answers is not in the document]
jlapeyre: Done

jlapeyre (commented):
> Hello, a general comment. Basically all variable names are short and enigmatic. Would it be possible to refactor them such that they are more descriptive?

Which variable names in particular? For example `p, q, r, s`, and `b_r`, `a_pq` follow the equations in the paper(s). Maybe some others could be changed?

test/mappers/second_quantization/test_bksf_mapper.py
woodsp-ibm: Unless we want to use the same data across multiple tests, since this is just a few lines, personally I do not see any real benefit in having it in a separate file.

test/mappers/second_quantization/test_bksf_mapper.py
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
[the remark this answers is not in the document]
jlapeyre: Done

jlapeyre:
I saw a comment, I think by @pbark on the docstring for `_pauli_id`, but I am unable to find the comment now. I pushed a commit with a reworded docstring: "Return the identity for `SparsePauliOp` on `n_qubits` qubits."

test/mappers/second_quantization/test_bksf_mapper.py
jlapeyre: Yes, I agree with @woodsp-ibm. If it were a bit bigger I'l make it into some kind of artifact.

test/mappers/second_quantization/test_bksf_mapper.py
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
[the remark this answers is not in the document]
jlapeyre: I tried to factor out some of the computation. But, this adds to complexity and obscures the relation with the mathematical expressions. I agree this looks repetitious, but factoring out the computation is marginally worse, I think. Maybe someone else wants to weigh in.

jlapeyre:
> I am not sure if following names from the paper is the optimal choice if they are so enigmatic.

If you don't refer to the paper perhaps the variables are enigmatic. But, even more generally, if you follow any Q chemistry books or lit., the notation `a_pq` should be strongly reminiscent of Latex, i.e. `a_{pq}`, which is quite standard.

I copied the notation from the previous code. But, the previous code cited the paper that I, as well, am working from. This paper uses `h_{ijkl}` instead of `h_{pqrs}`. If anything, I think it might be a good idea to keep the notation, but change `pqrs` to `ijkl`. This would make it even easier to compare the code to the paper. Another option would be to find a paper that uses `pqrs` and follows  closely the same exposition. But, I don't know of such a paper. I suspect it might exist, since the previous code uses `pqrs`.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,521 @@
...
+
+
+def _unpack_term(
+    term_str: str, expand_number_op: bool = False
+) -> Tuple[Tuple[int, int, int], List]:
+    """
+    Return a tuple specifying the counts of kinds of operators in `term_str` and
+    a list of the factors and their indices in `term_str`.
+
+    The factors are represented by tuples of the form `(i, c)`, where `i` is and index
```
mrossinek, line 142:
```suggestion
    The factors are represented by tuples of the form `(i, c)`, where `i` is an index
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,521 @@
...
+            raise ValueError("unexpected number of number operators")
+    elif n_raise == 2 and n_lower == 2:
+        return "double_excitation"
+    else:
+        raise ValueError("unexpected number of operators")
+
+
+def number_of_modes(fer_op: FermionicOp) -> int:
+    """Return the number of modes (including identities) in each term `fer_op`"""
+    return len(fer_op.to_list()[0][0])
```
mrossinek, line 229: Isn't this just `FermionicOp.register_length`?

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,521 @@
...
+    given by `term.
+    """
+    return term[0]
+
+
+def operator_coefficient(term: tuple) -> float:
+    """
+    Return the coefficient of the multi-mode operator term extracted from a `FermionicOp`.
+    """
+    return term[1]
```
mrossinek, line 244: What is the purpose of these? Is this simply for verbosity?

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,521 @@
...
+
+
+## TODO: We may want a lower-triangular matrix. This may be the cause of the minus sign error.
```
mrossinek, line 247: What's the status on this?

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,521 @@
...
+
+
+def _add_edges_for_term(edge_matrix, term_str: str) -> None:
+    """
+    Add one, two, or no edges to `edge_matrix` as dictated by the operator `term_str`.
+    """
+    (n_number, n_raise, n_lower), facs = _unpack_term(term_str)
+    ttype = _interaction_type(n_number, n_raise, n_lower)
+    # For 'excitation' and 'number_excitation', create and edge between the `+` and `-`.
+    if ttype in ("excitation", "number_excitation"):
```
mrossinek, line 285: ypThis usage strengthens @dlasecki's suggestion for an Enum.

Also, I typically see Python codes using an underscore-prefix to differentiate variables from builtin-names. I.e. `_type` rather than `ttype`

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,521 @@
...
+    v = np.zeros(edge_list.shape[1])
+    w = np.copy(v)
+    v[qubit_position] = 1
+    qubit_op = Pauli((v, w))
+    return SparsePauliOp(qubit_op)
+
+
+def _add_sparse_pauli(qubit_op1: SparsePauliOp, qubit_op2: SparsePauliOp) -> SparsePauliOp:
+    """
+    Return `qubit_op1` and `qubit_op2`, except when either one is `None`.
```
mrossinek, line 382:
```suggestion
    Return `qubit_op1 + qubit_op2`, except when either one is `None`.
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,521 @@
...
+        return qubit_op1
+    else:
+        return qubit_op1 + qubit_op2
+
+
+def _to_physicist_index_order(facs: List) -> Tuple[List, int]:
+    """
+    Reorder the factors `facs` to be two raising operators followed by two lowering operators and
+    return the new factors and the phase incurred by the reordering. Note that `facs` are not in
+    chemists' order, but rather sorted by index with least index first.
```
mrossinek, line 398: This is normal ordering, correct?
@ikkoham What's the status on #88?
@jlapeyre Is that something you could make use of instead? Then we don't need to duplicate this logic.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,521 @@
...
+        phase = -1
+    elif ops == ["+", "-", "-", "+"]:
+        facs_out = [facs[0], facs[3], facs[1], facs[2]]
+        phase = 1
+    else:
+        raise ValueError("unexpected sequence of operators", facs)
+    return facs_out, phase
+
+
+class BravyiKitaevSFMapper(FermionicMapper):
```
mrossinek, line 427:
```suggestion
class BravyiKitaevSuperFastMapper(FermionicMapper):
```
We tend to stick to full name so I'd argue we do the same here.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,521 @@
...
+        sparse_pauli = sparse_pauli.simplify()
+        indices = sparse_pauli.table.argsort()
+        table = sparse_pauli.table[indices]
+        coeffs = sparse_pauli.coeffs[indices]
+        sorted_sparse_pauli = SparsePauliOp(table, coeffs)
+
+        return PauliSumOp(sorted_sparse_pauli)
+
+
+def _analyze_term(term_str: str) -> Tuple[str, List]:
```
mrossinek, line 456: Can we either put the main class of this file all the way at the front or end? Right now, it seems to be arbitrarily sandwiched between utility methods...

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,521 @@
...
+    Returns:
+       tuple: The first element is a string specifying the interaction type. See the method
+       `_interaction_type`. The second is a list of factors as returned by `_unpack_term`.
+    """
+    (n_number, n_raise, n_lower), facs = _unpack_term(term_str, expand_number_op=True)
+    ttype = _interaction_type(n_number, n_raise, n_lower)
+    return ttype, facs
+
+
+def _convert_operators(fer_op_qn: FermionicOp, edge_list: np.ndarray) -> SparsePauliOp:
```
mrossinek, line 476: This method could use some explanatory docstring.

releasenotes/notes/add-bksf-mapper-7826652bfa4403a6.yaml
```diff
@@ -0,0 +1,11 @@
+---
+features:
+  - |
+    Add :class:`qiskit_nature.mappers.second_quantization.BravyiKitaevSFMapper` implementing the
```
mrossinek, line 4: In case you do go with my suggested name change, be sure to update it here, too.

mrossinek (commented):
This looks like its on the right way :+1: I just have some comments of varying severity :slightly_smiling_face:

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Yes!

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Do you mean are the functions for verbosity? Or do you mean the doc strings are for verbosity?
I introduced the functions to abstract away the way the data is stored and to signal the meaning of the piece of data that is extracted.
These functions would be better as methods of `FermionicOp`.
Perhaps the doc strings are inserted into the online docs automatically, which is not what we want here... I should prepend the names with an underscore to emphasize that they are not part of the API of BKSF.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: This is a stale comment. I have now removed it

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: I agree `ttype` -> `_type` is better.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: The relation between #88 and the present ordering function is not clear to me.

qiskit_nature/mappers/second_quantization/bksf.py
mrossinek: Yes, I meant the functions themselves. I agree that these would make more sense as functions of the `FermionicOp`. These could be nicely implemented as [iterators](https://docs.python.org/3/library/stdtypes.html#typeiter) if you'd ask me :thinking:

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Done

releasenotes/notes/add-bksf-mapper-7826652bfa4403a6.yaml
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: done

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: done

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Both of these are done.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+    'number_excitation', 'double_excitation'.
+
+    Args:
+       `n_number`: the number of number operators
+       `n_raise`: the number of raising operators
+       `n_lower`: the number of lowering operators
+
+    Returns:
+      str: One of 'number', 'excitation', 'coulomb_exchange',
+      'number_excitation', 'double_excitation'.
```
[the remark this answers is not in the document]
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+      ValueError: if the numbers of operators don't describe a one- or two-body term from
+      an electronic Hamiltonian.
+    """
+    if n_raise == 0 and n_lower == 0:
+        if n_number == 1:
+            return "number"
+        elif n_number == 2:
+            return "coulomb_exchange"
+        else:
+            raise ValueError("unexpected number of number operators")
```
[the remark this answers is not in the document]
jlapeyre: Done here and in other places as well.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+    else:
+        raise ValueError("unexpected number of operators")
+
+
+def number_of_modes(fer_op: FermionicOp):
+    """Return the number of modes (including identities) in each term `fer_op`"""
+    return len(fer_op.to_list()[0][0])
+
+
+def operator_string(term: tuple):
```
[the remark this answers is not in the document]
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+    """
+    if i < j:
+        edge_matrix[i, j] = True
+    elif j < i:
+        edge_matrix[j, i] = True
+    else:
+        raise ValueError("expecting i != j")
+
+
+def _add_edges_for_term(edge_matrix, term_str: str):
```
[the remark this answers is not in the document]
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Yes, at least as it is used here, it seems that iterators would be a nice solution. Similarly, I really don't like things like this `if facs[0][0] == facs[1][0]`. Storing data in nested lists, tuples, and dicts and digging the pieces out manually makes for unreadable, brittle code.

jlapeyre: I prepended an underscore, so that we have `_operator_coefficient`. It should be replaced at some point by something implemented in `FermionicOp`. Can we consider this resolved?

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,511 @@
...
+    else:  # one commutation
+        final_coeff = -0.25
+    qubit_op = (final_coeff * h2_pqrs) * qubit_op
+    return qubit_op
+
+
+## SW2018 eq 35
+## Includes contributions from a h.c. pair
+def _excitation_operator(
+    edge_list: List, p: int, q: int, h1_pq: float
```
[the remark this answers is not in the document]
jlapeyre: Yes. Here, the type hint was changed. Elsewhere, the type of elements in `List` is now specified.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: IIRC I copied the alternative from older code and replaced it by the conditional, changing the old code to a comment. But, I think the old way is fine too. I have changed the code to use `int(not(i_i))`.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Done. This was a good idea.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,536 @@
...
+        return qubit_op1 + qubit_op2
+
+
+def _analyze_term(term_str: str) -> Tuple[TermType, List]:
+    """
+    Return the type of interaction represented by `term_str` and
+    a list of the factors and their indices in `term_str`.
+
+    The types of interaction are NUMBER, EXCITATION, COULOMB_EXCHANGE, NUMBER_EXCITATION,
+    DOUBLE_EXCITATION.
```
mrossinek, line 148: Maybe refer to `TermType` rather than duplicate the Enum values here.

qiskit_nature/mappers/second_quantization/bksf.py
mrossinek: @ikkoham could you please take a look at this? Maybe this is clearer to you as the author of #88

mrossinek (commented):
Some final comments but other than that this LGTM :+1:

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: I pushed a commit that may do you are thinking of here.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    Add an edge from lesser index to greater. This maintains the upper triangular structure.
+    """
+    if i == j:
+        raise ValueError("expecting i != j")
+    edge_matrix[min(i, j), max(i, j)] = True
+
+
+def _add_edges_for_term(edge_matrix: np.ndarray, term_str: str) -> None:
+    """
+    Add one, two, or no edges to `edge_matrix` as dictated by the operator `term_str`.
```
mrossinek, line 405: Missing `Args` and `Raises` documentation

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    n_modes = fer_op.register_length
+    edge_matrix = np.zeros((n_modes, n_modes), dtype=bool)
+    for term in fer_op.to_list():
+        _add_edges_for_term(edge_matrix, _operator_string(term))
+    return edge_matrix
+
+
+def _add_one_edge(edge_matrix: np.ndarray, i: int, j: int) -> None:
+    """
+    Add an edge from lesser index to greater. This maintains the upper triangular structure.
```
mrossinek, line 396: Missing `Args` and `Raises` documentation

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    edge_matrix[min(i, j), max(i, j)] = True
+
+
+def _add_edges_for_term(edge_matrix: np.ndarray, term_str: str) -> None:
+    """
+    Add one, two, or no edges to `edge_matrix` as dictated by the operator `term_str`.
+    """
+    (n_number, n_raise, n_lower), facs = _unpack_term(term_str)
+    _type = _interaction_type(n_number, n_raise, n_lower)
+    # For EXCITATION and NUMBER_EXCITATION, create and edge between the `+` and `-`.
```
mrossinek, line 409:
```suggestion
    # For EXCITATION and NUMBER_EXCITATION, create an edge between the `+` and `-`.
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    if _type in (TermType.EXCITATION, TermType.NUMBER_EXCITATION):
+        inds = [i for (i, c) in facs if c in "+-"]
+        if len(inds) != 2:
+            raise ValueError("wrong number or raising and lowering: ", len(inds))
+        _add_one_edge(edge_matrix, *inds)
+    # For `double_excitation` create an edge between the two `+`s and edge between the two `-`s.
```
mrossinek, line 415:
```suggestion
    # For `double_excitation` create an edge between the two `+`s and an edge between the two `-`s.
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    elif _type == TermType.DOUBLE_EXCITATION:
+        raise_inds = [i for (i, c) in facs if c == "+"]
+        lower_inds = [i for (i, c) in facs if c == "-"]
+        _add_one_edge(edge_matrix, *raise_inds)
+        _add_one_edge(edge_matrix, *lower_inds)
+
+
+def bksf_edge_list_fermionic_op(ferm_op: FermionicOp) -> np.ndarray:
```
mrossinek, line 423: Why are the following methods public and not private?
- `bksf_edge_list_fermionic_op`
- `edge_operator_aij`
- `edge_operator_bi`

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    """
+    Construct edge list required for the bksf algorithm.
+
+    Args:
+        ferm_op: the fermionic operator in the second quantized form
+
+    Returns:
+        numpy.ndarray: edge_list, a 2xE matrix, where E is total number of edge
```
mrossinek, line 431:
```suggestion
        numpy.ndarray: edge_list, a 2xE matrix, where E is total number of edges
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    elif _type == TermType.DOUBLE_EXCITATION:
+        raise_inds = [i for (i, c) in facs if c == "+"]
+        lower_inds = [i for (i, c) in facs if c == "-"]
+        _add_one_edge(edge_matrix, *raise_inds)
+        _add_one_edge(edge_matrix, *lower_inds)
+
+
+def bksf_edge_list_fermionic_op(ferm_op: FermionicOp) -> np.ndarray:
+    """
+    Construct edge list required for the bksf algorithm.
```
mrossinek, line 425: I would stick to `BKSF` in all-caps everywhere.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+def edge_operator_aij(edge_list: np.ndarray, i: int, j: int) -> SparsePauliOp:
+    """Calculate the edge operator A_ij.
+
+    The definitions used here are consistent with arXiv:quant-ph/0003137
+
+    Args:
+        edge_list (numpy.ndarray): a 2xE matrix, where E is total number of edge
+                                    and each pair denotes (from, to)
+        i (int): specifying the edge operator A
+        j (int): specifying the edge operator A
```
mrossinek, line 448:
```suggestion
        edge_list: a 2xE matrix, where E is total number of edges
                                    and each pair denotes (from, to)
        i: specifying the edge operator A
        j: specifying the edge operator A
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+
+def edge_operator_bi(edge_list: np.ndarray, i: int) -> SparsePauliOp:
+    """Calculate the edge operator B_i.
+
+    The definitions used here are consistent with arXiv:quant-ph/0003137
+
+    Args:
+        edge_list (numpy.ndarray): a 2xE matrix, where E is total number of edge
+                                    and each pair denotes (from, to)
+        i (int): index for specifying the edge operator B.
```
mrossinek, line 488:
```suggestion
        edge_list: a 2xE matrix, where E is total number of edges
                                    and each pair denotes (from, to)
        i: index for specifying the edge operator B.
```

qiskit_nature/mappers/second_quantization/bksf.py
mrossinek: What happened to this discussion?

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    elif q == s:
+        b_r = edge_operator_bi(edge_list, r)
+        a_pr = edge_operator_aij(edge_list, p, r)
+        a_pr = -a_pr if r < p else a_pr
+        qubit_op = (a_pr * b_r + b_p * a_pr) * (id_op - b_q)
+        final_coeff = 1j * 0.25
+    else:
+        raise ValueError(f"unexpected sequence of indices: {p}, {q}, {r}, {s}")
+    qubit_op = (final_coeff * h2_pqrs) * qubit_op
+    return qubit_op
```
mrossinek, line 285:
```suggestion
    return (final_coeff * h2_pqrs) * qubit_op
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+        + b_p * b_r
+        + b_p * b_s
+        + b_q * b_r
+        + b_q * b_s
+        - b_r * b_s
+        + b_p * b_q * b_r * b_s  ## Agrees with SW2018 eq 37 and OpenFermion. Aqua had `-`.
+    )
+    final_coeff = 0.125
+    qubit_op = (final_coeff * h2_pqrs) * qubit_op
+    return qubit_op
```
mrossinek, line 249:
```suggestion
    return (final_coeff * h2_pqrs) * qubit_op
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+def _excitation_operator(  # pylint: disable=invalid-name
+    edge_list: np.ndarray, p: int, q: int, h1_pq: float
+) -> SparsePauliOp:
+    if p >= q:
+        raise ValueError("Expected p < q, got p = ", p, ", q = ", q)
+    b_a = edge_operator_bi(edge_list, p)
+    b_b = edge_operator_bi(edge_list, q)
+    a_ab = edge_operator_aij(edge_list, p, q)
+    qubit_op = (-1j * 0.5 * h1_pq) * ((b_b & a_ab) + (a_ab & b_a))
+    return qubit_op
```
mrossinek, line 220:
```suggestion
    return (-1j * 0.5 * h1_pq) * ((b_b & a_ab) + (a_ab & b_a))
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    b_p = edge_operator_bi(edge_list, p)
+    b_q = edge_operator_bi(edge_list, q)
+    id_op = _pauli_id(edge_list.shape[1])
+    qubit_op = (id_op - b_p) * (id_op - b_q)
+    if p == s:  # two commutations
+        final_coeff = 0.25
+    else:  # one commutation
+        final_coeff = -0.25
+    qubit_op = (final_coeff * h2_pqrs) * qubit_op
+    return qubit_op
```
mrossinek, line 206:
```suggestion
    return (final_coeff * h2_pqrs) * qubit_op
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    return SparsePauliOp(Pauli((np.zeros(n_qubits, dtype=bool), np.zeros(n_qubits, dtype=bool))))
+
+
+def _number_operator(  # pylint: disable=invalid-name
+    edge_list: np.ndarray, p: int, h1_pq: float
+) -> SparsePauliOp:
+    b_p = edge_operator_bi(edge_list, p)
+    id_op = _pauli_id(edge_list.shape[1])
+    qubit_op = (0.5 * h1_pq) * (id_op - b_p)  # SW2018 eq 33
+    return qubit_op
```
mrossinek, line 190:
```suggestion
    return (0.5 * h1_pq) * (id_op - b_p)  # SW2018 eq 33
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+                raise ValueError("Unknown interaction: ", term_type)
+
+    return sparse_pauli
+
+
+def _add_sparse_pauli(qubit_op1: SparsePauliOp, qubit_op2: SparsePauliOp) -> SparsePauliOp:
+    """
+    Return `qubit_op1` + `qubit_op2`, except when either one is `None`.
+    In the latter case, return the one that is not `None`. In other words, assume
+    `None` signifies the additive identity.
```
mrossinek, line 135: Missing `Args` documentation

mrossinek (changes requested):
This is in very good shape, thanks @jlapeyre! I still have some comments but most of them are really minor :+1:

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""The Bravyi-Kitaev Super Fast Mapper."""
```
pbark, line 13:
```suggestion
"""The Bravyi-Kitaev Super Fast(BKSF) Mapper."""
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+        coeffs = sparse_pauli.coeffs[indices]
+        sorted_sparse_pauli = SparsePauliOp(table, coeffs)
+
+        return PauliSumOp(sorted_sparse_pauli)
+
+
+class TermType(Enum):
+    """Denotes the type of interaction of a Fermionic operator.
+    Allowed types of interaction are: NUMBER, EXCITATION, COULOMB_EXCHANGE, NUMBER_EXCITATION,
+    DOUBLE_EXCITATION.
```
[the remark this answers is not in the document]
pbark: I agree with @mrossinek here. Also I think is counter-intuitive the following enumeration on the type of terms.

qiskit_nature/mappers/second_quantization/bksf.py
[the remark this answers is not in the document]
pbark: I think here makes sense not to extract the functionality so that one can read and follow the reference of the mapping. I think this duplication of code increase readability.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+def _unpack_term(
+    term_str: str, expand_number_op: bool = False
+) -> Tuple[Tuple[int, int, int], List[Tuple[int, str]]]:
+    """
+    Return a tuple specifying the counts of kinds of operators in `term_str` and
+    a list of the factors and their indices in `term_str`.
+
+    The factors are represented by tuples of the form `(i, c)`, where `i` is an index
+    and `c` is a character.
+    Allowed characters in `term_str` are 'N+-I`.
```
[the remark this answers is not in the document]
pbark: @mrossinek I think E is not directly supported and should be mapped as `-+`. @jlapeyre please correct me if I am wrong.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    The returned tuple contains counts for `N`, `+`, and `-`, in that order. Identity operators
+    are ignored.
+
+    Args:
+       `term_str`: a string of characters in `+-NI`.
```
pbark, line 302: for consistency
```suggestion
       `term_str`: a string of characters in `N+-I`.
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+       `expand_number_op`: if `True`, number operators are expanded to `(i, '+')`, `(i, '-')`
+         in the returned list of factors.
+
+    Returns:
+       tuple: A tuple of two elements. First, a tuple of three integers giving the number of
+       number-, raising-, and -lowering operators. Second a list of factors represented by
```
pbark, line 308: consistency with below
```suggestion
       number, raising, and lowering operators. Second a list of factors represented by
```

qiskit_nature/mappers/second_quantization/bksf.py
pbark: The same comment of @mrossinek applies for many private methods. I am not sure if it is needed for private methods but would be good to include it.

qiskit_nature/mappers/second_quantization/bksf.py
pbark: This indeed is the normal ordering of a fermionic operator. We should address this.

pbark (changes requested):
This looks almost ready. There are a couple of things that still need to be addressed, but it mostly LGTM! Thank @jlapeyre!

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Ok. I'll take another look into it.

qiskit_nature/mappers/second_quantization/bksf.py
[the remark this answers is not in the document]
jlapeyre: I think I assumed that operators constructed from electronic structure packages do not need `E`, so I did not support it. I mean, it doesn't occur in the handful of Hamiltonians that I tested. I'm pretty sure that our translating code will never insert an `E`. But, I suppose someone could construct such a Hamiltonian in another way. I could support `E` and then write a low-level test for it. I guess I favor supporting it only if the need arises. But, if you all think it's better to include it now, that's ok, too.

qiskit_nature/mappers/second_quantization/bksf.py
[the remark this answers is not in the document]
jlapeyre: I did not intend that documentation for these interaction term types should be exposed to users. I am not sure how to control what documentation is automatically included in API documentation. But, I do want the term types and their relation to the literature to be very clear when reading or working on the code. I think all the repetition of the types in doc strings was a hold-over from when I used strings rather than an enum.

Perhaps repeating the types of the Enum in the doc string is not useful. It would be more informative to have something like: "Denotes the type of interaction of a Fermionic operator. The names of the types follow the usage of Reference [1] above". If the reader reads `TermType` in the methods below, she can look at the definition of `TermType` to find the meaning.

> I think is counter-intuitive the following enumeration on the type of terms.

I'm not sure what this means. Do you mean it's better not to explicitly number them?

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    """
+
+    NUMBER = 1
+    EXCITATION = 2
+    DOUBLE_EXCITATION = 3
+    NUMBER_EXCITATION = 4
+    COULOMB_EXCHANGE = 5
+
+
+def _convert_operators(ferm_op: FermionicOp, edge_list: np.ndarray) -> SparsePauliOp:
```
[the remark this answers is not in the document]
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+    """
+    if qubit_op1 is None:
+        return qubit_op2
+    elif qubit_op2 is None:
+        return qubit_op1
+    else:
+        return qubit_op1 + qubit_op2
+
+
+def _analyze_term(term_str: str) -> Tuple[TermType, List]:
```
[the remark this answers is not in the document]
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
ikkoham: Sorry for late. I did not notice the mention. I have updated #88. Previously, special class was needed to represent the normal order, but now we have the sparse label, I improved the implementation using the sparse label.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+        if n_number == 1:
+            return TermType.NUMBER_EXCITATION
+        elif n_number == 0:
+            return TermType.EXCITATION
+        else:
+            raise ValueError("unexpected number of number operators: ", n_number)
+    elif n_raise == 2 and n_lower == 2:
+        return TermType.DOUBLE_EXCITATION
+    else:
+        raise ValueError("unexpected number of operators: ", n_number)
```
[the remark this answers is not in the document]
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: I fixed this and made many other changes and additions to doc strings to conform to the google style guide as required in *CONTRIBUTING.md*

qiskit_nature/mappers/second_quantization/bksf.py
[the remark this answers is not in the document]
mrossinek: I was just surprised to not see it included. If it does not arise we can add it once needed. What error would currently be thrown when `E` is encountered? Is that a meaningful error message or something cryptic from somewhere deep down in the stack?

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,583 @@
...
+from .fermionic_mapper import FermionicMapper
+
+
+class BravyiKitaevSuperFastMapper(FermionicMapper):
+    """The Bravyi-Kitaev super fast fermion-to-qubit mapping.
+
+    This implementation follows closely Reference [1].
+
+       References:
+           [1]: Bravyi-Kitaev Superfast simulation of electronic structure on a quantum computer
```
mrossinek, line 31: This appears to be the intended way of typesetting `Superfast`. Maybe we should stick to that throughout the docstrings? E.g. a few lines higher we instead write `super fast`...

qiskit_nature/mappers/second_quantization/bksf.py
mrossinek: Thanks! I know that most of these are private methods but I am still seeing that some of them now still lack `Returns` documentation.
Maybe we can leave them for now (since most are self-explanatory and private anyways) but we should check inquire about a tool to detect the lack of these docstring aspects (as you also wrote offline).

mrossinek (commented):
Except for the few remaining open comments I think this looks good! If @jlapeyre can address those we can merge this soon :tada:

qiskit_nature/mappers/second_quantization/bksf.py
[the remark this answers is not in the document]
jlapeyre: I'm not sure I fully understand the comments from @mrossinek and @pbark. But, I have removed all documentation of the allowed values/labels of the enum from the file. The reasoning is that the enum is not exposed in any API and is only used in the file. The list of labels is self documenting. We use type annotations, so whereever `TermType` appears, one can check the definition of the enum to get the allowed values.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: I added an "Args:" section here (pylint did not complain about the absence, btw.) The way I interpret the google style guide, the "Returns:" section should *not* be included here (or at least is not required).

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: > applies for many private methods
Yes. Since you made these comments, I have read the style guide on docstrings, and revised all of the docstrings. However, this one in particular does not yet conform.

qiskit_nature/mappers/second_quantization/bksf.py
mrossinek: Yes `Returns` is not needed since it returns `None`. But did you add the `Raises` section?

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Fixed. This was a holdover from the aqua code.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Done

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: I did a quick search and found examples of BKSF referred to as "superfast", "super fast" and "super-fast". The author of that paper (Whitfield) uses "super-fast" on his webpage and in a grant application.
I prefer "super-fast" since it follows standard hyphenation rules and coining the word "superfast" seems like a bit of overreach, even in the physics world.

What we have now, "super fast", is the worst of the three. For the moment, I am changing it to "super-fast".

qiskit_nature/mappers/second_quantization/bksf.py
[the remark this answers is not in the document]
mrossinek: What I meant was that I did not like the duplication of the Enum's contents in its own docstring. At least not in that format because it did not really add any information/explanation about the different types.

mrossinek (approved):
This looks good to me now! Thanks @jlapeyre :tada:

Let's also await a second review on this one before merging :+1:

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,591 @@
...
+                Kanav Setia and James D Whitfield
+                JCP Volume 148, Issue 16 - Published April 2018
+                arXiv:1712.00446
+    """
+
```
woodsp-ibm, line 36: I think this needs an init here as below; otherwise it ends up allowing the 2 qubit reduction setting to be changed via base class. Maybe it was a mistake to default it there in the base class.
```
    def __init__(self):
        """The BKSF mapping."""
        super().__init__(allows_two_qubit_reduction=False)
``` 

Maybe this will help with the docs formatting too - as it stands it could do with improving around the way the references come out as it seems a bit too much

image

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,591 @@
...
+
+def _convert_operator(ferm_op: FermionicOp, edge_list: np.ndarray) -> SparsePauliOp:
+    """Convert a fermionic operator together with qubit-connectivity graph to a Pauli operator.
+
+    This is the heart of the implementation of BKSF mapping. The connectivity graph must be
+    computed before this method is called. The returned Pauli operator must be sorted and simplified.
+
+    Args:
+      `ferm_op`: The fermionic operator to convert.
```
woodsp-ibm, line 71: Is there are reason to quote the names of the args here ? e.g. 
```
`ferm_op`: ...
```
Normally this would be 
```
ferm_op: ...
```

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,591 @@
...
+    final_coeff = 0.125
+    return (final_coeff * h2_pqrs) * qubit_op
+
+
+def _number_excitation(  # pylint: disable=invalid-name
+    edge_list: np.ndarray, p: int, q: int, r: int, s: int, h2_pqrs: float
+) -> SparsePauliOp:
+    """Map a number-excitation operator to a Pauli operator.
+
+    Exactly two of the indices p, q, r, s must be equal. These represent
```
woodsp-ibm, line 294: In looking at `q` below I guess this statement here about two of them being equal eliminates the pair p+q; I guess its fine without having that constraint listed here too in this pairing - assuming this is correct.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,591 @@
...
+    a list of the factors and their indices in `term_str`.
+
+    The factors are represented by tuples of the form `(i, c)`, where `i` is an index
+    and `c` is a character.
+    Allowed characters in `term_str` are 'N+-I`.
+    The returned tuple contains counts for `N`, `+`, and `-`, in that order. Identity operators
+    are ignored.
+
+    Args:
+       `term_str`: a string of characters in `N+-I`.
```
woodsp-ibm, line 354: Quoted args again

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,591 @@
...
+def _interaction_type(n_number: int, n_raise: int, n_lower: int) -> TermType:
+    """Return a `TermType` instance describing the type of interaction given the number of
+    number, raising, and lowering operators.
+
+    The number number operators must be 1 or 2. The number of raising operators
+    must be equal to 0, 1 or 2, and the number of raising operators must be equal
+    to the number of lowering operators.
+
+    Args:
+       `n_number`: the number of number operators
```
woodsp-ibm, line 397: And here - but since we don't emit docs for private methods presently whether this would change the outcome in the html or not I am uncertain.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,591 @@
...
+    The definitions used here are consistent with arXiv:quant-ph/0003137
+
+    Args:
+        edge_list: a 2xE matrix, where E is total number of edges
+                                    and each pair denotes (from, to)
+        i: specifying the edge operator A
+        j: specifying the edge operator A
+
+    Returns:
+        Pauli: qubit operator
```
woodsp-ibm, line 511: Normally the return type is not given here - in this case it seems to conflict with what the method states i.e. SparsePauliOp

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,591 @@
...
+
+    The definitions used here are consistent with arXiv:quant-ph/0003137
+
+    Args:
+        edge_list: a 2xE matrix, where E is total number of edges
+                                    and each pair denotes (from, to)
+        i: index for specifying the edge operator B.
+
+    Returns:
+        Pauli: qubit operator
```
woodsp-ibm, line 551: Similar comment to above

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -0,0 +1,532 @@
...
+            return TermType.EXCITATION
+        else:
+            raise ValueError("unexpected number of number operators: ", n_number)
+    elif n_raise == 2 and n_lower == 2:
+        return TermType.DOUBLE_EXCITATION
+    else:
+        raise ValueError("unexpected number of operators: ", n_number)
+
+
+def _get_adjacency_matrix(fer_op: FermionicOp) -> np.ndarray:
```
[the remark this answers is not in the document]
jlapeyre: After a little investigating, it looks like this may be a feature that will land soon, or at best, very recently. Or, there may be a few packages and modules that you can install as a workaround. I'm inclined to wait until this feature is better supported.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Yes, I think the latter is the usual way to do it.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Still a good idea to follow the convention.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: * "qubit operator" is too generic. Better might be "the edge operator A_ij". I did not say "the qubit operator as a `SparsePauliOp`, because I would then be repeating information that's already in the type annotation.
* Maybe a good solution that satisfies the letter and spirit of the docstring style guide is "Return the edge operator A_ij" as the first line of the docstring. Then omit the "Returns:" section. And also add a pylint hint to disable the complaint about the missing section.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Yes, it's correct. I could be even more specific about the requirement.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: I revised the docstring according to my comment above.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: See my comments above. I did the same revision here.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: Yes, the `__init__` method is necessary. Apparently, `allows_two_qubit_reduction` is semantically a class property, but it is implemented on a per-instance basis.

qiskit_nature/mappers/second_quantization/bksf.py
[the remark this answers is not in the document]
jlapeyre: * An exception is raised in this very method
* I changed the message from "Unexpected operator..." to "Unsupported operator...", which is slightly better.

btw. as we know by now the docstring style guide requires that we do not document this exception.

qiskit_nature/mappers/second_quantization/bksf.py
jlapeyre: I tested #88. This indeed looks similar to what I have done here. However, the sparse representation of `FermionicOp` is a space separated string. I would probably need to decompose it into convenient data structures, lists and tuples, as I have done here.

Apparently, there are no methods in the class to parse the string into data structures. Is this correct?

mrossinek (dismissed):
LGTM, thanks @jlapeyre! Let's wait for another approving review before merging :slightly_smiling_face:

qiskit_nature/mappers/second_quantization/bksf.py
pbark: tagging also @ikkoham here. I think #88 is ready to be merged. So let's clarify this open point so that we finalize this also. Personally I think we can add a TODO as a reminder here and possibly address it in the future, but maybe it is worthy to do now.

pbark (commented):
I think that his now LGTM. We need to address the normal ordering of the fermionic operators and the this should be done. Thanks @jlapeyre and @mrossinek @ikkoham @dlasecki for the reviews!

pbark (approved):
For the time being we can keep the functionality of normal ordering as it is and we can add a TODO to address this in the future. If  @jlapeyre you are fine with this we can merge it.  If this is a really minor edit that you can do soon let's address this now.

jlapeyre:
@pbark I agree that it's okay to merge as is.
No, it's not a minor edit. The normal ordering is the same, but the data structure is different. It would require conversion. Not, a big chore, but requires a little thought: There is the question of whether to put the conversion in the `FermionicOp` class, or in this module.
The choice does not affect the user-facing API.

pbark:
@jlapeyre then we are fine to merge it as is. I resolved some minor conflicts. Can you please open an issue for future development where we will use the features introduced in PR #88 and have the follow up discussion to your comment above? Please document all your findings for future reference! Thanks a lot! Now this LGTM!

pbark (approved):

mrossinek (approved):
Thanks @jlapeyre for all the work you put into this! :tada:

PR #238, opened 2021-06-16
Fix #233 (Remove hardcoded "I" * int)

### LINKED ISSUE #233: `FermionicOp.set_label_display_mode` affects construction of operator
IIUC,  `FermionicOp.set_label_display_mode` should only affect printing. But, in fact it seems to affect interpreting labels in some way.

This works fine:
```python
from qiskit_nature.drivers import PySCFDriver, UnitsType
from qiskit_nature.problems.second_quantization.electronic.builders.fermionic_op_builder import build_ferm_op_from_ints

def get_molecule():
    h2 = 'H .0 .0 0.7414; H .0 .0 .0'
    basis0 = "sto3g"

    driver = PySCFDriver(atom=h2,
                         unit=UnitsType.ANGSTROM,
                         charge=0,
                         spin=0,
                         basis=basis0)
    molecule = driver.run()
    return molecule

def get_qkn_op(molecule):
    one_body = molecule.one_body_integrals
    two_body = molecule.two_body_integrals
    fop_qkn = build_ferm_op_from_ints(one_body, two_body)
    return fop_qkn

FermionicOp.set_label_display_mode('dense')

molecule = get_molecule()
fop = get_qkn_op(molecule)
```

However, if I change the string `'dense'` to `'sparse'`, then the script raises an error like this:
```python
~/code/github/Qiskit/oneenv/qiskit-nature/qiskit_nature/problems/second_quantization/electronic/builders/fermionic_op_builder.py in _build_ferm_op_helper(one_body_integrals, two_body_integrals)
     85     #  initial labels prevents from an empty labels list when building a FermionicOp)
     86     base_ops_labels.append(initial_label_with_ceoff)
---> 87     fermionic_op = FermionicOp(base_ops_labels)
     88 
     89     return fermionic_op

~/code/github/Qiskit/oneenv/qiskit-nature/qiskit_nature/operators/second_quantization/fermionic_op.py in __init__(self, data, register_length)
    188         ]
    189         if invalid_labels:
--> 190             raise ValueError(f"Invalid labels for sparse labels are given: {invalid_labels}")
    191 
    192         self._register_length = (

ValueError: Invalid labels for sparse labels are given: ['IIII']
```

#### Versions

qiskit_nature: a0f4c7bb3
qiskit-terra: 0df56b345b

ikkoham:
### Summary

Fix #233

### Details and comments

I am not sure the TODO comment. Is my fix correct?


mrossinek (commented):
Ah I see where this issue comes from now.. @jlapeyre could you check whether this works?
FWIW this is exactly the reason why I introduced `FermionicOp.one` and `FermionicOp.zero` in #220

They will ensure there is exactly one documented way how to construct these operators. In your comment (https://github.com/Qiskit/qiskit-nature/pull/220#discussion_r650685107) you said you don't like multiple ways to achieve the same goal but immediately after provide multiple supported methods. This kind of contradicts itself. Thus, having one method exposed as a classmethod (a class-level utility constructor) users see and know to use that way (and ideally only that way).

All that said, if @jlapeyre could verify this fix we can merge this as is *or* we can discuss whether to add these classmethods already via this PR.

jlapeyre:
> FWIW this is exactly the reason why I introduced FermionicOp.one and FermionicOp.zero

Agree. There are many reasons to do this.

ikkoham:
In many cases, Zero operator should not be created. `int` and `float` zeros will do like:
```python
fer = 0
for _ in some:
    fer += FermionicOp("some")
```
So in many cases, 0 is enough. I am concerned that if it is easy to create zero operators, users will create useless zero operators.

jlapeyre:
I verify that this branch fixes the problem in #233

jlapeyre:
> All that said, if @jlapeyre could verify this fix we can merge this as is or we can discuss whether to add these classmethods already via this PR.

This bug doesn't seem to be causing much trouble at the moment. On the other hand, this fix is very simple. I'm ok with committing this as is, and then discussing the zero and/or one methods further. It's something that I thought was missing when I first saw qiskit. But, perhaps the pros and cons could be articulated in a more organized way.

mrossinek (approved):

pbark (approved):

mrossinek:
What is the current status of this? The conflict with `main` arose from the erroneous deletion of the `fermionic_op_builder`. It will be re-added by #263 but I am still not sure whether this fix is still needed. @ikkoham @jlapeyre any thoughts?

ikkoham:
This should be added after `fermionic_op_builder` is re-added asap (or You can add it in your PR #263).

ikkoham:
Close because this is fixed by #263

PR #240, opened 2021-06-16
Add deprecation method template

manoelmarques:
### Summary
This adds warnings method and decorators to simplify deprecation messages as already done in Optimization.
The deprecation msg will only show the first time it is hit.

Also removed unused loggers in files.

### DIFF
--- a/qiskit_nature/algorithms/excited_states_solvers/excited_states_eigensolver.py
+++ b/qiskit_nature/algorithms/excited_states_solvers/excited_states_eigensolver.py
@@ -12,7 +12,6 @@
 
 """The calculation of excited states via an Eigensolver algorithm"""
 
-import logging
 from typing import List, Union, Optional
 
 from qiskit.algorithms import Eigensolver
@@ -25,8 +24,6 @@ from qiskit_nature.results import EigenstateResult
 
 from .excited_states_solver import ExcitedStatesSolver
 from .eigensolver_factories import EigensolverFactory
-
-logger = logging.getLogger(__name__)
 
 
 class ExcitedStatesEigensolver(ExcitedStatesSolver):
--- a/qiskit_nature/deprecation.py
+++ b/qiskit_nature/deprecation.py
@@ -1,0 +1,288 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Contains the Deprecation message methods."""
+
+from abc import abstractmethod
+import warnings
+import functools
+from typing import NamedTuple, Optional, Callable, Dict, Set
+from enum import Enum, EnumMeta
+
+
+class DeprecatedEnum(Enum):
+    """
+    Shows deprecate message whenever member is accessed
+    """
+
+    def __new__(cls, value, *args):
+        member = object.__new__(cls)
+        member._value_ = value
+        member._args = args
+        member._show_deprecate = member.deprecate
+        return member
+
+    @abstractmethod
+    def deprecate(self):
+        """show deprecate message"""
+        pass
+
+
+class DeprecatedEnumMeta(EnumMeta):
+    """
+    Shows deprecate message whenever member is accessed
+    """
+
+    def __getattribute__(cls, name):
+        obj = super().__getattribute__(name)
+        if isinstance(obj, DeprecatedEnum) and obj._show_deprecate:
+            obj._show_deprecate()
+        return obj
+
+    def __getitem__(cls, name):
+        member = super().__getitem__(name)
+        if member._show_deprecate:
+            member._show_deprecate()
+        return member
+
+    # pylint: disable=redefined-builtin
+    def __call__(cls, value, names=None, *, module=None, qualname=None, type=None, start=1):
+        obj = super().__call__(
+            value, names, module=module, qualname=qualname, type=type, start=start
+        )
+        if isinstance(obj, DeprecatedEnum) and obj._show_deprecate:
+            obj._show_deprecate()
+        return obj
+
+
+class DeprecatedType(Enum):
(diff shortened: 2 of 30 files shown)


qiskit_nature/algorithms/excited_states_solvers/excited_states_eigensolver.py
```diff
@@ -26,8 +25,6 @@
 from .excited_states_solver import ExcitedStatesSolver
 from .eigensolver_factories import EigensolverFactory
 
-logger = logging.getLogger(__name__)
```
mrossinek, line 29: I guess you are simply removing these because they are unused, right? I am fine with that for now since we are likely to re-evaluate the logging situation anyways.

qiskit_nature/drivers/base_driver.py
```diff
@@ -18,10 +18,10 @@
...
 
 from enum import Enum, EnumMeta
-import warnings
 
 from .molecule import Molecule
 from ..exceptions import QiskitNatureError
+from ..deprecation import DeprecatedType, warn_deprecated_same_type_name
 
 
 class DeprecatedEnum(Enum):
```
mrossinek, line 27: Now that we have `qiskit_nature.deprecation`, this class and the meta-classes below would seem to fit there better, too, don't you think?

qiskit_nature/deprecation.py
```diff
@@ -0,0 +1,242 @@
...
+    # skip if it was already added
+    obj = _DeprecatedObject(version, old_type, old_name, new_type, new_name, additional_msg)
+    if obj in _DEPRECATED_OBJECTS:
+        return
+
+    _DEPRECATED_OBJECTS.add(obj)
+
+    msg = (
+        f"The {old_name} {old_type.value} is deprecated as of version {version} "
+        "and will be removed no sooner than 3 months after the release"
```
mrossinek, line 73: Maybe just a general question, but this seems like a good place to ask: where do these 3 months come from? Is this because of the approximate 3-month release cycle? Wouldn't it make more sense to tie the deprecation to the next release (e.g. `0.{n+1}.0` if deprecated with release of `0.{n}.0`). Given our dev cycles this will roughly coincide with 3 months but it will be *much* easier to keep track off and will also not require users to know when version `0.{n}.0` actually came out in order to know when this will be removed..

mrossinek (commented):
Thanks @manoelmarques! This looks mostly good to me :+1: I just left some minor and/or general comments

qiskit_nature/algorithms/excited_states_solvers/excited_states_eigensolver.py
manoelmarques: Yes, the unused loggers have been removed from Terra and I believe Optimization too so I went ahead and removed it in Nature..

qiskit_nature/drivers/base_driver.py
manoelmarques: You are right, I will move those.

qiskit_nature/deprecation.py
manoelmarques: I believe it is a standard set initially by Terra. I agree that tying the schedule to some release after seems better than an arbitrary timeframe. Maybe @woodsp-ibm would have more to say about it.

pbark: I agree with @mrossinek. Setting deprecation based on releases is reasonable and easier for people to follow.

qiskit_nature/drivers/base_driver.py
manoelmarques: I moved them

qiskit_nature/deprecation.py
woodsp-ibm: https://qiskit.org/documentation/contributing_to_qiskit.html#deprecation-policy

woodsp-ibm: Given the release when it is deprecated in there, which can be looked up and its date found, computing at least 3 months from that can be done. Ideally users should transition to the new way asap but the idea here is that get some time to do that. Saying in some future release by version people may have no idea when that is coming out - also plans change. In the end if you get deprecated messages you should be changing asap, if you procrastinate then things will just break when its removed. This clients see sometimes if they have some old code that worked and then go back to it after a few months during which time code it uses has since been both deprecated and removed after the time was up.

manoelmarques: Given that this is a documented Qiskit policy, I believe we need to comply with it then.

mrossinek: > which can be looked up and its date found
I agree, but I think that this is unnecessary manual.

> may have no idea when that is coming out - also plans change
True, but I don't think this really affects what I proposed above

> This clients see sometimes if they have some old code that worked...
I actually think that having the version number as part of the deprecation warning would actually help with this situation as it makes clearer which versions to look for in the Changelog.

> https://qiskit.org/documentation/contributing_to_qiskit.html#deprecation-policy
Thanks! I see, if this is done the same everywhere as is, we should probably stick to it. I just wanted to bring it up since I personally would find it much more sensible, with the removal pinned to a version number. We could even include that link in the deprecation message..

woodsp-ibm: Its consistent with Qiskit. Maybe having a future date in there would be better - e.g. we know the date of Python 3.6 EOL so its easy to deal with. By plans changing I meant its harder to predict the version we will be 3 months out to put in the message - if we decide to bring the next one release sooner then it kinda messes things up in that regard. But we would know just ahead of release when 3 months out was to set a date for all deprecation messages added that release to use. But unless Qiskit overall changes I think though we should stick to the current way/message and be consistent.

ikkoham:
Thank you. These are really useful tools. But, is it difficult to implement these in Terra? (i.e. I want to use from Terra)

manoelmarques:
@ikkoham Terra has two decorators, you need to give the whole message though: https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/utils/deprecation.py

qiskit_nature/deprecation.py
mrossinek: Yes let's leave it as is, since it appears to be Qiskit standard.

Just as a final comment to clarify: my suggestion was to make the deprecation such that we remove it at the next feature release **regardless** of when that is. Usually this fits well with 3 months but if it happens after 1 month then the user just needs to hurry up. Since this is really not in-line with Qiskit's deprecation policy I retract this proposal.

mrossinek (approved):

PR #248, opened 2021-06-23
Pin H5Py

mrossinek:
### Summary

H5Py v3.3.0 finally removed support for the deprecated default file
modes. Unfortunately, PySCF has not yet put out a release respecting
this change. In the meantime, we can pin H5Py below 3.3 to ensure our
development can continue until PySCF catches up.


mrossinek:
The mypy fixes are the required changes for Nature as done for Optimization here: https://github.com/Qiskit/qiskit-optimization/pull/176

manoelmarques (approved):

PR #262, opened 2021-07-05
Improve Windows compatibility (pathlib) for psi4

### LINKED ISSUE #260: Unicode Error when launching PSI4 driver (using Windows)
When using psi4 as driver for my vqe calculations, I get the following error:
```
  File "C:\Users\poc\Anaconda3\envs\qiskit-0-27-psi4-jupyter\Scripts\psi4-script.py", line 336, in <module>
    exec(content)

  File "<string>", line 31
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape

Printing out the relevant lines from the Psithon --> Python processed input file:


!----------------------------------------------------------------------------------!
!                                                                                  !
!  (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3:       !
!     truncated \UXXXXXXXX escape                                                  !
!                                                                                  !
!----------------------------------------------------------------------------------!

  File "c:\Users\poc\Documents\qiskit_scripts\groundstatesolver.py", line 336, in <module>
    compare_mappers()
  File "c:\Users\poc\Documents\qiskit_scripts\groundstatesolver.py", line 206, in compare_mappers
    second_q_op = es_problem.second_q_ops()
  File "C:\Users\poc\Anaconda3\envs\qiskit-0-27-psi4-jupyter\lib\site-packages\qiskit_nature\problems\second_quantization\electronic\electronic_structure_problem.py", line 64, in second_q_ops
    self._molecule_data = cast(QMolecule, self.driver.run())
  File "C:\Users\poc\Anaconda3\envs\qiskit-0-27-psi4-jupyter\lib\site-packages\qiskit_nature\drivers\psi4d\psi4driver.py", line 160, in run
    PSI4Driver._run_psi4(input_file, output_file)
  File "C:\Users\poc\Anaconda3\envs\qiskit-0-27-psi4-jupyter\lib\site-packages\qiskit_nature\drivers\psi4d\psi4driver.py", line 215, in _run_psi4
    raise QiskitNatureError('{} process return code {}\n{}'.format(
qiskit_nature.exceptions.QiskitNatureError: 'psi4 process return code 1\nTraceback (most recent call last):\n  File "C:\\Users\\poc\\Anaconda3\\envs\\qiskit-0-27-psi4-jupyter\\Scripts\\psi4-script.py", line 336, in <module>\n    exec(content)\n\n  File "<string>", line 31\nSyntaxError: (unicode error) \'unicodeescape\' codec can\'t decode bytes in position 2-3: truncated \\UXXXXXXXX escape\n\nPrinting out the relevant lines from the Psithon --> Python processed input file:\n\n\n!----------------------------------------------------------------------------------!\n!                                                                                  !\n!  (unicode error) \'unicodeescape\' codec can\'t decode bytes in position 2-3:       !\n!     truncated \\UXXXXXXXX escape                                                  !\n!                                                                                  !\n!----------------------------------------------------------------------------------!\n\n'
```

With the kind help of some psi4 developers I could track down it to these lines:
https://github.com/Qiskit/qiskit-nature/blob/c7fa055cfba2dc08daef0111e569ae5eb9bce10b/qiskit_nature/drivers/psi4d/psi4driver.py#L123-L133

I was able to get rid of the problem via changing above lines to this:
```python
        qiskit_nature_directory = os.path.abspath(os.path.join(psi4d_directory, "../.."))

        molecule = QMolecule()

        input_text = cfg + "\n"
        input_text += "import sys\n"
        syspath = "['" + qiskit_nature_directory.replace("\\", "\\\\") + "','" + "','".join(p.replace("\\", "\\\\") for p in sys.path) + "']"

        input_text += "sys.path = " + syspath + " + sys.path\n"
        input_text += "from qiskit_nature.drivers.qmolecule import QMolecule\n"
        input_text += "_q_molecule = QMolecule('" + molecule.filename.replace("\\", "\\\\") + "')\n"
```
Unfortunately this fix is not cross-platform compatible but quite windows-specific.
So, any help finding a nice fix for this bug or even a simplified/less error-prone way to call psi4 would be appreciated.

cpossel:
### Summary
Closes #260

Due to difficulties with backslashes in Windows path strings psi4 raised an Unicode Error, which is fixed by this approach. 

### Details and comments

The bug with the backslashes was fixed via replacing all backslashes in Windows path strings by forward slashes.
psi4, to which the string is sent for execution, can handle the forward slashes in Windows paths. So, compatibility for both Unix and Windows paths is ensured.
Switching to pathlib objects wasn't directly necessary to fix this bug, but makes things more consistent and easier for future Windows/Unix path issues.

### DIFF
--- a/qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
+++ b/qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
@@ -121,7 +121,7 @@ class PSI4Driver(FermionicDriver):
 
         input_text = cfg + "\n"
         input_text += "import sys\n"
-        syspath = "['" + str(qiskit_nature_directory).replace("\\", "/") + "','" 
+        syspath = "['" + str(qiskit_nature_directory).replace("\\", "/") + "','"
         syspath += "','".join(p.replace("\\", "/") for p in sys.path) + "']"
 
         input_text += "sys.path = " + syspath + " + sys.path\n"
--- a/qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
+++ b/qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
@@ -121,12 +121,17 @@ class PSI4Driver(FermionicDriver):
 
         input_text = cfg + "\n"
         input_text += "import sys\n"
-        syspath = "['" + str(qiskit_nature_directory).replace("\\", "/") + "','"
-        syspath += "','".join(p.replace("\\", "/") for p in sys.path) + "']"
+        syspath = (
+            "['"
+            + qiskit_nature_directory.as_posix()
+            + "','"
+            + "','".join(Path(p).as_posix() for p in sys.path)
+            + "']"
+        )
 
         input_text += "sys.path = " + syspath + " + sys.path\n"
-        input_text += "from qiskit_nature.drivers.second_quantization.qmolecule import QMolecule\n"
-        input_text += '_q_molecule = QMolecule("{0}")\n'.format(molecule.filename.replace("\\", "/"))
+        input_text += "from qiskit_nature.drivers.qmolecule import QMolecule\n"
+        input_text += '_q_molecule = QMolecule("{0}")\n'.format(Path(molecule.filename).as_posix())
 
         with open(template_file, "r") as file:
             input_text += file.read()
--- a/qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
+++ b/qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
@@ -17,6 +17,7 @@ import os
 import subprocess
 import sys
 import tempfile
+from pathlib import Path
 from shutil import which
 from typing import Union, List, Optional
 
@@ -112,19 +113,19 @@ class PSI4Driver(FermionicDriver):
         else:
             cfg = self._config
 
-        psi4d_directory = os.path.dirname(os.path.realpath(__file__))
-        template_file = psi4d_directory + "/_template.txt"
-        qiskit_nature_directory = os.path.abspath(os.path.join(psi4d_directory, "../.."))
+        psi4d_directory = Path(__file__).resolve().parent
+        template_file = psi4d_directory.joinpath("_template.txt")
+        qiskit_nature_directory = psi4d_directory.parent.parent
 
         molecule = QMolecule()
 
         input_text = cfg + "\n"
         input_text += "import sys\n"
-        syspath = "['" + qiskit_nature_directory + "','" + "','".join(sys.path) + "']"
+        syspath = "['" + str(qiskit_nature_directory).replace("\\", "/") + "','" + "','".join(p.replace("\\", "/") for p in sys.path) + "']"
 
         input_text += "sys.path = " + syspath + " + sys.path\n"
         input_text += "from qiskit_nature.drivers.second_quantization.qmolecule import QMolecule\n"
-        input_text += '_q_molecule = QMolecule("{0}")\n'.format(molecule.filename)
+        input_text += '_q_molecule = QMolecule("{0}")\n'.format(molecule.filename.replace("\\", "/"))
 
         with open(template_file, "r") as file:
             input_text += file.read()
--- a/qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
+++ b/qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
@@ -130,7 +130,7 @@ class PSI4Driver(FermionicDriver):
         )
 
         input_text += "sys.path = " + syspath + " + sys.path\n"
-        input_text += "from qiskit_nature.drivers.qmolecule import QMolecule\n"
+        input_text += "from qiskit_nature.drivers.second_quantization.qmolecule import QMolecule\n"
         input_text += '_q_molecule = QMolecule("{0}")\n'.format(Path(molecule.filename).as_posix())
 
         with open(template_file, "r") as file:
--- a/qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
+++ b/qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
@@ -121,7 +121,8 @@ class PSI4Driver(FermionicDriver):
 
         input_text = cfg + "\n"
         input_text += "import sys\n"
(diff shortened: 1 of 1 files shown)


qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
```diff
@@ -112,19 +113,19 @@ def run(self) -> QMolecule:
...
+        psi4d_directory = Path(__file__).resolve().parent
+        template_file = psi4d_directory.joinpath("_template.txt")
+        qiskit_nature_directory = psi4d_directory.parent.parent
 
         molecule = QMolecule()
 
         input_text = cfg + "\n"
         input_text += "import sys\n"
-        syspath = "['" + qiskit_nature_directory + "','" + "','".join(sys.path) + "']"
+        syspath = "['" + str(qiskit_nature_directory).replace("\\", "/") + "','" + "','".join(p.replace("\\", "/") for p in sys.path) + "']"
```
mrossinek, line 124: Is the `replace` call really necessary now? I would assume that the string-representation of `pathlib.Path` should work across platforms. Since we iterate all paths in `sys.path` anyways, rather than doing the replace you could do something like `[str(Path(p)) for p in sys.path]`. This would need to be tested on a Windows machine though..

cpossel: I did the test with `str(Path(p))` without `replace` on a Windows machine and still got the error.
Actually I don't really understand why the error still occurs since it should correctly create the string with escaped backslashes `\\`. It seems there is other stuff happening due to the way psi4 is invoked (save the string to a file, let psi4 read the file and execute it's content via `exec` command)...
It would be possible to use `repr(Path(p))` but then importing also `WindowsPath` from pathlib would be necessary and I think it's less readable/more complicated than the `replace` approach.

mrossinek: Ah yes I believe you are right about it having to do with how Psi4 is being called. The fact, that you are replacing the `\\` with `/` essentially makes a Windows path look like a Unix path.. I don't fully understand why this is necessary though and I don't have a Windows machine to test this myself to see what is going on.
Maybe @woodsp-ibm can help with this?

cpossel: If that info helps: I also tested with `replace("\\", "\\\\")` instead of `replace("\\", "/")` basically making it a kind of double-escaped backslash. The rawstring then contains stuff like `C:\\\\User` and that works fine, too. Though that looks pretty weird. That's why I thought the `/` Unix slash works nicer without completely understanding why the normaly escaped backslash `\\` doesn't work.

mrossinek: Thanks that information helped indeed! :slightly_smiling_face: What is going on here is the following:
The paths which we are dealing with here are written to an input file which gets read by Psi4. When reading paths from a string, Windows paths can actually be written as Posix paths (i.e. with `/` rather than `\\`). `\\\\` works too, because that is equivalent to two escaped `\`.

With all that said, I found this method: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix
Maybe we can use that instead of `replace`?

mrossinek: Also, in terms of CI, be sure to run `make black` before committing. That will take care of formatting for you :+1:

cpossel: 👍 `as_posix` is indeed a much nicer solution. And probably more robust if the interface to psi4 ever changes.
I changed it accordingly and did run `black` beforewards (which I had never used before, so I hope it's fine now).

qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
```diff
@@ -112,19 +113,25 @@ def run(self) -> QMolecule:
...
+            + qiskit_nature_directory.as_posix()
+            + "','"
+            + "','".join(Path(p).as_posix() for p in sys.path)
+            + "']"
+        )
 
         input_text += "sys.path = " + syspath + " + sys.path\n"
-        input_text += "from qiskit_nature.drivers.second_quantization.qmolecule import QMolecule\n"
-        input_text += '_q_molecule = QMolecule("{0}")\n'.format(molecule.filename)
+        input_text += "from qiskit_nature.drivers.qmolecule import QMolecule\n"
```
mrossinek, line 133:
```suggestion
        input_text += "from qiskit_nature.drivers.second_quantization.qmolecule import QMolecule\n"
```
I am not quite sure why you removed this, but other than that, this PR looks good to me now :+1:

mrossinek (changes requested):

qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
cpossel: It was actually a simple copy-paste mistake from the deprecated `qiskit_nature/drivers/psi4d/psi4driver.py` which I had used for some testing. Good that you saw it.

mrossinek (approved):
LGTM :+1: Thanks @cpossel for working on this!

cpossel:
👍 @mrossinek Thanks for your valuable comments.

I guess with that I can close the related issue [https://github.com/Qiskit/qiskit-nature/issues/260](url).
Do I also need to close this PR or is it done automatically when merging?

mrossinek:
The PR will be closed once it gets merged. No need to do that manually :+1: In fact, closing it now would prevent proper merging..
I will await one more review from either @pbark or @woodsp-ibm before merging, though

pbark (approved):
We can merge this after @woodsp-ibm review and make it to the release.

mrossinek:
Actually, we should probably apply an identical fix to `qiskit_nature/drivers/psi4d/psi4driver.py`.
Only if we do that, will we be able to backport this and include it in a the upcoming bugfix release.

@manoelmarques @woodsp-ibm @pbark

woodsp-ibm:
I guess we need to do this backport manually and as you say `we should probably apply an identical fix to qiskit_nature/drivers/psi4d/psi4driver.py.` since its just that one where the fix is needed.

manoelmarques (approved):

PR #270, opened 2021-07-09
W9015: "provider" missing in parameter documentation 

nalraajan:
### Summary
W9015: "provider" missing in parameter documentation 

### Details and comments
The provider parameter of _init_ function is not documented


woodsp-ibm:
Hi, thanks for fixing this. Someone commented the other day in Qiskit Slack workspace (https://qiskit.slack.com/archives/CB6C24TPB/p1625581496374900). They were also confused by the fact that `provider` is optional, less so by `backend` I think, though the same really applies, I think what might be helpful in the documentation too is to state that they are optional here to allow the program to be created without either, but both must be provided later, via their respective setters, in order that the program will run. Failure to do so will raise an error that it is not set.

nalraajan:
Thank you @woodsp-ibm. I will fix these kind of issues in code base and raise the PR for same.

pbark (approved):

pbark:
Hi @nalraajan! LGTM! In view that changes mentioned from @woodsp-ibm are pending in other repos, I think we can merge this PR now! Thank you!

PR #273, opened 2021-07-14
Ground state solvers notebook: remove draw statement without output

eendebakpt:
### Summary

The `TwoLocal.draw` statement in the notebook has no effect.

### Details and comments

 Since the number of qubits has not been specified yet, the draw has no output. It is then better to remove the statement.


mrossinek:
Hi @eendebakpt! Thank you for your contribution to Qiskit Nature!

I would argue that, for this example to make sense, we should rather make sure that the `draw` statement becomes useful, instead of removing it. The reason why it does not do anything right now, is that in the example we never set the number of qubits. I do not know the correct number of qubits from the top of my head but would you be willing to find that out and set it like so:
```python
tl_circuit.num_qubits = <insert number here>
```
After that, calling `draw` _should_ work as expected..

eendebakpt:
@mrossinek We can get the `TwoLocal` to draw, but I don't think it adds much:
image

mrossinek:
Try decomposing the circuit first:
```python
tl_circuit.decompose().draw(output="mpl")
```

woodsp-ibm:
Setting num_qubits just to be able to draw the ansatz is a bad idea. It used to have some number set here, probably as a legacy from before TwoLocal was a BlueprintCircuit and adaptable by the number of qubits. And others had noted that in the past which is why it was removed - but it seems the draw was overlooked. The whole idea of not supplying a number of qubits here is that user most likely does not know (hence Blueprint ccts where it can be set when we do) so this showed how you would actually use  VQE, and it not 'hacking'  things for the purpose of drawing. So for a tutorial on how to do ground state we should not be setting that.

Either we remove the draw() as was originally done - or we can draw the anstaz later after VQE has used it, when it will have the number of qubits has been set, if that's felt it really useful to see this.

mrossinek:
Currently, `tl_circuit` is supplied to `another_solver` which is _never_ used in this tutorial. Thus, if we do not remove the `draw()` call, then we must add another call which actually uses this solver.

woodsp-ibm:
I would just get rid of the draw() as was originally proposed - the tutorial is more about how to compute a ground state and bring the bits together to do that. Maybe it would be nice to run the alternate direct VQE too so the user sees things work that way as well - but its a bit more of a change  to have things flow nicely in the tutorial I think - to me it really depended how much it was wanted to see the ansatz drawn. This did not really enhance the main goal of the tutorial to me so not drawing it was fine.

eendebakpt:
@mrossinek Based on comments on this PR I removed the `draw` statement again.

mrossinek:
> @mrossinek Based on comments on this PR I removed the `draw` statement again.

Yep, thanks! Sorry about the back-and-forth..

mrossinek (approved):

PR #280, opened 2021-07-19
Fix mapper documentation formatting

### LINKED ISSUE #278: Error in docstring of Second Quantization Qubit Mappers
# **Problem**
The documentation for a bunch of the second quantization mappers is not being rendered correctly, as seen in the docs for the [Jordan Wigner](https://qiskit.org/documentation/nature/stubs/qiskit_nature.mappers.second_quantization.JordanWignerMapper.html#qiskit_nature.mappers.second_quantization.JordanWignerMapper) mapper. 

bugmap

# **Origin**
The docstring for these mappers is inherited from [```QubitMapper```](https://github.com/Qiskit/qiskit-nature/blob/fc8a649f3baa594526599ce4d99e034fa64ecbdc/qiskit_nature/mappers/second_quantization/qubit_mapper.py#L27)
```python
class QubitMapper(ABC):
    """The interface for implementing methods which map from a `SecondQuantizedOp` to a
    qubit operator in the form of a `PauliSumOp`.
    """

    def __init__(self, allows_two_qubit_reduction: bool = False):
        """
        Args:
            allows_two_qubit_reduction: Set if mapper will create known symmetry such that the
            number of qubits in the mapped operator can be reduced accordingly.
        """
        self._allows_two_qubit_reduction = allows_two_qubit_reduction

```
which looks almost like what is seen in the docs. Apidoc is misunderstanding the word _number_ as an arg datatype
# **Fix**
To fix this, the appropriate format must be given to ```QubitMapper```'s docstring.

woodsp-ibm:
Fixes #278

Fix formatting of mapper documentation. Also since concrete class constructors were not supposed to expose the base class constructor, but pass up a fixed value of whether it allowed 2q reduction or not, added empty constructor to derived classes and called super() accordingly. This was also necessary to fix base class constructor documentation from incorrectly becoming its docstring - which it did and would fail to format correctly around args. As the class and init doctrings are merged and I needed an init docstring in order to avoid Sphinx errors, since there was only a very short class docstring I moved it to the init. That created pylint errors about missing docstring so since we do not need to see the same thing twice, if I copied it in there, and its not obvious what to say more, I just suppressed the errors for the 5 mapper classes involved.

Also takes care of some other formatting where things were labelled class: with only one colon.


pbark (approved):
Thanks for taking care of this @woodsp-ibm ! LGTM!

mrossinek (approved):

eendebakpt:
@woodsp-ibm I was using the `BravyiKitaevMapper` with option `allows_two_qubit_reduction=True`. With this PR that option is not available any more. I do not completely understand the reasoning in the comment of  the PR. Can we add the option again, or is there another way to use the `BravyiKitaevMapper` with two qubit reduction?

woodsp-ibm:
It was not intended to allow an end user to change the allows_two_qubit_reduction since thats an intrinsic property of the mapper itself. The base class was given a way to set it from the parent but by having the default there and not overriding the constructor it ended up allowing it to be set externally and was documented that way too.

Now the two qubit reduction removes symmetries that are known to be there from the Parity Mapping. I am not sure its guaranteed to work with BK mapper.

PR #283, opened 2021-07-21
Fix #279

### LINKED ISSUE #279: Failed to import qiskit_nature.transformers FreezeCoreTransformer
The `transformers` were relocated under `second_quantization` and some deprecation logic was put in place to allow them to be used from the version 0.1.0 location with a deprecation warning. I was looking at some code which was posted describing an issue, written to 0.1.0, and it fails to import the transformers.
```
>>> from qiskit_nature.transformers import FreezeCoreTransformer, ActiveSpaceTransformer
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'FreezeCoreTransformer'
```

I do not see any unit tests using the code from the 0.1.0 location to ensure it works. And its not used by any notebooks either where we should catch any compatibility issues.

mrossinek:
### Summary

Closes #279

### Details and comments

I am currently unable to reproduce #279 locally. Hopefully the CI will catch more errors, now that the unittests are duplicated to also test the legacy code location.

### DIFF
--- a/qiskit_nature/transformers/__init__.py
+++ b/qiskit_nature/transformers/__init__.py
@@ -23,28 +23,10 @@ Transformers (:mod:`qiskit_nature.transformers`)
 
 """
 
-from importlib import import_module
-from ..deprecation import DeprecatedType, warn_deprecated_same_type_name
+from .active_space_transformer import ActiveSpaceTransformer
+from .freeze_core_transformer import FreezeCoreTransformer
 
-deprecated_names = [
+__all__ = [
     "ActiveSpaceTransformer",
-    "BaseTransformer",
     "FreezeCoreTransformer",
 ]
-
-
-def __getattr__(name):
-    if name in deprecated_names:
-        warn_deprecated_same_type_name(
-            "0.2.0",
-            DeprecatedType.PROPERTY,
-            name,
-            f"from {__name__}.second_quantization as a direct replacement",
-        )
-        module = import_module(".second_quantization", __name__)
-        return getattr(module, name)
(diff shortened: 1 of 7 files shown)


woodsp-ibm:
When I encountered the issue referenced it was under Python 3.6. The unit test here under 3.6 seems to fail similarly from what I can see.

woodsp-ibm:
I see rather than copy the code wholesale you did a new class that subclasses the one that was moved. As long as the new function does not change (I am thinking of discussion where transformers are more applied to properties) I guess that works too.

mrossinek:
> I see rather than copy the code wholesale you did a new class that subclasses the one that was moved. As long as the new function does not change (I am thinking of discussion where transformers are more applied to properties) I guess that works too.

Yes the change of interface will need to be handled in #263. I just noticed that (while I do keep the legacy transformers which take a `QMolecule` as input, I do not keep them as also returning a `QMolecule`. We may need to rethink that before merging that PR.

woodsp-ibm:
>  We may need to rethink that before merging that PR.

Users existing code - where they are using public APIs - ought to continue to work while they have time to transition to the new way. Of course this may have implications to our stack around existing transformers being passed in versus new ones.

mrossinek:
Do you want to merge this PR as is anyways? Or shall I directly update #263 and include the fix for #279 there?

woodsp-ibm:
I think this is an improvement over what is there currently so merging it seems fine. It may not be the final solution, that's fine

woodsp-ibm (approved):

PR #284, opened 2021-07-21
Add link to repo tutorial on html tutorial prolog

manoelmarques:
### Summary
When building the tutorial html, create a prolog with the link back to the tutorial in github.
This link will always point to the stable branch so if building from the main branch, the link will be broken.
That should not be a problem as we always upload docs from stable.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -63,17 +63,18 @@ rst_prolog = """
 nbsphinx_prolog = """
 {% set docname = env.doc2path(env.docname, base=None) %}
 .. only:: html
-
+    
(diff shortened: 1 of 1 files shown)


woodsp-ibm (approved):
So the link, when built off main, ends up pointing to where they will be at the next stable version; and what we publish to qiskit documentation from stable will be correct and point to stable. It looks fine to me with a link there (rather than some generic msg), a link like the core qiskit doc tutorials do, whereby an end user can easily go get the source notebook to run themselves if they so wish.

PR #287, opened 2021-07-23
Molecule Driver

### LINKED ISSUE #165: Extract `MoleculeDriver`

As part of #147 I propose to extract the common `Molecule`-based driver initialization into a separate driver implementation. As suggested in the initial proposal, this could look similar to the following:
```python
class MoleculeDriver:
   def __init__(self, molecule, driver_type, driver_kwargs):
      # intitialize stuff

   def run(self):
      driver_cls = getattr(qiskit_nature.drivers, self._driver_type)
      driver = driver_cls.from_molecule(self._molecule)
      for key, value in self._driver_kwargs:
         setattr(driver, key, value)
      return driver.run()
```
More details will need to be figured out as we proceed with the actual implementation in the upcoming weeks.

manoelmarques:
### Summary
Fixes #165

### Details and comments
The native drivers were refactored to separate the molecule handling code and have a class factory method `from_molecule`.

Classes `FermionicMoleculeDriver` and `BosonicMoleculeDriver` created.

The drivers were updated to use MissingOptionalLibraryError when not installed. The existent unit tests were simplified with the decorator `requires_extra_library`.


docs/tutorials/03_ground_state_solvers.ipynb
```diff
@@ -22,21 +22,37 @@
...
+     ]
+    },
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/manoel/opt/anaconda3/envs/Qiskitenv/lib/python3.8/site-packages/pyscf/lib/misc.py:46: H5pyDeprecationWarning: Using default_file_mode other than 'r' is deprecated. Pass the mode to h5py.File() instead.\n",
+      "  h5py.get_config().default_file_mode = 'a'\n"
+     ]
+    }
```
mrossinek, line 42: Since this h5py deprecation from within PySCF is a known issue (#248, https://github.com/pyscf/pyscf/issues/1003) I am wondering whether we should catch it (or at least locally in this tutorial)? :thinking:

docs/tutorials/03_ground_state_solvers.ipynb
```diff
@@ -241,15 +257,15 @@
...
+      "* Vibrational ground state energy (cm^-1): (1e-11+0j)\n",
       "The number of occupied modals is\n",
       "- Mode 0: [0.0, 0.0, 0.0, 0.0]\n",
       "\n",
       "\n",
       "\n",
       "=== GROUND STATE ENERGY ===\n",
       " \n",
-      "* Vibrational ground state energy (cm^-1): (2553.464851175474+0j)\n",
+      "* Vibrational ground state energy (cm^-1): (2555.376153240112+0j)\n",
```
mrossinek, line 268: I am curious why this value changed so significantly.. I don't think it's this PRs fault but maybe we should look into that!

qiskit_nature/drivers/second_quantization/bosonic_molecule_driver.py
```diff
@@ -0,0 +1,105 @@
...
+
+class BosonicDriverType(Enum):
+    """Bosonic Driver Type."""
+
+    AUTO = "auto"
+    GAUSSIAN_FORCES = "gaussian_forces"
+
+    @staticmethod
+    def driver_class_from_type(driver_type: "BosonicDriverType"):
+        """Get driver class from driver type"""
```
mrossinek, line 37: Missing `Args` and `Raises` documentation
Missing returned mypy type

qiskit_nature/drivers/second_quantization/fermionic_driver.py
```diff
@@ -42,6 +43,30 @@ class FermionicDriver(BaseDriver):
     Base class for Qiskit Nature's fermionic drivers.
     """
 
+    @abstractmethod
```
mrossinek, line 46: Why do you declare this as abstract?

qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
```diff
@@ -45,8 +44,6 @@ def __init__(
         self,
         jcf: Union[str, List[str]] = B3YLP_JCF_DEFAULT,
         logfile: Optional[str] = None,
-        molecule: Optional[Molecule] = None,
-        basis: str = "sto-3g",
```
mrossinek, line 49: Can we remove these args in all drivers without proper deprecation?

qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
```diff
@@ -56,33 +53,66 @@ def __init__(
...
+    def to_driver_basis(basis: str) -> Any:
+        """convert basis to a driver acceptable basis"""
+        if basis == "sto3g":
+            return "sto-3g"
+        return basis
+
+    @staticmethod
+    def check_installed():
+        """check if installed"""
+        check_valid()
```
mrossinek, line 114: Why not move the code from `check_valid()` into here?

qiskit_nature/drivers/second_quantization/base_driver.py
```diff
@@ -78,9 +63,8 @@ def molecule(self) -> Optional[Molecule]:
     @molecule.setter
     def molecule(self, value: Molecule) -> None:
         """set molecule"""
-        if not self.supports_molecule:
-            raise QiskitNatureError("Driver doesn't support molecule.")
         self._molecule = value
+        self._supports_molecule = self.molecule is not None
```
mrossinek, line 67: I don't really understand the point of this line..
Do we even still need `supports_molecule` or can we simply rely on whether `from_molecule` is implemented?

mrossinek (commented):
Thanks @manoelmarques! I have some (mostly minor) comments :slightly_smiling_face: Otherwise this LGTM :+1:

qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
manoelmarques: Those drivers haven't been released yet. There is no need to deprecate unreleased code.

qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
manoelmarques: check_valid() is part of gaussian_utils and use local variables that are used also by another function there (run_g16).
Moving it to the two Gaussian drivers would mean copy all this code to each driver.

qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
mrossinek: Ah because we relocated them, right. But then we should adapt the deprecation messages in the old locations. Otherwise they suggest the drivers can be used as direct replacements (which now will no longer be the case, depending on the use case)

qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
mrossinek: Ah I hadn't noticed that. Thanks!

docs/tutorials/03_ground_state_solvers.ipynb
manoelmarques: In the issue above: "This has been fixed in Pyscf v2.0.0.a, although it's not on Pypi yet."
So they already addressed it and it should be in their next release.

qiskit_nature/drivers/second_quantization/bosonic_molecule_driver.py
manoelmarques: Added the missing docs and type

qiskit_nature/drivers/second_quantization/fermionic_driver.py
manoelmarques: Removed the abstract

qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
manoelmarques: Fixed deprecation msgs for the drivers that changed

qiskit_nature/drivers/second_quantization/base_driver.py
manoelmarques: Fixed supports_molecule as the above was bogus. Since other methods in Nature use argument of type BaseDriver and access the molecule property, it is still needed so that they fail when try to access the property for a driver that doesn't support it.

docs/tutorials/03_ground_state_solvers.ipynb
manoelmarques: I ran locally from MacOS but those are similar to results we have been getting from 'main'. Looking at the tutorials run from a recent nightly CI (Ubuntu):
```
Vibrational ground state energy (cm^-1): (2555.376153240098-0j)
```

docs/tutorials/03_ground_state_solvers.ipynb
mrossinek: I was just thinking that the tutorials could look cleaner in the meantime. v2.0.0a has been around for a while and I don't see a timeline for when 2.0.0 will be published on PyPI..

But if you think it's fine as is, then it's okay :+1:

docs/tutorials/03_ground_state_solvers.ipynb
mrossinek: Yeh this could have happened during the migration from Aqua. But I think nonetheless we should investigate when/why this differences started to occur. Maybe a bisection will tell :thinking: (But this is unrelated to this PR for sure!)

qiskit_nature/drivers/second_quantization/fermionic_molecule_driver.py
```diff
@@ -0,0 +1,169 @@
...
+
+    @property
+    def driver_kwargs(self) -> Optional[Dict[str, Any]]:
+        """return driver kwargs"""
+        return self._driver_kwargs
+
+    @driver_kwargs.setter
+    def driver_kwargs(self, value: Optional[Dict[str, Any]]) -> None:
+        """set driver kwargs"""
+        self._driver_kwargs = value
```
mrossinek, line 160: I am wondering whether it would make sense to define a small `MoleculeDriver` interface which would take care of all these properties (which are identical in here and in `BosonicMoleculeDriver`). This would also allow type checking against the interface in case we ever need to differentiate a `MoleculeDriver` from another one (say e.g. in the `BopesSampler`)

qiskit_nature/drivers/second_quantization/gaussiand/gaussiandriver.py
```diff
@@ -49,69 +49,77 @@ def __init__(
...
 
-    def run(self) -> QMolecule:
-        if self.molecule is not None:
-            cfg = self._from_molecule_to_str()
-        else:
-            cfg = self._config
+        return GaussianDriver(cfg1 + cfg2 + cfg3)
+
+    @staticmethod
+    def to_driver_basis(basis: str) -> Any:
```
mrossinek, line 110: Why doesn't this return a `str`?

qiskit_nature/drivers/second_quantization/pyquanted/pyquantedriver.py
```diff
@@ -37,6 +39,14 @@ class BasisType(Enum):
     B631G = "6-31g"
     B631GSS = "6-31g**"
 
+    @staticmethod
+    def type_from_string(basis: str):
```
mrossinek, line 43: What's the return type here?

mrossinek (commented):
Some minor comments but otherwise this is looking good :+1:

qiskit_nature/drivers/second_quantization/fermionic_molecule_driver.py
manoelmarques: I had discussed that with Steve but it would need to be a separate interface not derived from BaseDriver. so it would be 2 base classes for those and you wouldn't be able to use as a typehint anyway because it is not a BaseDriver

qiskit_nature/drivers/second_quantization/gaussiand/gaussiandriver.py
manoelmarques: Added docstring and typehints

qiskit_nature/drivers/second_quantization/pyquanted/pyquantedriver.py
manoelmarques: Added docstring and typehints

qiskit_nature/drivers/second_quantization/fermionic_molecule_driver.py
mrossinek: What is the problem with using two base classes?

qiskit_nature/drivers/second_quantization/electronic_structure_driver.py
```diff
@@ -37,9 +37,9 @@ class MethodType(Enum):
     UKS = "uks"
 
 
-class FermionicDriver(BaseDriver):
+class ElectronicStructureDriver(BaseDriver):
```
mrossinek, line 40: I think the deprecation of `qiskit_nature.drivers.fermionic_driver.FermionicDriver` is still missing. Same for the `BosonicDriver` on that level.

qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class ElectronicStructureDriverType(Enum):
+    """Electronic structure Driver Type."""
+
+    AUTO = "auto"
+    PYSCF = "pyscf"
+    PSI4 = "psi4"
+    PYQUANTE = "pyquante"
+    GAUSSIAN = "gaussian"
```
mrossinek, line 39: I am not sure whether that will work but rather than having the values be strings, could we not set the classes here directly?
Below you could then do something along the lines of:
```python
if driver_type == ElectronicStructureDriverType.AUTO:
    ...
else:
    driver_class = driver_type.value
    driver_class.check_installed()
```

qiskit_nature/drivers/second_quantization/fermionic_molecule_driver.py
mrossinek: The idea was also whether we could extract the duplication of the attributes like `basis`, `method`, etc.
But I guess that is fine as is..

qiskit_nature/drivers/second_quantization/electronic_structure_driver.py
manoelmarques: I added the deprecations

qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
mrossinek: Thanks! I see now that I was simplifying it a bit too much because my proposal would have involved importing all drivers :sweat_smile: But the current approach looks good to me as it simplifies future addition of drivers :+1:

mrossinek (dismissed):
I think this looks good to me now! Thanks @manoelmarques :+1: :tada: 

Before merging let's await a second review though

qiskit_nature/drivers/second_quantization/gaussiand/gaussiandriver.py
```diff
@@ -49,69 +49,83 @@ def __init__(
...
+            basis: The basis set to be used
+        Returns:
+            driver acceptable basis
+        """
+        if basis == "sto3g":
+            return "sto-3g"
+        return basis
 
+    @staticmethod
+    def check_installed():
```
woodsp-ibm, line 123: Since check_valid is `-> None` should the return type here be that too otherwise this almost feels like a boolean. Maybe it should list what it does as a failure with a Raises here? (Same comment I think applies to other places that are similar)

qiskit_nature/drivers/second_quantization/pyscfd/pyscfdriver.py
```diff
@@ -209,6 +200,26 @@ def spin(self, spin: int) -> None:
...
+        return self._basis
+
+    @basis.setter
+    def basis(self, value: str) -> None:
+        """set basis"""
+        self._basis = value
+
+    @property
+    def method(self) -> MethodType:
+        """return Hartree-Fock method"""
```
woodsp-ibm, line 215: For PySCF it now supports DFT methods too. We can shorten it to say `Returns method` or have it be a be more than stating what the code already tells you by saying `Returns Hartree-Fock/Kohn-Sham method`  @mrossinek any thought

woodsp-ibm:
Aside from those minor comments it looks good to me.

qiskit_nature/drivers/second_quantization/pyscfd/pyscfdriver.py
manoelmarques: I added the typehint return and completed the docstrings with the exception

qiskit_nature/drivers/second_quantization/gaussiand/gaussiandriver.py
manoelmarques: I added the typehint return and completed the docstring with the exception

qiskit_nature/drivers/second_quantization/pyscfd/pyscfdriver.py
mrossinek: Yes that is true. I think the suggested docstring should be fine :+1:

manoelmarques: Changed the docstring

mrossinek (approved):
This LGTM! Thanks, @manoelmarques! :slightly_smiling_face: 

Edit: well I should have checked the CI first.. :see_no_evil: I will re-approve once the spelling is fixed :+1:

mrossinek (approved):

PR #290, opened 2021-07-28
Allow `Optimizer`s as input to `VQEProgram`

### LINKED ISSUE #289: `VQEProgram` to support optimizers classes for optimizers instead of dictionaries

Since Terra 0.18 optimizers are serializable and supported by the runtime VQE script. The VQEProgram must be updated to support them, too. Currently it stil expects the optimizer as dictionary.

Cryoris:
### Summary

Fixes #289.

### Details and comments

Dictionaries are still supported since some tutorials first need to be updated to use optimizers.


woodsp-ibm:
So this is a feature rather than any bug fix then. I see you put a high priority label on the issue which is why I am asking. As a feature this would go out in the next release unless somehow its creating critical problems/issues out there now. I assume you will be updating the optimization repo equivalently. If we need these to stay the same out there so users can use the one in Nature and/or the one in Optimization the same way then I guess things need to be coordinated/sorted out, since I presently would have expected a Nature release in advance of the next Optimization one - unless we start talking bug fix releases.

releasenotes/notes/vqeprogram-optimizers-cd23307c10fcd15b.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
```
Cryoris, line 2:
```suggestion
fixes:
```

Cryoris:
It would be better to get this out before the next feature release, right, so it seems more like a bugfix 😉 I synced with @pbark and looks like we can do a minor release soon.

pbark (approved):
This LGTM!

mrossinek (approved):

woodsp-ibm (approved):

PR #292, opened 2021-07-29
[Stable] Prepare 0.1.5 release

manoelmarques:


manoelmarques:
I added a `hold` status because would like to release Optimization together with this.

woodsp-ibm (approved):

PR #297, opened 2021-08-02
ElectronicStructureMoleculeDriver checks if driver supports method

manoelmarques:
### Summary
Before creating a driver, the Molecule Driver will check the given method and raise a proper exception.
This fixes the case of AUTO option that should chose the first installed driver that supports the given method


qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
```diff
@@ -38,34 +39,43 @@ class ElectronicStructureDriverType(Enum):
...
+                        driver_class = ElectronicStructureDriverType.driver_class_from_type(
+                            item, method
+                        )
                         break
                     except MissingOptionalLibraryError as ex:
-                        if missing_error is None:
-                            missing_error = ex
+                        if error is None:
+                            error = ex
+                    except NotSupportedMethodError as ex:
```
mrossinek, line 74: You could also put these on one line if you wanted to:
```python
except (MissingOptionalLibraryError, NotSupportedMethodError) as ex:
```

qiskit_nature/drivers/second_quantization/gaussiand/gaussiandriver.py
```diff
@@ -129,6 +131,19 @@ def check_installed() -> None:
...
+        """
+        Checks that Gaussian supports this method.
+        Args:
+            method: Method type
+
+        Raises:
+            NotSupportedMethodError: If method not supported.
+        """
+        if method in [MethodType.RKS, MethodType.ROKS, MethodType.UKS]:
+            raise NotSupportedMethodError(f"Invalid Gaussian method {method.value}.")
```
mrossinek, line 145: I am wondering what the more scalable approach is. Although I find it unlikely that there will be many more methods added in the future, _if_ we do add another one, in this current implementation you would have to go into each driver and add the new method to this list (because most likely most drivers won't have it).
Thus, it may be favorable to check the method which _are_ supported and raise an error if the current method is not in that set. In this way, future updates are likely fewer in number :thinking: 

The same can obviously be said for all other drivers, too.

qiskit_nature/exceptions.py
```diff
@@ -19,3 +19,9 @@ class QiskitNatureError(QiskitError):
     """Class for errors returned by Qiskit's Nature module."""
 
     pass
+
+
+class NotSupportedMethodError(QiskitNatureError):
```
mrossinek, line 24: Just an opinion: I think `UnsupportMethodError` could read more nicely. What do you think?

mrossinek (changes requested):
Thanks @manoelmarques! I left some minor comments as usual :slightly_smiling_face:

mrossinek:
Ah and I had one more thought: maybe we can add some unittests which ensure that unsupported methods get caught properly

qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
manoelmarques: Changed it, thanks.

qiskit_nature/drivers/second_quantization/gaussiand/gaussiandriver.py
manoelmarques: Changed.

qiskit_nature/exceptions.py
manoelmarques: Renamed

mrossinek (commented):
This looks basically good to me. However, I just noticed that we never included a reno in the `MoleculeDriver` PR.. Shall we add one here, instead?

mrossinek (approved):
Awesome, thanks @manoelmarques!

Issue #302, opened 2021-08-04
HDF5 integration for the `Property` framework


mrossinek:
### What is the expected behavior?

The `Property` framework will be mostly finalized and properly integrated with the drivers by closing #264 (PR link TBD).
Due to time constraints, the HDF5 integration will not become part of Nature's upcoming 0.2.0 release. The most important reason that it will require some more planning, which should be done in this issue (to be updated later).

For the time being, the (soon to be) deprecated `qiskit_nature.drivers.QMolecule` object can be stored and loaded to/from HDF5 files and subsequently converted to the `Property` framework as follows:
```python
from qiskit_nature.drivers import QMolecule
q_mol = QMolecule("path/to/file.hdf5")
q_mol.load()

from qiskit_nature.properties.second_quantization.electronic import ElectronicStructureDriverResult
property = ElectronicStructureDriverResult.from_legacy_driver_result(q_mol)
```
The `HDF5Driver` from the `qiskit_nature.drivers.second_quantization` module may be used for the exact same purpose as shown above.

PR #303, opened 2021-08-04
[WIP] Property Finalization

### LINKED ISSUE #148: QMolecule Design

### LINKED ISSUE #264: Leverage modular `DriverResult` directly on `Driver` level

The third (and final) step of #148 suggests to replace the return objects for all `qiskit_nature.drivers` with the new modular `Property` framework. I.e.:
```python
class BaseDriver:
    def run(self) -> DriverResult:
        ...

class FermionicDriver:
    def run(self) -> ElectronicDriverResult:
        ...

class BosonicDriver:
    def run(self) -> VibrationalDriverResult:
        ...
```

:warning: Names not final!

The implementation of these `CompositeProperty` objects (as proposed in #263) allows a straight forward migration using the `from_legacy_driver_result` methods.

The aforementioned PR also lists some more details which need to be addressed as part of closing this issue.

mrossinek:
### Summary

Closes #264
Closes #148 

### Details and comments

Task list partially carried over from #263 
- [x] direct integration of `DriverResult` with drivers (i.e. do not return `QMolecule` as intermediate storage container)
- [ ] deprecation of `BaseProblem.molecule_data*` properties
- ~implementation of HDF5 interface (storing and loading similar to `QMolecule`)~ (cf. #302)
- [ ] release note
- [ ] tutorial (if time permits) [WIP]
- [ ] documentation updates:
  - [ ] https://github.com/Qiskit/qiskit-nature/pull/263#discussion_r677840007
  - [ ] https://github.com/Qiskit/qiskit-nature/pull/263#discussion_r677711651
  - [ ] generally double check correct contents and rendering


qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
```diff
@@ -67,15 +68,18 @@ def second_q_ops(self) -> List[SecondQuantizedOp]:
...
                 ElectronicStructureDriverResult.from_legacy_driver_result(qmol_transformed)
             )
+        elif isinstance(driver_result, QMolecule):
+            self._properties_transformed = (
+                ElectronicStructureDriverResult.from_legacy_driver_result(driver_result)
+            )
         else:
-            prop = ElectronicStructureDriverResult.from_legacy_driver_result(self._molecule_data)
-            self._properties_transformed = self._transform(prop)
+            self._properties_transformed = self._transform(driver_result)
```
mrossinek, line 82: This still needs to be cleaned up. In order for that to happen I need to figure out how to properly deprecate the `BaseProblem.molecule_data*`..
@woodsp-ibm @manoelmarques do you have some ideas on how to proceed with that?

Since they are part of the public API I am unsure if I can just change/delete them. They won't exist in their format any longer and they were really just auxiliary access points to the driver results (which are now `Property` objects).
In #263 I added the `BaseProblem.roperties_transformed` attribute as a temporary workaround. Ideally I think we should have something like this:
```python
class BaseProblem:
    @property
    def driver_result(self) -> GroupedProperty:
        return self._driver_result

    @property
    def transformed_driver_result(self) -> GroupedProperty:
        return self._transformed_driver_result
```
Please let me know your opinions :slightly_smiling_face:

qiskit_nature/problems/second_quantization/vibrational/vibrational_structure_problem.py
```diff
@@ -62,9 +63,22 @@ def second_q_ops(self) -> List[SecondQuantizedOp]:
...
+        elif isinstance(driver_result, WatsonHamiltonian):
+            properties_transformed = VibrationalStructureDriverResult.from_legacy_driver_result(
+                driver_result
+            )
+        else:
+            properties_transformed = self._transform(driver_result)
+
+        self._properties_transformed = cast(
+            VibrationalStructureDriverResult, properties_transformed
+        )
```
mrossinek, line 81: The same as in the electronic case also applies here

qiskit_nature/properties/second_quantization/driver_metadata.py
```diff
@@ -29,3 +29,5 @@ def __init__(self, program: str, version: str, config: str) -> None:
         self.program = program
         self.version = version
         self.config = config
+
+    # TODO: implement __str__
```
mrossinek, line 33: Will be part of this PR

qiskit_nature/properties/second_quantization/electronic/angular_momentum.py
```diff
@@ -49,7 +47,7 @@ def __str__(self) -> str:
         return "\n".join(string)
 
     @classmethod
-    def from_legacy_driver_result(cls, result: LegacyDriverResult) -> "AngularMomentum":
+    def from_legacy_driver_result(cls, result: Any) -> "AngularMomentum":
```
mrossinek, line 50: Change `Any` -> `Type`? Need to double check

qiskit_nature/properties/second_quantization/electronic/angular_momentum.py
```diff
@@ -62,7 +60,10 @@ def from_legacy_driver_result(cls, result: LegacyDriverResult) -> "AngularMoment
         Raises:
             QiskitNatureError: if a WatsonHamiltonian is provided.
         """
-        cls._validate_input_type(result, LegacyElectronicStructureDriverResult)
+        # pylint: disable=import-outside-toplevel
+        from qiskit_nature.drivers import QMolecule
```
mrossinek, line 64: I need to double check whether the "outside-toplevel" imports are still necessary (I was running into some cyclic import issues before).
In any case, I still need to wrap the import in all occurrences with filters for the `DeprecationWarning`s

qiskit_nature/properties/second_quantization/electronic/bases/electronic_basis_transform.py
```diff
@@ -45,3 +45,5 @@ def __init__(
         self.final_basis = final_basis
         self.coeff_alpha = coeff_alpha
         self.coeff_beta = coeff_alpha if coeff_beta is None else coeff_beta
+
+    # TODO: implement __str__
```
mrossinek, line 49: Will be part of this PR

qiskit_nature/properties/second_quantization/electronic/electronic_structure_driver_result.py
```diff
@@ -28,6 +25,9 @@
 from .particle_number import ParticleNumber
 from .types import GroupedElectronicProperty
 
+if TYPE_CHECKING:
+    from qiskit_nature.molecule import Molecule
```
mrossinek, line 29: I need to double check if the cyclic imports still occur if this is not wrapped

qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals.py
```diff
@@ -86,6 +86,7 @@ def __str__(self) -> str:
 
     @staticmethod
     def _render_matrix_as_sparse_list(matrix) -> List[str]:
+        # TODO: handle all-zero matrix
```
mrossinek, line 89: Will be part of this PR

test/drivers/second_quantization/fcidumpd/test_driver_methods_fcidump.py
```diff
@@ -99,6 +99,7 @@ def test_oh_with_atoms(self):
         self._assert_energy(result, "oh")
 
 
+@unittest.skip("Until the Property framework supports log")
```
mrossinek, line 102: The `log` integration should still become part of this PR

test/drivers/second_quantization/hdf5d/test_driver_hdf5_save.py
```diff
@@ -21,6 +21,7 @@
 from qiskit_nature.drivers.second_quantization import HDF5Driver
 
 
+@unittest.skip("Until the Property framework supports HDF5 storage")
```
mrossinek, line 24: FYI: #302

mrossinek (commented):
Since I cannot comment on a changed binary file:
The file `test/transformers/second_quantization/electronic/LiH_sto3g_reduced.hdf5` had to be updated because the stored `QMolecule` in that file was missing its `num_molecular_orbitals` attribute.

qiskit_nature/properties/second_quantization/electronic/angular_momentum.py
mrossinek: This is indeed no longer necessary

qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
manoelmarques: You can’t delete the properties if you intend to deprecate. Is it possible to get the molecule data by other means to keep the properties or is it lost ?

mrossinek: `molecule_data` used to be `QMolecule` or `WatsonHamiltonian`. Since these are now deprecated, they will no longer occur within the stack, if a user uses one of the new drivers in `qiskit_nature.drivers.second_quantization`.
I could deprecate the properties and change their type to be `Optional[...]` but I am not sure if that is allowed? [Technically they probably should have been that though, because the `__init__` sets them to `None`...]

And to be clear: the deprecated drivers are still supported by this stack but right now they will not store these properties (which I could change ofcourse) :+1:

woodsp-ibm: It would be nice if using things with the existing drivers, for users that have code to what is currently released, works the same albeit deprecated. Switching to the newer drivers then other code for properties can come about. At that level things can be more separated - in the problem I guess it has the unenviable task of in the interim allowing the old way to work and supporting the new way too if thats indeed doable.

mrossinek: I have updated the code and we now have the following:
* `BaseProblem.molecule_data*` are deprecated in favor of `BaseProblem.grouped_property*`

We have some internal logic as follows:

#### Legacy Driver + Legacy Transformers
Works as expected. Deprecation warnings will be raised by those classes themselvers.
The following properties will be set:
* `molecule_data`
* `molecule_data_transformed`
* `grouped_property`
* `grouped_property_transformed`

#### Legacy Driver + New Transformers
Works. A `UserWarning` will be raised, suggesting to look into the new drivers.
The following properties will be set:
* `molecule_data`
* ~~`molecule_data_transformed`~~ (remains `None`)
* `grouped_property`
* `grouped_property_transformed`

#### New Driver + Legacy Transformers
Impossible! A `QiskitNatureError` will be raised.

#### New Driver + New Transfomrers
Works as expected.
The following properties will be set:
* ~~`molecule_data`~~ (remains `None`)
* ~~`molecule_data_transformed`~~ (remains `None`)
* `grouped_property`
* `grouped_property_transformed`


Finally, if a user mixes legacy and new transformers, a `QiskitNatureError` will
be raised, too.

mrossinek:
I am marking this as ready for review. Early next week, I am going to tackle the outstanding tasks:
* [ ] reno
* [ ] documentation completion + double checking
* [ ] tutorial completion

qiskit_nature/drivers/second_quantization/pyscfd/pyscfdriver.py
```diff
@@ -518,32 +535,46 @@ def run_pyscf(self) -> None:
...
+
+        return driver_result
+
+    def _populate_driver_result_molecule(
+        self, driver_result: ElectronicStructureDriverResult
+    ) -> None:
+        coords = self._mol.atom_coords()
+        geometry = []
+        for i, xyz in enumerate(coords):
+            geometry.append((self._mol.atom_pure_symbol(i), list(xyz)))
```
dlasecki, line 566:
```suggestion
       geometry = [(self._mol.atom_pure_symbol(i), list(xyz)) for i, xyz in enumerate(coords)]
```

qiskit_nature/drivers/second_quantization/pyscfd/pyscfdriver.py
```diff
@@ -610,78 +616,77 @@ def _populate_q_molecule_orbitals(self, q_mol: QMolecule) -> None:
...
+        electronic_energy.overlap = OneBodyElectronicIntegrals(
+            ElectronicBasis.AO,
+            (self._calc.get_ovlp(), None),
+        )
+
+        orbs_energy, orbs_energy_b = self._extract_mo_data("mo_energy")
+        if orbs_energy_b is not None:
+            electronic_energy.orbital_energies = np.asarray((orbs_energy, orbs_energy_b))
+        else:
+            electronic_energy.orbital_energies = np.asarray(orbs_energy)
```
dlasecki, line 679:
```suggestion
       orbital_energies = (orbs_energy, orbs_energy_b) if orbs_energy_b is not None else orbs_energy
       electronic_energy.orbital_energies = np.asarray(orbital_energies)
```

qiskit_nature/properties/second_quantization/electronic/bases/electronic_basis_transform.py
```diff
@@ -45,3 +45,21 @@ def __init__(
...
+        string += self._render_coefficients(self.coeff_beta)
+        return "\n".join(string)
+
+    @staticmethod
+    def _render_coefficients(coeffs) -> List[str]:
+        string = []
+        nonzero = coeffs.nonzero()
+        for value, *indices in zip(coeffs[nonzero], *nonzero):
+            string += [f"\t{indices} = {value}"]
+        return string
```
dlasecki, line 65:
```suggestion
        nonzero = coeffs.nonzero()
        return [f"\t{indices} = {value}" for value, *indices in zip(coeffs[nonzero], *nonzero)]
```

qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals.py
```diff
@@ -80,8 +80,12 @@ def __str__(self) -> str:
             string += self._render_matrix_as_sparse_list(self._matrices)
         else:
             for title, mat in zip(self._matrix_representations, self._matrices):
+                rendered_matrix = self._render_matrix_as_sparse_list(mat)
                 string += [f"\t{title}"]
-                string += self._render_matrix_as_sparse_list(mat)
+                if not rendered_matrix:
+                    string[-1] += " is all zero"
```
dlasecki, line 86: Is it ever possible that string is []?

mrossinek: No this is not possible. See 2 lines above where I do `string += [f"\t{title}"]`. This ensures that `string[-1]` is always existent.

qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
```diff
@@ -146,7 +147,7 @@ def check_method_supported(method: MethodType) -> None:
         if method not in [MethodType.RHF, MethodType.ROHF, MethodType.UHF]:
             raise UnsupportMethodError(f"Invalid PSI4 method {method.value}.")
 
-    def run(self) -> QMolecule:
+    def run(self) -> ElectronicStructureDriverResult:
```
pbark, line 150: Docstring missing

qiskit_nature/properties/second_quantization/driver_metadata.py
```diff
@@ -29,3 +29,11 @@ def __init__(self, program: str, version: str, config: str) -> None:
         self.program = program
         self.version = version
         self.config = config
+
+    def __str__(self) -> str:
```
pbark, line 33: this is useful and we needed it for logging the driver info. Thanks

pbark (dismissed):
There are some minor changes needed, but for me this is almost ready. Thanks for the great work @mrossinek

qiskit_nature/drivers/second_quantization/fcidumpd/fcidumpdriver.py
```diff
@@ -12,11 +12,21 @@
...
+    ElectronicEnergy,
+    ParticleNumber,
+)
+from qiskit_nature.properties.second_quantization.electronic.bases import ElectronicBasis
+from qiskit_nature.properties.second_quantization.electronic.integrals import (
+    OneBodyElectronicIntegrals,
+    TwoBodyElectronicIntegrals,
+)
+
+from ...qmolecule import QMolecule
```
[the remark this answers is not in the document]
mrossinek: Thanks! It appears to be left over from some way of specifying a list of atoms for this driver. I will remove it for now.

qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
```diff
@@ -138,10 +140,11 @@ def check_installed() -> None:
         """
         check_valid()
 
-    def run(self) -> WatsonHamiltonian:
+    def run(self) -> VibrationalStructureDriverResult:
```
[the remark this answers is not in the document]
mrossinek: These are inherited from the base class. I don't think we need to duplicate them. In some cases where additional information get's added (like fcidump or hdf5) they have been overwritten in the respective class.

qiskit_nature/drivers/second_quantization/fcidumpd/fcidumpdriver.py
```diff
@@ -103,19 +106,24 @@ def dump(
 
         Args:
             outpath: Path to the output file.
-            q_mol: QMolecule data to be dumped. It is assumed that the nuclear_repulsion_energy in
-                this QMolecule instance contains the inactive core energy.
+            driver_result: The ElectronicStructureDriverResult to be dumped. It is assumed that the
+                nuclear_repulsion_energy in its ElectronicEnergy property contains the inactive core
+                energy.
```
dlasecki, line 111:
```suggestion
            driver_result: The ElectronicStructureDriverResult to be dumped. It is assumed that the
                nuclear_repulsion_energy contains the inactive core
                energy in its ElectronicEnergy property .
```

qiskit_nature/drivers/second_quantization/psi4d/psi4driver.py
```diff
@@ -166,7 +167,7 @@ def run(self) -> QMolecule:
         )
 
         input_text += "sys.path = " + syspath + " + sys.path\n"
-        input_text += "from qiskit_nature.drivers.second_quantization.qmolecule import QMolecule\n"
+        input_text += "from qiskit_nature.drivers.qmolecule import QMolecule\n"
         input_text += '_q_molecule = QMolecule("{0}")\n'.format(Path(molecule.filename).as_posix())
```
dlasecki, line 171: Can we do it through join?

qiskit_nature/problems/second_quantization/vibrational/vibrational_structure_problem.py
```diff
@@ -47,9 +48,9 @@ def __init__(
         """
         Args:
             bosonic_driver: A bosonic driver encoding the molecule information.
-            transformers: A list of transformations to be applied to the molecule.
             num_modals: the number of modals per mode.
             truncation_order: order at which an n-body expansion is truncated
+            transformers: A list of transformations to be applied to the driver result.
```
dlasecki, line 53: Inconsistent letter capitalization.

qiskit_nature/properties/second_quantization/electronic/dipole_moment.py
```diff
@@ -100,11 +101,14 @@ def __init__(
...
         reverse_dipole_sign: bool = False,
-    ):
+    ) -> None:
         """
         Args:
             dipole_axes: a dictionary mapping Cartesian axes to DipoleMoment properties.
             dipole_shift: an optional dictionary of named dipole shifts.
+            nuclear_dipole_moment: the optional nuclear dipole moment.
+            reverse_dipole_sign: whether the sign of the electronic dipole components needs to be
+                reversed in order to match the nuclear dipole moment direction.
```
dlasecki, line 111:
```suggestion
            reverse_dipole_sign: indicates whether the sign of the electronic dipole components needs to be
                reversed in order to match the nuclear dipole moment direction.
```

qiskit_nature/properties/second_quantization/electronic/particle_number.py
```diff
@@ -33,16 +33,22 @@ class ParticleNumber(ElectronicProperty):
     Note that this Property serves a two purposes:
         1. it stores the expected number of electrons (`self.num_particles`)
         2. it is used to evaluate the measured number of electrons via auxiliary operators.
-           If this measured number does not match the expected number a warning will be logged.
+           If this measured number does not match the expected number it will be logged on the INFO
+           level.
```
dlasecki, line 37:
```suggestion
           If this measured number does not match the expected number, it will be logged on the INFO
           level.
```

qiskit_nature/drivers/molecule.py
```diff
@@ -70,6 +70,19 @@ def __init__(
 
         self._perturbations = None  # type: Optional[List[float]]
 
+    def __str__(self) -> str:
```
woodsp-ibm, line 73: It might have been convenient for this to simply have been the standard xyz format for a molecule definition - though that would have meant not having masses as I don't think that's part of it - or at least I guess it could be appended afterwards. One reason for mentioning using that is then the extra labelling for geometry, multiplicity etc would not be needed, since its standard, and this whole output would be more compact.

woodsp-ibm:
What do you intend with the PSi4 driver? The method of getting the data back across processes was the QMolecule which had no dependence on any else in Qiskit. At present its marked deprecated - depending on where you want to go with PSi4 driver maybe an easy solution is to have a copy (non-deprecated) of the QMolecule in with PSI4 driver so it uses that 'private' copy internally to the driver.

woodsp-ibm:
This formatting issue I guess was there from before - The GroupedProperty interpret() method the :class: is missing the `` around the classname  
```
def interpret(self, result: EigenstateResult) -> None:
        """Interprets an :class:~qiskit_nature.result.EigenstateResult in this property's context.
```     
i.e. should be
```
            Interprets an :class:`~qiskit_nature.result.EigenstateResult` in this property's context.
```

woodsp-ibm:
(I was looking through the html output hence the above and this). In Property it has text with underscores `_observables_` to presumably do italic - think with Sphinx this needs to be asterisks 
```
(i.e. the user can evaluate custom _observables_ by writing a class which can generate an operator out of a given set of raw data).
```

qiskit_nature/properties/second_quantization/second_quantized_property.py
```diff
@@ -13,66 +13,54 @@
...
+LegacyDriverResult = Union[QMolecule, WatsonHamiltonian]
 
 
 class SecondQuantizedProperty(Property):
     """The SecondQuantizedProperty base class.
 
-    A second-quantization property provides the logic to transform a raw data (as e.g. produced by a
+    A second-quantization property provides the logic to transform raw data (as e.g. produced by a
     `qiskit_nature.second_quantization.drivers.BaseDriver`) into a
     `qiskit_nature.operators.second_quantization.SecondQuantizedOp`.
```
woodsp-ibm, line 33:
```suggestion
    A second-quantization property provides the logic to transform raw data (as e.g. produced by a
    :class:`~qiskit_nature.second_quantization.drivers.BaseDriver`) into a
    :class:`~qiskit_nature.operators.second_quantization.SecondQuantizedOp`.
```
So these turn out as links in the doc rather than just highlighted text - and limited now to just the class name since it does not need to be so verbose with a link I think.

woodsp-ibm:
In second_quantization.electronic init file it has this
```
which constructs the primary Hamiltonian whose solution is the goal of the Quantum Algorithm.
The following auxiliary properties will be evaluated by default to provide further details of the
solution:
```
> will be evaluated by default

Is not strictly true - they are only evaluated (ie expectation values computed for them) when the underlying algorithm supports auxiliary operators.

qiskit_nature/properties/second_quantization/electronic/bases/electronic_basis.py
```diff
@@ -20,7 +20,7 @@ class ElectronicBasis(Enum):
 
     This ``Enum`` simply names the available electronic bases. The ``SO`` basis is the _special_
     basis into which an ``ElectronicEnergy`` must map its integrals before being able to perform the
```
woodsp-ibm, line 22:
```suggestion
    This ``Enum`` simply names the available electronic bases. The ``SO`` basis is the *special*
    basis into which an ``ElectronicEnergy`` must map its integrals before being able to perform the
```
I think that was intended to come out in italic font - while the underscore is interpreted that way for markdown text it does not do it for Sphinx

qiskit_nature/properties/second_quantization/second_quantized_property.py
woodsp-ibm: I see there are other places in the text that would benefit from being links rather than quoted text as they refer to classes etc that it would be nice to navigate too. Since the text existed from before I cannot annotate it like above as its too far from any changes. And rather than try to include them here I will just note this. It would be nice to have them as links but it can always be improved later if it cannot be done now

qiskit_nature/properties/second_quantization/vibrational/bases/vibrational_basis.py
```diff
@@ -39,7 +39,7 @@ def __init__(
 
     @property
     def num_modals_per_mode(self) -> List[int]:
-        """Returns the num_modals_per_mode."""
+        """Returns the number of modals per mode."""
```
woodsp-ibm, line 42: In the class documentation a few lines above this

> Refer to the documentation of ``qiskit_nature.properties.vibrational.integrals`` for more details.

This would be nice to have as a link to the module since the user for when the user is already reading the docs

qiskit_nature/drivers/second_quantization/__init__.py
```diff
@@ -25,8 +25,7 @@
 already using the underlying chemistry program or library will be familiar with. The driver is
 used to compute some intermediate data, which later will be used to form the input to an
 algorithm.  Such intermediate data, is populated into a
-:class:`~qiskit_nature.drivers.second_quantization.QMolecule` object and includes the following for
-example:
+:class:`~qiskit_nature.drivers.QMolecule` object and includes the following for example:
```
woodsp-ibm, line 28: I am under the impression this is deprecated and going away and the drivers here emit some grouped property result now. This text I think ought to be updated in that regard etc.

qiskit_nature/drivers/molecule.py
mrossinek: According to [Wikipedia](https://en.wikipedia.org/wiki/XYZ_file_format) the XYZ format has neither support for masses nor spins or charges. It links to an extended format but that requires information like cell vectors (which we do not have or which does not always apply).

Thus, I am not 100% sure what you are saying: I agree, that the xyz format could have been useful but we cannot, in any way, add all the information which we have here. Besides: the `geometry` section is formatted as XYZ with only the top line indicating the number of atoms missing. But that can be easily added by a user.

Is this okay as is for now or what concretely would you like me to change?

mrossinek:
> What do you intend with the PSi4 driver? The method of getting the data back across processes was the QMolecule which had no dependence on any else in Qiskit. At present its marked deprecated - depending on where you want to go with PSi4 driver maybe an easy solution is to have a copy (non-deprecated) of the QMolecule in with PSI4 driver so it uses that 'private' copy internally to the driver.

The PSi4 driver is awaiting a refactoring as part of the Epic #166 (same like some other drivers). It depends on the HDF5 driver whose refactoring needs further discussion (#302). I intentionally want to keep using the deprecated QMolecule here. There is no benefit in copying the the QMolecule to an internal class.
I can add an improved warning message to the driver stating that it will undergo a pending refactoring in Qiskit Nature 0.3.0 if you like. Same with the other drivers.

qiskit_nature/properties/second_quantization/second_quantized_property.py
mrossinek: I have iterated through all files in the `qiskit_nature.properties` submodule looking for "`" characters.
I think the situation should be generally better now :+1:

qiskit_nature/drivers/molecule.py
woodsp-ibm: I guess I was more thinking of the xyz in the context drivers do it e.g
```
0 1
H 0 0 0
H 0 0 0.75
```
and yes I was saying that does not include mass. But having seen the Property notebook and the formatting there I guess the 'extra' verbosity in the str here fits in general with what you have here. So lets go with this then as you have it.

woodsp-ibm:
In the notebook this warning message is emitted - as we use the default value for display_format. As this message is only emitted when we rely on the default can we explicitly pass in the value we want (maybe we need to change it internally to sparse later but I think thats better than having warnings come out by default from our internal use of our code)

> /home/runner/work/qiskit-nature/qiskit-nature/qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals.py:218: UserWarning: The default value for `display_format` will be changed from 'dense'to 'sparse' in version 0.3.0. Once that happens, you must specifydisplay_format='dense' directly.

(PS. I fixed for the formatting of the message in #330)

woodsp-ibm:
In the notebook ParticleNumber prints out this
```
ParticleNumber:
        4 SOs
        1 alpha electrons: [1. 0.]
        1 beta electrons: [1. 0.]
```
I had to look at the code to figure what the [1. 0.] was for - with the example it looks not so far from the tuple that is used to define the num particles (1, 1) given the lengths are the same. I see now its the orbital occupation. It might be helpful to say a little more e.g. something like this perhaps, though I agree its a bit verbose; maybe there is a better way to format it and say what things are. 
```
ParticleNumber:
        4 SOs
        1 alpha electrons:
              orbital occupation: [1. 0.]
        1 beta electrons:
             orbital occupation: [1. 0.]
```

woodsp-ibm:
In looking at the strings, particularly the integrals, should we be truncating the output by default, like numpy does with its arrays. You example in the notebook is small. By the time the integrals are more real-world it seems the formatted string is going to be huge. Having a summary like the size/shape and then some of content seems more practical. With access to data it can be printed anyway the user wants, but the str here should be a useful summary/overview of that data - given sizes of things dumping out all the data seems too much.

docs/tutorials/08_property_framework.ipynb
```diff
@@ -0,0 +1,1469 @@
...
+   "cell_type": "markdown",
+   "id": "c7c1b39e",
+   "metadata": {},
+   "source": [
+    "As you can see, the first argument simply specifies the basis of the integrals.\n",
+    "The second argument requires a little further explanation:\n",
+    "\n",
+    "1. In the case of the `AO` or `MO` basis, this argument **must** be a pair of numpy arrays, where the first and second one are the alpha- and beta-spin integrals, respectively.\n",
+    "\n",
+    "**Note:** The second argument may be `None`, in which case the alpha-spin integrals are used instead.\n",
```
woodsp-ibm, line 274:
```suggestion
    "**Note:** The second argument may be `None`, for the case where the beta-spin integrals are the same as the alpha-spin integrals, so there is not need to provide the same values twice.\n",
```
I know the intent of the text, wondering if the above is a better way of saying the intent etc.

docs/tutorials/08_property_framework.ipynb
```diff
@@ -0,0 +1,1469 @@
...
+   "source": [
+    "import numpy as np"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ed023bb1",
+   "metadata": {},
+   "source": [
+    "Since version 0.2.0 Qiskit Nature includes the _Property_ framework. This framework replaces the legacy driver return types like `QMolecule` and `WatsonHamiltonian` with a more modular and extensible approach."
```
woodsp-ibm, line 26:
```suggestion
    "Qiskit Nature 0.2.0 introduces the _Property_ framework. This framework replaces the legacy driver return types like `QMolecule` and `WatsonHamiltonian` with a more modular and extensible approach."
```
I just changed the introduction sentence - think that reads a little easier.

mrossinek:
I have added a truncation method similar to what we did for the `FermionicOp`. However, since we are explicitly listing the matrices as a sparse list, this truncation sets the number of entries to be displayed (i.e. the number of lines for each matrix) rather than limiting the total number of chars to be printed. I think this makes the most sense.

qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals.py
```diff
@@ -97,10 +109,31 @@ def __str__(self) -> str:
...
 
+    @staticmethod
+    def set_truncation(max_num_entries: int) -> None:
+        """Set the maximum number of integral values to display before truncation.
+
+        Args:
+            max_num_entries: the maximum number of entries.
+
+        .. note::
+            Truncation will be disabled if `max_num)lines` is set to 0.
```
woodsp-ibm, line 133:
```suggestion
            Truncation will be disabled if `max_num_entries` is set to 0.
```
I imagine this is what it should be

qiskit_nature/properties/second_quantization/vibrational/integrals/vibrational_integrals.py
```diff
@@ -75,10 +82,31 @@ def integrals(self, integrals: List[Tuple[float, Tuple[int, ...]]]) -> None:
...
 
+    @staticmethod
+    def set_truncation(max_num_entries: int) -> None:
+        """Set the maximum number of integral values to display before truncation.
+
+        Args:
+            max_num_entries: the maximum number of entries.
+
+        .. note::
+            Truncation will be disabled if `max_num)lines` is set to 0.
```
woodsp-ibm, line 106:
```suggestion
            Truncation will be disabled if `max_num_entries` is set to 0.
```

qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals.py
mrossinek: Yes you're right! I had originally used `max_num_lines` as the argument name but apparently I even misspelled that :sweat_smile:

woodsp-ibm (approved):

PR #304, opened 2021-08-04
Deprecate EvolvedOperatorAnsatz

ikkoham:
### Summary

Close https://github.com/Qiskit/qiskit-nature/issues/117.


qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -247,6 +248,8 @@ def _build(self) -> None:
         if self._data is not None:
             return
 
+        self.operators: Optional[Union[OperatorBase, QuantumCircuit, list]]
```
woodsp-ibm, line 251: I assume this typehint is here as it lacking on the EvolvedOperatorAnsatz in Terra. (similarly in the other changed file)

qiskit_nature/circuit/library/ansatzes/evolved_operator_ansatz.py
```diff
@@ -44,6 +44,14 @@ def __init__(
...
             initial_state: A `QuantumCircuit` object to prepend to the circuit.
         """
+        warnings.warn(
+            "This EvolvedOperatorAnsatz is deprecated as of 0.2.0, "
+            "and will be removed no earlier than 3 months after the release. "
+            "You should use the qiskit.circuit.library.EvolvedOperatorAnsatz "
+            "as a direct replacement instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
mrossinek, line 54: You may want to look into using `qiskit_nature.deprecation.warn_deprecated_same_type_name` instead of this manual warning :slightly_smiling_face:

qiskit_nature/circuit/library/ansatzes/ucc.py
ikkoham: Yes. Should I fix this in Terra?

qiskit_nature/circuit/library/ansatzes/evolved_operator_ansatz.py
ikkoham: Oh... Thanks. I didn't notice it.

qiskit_nature/circuit/library/ansatzes/evolved_operator_ansatz.py
```diff
@@ -44,6 +45,11 @@ def __init__(
             name: The name of the circuit.
             initial_state: A `QuantumCircuit` object to prepend to the circuit.
         """
+        warn_deprecated_same_type_name(
+            "0.2.0",
+            DeprecatedType.CLASS,
+            "qiskit.circuit.library.EvolvedOperatorAnsatz",
+        )
```
mrossinek, line 52: This works a little bit differently. Right now, the following `DeprecationWarning` will be raised by your code:
```python
<stdin>:1: DeprecationWarning: The qiskit.circuit.library.EvolvedOperatorAnsatz class is deprecated as of version 0.2.0 and will be removed no sooner than 3 months after the release. Instead use the qiskit.circuit.library.EvolvedOperatorAnsatz class.
```

Thus, you need to change the above to the following:
```suggestion
warn_deprecated_same_type_name(
            "0.2.0",
            DeprecatedType.CLASS,
            "EvolvedOperatorAnsatz",
            "from qiskit.circuit.library as a direct replacement",
        )
```
This will then produce the expected message:
```python
<stdin>:1: DeprecationWarning: The EvolvedOperatorAnsatz class is deprecated as of version 0.2.0 and will be removed no sooner than 3 months after the release. Instead use the EvolvedOperatorAnsatz class from qiskit.circuit.library as a direct replacement.
```

ikkoham: Thank you very much. your warning message must be better.

qiskit_nature/circuit/library/ansatzes/ucc.py
woodsp-ibm: In my mind, since the Terra version we derive from defines operators and what the types are I think the typehint should be there. That module does have typehints on some methods, and not on others, such as operators. And in that regard it is different than what is here which looks like hints were removed in places - @Cryoris any comment/thought on this?

Cryoris: We currently can't put the typehint in the circuit library due to circular imports between the circuit library and opflow

mrossinek (approved):
Thanks @ikkoham! This LGTM :slightly_smiling_face:

pbark (approved):
Thanks @ikkoham

Issue #309, opened 2021-08-07
Protein Folding API - Lists plus their lengths


woodsp-ibm:
To follow on the discussion on the main PR #231, in regards of its public API, ahead of the function being released. In particular around passing both a List, which has a length property, as well its length separately - see comments in that PR here https://github.com/Qiskit/qiskit-nature/pull/231#issuecomment-894307986 and  https://github.com/Qiskit/qiskit-nature/pull/231#issuecomment-894320719

In summary the current API looks like this (example in the docs)
```
main_chain_residue_seq = ["S", "A", "A", "S", "S", "A", "S", "A", "A", "G"]
main_chain_len = 10
side_chain_lens = [0, 0, 1, 1, 1, 1, 1, 1, 1, 0]
side_chain_residue_sequences = [None, None, "A", "A", "A", "A", "A", "A", "S", None]

peptide = Peptide(
    main_chain_len, main_chain_residue_seq, side_chain_lens, side_chain_residue_sequences
)
```
where the lengths are passed as well as the lists themselves. Since the lengths can be directly read from the Lists it seems like passing them explicitly is an extra burden on an end user. Rather it seems simpler to pass
```
main_chain_residue_seq = ["S", "A", "A", "S", "S", "A", "S", "A", "A", "G"]
side_chain_residue_sequences = [None, None, "A", "A", "A", "A", "A", "A", "S", None]

peptide = Peptide(
    main_chain_residue_seq, side_chain_residue_sequences
)
```

Also, if a user is entering such sequences manually would not a simple String suffice? I can see if the input is being generated programmatically then a mutable List lends itself to being more easily created/changed that a non-mutable String. If String were supported it would look more like
```
main_chain_residue_seq = "SAASSASAAG"
side_chain_residue_sequences = ["", "", "A", "A", "A", "A", "A", "A", "S", ""]

peptide = Peptide(
    main_chain_residue_seq, side_chain_residue_sequences
)
```
Where I also eliminated None from the interface by allowing an empty string to placed there. In the sidechains a string of max len 1 can be checked since that is all that is supported today. In the future these side strings could grow in length when supported.

Now of course either a List or a String (empty or not) could be supported for more flexibility. In summary though the main conversation, in my mind, is around explicit lengths, with the representation additional since this whole discussion is for me is around ease of use etc.

mrossinek:
[...]

Issue #252, opened 2021-06-29
Unlock a side chain at the 2nd main bead.


dlasecki:
Currently, it is forbidden to attach a side chain to the 2nd main bead. By geometry arguments, it allows for saving 1 qubit at index 6 in an optimization problem.

The goal of this issue is to allow having a side chain attached to the 2nd main bead at the expense of not saving the qubit at index 6. If a provided protein does not have a side chain attached to the 2nd main bead, the qubit at index 6 should still be saved.

PR #310, opened 2021-08-10
Protein folding - chain lengths.

dlasecki:
### Summary

Solves #309.

### DIFF
--- a/qiskit_nature/problems/sampling/protein_folding/__init__.py
+++ b/qiskit_nature/problems/sampling/protein_folding/__init__.py
@@ -13,19 +13,25 @@
 Protein Folding Problems (:mod:`qiskit_nature.problems.sampling.protein_folding`)
 =================================================================================
 The protein to be folded is defined in a Peptide class. Each peptide consists of one and only one
-main chain and optionally several side chains. Side chains cannot be attached to first, second or
-last main bead. First and last main beads with a side chain can be modeled by elongating the
-main chain with corresponding side chains. The assumption of no side chain attached to a second
-main bead allows for saving 1 qubit in the qubit encoding of a main chain without the loss of
-generality due to symmetry arguments (see the paper cited below). Allowing a side chain at a
-second bead might be supported in the future at the expense of an additional qubit required. Each
-chain consists of beads that encode information about the turn that follows to another main bead
-(in case of main beads) or into a side bead (in case of side beads). Moreover, each bead is
+main chain and optionally several side chains. Side chains cannot be attached to the first,
+second or last main bead. First and last main beads with a side chain can be modeled by
+elongating the main chain with corresponding side chains. The assumption of no side chain
+attached to a second main bead allows for saving 1 qubit in the qubit encoding of a main chain
+without loss of generality due to symmetry arguments (see the paper cited below). Allowing a side
+chain at a second bead might be supported in the future at the expense of an additional qubit
+required.
+
+Each chain consists of beads that encode information about the turn that follows to another main
+bead (in case of main beads) or into a side bead (in case of side beads). Moreover, each bead is
 characterized by a letter which encodes its residue sequence and defines the energy of
-interactions with other beads (unless interactions are random). Currently, only interactions
-involving first nearest neighbors are supported. Each side chain is attached to one and only one
-main bead. Currently, only side chains of length 1 (i.e. with 1 bead) are supported which is a
+interactions with other beads (unless interactions are random).
+
+Currently, only interactions involving first nearest neighbors are supported. Each side chain is
+attached to one and only one main bead.
+
+At the moment, only side chains of length 1 (i.e. with 1 bead) are supported which is a
 simplifying assumption. A generalization of this approach is for future investigation.
+
 Constraints on feasible folds are incorporated in the objective function using penalty terms
 whose importance is regulated by parameters in the PenaltyParameters class.
 
--- a/qiskit_nature/problems/sampling/protein_folding/__init__.py
+++ b/qiskit_nature/problems/sampling/protein_folding/__init__.py
@@ -14,14 +14,17 @@ Protein Folding Problems (:mod:`qiskit_nature.problems.sampling.protein_folding`
 =================================================================================
 The protein to be folded is defined in a Peptide class. Each peptide consists of one and only one
 main chain and optionally several side chains. Side chains cannot be attached to first, second or
-last main bead which is an assumption of the algorithm without loss of generality (see the paper
-cited below). Each chain consists of beads that encode information about the turn that follows
-to another main bead (in case of main beads) or into a side bead (in case of side beads).
-Moreover, each bead is characterized by a letter which encodes its residue sequence which defines
-the energy of interactions with other beads (unless interactions are random). Currently,
-only interactions involving first nearest neighbors are supported. Each side chain is
-attached to one and only one main bead. Currently, only side chains of length 1 (i.e. with 1
-bead) are supported which is a simplifying assumption. A generalization of this approach is
+last main bead. First and last main beads with a side chain can be modeled by elongating the
+main chain with corresponding side chains. The assumption of no side chain attached to a second
+main bead allows for saving 1 qubit without the loss of generality due to symmetry arguments (see
+the paper cited below). Allowing a side chain at a second bead might be supported in the future at
+the expense of an additional qubit required. Each chain consists of beads that encode information
+about the turn that follows to another main bead (in case of main beads) or into a side bead (in
+case of side beads). Moreover, each bead is characterized by a letter which encodes its residue
+sequence which defines the energy of interactions with other beads (unless interactions are
+random). Currently, only interactions involving first nearest neighbors are supported. Each side
+chain is attached to one and only one main bead. Currently, only side chains of length 1 (i.e.
+with 1 bead) are supported which is a simplifying assumption. A generalization of this approach is
 for future investigation. Constraints on feasible folds are incorporated in the objective function
 using penalty terms whose importance is regulated by parameters in the PenaltyParameters class.
 
--- a/qiskit_nature/problems/sampling/protein_folding/__init__.py
+++ b/qiskit_nature/problems/sampling/protein_folding/__init__.py
@@ -16,17 +16,18 @@ The protein to be folded is defined in a Peptide class. Each peptide consists of
 main chain and optionally several side chains. Side chains cannot be attached to first, second or
 last main bead. First and last main beads with a side chain can be modeled by elongating the
 main chain with corresponding side chains. The assumption of no side chain attached to a second
-main bead allows for saving 1 qubit without the loss of generality due to symmetry arguments (see
-the paper cited below). Allowing a side chain at a second bead might be supported in the future at
-the expense of an additional qubit required. Each chain consists of beads that encode information
-about the turn that follows to another main bead (in case of main beads) or into a side bead (in
-case of side beads). Moreover, each bead is characterized by a letter which encodes its residue
-sequence and defines the energy of interactions with other beads (unless interactions are
-random). Currently, only interactions involving first nearest neighbors are supported. Each side
-chain is attached to one and only one main bead. Currently, only side chains of length 1 (i.e.
-with 1 bead) are supported which is a simplifying assumption. A generalization of this approach is
-for future investigation. Constraints on feasible folds are incorporated in the objective function
-using penalty terms whose importance is regulated by parameters in the PenaltyParameters class.
+main bead allows for saving 1 qubit in the qubit encoding of a main chain without the loss of
+generality due to symmetry arguments (see the paper cited below). Allowing a side chain at a
+second bead might be supported in the future at the expense of an additional qubit required. Each
+chain consists of beads that encode information about the turn that follows to another main bead
+(in case of main beads) or into a side bead (in case of side beads). Moreover, each bead is
+characterized by a letter which encodes its residue sequence and defines the energy of
+interactions with other beads (unless interactions are random). Currently, only interactions
+involving first nearest neighbors are supported. Each side chain is attached to one and only one
+main bead. Currently, only side chains of length 1 (i.e. with 1 bead) are supported which is a
+simplifying assumption. A generalization of this approach is for future investigation.
+Constraints on feasible folds are incorporated in the objective function using penalty terms
+whose importance is regulated by parameters in the PenaltyParameters class.
 
 In the final operator for the problem qubit registers have the following meaning:
 (interactions qubits) tensored with (conformation qubits),
--- a/qiskit_nature/problems/sampling/protein_folding/__init__.py
+++ b/qiskit_nature/problems/sampling/protein_folding/__init__.py
@@ -21,7 +21,7 @@ the paper cited below). Allowing a side chain at a second bead might be supporte
 the expense of an additional qubit required. Each chain consists of beads that encode information
 about the turn that follows to another main bead (in case of main beads) or into a side bead (in
 case of side beads). Moreover, each bead is characterized by a letter which encodes its residue
-sequence which defines the energy of interactions with other beads (unless interactions are
+sequence and defines the energy of interactions with other beads (unless interactions are
 random). Currently, only interactions involving first nearest neighbors are supported. Each side
 chain is attached to one and only one main bead. Currently, only side chains of length 1 (i.e.
 with 1 bead) are supported which is a simplifying assumption. A generalization of this approach is
--- a/qiskit_nature/problems/sampling/protein_folding/interactions/interaction.py
+++ b/qiskit_nature/problems/sampling/protein_folding/interactions/interaction.py
@@ -19,12 +19,11 @@ class Interaction(ABC):
     """An abstract class defining an interaction between beads of a peptide."""
 
     @abstractmethod
-    def calculate_energy_matrix(self, chain_len: int, residue_sequence: List[str]) -> np.ndarray:
+    def calculate_energy_matrix(self, residue_sequence: str) -> np.ndarray:
         """
         Calculates an energy matrix for a particular interaction type.
 
         Args:
-            chain_len: Length of a protein chain.
             residue_sequence: A list that contains characters defining residues for a chain of
                             proteins.
 
--- a/qiskit_nature/problems/sampling/protein_folding/interactions/interaction.py
+++ b/qiskit_nature/problems/sampling/protein_folding/interactions/interaction.py
@@ -20,12 +20,12 @@ class Interaction(ABC):
     """An abstract class defining an interaction between beads of a peptide."""
 
     @abstractmethod
-    def calculate_energy_matrix(self, residue_sequence: Union[List[str], str]) -> np.ndarray:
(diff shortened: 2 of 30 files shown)


qiskit_nature/problems/sampling/protein_folding/interactions/interaction.py
```diff
@@ -19,12 +19,11 @@ class Interaction(ABC):
...
     @abstractmethod
-    def calculate_energy_matrix(self, chain_len: int, residue_sequence: List[str]) -> np.ndarray:
+    def calculate_energy_matrix(self, residue_sequence: str) -> np.ndarray:
         """
         Calculates an energy matrix for a particular interaction type.
 
         Args:
-            chain_len: Length of a protein chain.
             residue_sequence: A list that contains characters defining residues for a chain of
                             proteins.
```
mrossinek, line 28: This docstring (and probably others, too) will need updating

qiskit_nature/problems/sampling/protein_folding/interactions/random_interaction.py
```diff
@@ -21,19 +21,18 @@
 class RandomInteraction(Interaction):
     """A class defining a random interaction between beads of a peptide."""
 
-    def calculate_energy_matrix(
-        self, chain_len: int, residue_sequence: List[str] = None
-    ) -> np.ndarray:
+    # TODO only needs length...
```
mrossinek, line 24: Do you want to address this as part of this PR or in the future?

qiskit_nature/problems/sampling/protein_folding/peptide/beads/base_bead.py
```diff
@@ -37,7 +37,7 @@ def __init__(
         Args:
             chain_type: Type of the chain, either "main_chain" or "side_chain".
             main_index: index of the bead on the main chain in a peptide.
-            residue_type: A character representing the type of a residue for the bead. None in
+            residue_type: A character representing the type of a residue for the bead. "" in
```
mrossinek, line 40: Maybe write `an empty string` instead of `""`? Same elsewhere (presumably..)

qiskit_nature/problems/sampling/protein_folding/peptide/beads/base_bead.py
```diff
@@ -67,7 +67,7 @@ def __init__(
         self._residue_type = residue_type
         _validate_residue_symbol(residue_type)
         self._turn_qubits = turn_qubits
-        if self._residue_type is not None and self.turn_qubits is not None:
+        if self._residue_type != "" and self.turn_qubits is not None:
```
mrossinek, line 70: I don't have a strong opinion on this, but `if self._residue_type` is probably more Pythonic than `!= ""`

mrossinek (commented):
Thanks @dlasecki for the quick follow-up PR! :slightly_smiling_face: 
I did a quick review with some minor comments but in general this looks like a good approach!

qiskit_nature/problems/sampling/protein_folding/interactions/random_interaction.py
dlasecki: Ideally we could solve it immediately, but I am not sure what is the best approach here.

pbark: I have exactly the same question as @mrossinek  of the type change. In the previous version one would define the residues as `List[str]`. How can one define it in a single `Str` ?

qiskit_nature/problems/sampling/protein_folding/peptide/chains/main_chain.py
```diff
@@ -26,77 +26,64 @@ class MainChain(BaseChain):
 
     def __init__(
         self,
-        main_chain_len: int,
-        main_chain_residue_sequence: List[str],
-        side_chain_lens: List[int],
-        side_chain_residue_sequences: List[Optional[str]],
+        main_chain_residue_sequence: str,
```
pbark, line 29: Should this also be a `List(str)` or not?

qiskit_nature/problems/sampling/protein_folding/protein_folding_problem.py
```diff
@@ -28,14 +28,10 @@ class ProteinFoldingProblem(SamplingProblem):
     .. code-block:: python
 
         penalty_terms = PenaltyParameters(15, 15, 15)
-        main_chain_residue_seq = ["S", "A", "A", "S", "S", "A", "S", "A", "A", "G"]
-        main_chain_len = 10
-        side_chain_lens = [0, 0, 1, 1, 1, 1, 1, 1, 1, 0]
-        side_chain_residue_sequences = [None, None, "A", "A", "A", "A", "A", "A", "S", None]
+        main_chain_residue_seq = "SAASSASAAG"
```
pbark, line 31: would this also work with `["S","A","A","S",...]` ?

pbark (commented):
I think this looks good. Just a small remark that I want to make sure that is properly addressed and then we can merge this.

qiskit_nature/problems/sampling/protein_folding/peptide/chains/main_chain.py
```diff
@@ -105,57 +92,30 @@ def _build_main_chain(
...
     ) -> None:
-        if side_chain_lens is not None and main_chain_len != len(side_chain_lens):
-            raise InvalidSizeException(
-                f"The length of list of side chain lengths provided: {len(side_chain_lens)}, "
-                f"does not equal the length of the main chain: {main_chain_len}"
-            )
-        if side_chain_lens is not None and main_chain_len != len(side_chain_residue_sequences):
+        if main_chain_len != len(side_chain_residue_sequences):
             raise InvalidSizeException(
                 f"The length of list of main chain residue sequences: "
```
woodsp-ibm, line 106:
```suggestion
        if main_chain_len != len(side_chain_residue_sequences):
            raise InvalidSizeException(
                f"The length of list of side chain residue sequences: "
```
I think the message should say `side` chain.... not `main` chain right

qiskit_nature/problems/sampling/protein_folding/peptide/chains/main_chain.py
```diff
@@ -105,57 +92,30 @@ def _build_main_chain(
...
         if side_chain_residue_sequences is not None and (
-            side_chain_residue_sequences[0] is not None
-            or side_chain_residue_sequences[1] is not None
-            or side_chain_residue_sequences[-1] is not None
+            side_chain_residue_sequences[0] != ""
+            or side_chain_residue_sequences[1] != ""
+            or side_chain_residue_sequences[-1] != ""
         ):
             raise InvalidSideChainException(
                 "First, second and last main beads are not allowed to have a side chain. Non-None "
```
woodsp-ibm, line 121: Is this limitation of 1st, 2nd and last not being permitted a side chain exposed when you create a Peptide on its constructor. If I follow things down it seems that it just builds a MainChain and will fail if this is not the case. Looking at the example code, I guess I assumed that the absence of side chains in those places was just because it was an example, not that it was required as such. Is this something likely to change in the future? If not why would we want to user to pass these - would it make sense to define the side chain sequences as things that can be set - i.e length main chain - 3 so the first side chain in the sequence applies to main bead 3 (assuming indexes start at 1). Maybe passing them is clearer in that then each main has a side chain entry. But I think the behavior of 1st/2nd and last needs to be surfaced.

qiskit_nature/problems/sampling/protein_folding/peptide/chains/main_chain.py
mrossinek: No that is the proposal of #309: rather than deal with a list of single-character strings, a simple string itself is much more reasonable. This should also resolve your comment below.

qiskit_nature/problems/sampling/protein_folding/peptide/chains/main_chain.py
mrossinek: There are multiple aspects here:

1. 1st, 2nd and last not allowed: this is clearly documented in the Peptide's docstrings. The docstrings states that "this can be done without loss of generality" and refers to the paper for more information.

2. changing the side chain sequence: I don't think, that this is a good idea. Simply in terms of usability, I find it difficult to justify why a list `["a", "b", "c"]` would attach to some main sequence `"ABCDEF"` at the positions `a -> C`, `b -> D` and `c -> E`. That is both, not easy to document or explain and also just confusing. I would then rather see a good warning for the user.

3. Unittests: I am not sure if this exists already, but arguably there should be some unittests which assert the correct warnings being raised in edge cases similar to this.

qiskit_nature/problems/sampling/protein_folding/interactions/random_interaction.py
dlasecki: Ok, for a moment I got confused and thought that a residue provided might potentially be coming from a main chain or a side chain. It might be the case in the future when extended to longer side chain but currently we only rely on the main chain. I will revert this change to avoid confusion for now.

qiskit_nature/problems/sampling/protein_folding/peptide/beads/base_bead.py
dlasecki: I corrected in other places and somehow skipped this one. My bad.

qiskit_nature/problems/sampling/protein_folding/peptide/chains/main_chain.py
dlasecki: I agree with Max and that was my motivation. 

Not crucial to this discussion but there is an issue to allow side chains on the second main bead at an expense of an additional qubit required for encoding the problem (#252) 

Thanks to Max I realized that my docs regarding this need to be fixed. I made too big of a shorthand in writing that actually result in a false claim.

qiskit_nature/problems/sampling/protein_folding/interactions/random_interaction.py
dlasecki: @mrossinek The problem is that the random interaction does not use residues. It only needs the length of a residue. Ideally, it would only accept the length to avoid confusion that changing a residue would change the result somehow... 
But then it does not conform to the abstract method definition in the Interaction class...

But maybe it is not too bad...

mrossinek: Yes I see. I think this is a very special case that the RandomInteraction does not make use of all the information given to it (i.e. it only cares about the length of the list and not its actual contents). This can happen and I don't think it's a problem.

And since we have a clear abstract definition, there is nothing we can do about it :upside_down_face: Simply keep the method definition, get the length from the list and ignore the rest. You can add a comment that this method does not care about the contents but other than that I would not change anything

qiskit_nature/problems/sampling/protein_folding/peptide/chains/main_chain.py
dlasecki: We do have unit tests for that:
```
    def test_main_chain_illegal_side_chain_first(self):
        """Tests that an exception is thrown in case of illegal side chain."""
        main_chain_residue_seq = "SAAA"
        side_chain_residue_sequences = ["A", "", "S", ""]
        with self.assertRaises(InvalidSideChainException):
            _ = MainChain(
                main_chain_residue_seq,
                side_chain_residue_sequences,
            )

    def test_main_chain_illegal_side_chain_second(self):
        """Tests that an exception is thrown in case of illegal side chain."""
        main_chain_residue_seq = "SAAA"
        side_chain_residue_sequences = ["", "A", "S", ""]
        with self.assertRaises(InvalidSideChainException):
            _ = MainChain(
                main_chain_residue_seq,
                side_chain_residue_sequences,
            )

    def test_main_chain_illegal_side_chain_last(self):
        """Tests that an exception is thrown in case of illegal side chain."""
        main_chain_residue_seq = "SAAA"
        side_chain_residue_sequences = ["", "", "A", "S"]
        with self.assertRaises(InvalidSideChainException):
            _ = MainChain(
                main_chain_residue_seq,
                side_chain_residue_sequences,
            )
```

woodsp-ibm: >  1st, 2nd and last not allowed: this is clearly documented in the Peptide's docstrings. The docstrings states that "this can be done without loss of generality" and refers to the paper for more information.

Where are you looking to see that - this is what I see in Peptide where I do not see the text you refer to
```
class Peptide:
    """A class containing information about beads and chains of a protein."""

    def __init__(
        self,
        main_chain_residue_sequence: str,
        side_chain_residue_sequences: List[str],
    ):
        """
        Args:
            main_chain_residue_sequence: String of characters that define residues for a main
                                        chain. Valid residue types are [A, C, D, E, F, G, H, I,
                                        K, L, M, N, P, Q, R, S, T, V, W, Y].
            side_chain_residue_sequences: List of characters that define residues for all side
                                        beads. Empty string if a side bead does not exist. Valid
                                        residue types are [A, C, D, E, F, G, H, I, K, L, M, N, P,
                                        Q, R, S, T, V, W, Y].
        """
```

mrossinek: Sorry, it was indeed not in the Peptide but in the module-level docstring: https://github.com/Qiskit/qiskit-nature/blob/main/qiskit_nature/problems/sampling/protein_folding/__init__.py#L16-L18

woodsp-ibm: Update: I found the text you refer to - it occurred to me after writing the above it might be in the overview text in the init file - and yes it was here qiskit_nature/problems/sampling/protein_folding/__init__.py

I still think PepTide class itself thought should make this clear when going to its API ref.

mrossinek: Sounds good! @dlasecki could you copy that into the docstring of Peptide?

woodsp-ibm: That comment sequence just above looks weird now - I tell you I figured it out after told me where it was! I had left the page open and perhaps it had not refreshed when I added the update after the prior comment - oh well!

dlasecki: Do you mean duplicating what's in the init file to the Peptide docstring? Not sure about it... Maybe we could split it somehow and avoid duplication...

mrossinek: I agree with Steve that this needs to go into the `Peptide` docstring. Whether we duplicate it on the module level or not is fine either way to me

woodsp-ibm: If the more descriptive overall text stays in init, that would seem fine to me. I think here it should at least note the restriction on 1st, 2nd and last and perhaps link back to the main descriptive text for a higher level view/more information overall. It does not have to duplicate the entire text but there should be at least enough in PepTide to make them aware of the restriction/behavior if they first arrive at this class.

mrossinek: Ah I did not want to say to copy the entire module level docstring. I would have only copied the sentence referring to the side chain restrictions.

dlasecki: Updated the docs.

qiskit_nature/problems/sampling/protein_folding/__init__.py
```diff
@@ -14,14 +14,17 @@
 =================================================================================
 The protein to be folded is defined in a Peptide class. Each peptide consists of one and only one
 main chain and optionally several side chains. Side chains cannot be attached to first, second or
```
mrossinek, line 16:
```suggestion
main chain and optionally several side chains. Side chains cannot be attached to the first, second or
```
This will likely break `style` and `lint`...

qiskit_nature/problems/sampling/protein_folding/__init__.py
```diff
@@ -14,14 +14,17 @@
...
-cited below). Each chain consists of beads that encode information about the turn that follows
-to another main bead (in case of main beads) or into a side bead (in case of side beads).
-Moreover, each bead is characterized by a letter which encodes its residue sequence which defines
-the energy of interactions with other beads (unless interactions are random). Currently,
-only interactions involving first nearest neighbors are supported. Each side chain is
-attached to one and only one main bead. Currently, only side chains of length 1 (i.e. with 1
-bead) are supported which is a simplifying assumption. A generalization of this approach is
+last main bead. First and last main beads with a side chain can be modeled by elongating the
+main chain with corresponding side chains. The assumption of no side chain attached to a second
+main bead allows for saving 1 qubit without the loss of generality due to symmetry arguments (see
```
mrossinek:
```suggestion
main bead allows for saving 1 qubit without loss of generality due to symmetry arguments (see
```

qiskit_nature/problems/sampling/protein_folding/__init__.py
```diff
@@ -14,14 +14,17 @@
...
+the paper cited below). Allowing a side chain at a second bead might be supported in the future at
+the expense of an additional qubit required. Each chain consists of beads that encode information
+about the turn that follows to another main bead (in case of main beads) or into a side bead (in
+case of side beads). Moreover, each bead is characterized by a letter which encodes its residue
+sequence and defines the energy of interactions with other beads (unless interactions are
+random). Currently, only interactions involving first nearest neighbors are supported. Each side
```
mrossinek: Just a general comment: right now, this is all one big paragraph. Maybe this will be more readable with some line breaks,

qiskit_nature/problems/sampling/protein_folding/residue_validator.py
```diff
@@ -10,19 +10,20 @@
...
-from typing import List
+from typing import List, Union
 
 from .exceptions.invalid_residue_exception import (
     InvalidResidueException,
 )
 
 
-def _validate_residue_sequence(residue_sequence: List[str]):
+def _validate_residue_sequence(residue_sequence: Union[List[str], str]):
```
mrossinek, line 20: Why does this still need to support `List[str]`?

qiskit_nature/problems/sampling/protein_folding/peptide/chains/main_chain.py
```diff
@@ -26,77 +26,64 @@ class MainChain(BaseChain):
...
+            main_chain_residue_sequence: String of characters that define residues for a main chain.
+            side_chain_residue_sequences: List of characters that define residues for all side
+                                        beads. Empty string if a side bead does not exist.
 
         Returns:
             An instance of a MainChain class.
 
         Raises:
-            InvalidSizeException: when the length of list of side chain lengths provided does not
+            InvalidSizeException: When the length of list of side chain lengths provided does not
```
mrossinek, line 73: Since you changed it above, maybe do the same here. I am not sure about other occurrences.

mrossinek:
Other than this comment I think this looks good to me :+1: 
You may want to merge `main` back in though, too

mrossinek (approved):
LGTM! Let's await another review before merging though :+1:

pbark (approved):
This also LGTM! Thanks for the quick fix!

Issue #234, opened 2021-06-16
Big performance regression with new `FermionicOp`


jlapeyre:
I am using this code:
```python

from qiskit_nature.drivers import PySCFDriver, UnitsType
from qiskit_nature.problems.second_quantization.electronic.builders.fermionic_op_builder import build_ferm_op_from_ints
from qiskit_nature.operators.second_quantization import FermionicOp

h2o = "O 0.0 0.0 0.0;H 0.757 0.586 0.0;H -0.757 0.586 0.0"
h2 = 'H .0 .0 0.7414; H .0 .0 .0'
lih = 'Li .0 .0 0.7414; H .0 .0 .0'
basis0 = "sto3g"
basis1 = "631g"

def get_molecule():

    driver = PySCFDriver(atom=h2o,
                         unit=UnitsType.ANGSTROM,
                         charge=0,
                         spin=0,
                         basis=basis0)
    molecule = driver.run()
    return molecule

def get_qkn_op(molecule):
    one_body = molecule.one_body_integrals
    two_body = molecule.two_body_integrals
    fop_qkn = build_ferm_op_from_ints(one_body, two_body)
    return fop_qkn

molecule = get_molecule()
```
Then, after the change of internals, with a0f4c7bb , I get this:
```python
%time fop = get_qkn_op(molecule)
CPU times: user 12.2 s, sys: 58.4 ms, total: 12.3 s
Wall time: 4.87 s
```

However, before the rewrite of internals 9f75adc, I get this:
```python
%time fop = get_qkn_op(molecule)
CPU times: user 316 ms, sys: 1.83 ms, total: 317 ms
Wall time: 318 ms
```

If I use `basis1` (631g) above, then the time before the change is 4s. The time after the change is longer than several minutes; I killed the process.

mrossinek:
I can confirm that 8645ee35fa40d1c4d67d71e8a1c0e766f8af324b is the first commit on `main` which experiences this drastic performance degradation. @ikkoham could you please also take a look at this one?

ikkoham:
hmm... Coexistence of dense and sparse labels causes the performance regression. I think we should separate the internal structure and API as I first proposed. But, I will try to improve the performance when I do #224 .

jlapeyre:
@mrossinek . I did not identify the first release that introduces the regression. I assumed it was the introduction of the new structure.

mrossinek:
> Coexistence of dense and sparse labels causes the performance regression

[...]

PR #314, opened 2021-08-11
Remove LegacyFermionicOp

### LINKED ISSUE #224: Removal of Legacy FermionicOp
#201 reworked the internal data of FermionicOp and the original is being kept for a short while as operators/second_quantization/legacy_fermionic_op.py - but the intent is to remove that ahead of the next release. This is just to track that so we ensure its done ahead of the release.

ikkoham:
### Summary

This PR Resolves #224, #234.

I took a benchmark for #234

main 0d2cc5f
```python
%time fop = get_qkn_op(molecule)
CPU times: user 37.6 s, sys: 286 ms, total: 37.9 s
Wall time: 38.3 s
```

92951ca in this PR
```python
%time fop = get_qkn_op(molecule)
CPU times: user 445 ms, sys: 14.2 ms, total: 459 ms
Wall time: 472 ms
```


mrossinek (commented):
Thanks @ikkoham for looking into this! I think I stumbled on two remaining issues:

1. `compose` does not preserve the label type:
```python
sparse_plus = FermionicOp("+_1", register_length=4, sparse_label=True)
sparse_minus = FermionicOp("-_1", register_length=4, sparse_label=True)
print(sparse_plus @ sparse_minus)

>>> FermionicOp([('INII', (1+0j))], register_length=4, sparse_label=False)
```
This should remain sparse, same as it does during addition.

2. Going back to [this example](https://gist.github.com/mrossinek/3cc65837d2a998856f202ef341e7f455) from back when we were investigating the particle-hole transformer, the example at the end:
```python
import numpy as np
from functools import reduce

ijkl, operators = (0, 0, 1, 0), ('-', '+', '-', '+')
iklj, operators_reordered = zip(*np.asarray(list(zip(ijkl, operators)))[[0, 2, 3, 1]])
op = reduce(lambda a, b: a @ b, [FermionicOp(f'{op}_{idx}', register_length=4, sparse_label=True)
                                 for idx, op in zip(ijkl, operators)])
print(op)
```
still will not work, even if I hardcode the `compose` method to use sparse labels (i.e. fixing 1. locally). The problem appears to be that `to_normal_order` internally converts to dense labels. Is that really necessary?

ikkoham:
@mrossinek Thanks. I have fixed 1 in 4014cf2, and the output of 2 becomes
```
-_0 +_0 -_1 +_0 * (1+0j)
```.
Is this desired?

**Comment by pbark** [comment]
Thanks @ikkoham ! Please also add a reno.

**Comment by ikkoham** [comment]
@pbark Thanks. What types of reno is required? Upgrade?
If necessary, I'll add https://github.com/Qiskit/qiskit-nature/blob/main/releasenotes/notes/add-fermionicop-label-mode-a1faa97811dac9b4.yaml.

**Comment by pbark** [comment]
@ikkoham I think what you propose is fine! Thanks!

**Comment by mrossinek** [comment]
I don't quite understand the current behavior in this case:
```python
In [3]: FermionicOp("+_2", register_length=4)
Out[3]: FermionicOp([('II+I', (1+0j))], register_length=4, sparse_label=False)
```
It is somehow able to detect the sparse label and deal with it correctly, but why does it then convert it to dense labels? Is the `sparse_label` argument on `__init__` determining the expected input kind or is it specifying the internal storage representation format?

mrossinek:
Besides my question above I think this looks good now :+1: 
I dtug into this p/h example again because we want to use that for the mentorship program. I think we will have to take another look at `to_normal_order` then (but this does not need to happen now). The code is still doing an internal conversion to dense and back to sparse which is messing cases like the following up:
```python
 In [45]: op = reduce(lambda a, b: a @ b, [FermionicOp(f'{op}_{idx}', register_length=4, sparse_label=True) for idx, op in zip([0, 0, 0, 2], "-+-+")])

In [46]: str(op)
Out[46]: '-_0 +_0 -_0 +_2 * (1+0j)'

In [47]: str(op.to_normal_order())
Out[47]: '+_2 -_0 * (-1+0j)'
```
Where the last term should really also contain `+_0 +_2 -_0 -_0 * (-1+0j)`. I believe that this get's lost because of the `to_dense` conversion.

But this can be left for another time to be fixed.

ikkoham:
From anti-commutation relation, `-+- = -` holds. The reason for reverting back to dense is to simplify the calculation. Do you really need this conservation for p/h?

mrossinek:
> Do you really need this conservation for p/h?

As far as my understanding of that transformation goes we will need this, yes. However, this does not need to happen now. Let's focus on getting the release done and address this as part of the one.

Besides [this comment](https://github.com/Qiskit/qiskit-nature/pull/314#issuecomment-898261938) I think everything else about this PR is good to go. Could you please address that final aspect?

ikkoham:
Reply to the comment: https://github.com/Qiskit/qiskit-nature/pull/314#issuecomment-898261938

sparse label specifies the label format of the output. I might as well rename it, since the name is complicated. How about display_label_is_sparse ? a bit long?

qiskit_nature/mappers/second_quantization/qubit_mapper.py
```diff
@@ -107,7 +107,12 @@ def times_annihilation_op(position, pauli_table):
 
         # TODO to_list() is not an attribute of SecondQuantizedOp. Change the former to have this or
         #   change the signature above to take FermionicOp?
-        for label, coeff in second_q_op.to_list():
+        label_coeff_list = (
+            second_q_op.to_list(display_format="dense")
+            if isinstance(second_q_op, FermionicOp)
```
mrossinek, line 112: I don't think we should have this isinstance check. Instead, we should propagate the `display_format` argument up to the `SecondQuantizedOp` interface.
It is fine that implementations other than `FermionicOp` ignore it for now.

qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -406,7 +406,8 @@ def reduce(self, atol: Optional[float] = None, rtol: Optional[float] = None) ->
         new_coeff = coeff_list[non_zero]
         return SpinOp((new_array, new_coeff), spin=self.spin)
 
-    def to_list(self) -> List[Tuple[str, complex]]:
+    # pylint: disable=arguments-differ
```
mrossinek, line 409: As per my comment above, this should not be necessary. We can simply not use the argument.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -144,70 +199,100 @@ class FermionicOp(SecondQuantizedOp):
...
+        display_format: str = "dense",
     ):
         """
         Args:
             data: Input data for FermionicOp. The allowed data is label str,
                   tuple (label, coeff), or list [(label, coeff)].
             register_length: positive integer that represents the length of registers.
-            sparse_label: the label is represented by sparse mode.
+            display_format: If sparse, the label of output is represented by sparse.
+                            if dense, the label of output is represented by dense. (default: dense)
```
mrossinek, line 218:
```suggestion
            display_format: If sparse, the label is represented sparsely during output.
                            if dense, the label is represented densely during output. (default: dense)
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -144,70 +199,100 @@ class FermionicOp(SecondQuantizedOp):
...
 
         Raises:
             ValueError: given data is invalid value.
             TypeError: given data has invalid type.
         """
+        self.display_format: str
```
mrossinek, line 224: I think we can make this `self._display_format` and then add `@property` and `@property.setter` methods instead of `set_display_format`

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -282,26 +372,43 @@ def add(self, other: "FermionicOp") -> "FermionicOp":
...
+            self.display_format or other.display_format,
         )
 
-    def to_list(self) -> List[Tuple[str, complex]]:
+    def to_list(self, display_format: Optional[str] = None) -> List[Tuple[str, complex]]:
         """Returns the operators internal contents in list-format.
 
+        Args:
+            display_format: If sparse, the label of output is represented by sparse.
+                if dense, the label of output is represented by dense. (default: None)
```
mrossinek, line 383:
```suggestion
            display_format: If sparse, the label is represented sparsely during output.
                            if dense, the label is represented densely during output. (default: None)
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -144,70 +199,100 @@ class FermionicOp(SecondQuantizedOp):
...
+        data: Union[
+            str,
+            Tuple[str, complex],
+            List[Tuple[str, complex]],
+            List[Tuple[str, float]],
+            List[Tuple[_FermionLabel, complex]],
+        ],
         register_length: Optional[int] = None,
-        sparse_label: bool = False,
+        display_format: str = "dense",
```
mrossinek, line 210: Can we also add some sort of warning that the default of this is going to change to `sparse` in Qiskit Nature 0.3.0?
Either a `DeprecationWarning` or `UserWarning` should work I think.

mrossinek (commented):
Thanks for the changes @ikkoham. I still have some minor comments though

qiskit_nature/mappers/second_quantization/qubit_mapper.py
ikkoham: SpinOp and VibrationalOp do not have it. I considered adding the arguments to them, but I didn't think it's a good idea to add meaningless ones, so I left it like this for now.

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: So you recommend meaningless arguments. That is one direction. (# pylint: disable=unused-argument)

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: This is a remnant from when it was first a class variable. You are right, that is better.

qiskit_nature/operators/second_quantization/spin_op.py
mrossinek: Well since you add the argument on the interface anyways, I would rather make the code in the `QubitMapper` not case-specific.
Adding this exception here causes complexity in the `QubitMapper` *and* a lint exception here. Having an argument which is ignored here and in the vibrational case only causes a lint exception (which may even get remove in the future)

qiskit_nature/mappers/second_quantization/qubit_mapper.py
mrossinek: See below.

qiskit_nature/operators/second_quantization/spin_op.py
ikkoham: I reverted the interface. Since FermionicOp is an exception, I shouldn't have specified it in the super class. In the future, I think it would be better to be able to map sparse label directly.

mrossinek: Okay I see. I absolutely agree that we need to be able to map sparse labels directly in the future. Until then, let's use this workaround with the `isinstance` check inside of the `QubitMapper`. @woodsp-ibm is this okay with you, too?

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -141,73 +197,115 @@ class FermionicOp(SecondQuantizedOp):
...
-                raise TypeError(
-                    f"Data tuple must be (str, number), not ({type(data[0])}, {type(data[1])})."
+        if display_format is None:
+            display_format = "dense"
+            if FermionicOp._display_format_warn:
+                FermionicOp._display_format_warn = False
+                warnings.warn(
+                    'The display_format="dense" will be changed to "sparse" as default.'
+                    "as of version 0.2.0 no sooner than 3 months after the release. "
+                    "Instead use the display_format=dense explicitly.",
```
mrossinek, line 234:
```suggestion
                    "The default value for `display_format` will be changed from 'dense'"
                    "to 'sparse' in version 0.3.0. Once that happens, you must specify"
                    "display_format='dense' directly."
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -282,94 +385,113 @@ def add(self, other: "FermionicOp") -> "FermionicOp":
...
+    # pylint: disable=arguments-differ
+    def to_list(
+        self,
+        display_format: Optional[str] = None,
+    ) -> List[Tuple[str, complex]]:  # type: ignore
         """Returns the operators internal contents in list-format.
 
+        Args:
+            display_format: If sparse, the label is represented sparsely during output.
+                            if dense, the label is represented densely during output. (default: None)
```
mrossinek, line 400:
```suggestion
                            If dense, the label is represented densely during output.
                            The default value is ``None``, in which case ``self.display_format``
                            will be used.
```

mrossinek (commented):
Some final touch ups

qiskit_nature/operators/second_quantization/fermionic_op.py
woodsp-ibm: So this is basically a way to locally, for this method, override the main display_format property, to which it will otherwise default to for its formatting.

mrossinek: Let me propose a better change (will have to do so in a separate discussion to include the line above)

mrossinek, line 400:
```suggestion
            display_format: when specified this will overwrite ``self.display_format``. Can
                be either 'dense' or 'sparse'. See the class documentation for more details.
```

mrossinek (approved):
Thanks @ikkoham for the great work!

PR #315, opened 2021-08-11
Remove deprecated QubitConverter

woodsp-ibm:
QubitConvertor was moved to `convertors` out of `operators` in the first bug fix release, shortly after the original release, to avoid cyclic import problems. The module, though it was only out in the wild for a few days, was kept/deprecated in its former location in case it had been used in the interim. This PR removes it now from `operators` completely, ahead of the upcoming release, since its beyond the 3 month window.


mrossinek (approved):
Thanks @woodsp-ibm for taking care of this!

pbark (approved):

PR #316, opened 2021-08-12
Truncate str and repr in FermionicOp

ikkoham:
### Summary

I implemented the truncation of the output of str and repr.
https://github.com/Qiskit/qiskit-nature/pull/88#discussion_r678469775

The new output is
image


mrossinek (commented):
Thanks @ikkoham! Do you think you could add a very simple unittest for this?

woodsp-ibm:
In optimization there is a discussion over str and repr of QuadraticProgram and concern over str length which you have here taken care of with an extra truncation variable. I wonder in general if a short str, that was more of an overall summary of the FermionicOp, would be preferable - e.g. state the number of terms in the operator and register length e.g. `Fermionic Operator register length=6, number terms=672` and have the full str (that is the current str) as say a `details(truncation=0) -> str` method. This way you can still get the full list as a string if you want to print it - or use it to save to file or build another FermionicOp later - but by default its hopefully a useful enough summary. In the legacy operator implementation in Aqua we used to have a 'name' field (setter/getter etc) that allowed the creator to give it a name like 'H2 Hamiltonian`, `spin operator` etc which would be printed and if you had a bunch of these operators in say some List would allow you to annotate what they represented if that was useful to you. I would have said that would be part of the summary str too but I do not see such a field here.

ikkoham:
@woodsp-ibm Thanks. Displaying the register_length and # terms is very nice. I want to add them.
If you need full str, you can set the truncation 0 by `FermionicOp.set_truncation(0)`. 

As for name, it's not obvious about arithmetic operations. Should the name of A + B be A.name + B.name or A.name or B.name? This is the difference between operators and QuadraticProgram. If this problem is resolved, I can add the name field.

mrossinek:
> I am able to add some unittests. But I think many modules have no unit tests for __str__ and __repr__. What do you think?

Yes I see. I think it's fine then :+1:

mrossinek (dismissed):
Looks good to me! For context here are some examples:
```python
In [8]: print(FermionicOp([("+_0", 1)] * 100, register_length=4))
Fermionic Operator
register length=4, number terms=100
  (1+0j) * ( +III )
+ (1+0j) * ( +III )
+ (1+0j) * ( +III )
+ (1+0j) * ( +III )
+ (1+0j) * ( +III )
+ (1+0j) * ( +III )
+ (1+0j) * ( +III )
+ (1+0j) * ( +III )
+ (1+0j) * ( +III )
+ (1+0j) * ( +II ...

In [13]: print(FermionicOp([("+_0", 1)] * 2, register_length=4))
Fermionic Operator
register length=4, number terms=2
  (1+0j) * ( +III )
+ (1+0j) * ( +III )

In [14]: print(FermionicOp([("+_0", 1)] * 1, register_length=4))
(1+0j) * (+III)
```

I would like @woodsp-ibm to also take a look before merging this, though

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -226,12 +228,33 @@ def __repr__(self) -> str:
             ")"
         )
 
+    @classmethod
+    def set_truncation(cls, val):
+        """Set the max number of characters to display before truncation/
+        Args:
+            val (int): the number of characters.
```
woodsp-ibm, line 235:
```suggestion
    def set_truncation(cls, val: int) -> None:
        """Set the max number of characters to display before truncation/
        Args:
            val: the number of characters.
```
We normally have typehints on the method itself rather than on the params

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -141,6 +141,7 @@ class FermionicOp(SecondQuantizedOp):
             fermion += FermionicOp(somedata)
 
     """
+    _truncate = 200
```
woodsp-ibm, line 144: Should we have some class level comment that talks about the str truncation and the default value truncation is set to? It might be helpful I think/

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -217,7 +218,8 @@ def __repr__(self) -> str:
             data_str = f"'{data[0]}'"
         data_str = f"{data}"
 
-        # TODO truncate
+        if self._truncate and len(data_str) > self._truncate:
```
woodsp-ibm, line 221: I know Python allows access to class variables via self. I have always preferred to access them via the classname in situations like this so its clear and you don't slip and assign to self and create and instance var

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Thanks. I made a mistake because I've been developing a lot of modules recently.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -98,6 +98,26 @@ class FermionicOp(SecondQuantizedOp):
         `FermionicOp(list)`
           The list must be a list of valid tuples as explained above.
 
+    **Output or str and repr**
```
mrossinek, line 101:
```suggestion
    **Output of str and repr**
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -226,12 +247,33 @@ def __repr__(self) -> str:
             ")"
         )
 
+    @classmethod
+    def set_truncation(cls, val: int) -> None:
+        """Set the max number of characters to display before truncation/
+        Args:
+            val: the number of characters.
```
mrossinek, line 254:
```suggestion
        """Set the max number of characters to display before truncation.
        
        Args:
            val: the number of characters.
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -226,12 +247,33 @@ def __repr__(self) -> str:
...
-        return "  " + "\n+ ".join([f"{label} * {coeff}" for label, coeff in self.to_list()])
+            return f"{coeff} * ({label})"
+        pre = (
+            "Fermionic Operator\n"
+            f"register length={self.register_length}, number terms={len(self)}\n"
+        )
+        ret = "  " + "\n+ ".join(
+            [f"{coeff} * ( {label} )" if label else f"{coeff}" for label, coeff in self.to_list()]
+        )
+        if self._truncate and len(ret) > self._truncate:
```
woodsp-ibm, line 274:
```suggestion
        if FermionicOp._truncate and len(ret) > FermionicOp._truncate:
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -226,12 +247,33 @@ def __repr__(self) -> str:
...
+            ret = ret[0 : self._truncate - 4] + " ..."
```
woodsp-ibm, line 275:
```suggestion
            ret = ret[0 : FermionicOp._truncate - 4] + " ..."
```

woodsp-ibm: As init now refers to class var _truncate via class FermionicOp I think we should do that here too

mrossinek (approved):
Thanks @ikkoham and @woodsp-ibm!

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -226,12 +247,34 @@ def __repr__(self) -> str:
             ")"
         )
 
+    @classmethod
+    def set_truncation(cls, val: int) -> None:
+        """Set the max number of characters to display before truncation.
+
```
woodsp-ibm, line 253:
```suggestion

```

mrossinek (approved):

PR #318, opened 2021-08-12
Second bead with a side chain.

### LINKED ISSUE #252: Unlock a side chain at the 2nd main bead.

dlasecki:
### Summary

Resolves #252. (Edited from solves to one of the words recognized by github for [auto-closing](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

### Details and comments

This PR allows attaching a side chain at a second main bead (it needs an additional qubit for encoding a main chain). If no side chain is attached there, the qubit saving stays as before.


pbark (approved):
This LGTM! Thanks!

qiskit_nature/problems/sampling/protein_folding/__init__.py
```diff
@@ -13,13 +13,11 @@
...
-main chain and optionally several side chains. Side chains cannot be attached to the first,
-second or last main bead. First and last main beads with a side chain can be modeled by
-elongating the main chain with corresponding side chains. The assumption of no side chain
-attached to a second main bead allows for saving 1 qubit in the qubit encoding of a main chain
-without loss of generality due to symmetry arguments (see the paper cited below). Allowing a side
-chain at a second bead might be supported in the future at the expense of an additional qubit
-required.
+main chain and optionally several side chains. Side chains cannot be attached to the first or
+last main bead. First and last main beads with a side chain can be modeled by
+elongating the main chain with corresponding side chains. No side chain attached to a second main
```
woodsp-ibm, line 18: This is more a general comment - if this is how side chains are done for first and last bead could we not do this internally on behalf of the user - let them supply main and side chains and if there are side chains on first or last move to main chain. I am not suggesting changing anything here  in this PR - but it seemed like a place to comment in this regard.

woodsp-ibm:
Since you have this PR open, and this is from folding, perhaps you can address these errors that come out in CI in the deprecations roll-up (if you prefer to do this is a separate PR thats fine - I had mused over opening a new issue but since this is small....)

```
/home/runner/work/qiskit-nature/qiskit-nature/qiskit_nature/problems/sampling/protein_folding/bead_distances/distance_map_builder.py:112: DeprecationWarning: invalid escape sequence \s
/home/runner/work/qiskit-nature/qiskit-nature/qiskit_nature/problems/sampling/protein_folding/bead_distances/distance_map_builder.py:122: DeprecationWarning: invalid escape sequence \s
```
I think the """ just needs to be preceded by 'r' for raw string so it does not attempt to interpret things that way.

qiskit_nature/problems/sampling/protein_folding/__init__.py
dlasecki: It could be done. But we would need to make sure that user understands that a main chain (in our understanding) gets longer than what they defined. It is important for interpreting the result at the end - certain qubit registers correspond to main beads and others to side beads. We could perhaps reshuffle them accordingly but it would lead to more and more explanations and documentation on our side. Not sure if it is worth it...

 @pbark Would allowing side chains on terminal beads and handling it behind the scenes raise eyebrows among the target research community? We could handle this but perhaps it might lead to confusion? Please decide.

mrossinek: I think this might be a dangerous direction to go in because changing the user input can open up paths to many bugs/edge cases for unexpected behavior.
For the time being, I would strongly prefer having this properly documented as it is done now.

mrossinek (commented):
LGTM but I also suggest to quickly add the fix mentioned by Steve, above :+1:

qiskit_nature/problems/sampling/protein_folding/__init__.py
pbark: @dlasecki @woodsp-ibm I do not find treating the side chain of the first bead and last bead as enlongation to the main chain under the hood a good idea. This will only create a problem for the user. I see the point of Steve (and thanks for raising it) but I really believe that as it is documented now it is enough and the users should make the choice of adding beads in the main chain. At the same time changing it without exposing it will make the interpretation of results difficult and will only raise questions. The point of @mrossinek is exactly on the spot.

pbark (approved):
Now that the deprecation is fixed and the issue with the chain elongation is discussed I think we can merge this PR and start the work on the tutorial.  We can discuss the second point further if @woodsp-ibm wants to in an issue! LGTM!

mrossinek (approved):

qiskit_nature/problems/sampling/protein_folding/__init__.py
woodsp-ibm: That's fine - it would have to have been reversed for interpret etc. If you looked inside you would have to know what was done - i.e. 1st/last side beads attached to main to simplify the computation - and hence allow you to interpret qubits. From an external perspective of inputting the problem and looking at a result it should be transparent. Only if you peered under the hood would the behavior be visible. I guess I don't view it 'changing' the users input - just taking it and adapting it to the underlying algorithm - a user should not see it as changed. As such now its really asking the user to perhaps change their input/thinking around what are side beads versus main to facilitate the computation - which seems a little awkward/off to me, hence the comment.

PR #321, opened 2021-08-12
BopesSampler no longer uses deprecated VQE function

### LINKED ISSUE #242: BopesSampler uses deprecated method of VQE - needs to be updated
VQE was changed to deprecate a bunch of methods, that were around aspects of the result, to make it stateless, which now requires the result object to be used directly. AdaptVQE was updated in #222 but as that was progressed, after the Terra PR to make VQE stateless was merged, CI failed when BopeSampler accessed one of the deprecated properties. Turned out the decorators were in the wrong order leading it to return different than was expected - see #222 for more details in the comments. That was fixed an CI passed for #222, which has now been merged asap since it was blocking all other activity, but this leaves BopesSampler still needing to be updated to avoid the newly added deprecation of aspects of VQE. The failure was in the BopesSampler tutorial - I recall unit tests passing - indeed in the CI of #222, after it passed, I see no deprecation message in the rollup - so maybe the unit tests don't cover that code (arguably they should) - either way this need investigation and correction. Here is what was saved from the when it showed the message

```
qiskit_nature/algorithms/pes_samplers/bopes_sampler.py:196: DeprecationWarning: 
The VQE.optimal_params property is deprecated as of Qiskit Terra 0.18.0
and will be removed no sooner than 3 months after the releasedate.
This information is part of the returned result object and can be
queried as VQEResult.optimal_point.
  optimal_params = self._gss.solver.optimal_params  # type: ignore
  ```
  so its part of this code https://github.com/Qiskit/qiskit-nature/blob/3757845c601a65bc0dad7a6a4eae78ef7a235e2c/qiskit_nature/algorithms/pes_samplers/bopes_sampler.py#L193-L197

woodsp-ibm:
Fixes #242

Changed BopesSampler to access the optimal point from the result - rather than the now deprecated stateful method on VQE that used to expose that as well.

Also added a new unit test, based on the bopes sampler tutorial, but rather quicker, that uses VQE so it tests a number of paths in BopesSampler specific to VariationalAlgorithm classes like VQE, that were untested. This unit test, before the BopesSampler change, emitted the deprecation warning, but no longer does after said change. Current BopesSampler test, before this additional test was 64.84% coverage - this should improve that. 

Update: With this new test CI now shows 81.32% coverage for BopesSampler.

### DIFF
--- a/qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
+++ b/qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
@@ -232,7 +232,7 @@ class BOPESSampler:
         # Save optimal point to bootstrap
         if isinstance(self._gss.solver, VariationalAlgorithm):  # type: ignore
             # at every point evaluation, the optimal params are updated
-            optimal_params = self._gss.solver.optimal_params  # type: ignore
+            optimal_params = result.raw_result.optimal_point  # type: ignore
             self._points_optparams[point] = optimal_params
 
         return result
--- a/test/algorithms/pes_samplers/test_bopes_sampler.py
+++ b/test/algorithms/pes_samplers/test_bopes_sampler.py
@@ -14,12 +14,14 @@
 
 import unittest
 from functools import partial
+
 from test import QiskitNatureTestCase, requires_extra_library
 
 import numpy as np
 
-from qiskit.algorithms import NumPyMinimumEigensolver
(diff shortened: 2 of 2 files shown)


pbark (approved):
Thanks a lot @woodsp-ibm ! This LGTM!

mrossinek (approved):
Thanks @woodsp-ibm!

PR #326, opened 2021-08-13
Protein folding tutorial

### LINKED ISSUE #250: Create a protein folding tutorial.
The goal of this issue is to create a tutorial (Jupyter Notebook) that would demonstrate how to use the latest Protein Folding feature from Qiskit Nature for some examples of proteins.

pbark:
### Summary

closes #250 

### Details and comments

Protein folding tutorial. I named the tutorial as 09 since we wanted to keep 08 for the properties tutorial.


docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+   "source": [
+    "The structure and function of many natural and human-engineered\n",
+    "proteins is still only poorly understood. As a result, our understanding of\n",
+    "processes connected with protein folding, such as those encountered in\n",
+    "Alzheimer’s disease, vaccine development, and crop improvement\n",
+    "research, has remained limited.\n",
+    "\n",
+    "Unfolded polypeptides have a very large number of degrees of freedom\n",
+    "and thus an enormous number of potential conformations. For example, a\n",
+    "chain with 100 amino acids has on the order of 10 conformations. In\n",
```
mrossinek, line 32: > For example, a chain with 100 amino acids has on the order of 10 conformations.

This seems like it should be a greater number of conformations...

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+    "\n",
+    "<font size='2'>[1] https://en.wikipedia.org/wiki/Levinthal%27s_paradox </font>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3110941f",
+   "metadata": {},
+   "source": [
+    "The goal of this work is to determine the minimum energy configuration of a protein. Starting from a random configuration, the protein's structure is optimized to lower the energy. This can be achieved by encoding the protein folding problem into a qubit operator and ensuring that all physical constrains are satisfied. "
```
mrossinek, line 50: > configuration

You are mixing the use of `conformation` and `configuration`. But these are not interchangeable!

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+   "id": "f571d4a8",
+   "metadata": {},
+   "source": [
+    "The Hamiltonian of the system for a set of qubits $\\mathbf{q}=\\{\\mathbf{q}_{cf}, \\mathbf{q}_{in}\\}$ is \n",
+    "\n",
+    "$$H(\\mathbf{q}) = H_{gc}(\\mathbf{q}_{cf}) + H_{ch}(\\mathbf{q}_{cf}) + H_{in}(\\mathbf{q}) $$\n",
+    "\n",
+    "where \n",
+    "- $H_{gc}$ is the geometrical constrain term (governing the growth of the primary sequence of aminoacids without bifurcations\n",
+    "- $H_{ch}$ is the chirality constrain (enforcing the righ stereochemistry for the system)\n",
```
mrossinek, line 79:
```suggestion
    "- $H_{ch}$ is the chirality constrain (enforcing the right stereochemistry for the system)\n",
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "62d99a5f",
+   "metadata": {},
+   "source": [
+    "Further details about the used model and the encoding of the problem can be found in [2].\n",
+    "\n",
+    "<font size='2'>[2] A.Robert, P.Barkoutsos, S.Woerner and I.Tavernelli, Resource-efficient quantum algorithm for protein folding, NPJ Quantum Information, 2021 </font>"
```
mrossinek, line 90: I am not sure whether it is Github's rendering, but I do not see this in the output.

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+   "source": [
+    "### Protein Main Chain"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "350c031c",
+   "metadata": {},
+   "source": [
+    "The Protein consists of a main chain that is a linear chain of aminoacids. For the naming of different residues we use the one-letter code as defined in Reference [3]. Further details about the naming and the type of aminoacids can also be found in [4].\n",
```
mrossinek, line 121: > aminoacids

Above you wrote "amino acids" as two words. I would stick to that. Maybe do a search-and-replace in case I overlooked more occurrences.

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "bf140d37",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "main_chain_len = 7\n",
+    "main_chain_residue_seq = [\"A\", \"P\", \"R\", \"L\", \"R\", \"F\", \"Y\"]"
```
mrossinek, line 137: You no longer need to specify these separately (#309). I suggest to simply write:
```suggestion
    "main_chain = \"APRLRFY\"",
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+   "source": [
+    "### Side Chains"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d71b4571",
+   "metadata": {},
+   "source": [
+    "Beyond the main chain of the protein there may be aminoacids attached to the residues of the main chain. Our model allows for side chains of the maximum length of one. Enlongated side chains would require the introduction of additional penalty terms that are still under development. \n",
```
mrossinek, line 153:
```suggestion
    "Beyond the main chain of the protein there may be aminoacids attached to the residues of the main chain. Our model allows for side chains of the maximum length of one. Elongated side chains would require the introduction of additional penalty terms which are still under development. \n",
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+    "\n",
+    "For the definition of the side chain we provide a binary vector specifying the existence of a side chain and if a side chain is present we indicate the letter type of the residue."
```
mrossinek, line 155: This description is outdated.

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "4d148961",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "side_chain_lens = [0, 0, 0, 0, 0, 0, 0]\n",
+    "side_chain_residue_sequences = [None]*7"
```
mrossinek, line 166:
```suggestion
    "side_chains = [\"\"] * 7"
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "cf834eb2",
+   "metadata": {},
+   "source": [
+    "To ensure that all physical constrains are respected we introduce penalty functions. The different penalty terms used are: \n",
+    "- penalty_chiral: A penalty parameter used to impose the right chirality.\n",
+    "- penalty_back: A penalty parameter used to penalize turns along the same axis.\n",
```
mrossinek, line 215: Reading this I don't see how `back` and `axis` are related.. Is this supposed to be the `"backbone"` of the peptide?

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "id": "d37df517",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "peptide = Peptide(main_chain_len, main_chain_residue_seq, side_chain_lens, side_chain_residue_sequences)"
```
mrossinek, line 256: I believe this needs updating after closing #309

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
+   "source": [
+    "### Using VQE with CVAR expecation value for the solution of the problem"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0f2d3e2c",
+   "metadata": {},
+   "source": [
+    "The problem that we are tackling has now been mapped to an optimizaiton problem. We can use the Variational Quantum Eigensolver with Conditional Value at Risk (CVAR) expectation value for the solution of the problem and for finding the minimum configuration energy [6] . We follow the same approach as in Ref. [2] but here we use COBYLA for the classical optimization part. \n",
```
mrossinek, line 395: > The problem that we are tackling has now been mapped to an optimizaiton problem.

How? Where? By generating the `qubit_op`?

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,708 @@
...
...5AAAoqOjMXToUJw6dQrh4eEYNGgQunbtqlMcRPRgTICI6IGioqIwadIkrF27FrGxsWjVqhV69OghH2/dujV++uknlJWVVUt0MjIyUFBQgEcffVRrf4sWLfQ6hLyoqAgA8P333+ORRx7ROmZra6u17ejoqLU9ffp0xMfHY8WKFQgICIC9vT2ee+45lJWV6S2+qu5O9iRJglqtBgD06dMHV65cwQ8//ID4+Hj07NkTMTExWLFihUFiIbJUHAVGRA/0/PPPw8rKCtu2bcPWrVvx0ksvQZIk+fjIkSNRVFSETz75pNq1K1asgJ2dHYYPH663eDRJlkqlkvcFBQXB1tYWaWlpCAgI0Po0bdr0vvf7+eefMXbsWAwePBjBwcHw8fHB5cuXqz2z6vNq0qZNG/z888/V7h0UFKRD6QBPT09ERkbis88+w6pVq7BhwwadrieiB2MNEBE9kJOTE4YPH45Zs2ahoKAAY8eO1ToeFhaGKVOmYMaMGSgrK8OgQYNQXl6Ozz77DB9++CE2b94MDw8PrWtycnJQUlKitc/Dw6NWTWH+/v6QJAlxcXHo27cv7O3t4ezsjOnTp2Pq1KlQq9Xo3r078vPz8fPPP8PFxQWRkZH3vF/r1q2xY8cO9O/fH5IkYc6cOXKNjEbz5s1x+PBhjBgxAra2tmjcuHG1+8yYMQPPP/88QkND0atXL3z33XfYsWMH9u7d+8AyacydOxcdO3ZE27ZtUVpairi4OLRp06bW1xNRLZm6ExIRNQxHjx4VALQ69N5t48aNomPHjsLOzk4AEEqlUhw6dEjrHE1H5Jo+CQkJNd737k7QQgixcOFC4ePjIyRJkofBq9VqsWrVKvHYY48JhUIhPD09RUREhBzDvUagpaamiqefflrY29uLpk2bijVr1ogePXqIKVOmyOckJCSIkJAQYWtr+9DD4Hfu3Km1z9XVVcTGxgohhFi0aJFo06aNsLe3F+7u7mLgwIHi0qVL93jjRFRXkhAP6LlIRFQHly9fRo8ePRAWFobPP/8c1tbWpg6JiEjGPkBEZBDNmzfHwYMHERgYKC9hQURUX7AGiIiIiCwOa4CIiIjI4jABIiIiIovDBIiIiIgsDhMgIiIisjhMgIiIiMjiMAEiIiIii8MEiIiIiCwOEyAiIiKyOEyAiIiIyOL8PzAhpzTf9hrXAAAAAElFTkSuQmCC\n",
```
mrossinek, line 622: Seeing this image it looks like the VQE does an awful lot of unnecessary iterations. Can't we shorten it to ~50-75 cycles to keep the tutorial runtime down?

mrossinek (commented):
Just some comments :slightly_smiling_face:

docs/tutorials/09_Protein_Folding.ipynb
pbark: yes we can do it, but sometimes (due to cvar and qasm simulator you need more iterations so I would prefer to keep it to 200).

docs/tutorials/09_Protein_Folding.ipynb
woodsp-ibm: Should be `chirality constraint` I think ie `constraint` with the `t` - (constrain is a verb)

mrossinek: Thanks I overlooked that typo. I have updated my suggestion accordingly

docs/tutorials/09_Protein_Folding.ipynb
mrossinek: The tutorial should be run with the `algorithm_seed` specified. Otherwise it will keep changing it's values everytime its run by the CI. That is undesirable.

docs/tutorials/09_Protein_Folding.ipynb
pbark: I missed the superscript! Thanks!

docs/tutorials/09_Protein_Folding.ipynb
pbark: Thanks!

docs/tutorials/09_Protein_Folding.ipynb
pbark: I can see it locally. I guess is rendering.

mrossinek: We need to be careful and check how the Sphinx docs render it..

mrossinek: When the next CI passes you can download the generated docs as artefacts and take a look :+1:

woodsp-ibm: After CI has run download the 'tutorials3.8` Artifact from the CI build Details page and expand that. It has full html as would be published.

mrossinek:
@woodsp-ibm we are still seeing randomness in the produced results although we are setting `algorithm_globals` and the seeds for `QuantumInstance`. Do you know what we might be missing that could explain this?

woodsp-ibm:
> we are still seeing randomness in the produced results although we are setting algorithm_globals and the seeds for QuantumInstance. Do you know what we might be missing that could explain this?

That should be all that is needed - as long as no code anywhere is directly using np.random - which I think was the case for random interaction but was changed. I'll look a bit later too.

docs/tutorials/09_Protein_Folding.ipynb
woodsp-ibm: I think the changing behavior is because of the fact that the Paulis are grouped and the grouping algorithm was found to be non-deterministic in terms of the order of the groups - see Qiskit/retworkx#347. The issue has been closed but looking at dates it does not look like its in a released build yet. So I think in the interim we will have to accept that result varies (we noticed originally in some unit test we have which would occasionally fail and it turned out to be because the order of the groups varies (hence the ccts are executed in different order and we do not get quite the same outcome)

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "from qiskit.algorithms.optimizers import COBYLA\n",
+    "from qiskit.algorithms import NumPyMinimumEigensolver, VQE\n",
+    "from qiskit.opflow import PauliExpectation, CVaRExpectation\n",
+    "from qiskit import execute, Aer\n",
+    "\n",
+    "# set classical optimizer\n",
+    "optimizer = COBYLA(maxiter=50)\n",
+    "\n",
+    "# set variational ansatz\n",
+    "ansatz = RealAmplitudes(7, reps=1)\n",
```
woodsp-ibm, line 598:
```suggestion
    "ansatz = RealAmplitudes(reps=1)\n",
```
We should not need to set number of qubits when building out anzatzes. That was the whole rationale behind blueprint ccts to set it later once the size was known from the operator. VQE does this when one calls to compute the min eig value - and in this case its overwritten to '9' which is the size of the operator. So setting 7 here is misleading and confusing I think. Better not to do it at all - if one has to do it set it to the qubit operator size so its consistent - I would leave it unset though as this is how we would normally expect it to be used.

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "\n",
+    "#set the backend\n",
+    "backend_name = 'qasm_simulator' \n",
```
woodsp-ibm, line 601:
```suggestion
    "backend_name = 'aer_simulator' \n",
```
The name `qasm_simulator` I believe is now a legacy name for Aer.

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "\n",
+    "# initialize CVaR_alpha objective with alpha = 0.1\n",
+    "cvar_exp = CVaRExpectation(0.1, PauliExpectation())\n",
+    "\n",
+    "# initialize VQE using CVaR\n",
+    "vqe = VQE(expectation=cvar_exp, optimizer=optimizer, ansatz=ansatz, quantum_instance=backend,\n",
+    "          callback=store_intermediate_result)\n",
+    "\n",
+    "eigenvalue = vqe.compute_minimum_eigenvalue(qubit_op)\n",
+    "print(eigenvalue)"
```
woodsp-ibm, line 623: The name here seems to suggest it passes back the `eigenvalue` which can be seen from the print below that this is not the case. Maybe 'result' - or 'min_eig_result' or something else that says this is more than a single value might be a better name here.

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,720 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "62d99a5f",
+   "metadata": {},
+   "source": [
+    "Further details about the used model and the encoding of the problem can be found in [2].\n",
+    "\n",
+    "<font size='2'>[2] A.Robert, P.Barkoutsos, S.Woerner and I.Tavernelli, Resource-efficient quantum algorithm for protein folding, NPJ Quantum Information, 2021 </font>"
```
woodsp-ibm, line 90: Is there a reason why these references are not collected together at the end of the notebook in a references section? Having them in among the text means you have to go look for them a bit, although like the above its so close and I think is the only reference the pointer could be much more direct.
I had to search hard in the html for [4] - turns out it wrapped in a paragraph and tagged directly on the end of [3] so was not so easy to find. Having a references section with these as the bullets, if they are not wanted inline, would make it so much easier I think.

woodsp-ibm:
image

Its hard to comment inline in the notebook since a lot of text stays on one very long line. The above does not come out well in the html - sphinx needs extra blank line ahead of lists - otherwise as you see here, unlike the notebook rendering - it sees the text as part of the same paragraph.

And for the variable names here - would be be better to stay with 'penalty' rather than drop back to the older 'lambda' naming which was dropped.

docs/tutorials/09_Protein_Folding.ipynb
pbark: Good catch. This was a leftover from previous tests. Now is updated.

docs/tutorials/09_Protein_Folding.ipynb
pbark: I added a section with all the references. I think it is now easier to read.

docs/tutorials/09_Protein_Folding.ipynb
pbark: Update as suggested.

docs/tutorials/09_Protein_Folding.ipynb
pbark: Thanks for this.

pbark:
I believe I updated all list and bullets to render properly in the html. Thanks for pointing them out.

woodsp-ibm:
That reads a lot easier now, thanks.

* It would be nice though for links in the references where they refer to online papers etc.

* One other comment. In the section that does the computation it says this:

> We can use the Variational Quantum Eigensolver with Conditional Value at Risk (CVAR) expectation value 

 I.e `can use` - what about VQE with the regular PauliExpectation does that work at all (I know when investigating the reproducibility aspect I tried that but the min energy was very different if I recall). It might be nice for a summary here of the rationale of using CVar - some people seem to have an aversion to reading papers - so some highlight might be nice. How about QAOA? I assume whatever the answer is would be related to say why we do not see some classical reference value say by using the NumpyMinEigsolver.

.pylintdict
```diff
@@ -373,6 +386,7 @@ tavernelli
 taylor
 tdg
 tensored
+tetrahydral
```
mrossinek, line 389: I believe this should be `tetrahedral` instead.

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+   "source": [
+    "The structure and function of many natural and human-engineered\n",
+    "proteins is still only poorly understood. As a result, our understanding of\n",
+    "processes connected with protein folding, such as those encountered in\n",
+    "Alzheimer’s disease, vaccine development, and crop improvement\n",
+    "research, has remained limited.\n",
+    "\n",
+    "Unfolded polypeptides have a very large number of degrees of freedom\n",
+    "and thus an enormous number of potential conformations. For example, a\n",
+    "chain with 100 aminoacids has on the order of 10<sup>47</sup> conformations. In\n",
```
mrossinek, line 32: The superscript does not get rendered properly in Qiskit's documentation html output:
screenshot_1629098518

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+   "id": "29deb98b",
+   "metadata": {},
+   "source": [
+    "For the problem encoding we use: \n",
+    "\n",
+    "- Configuration qubits: qubits that are used to describe the configurations and the relative position of the different beads\n",
+    "\n",
+    "- Interaction qubits: qubits that encode interactions between the different aminoacids\n",
+    "\n",
+    "For our case we use a tetrahydral lattice (diamond shape lattice) where we encode the movement through the configuration qubits (see image below). \n",
```
mrossinek, line 61:
```suggestion
    "For our case we use a tetrahedral lattice (diamond shape lattice) where we encode the movement through the configuration qubits (see image below). \n",
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f571d4a8",
+   "metadata": {},
+   "source": [
+    "The Hamiltonian of the system for a set of qubits $\\mathbf{q}=\\{\\mathbf{q}_{cf}, \\mathbf{q}_{in}\\}$ is \n",
+    "\n",
+    "$$H(\\mathbf{q}) = H_{gc}(\\mathbf{q}_{cf}) + H_{ch}(\\mathbf{q}_{cf}) + H_{in}(\\mathbf{q}) $$\n",
```
mrossinek, line 73: Does `H_{in}` depend on the entire set of qubits (as it is currently written) or only on `q_{in}`?

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+    "Unfolded polypeptides have a very large number of degrees of freedom\n",
+    "and thus an enormous number of potential conformations. For example, a\n",
+    "chain with 100 aminoacids has on the order of 10<sup>47</sup> conformations. In\n",
+    "reality, however, many proteins fold to their native structure within\n",
+    "seconds. This is known as Levinthal’s paradox [1].\n",
+    "\n",
+    "The exponential growth of potential conformations with chain length\n",
+    "makes the problem intractable for classical computers. In the quantum\n",
+    "framework, our resource-efficient algorithm scales linearly with\n",
+    "the number of amino acids N."
```
mrossinek, line 39: To conform with the rest of the tutorial:
```suggestion
    "the number of aminoacids N."
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+   "source": [
+    "### Side Chains"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d71b4571",
+   "metadata": {},
+   "source": [
+    "Beyond the main chain of the protein there may be aminoacids attached to the residues of the main chain. Our model allows for side chains of the maximum length of one. Elongated side chains would require the introduction of additional penalty terms which are still under development. "
```
mrossinek, line 152: Maybe add a sentence that in this simple example we don't specify any side chains. Otherwise the following code block is not really connected to the text.

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "4d148961",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "side_chains = ['']*7"
```
mrossinek, line 162: To conform with Qiskit's code conventions:
```suggestion
    "side_chains = [""]*7"
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "id": "d37df517",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "peptide = Peptide(main_chain,side_chains)"
```
mrossinek, line 253:
```suggestion
    "peptide = Peptide(main_chain, side_chains)"
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+   "source": [
+    "### Using VQE with CVAR expectation value for the solution of the problem"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0f2d3e2c",
+   "metadata": {},
+   "source": [
+    "The problem that we are tackling has now implement all the physical constraints and has a diagonal Hamiltonian. We can use the Variational Quantum Eigensolver with Conditional Value at Risk (CVAR) expectation value for the solution of the problem and for finding the minimum configuration energy [6] . We follow the same approach as in Ref. [2] but here we use COBYLA for the classical optimization part. "
```
mrossinek, line 392:
```suggestion
    "The problem that we are tackling has now implemented all the physical constraints and has a diagonal Hamiltonian. We can use the Variational Quantum Eigensolver with Conditional Value at Risk (CVaR) expectation values for the solution of the problem and for finding the minimum configuration energy [6] . We follow the same approach as in Ref. [2] but here we use COBYLA for the classical optimization part. "
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+   "source": [
+    "print(qubit_op)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f170cad7",
+   "metadata": {},
+   "source": [
+    "### Using VQE with CVAR expectation value for the solution of the problem"
```
mrossinek, line 384:
```suggestion
    "### Using VQE with CVaR expectation value for the solution of the problem"
```

mrossinek (commented):
Some minor comments. I agree with @woodsp-ibm that the references could contain links to the online versions of the respective papers.

docs/tutorials/09_Protein_Folding.ipynb
pbark: Yes H_in is depended in the full qubit registed since it requires knowledge of the structure of the beads.

mrossinek: Might I suggest to make this more explicit then? Something like this:
```suggestion
    "$$H(\\mathbf{q}) = H_{gc}(\\mathbf{q}_{cf}) + H_{ch}(\\mathbf{q}_{cf}) + H_{in}(\\mathbf{q}_{cf}, \\mathbf{q}_{in}) $$\n",
```

pbark: Yes H_in is depended in the full qubit registed since it requires knowledge of the structure of the beads.

woodsp-ibm:
In ref [3] there seems to look like a doi reference like was added to the others. It does not show as a a link though 
> Applied Chemistry. 31 (4): 641–645. doi:10.1351/pac197231040639. PMID 5080161.

For the `10^47` that could done with math there to look a bit nicer - the 100 could be done with math too so the fonts look the same since the numbers are connected as such.

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+    "proteins is still only poorly understood. As a result, our understanding of\n",
+    "processes connected with protein folding, such as those encountered in\n",
+    "Alzheimer’s disease, vaccine development, and crop improvement\n",
+    "research, has remained limited.\n",
+    "\n",
+    "Unfolded polypeptides have a very large number of degrees of freedom\n",
+    "and thus an enormous number of potential conformations. For example, a\n",
+    "chain with 100 aminoacids has on the order of 10^47 conformations. In\n",
+    "reality, however, many proteins fold to their native structure within\n",
+    "seconds. This is known as Levinthal’s paradox [1].\n",
```
dlasecki, line 34:
```suggestion
    "seconds. This is known as the Levinthal’s paradox [1].\n",
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+   "id": "50dc1324",
+   "metadata": {},
+   "source": [
+    "The Hamiltonian of the system for a set of qubits $\\mathbf{q}=\\{\\mathbf{q}_{cf}, \\mathbf{q}_{in}\\}$ is \n",
+    "\n",
+    "$$H(\\mathbf{q}) = H_{gc}(\\mathbf{q}_{cf}) + H_{ch}(\\mathbf{q}_{cf}) + H_{in}(\\mathbf{q}) $$\n",
+    "\n",
+    "where \n",
+    "\n",
+    "- $H_{gc}$ is the geometrical constraint term (governing the growth of the primary sequence of aminoacids without bifurcations\n",
```
dlasecki, line 77:
```suggestion
    "- $H_{gc}$ is the geometrical constraint term (governing the growth of the primary sequence of aminoacids without bifurcations)\n",
```

docs/tutorials/09_Protein_Folding.ipynb
```diff
@@ -0,0 +1,824 @@
...
+   "execution_count": 5,
+   "id": "d96678c3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "penalty_back = 10\n",
+    "penalty_chiral = 10\n",
+    "penalty_1 = 10\n",
+    "\n",
+    "penalty_terms = PenaltyParameters(penalty_chiral, penalty_back, penalty_1)"
```
dlasecki, line 227: We could potentially make use of default values which are also 10. But perhaps it is better to be explicit in the tutorial...
```suggestion
    "\n",
    "penalty_terms = PenaltyParameters()"
```

dlasecki (changes requested):

docs/tutorials/09_Protein_Folding.ipynb
mrossinek: I actually don't think this article is needed

docs/tutorials/09_Protein_Folding.ipynb
mrossinek: I think the tutorial should be explicit here :+1:

docs/tutorials/09_Protein_Folding.ipynb
woodsp-ibm: Do you really put `the` there? For me it would sound very strange indeed.

woodsp-ibm (approved):

mrossinek (approved):
Thanks @pbark!

docs/tutorials/09_Protein_Folding.ipynb
dlasecki: Ok.

dlasecki (approved):

PR #329, opened 2021-08-16
`VQEProgram` should state it supports aux ops

Cryoris:
### Summary

The `VQEProgram` does support the evaluation of auxiliary operators at the final state, but the 
`supports_aux_operators` method previously returned `False` instead of `True`.

### Details and comments

Also refactored the tests minimally to avoid code duplication.

### DIFF
--- a/test/circuit/library/ansatzes/test_uccsd.py
+++ b/test/circuit/library/ansatzes/test_uccsd.py
@@ -106,5 +106,6 @@ class TestUCCSD(QiskitNatureTestCase):
             num_particles=num_particles,
             num_spin_orbitals=num_spin_orbitals,
         )
+        print(ansatz.draw())
 
         assert_ucc_like_ansatz(self, ansatz, num_spin_orbitals, expect)
--- a/test/circuit/library/ansatzes/test_uccsd.py
+++ b/test/circuit/library/ansatzes/test_uccsd.py
@@ -106,6 +106,5 @@ class TestUCCSD(QiskitNatureTestCase):
             num_particles=num_particles,
             num_spin_orbitals=num_spin_orbitals,
         )
-        print(ansatz.draw())
 
         assert_ucc_like_ansatz(self, ansatz, num_spin_orbitals, expect)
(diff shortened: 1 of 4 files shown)


test/circuit/library/ansatzes/test_uccsd.py
```diff
@@ -106,5 +106,6 @@ def test_uccsd_ansatz(self, num_spin_orbitals, num_particles, expect):
             num_particles=num_particles,
             num_spin_orbitals=num_spin_orbitals,
         )
+        print(ansatz.draw())
```
mrossinek, line 109: I don't think this should be here

test/runtime/test_vqeprogram.py
```diff
@@ -36,25 +39,42 @@ def setUp(self):
...
+        vqe, operator = self.get_standard_program()
+        vqe.optimizer = optimizer
         result = vqe.compute_minimum_eigenvalue(operator)
 
         self.assertIsInstance(result, VQEResult)
 
+    def test_supports_aux_ops(self):
+        """Test the VQEProgram says it supports aux operators."""
+        vqe, _ = self.get_standard_program()
+        self.assertTrue(vqe.supports_aux_operators)
```
mrossinek, line 69: Do you also want to actually test the aux_operators somewhere?

mrossinek (commented):
Just some minor comments :+1:

test/circuit/library/ansatzes/test_uccsd.py
Cryoris: Whoops 😬

test/runtime/test_vqeprogram.py
Cryoris: That would require running the VQE on runtime which we didn't include in the nature/optimization tests. We have tests in the runtime repo however, that do test this 🙂

Cryoris: Running some tests in nature would be good generally, but that requires some more planning (maybe nightly runs would be better since it runs on the IBM simulator on the server) and provider access etc.

mrossinek: Alright :+1:

mrossinek (approved):

woodsp-ibm (approved):
So this is the equivalent here to Qiskit/qiskit-optimization#223 it seems.

PR #330, opened 2021-08-17
Fix FermionicOp Warning msg format

woodsp-ibm:
Strings in a warning message that was added were missing blank spaces (I put at end of lines) so words like 'dense' and to, specify and display_format have space between them when the strings are concatenated - below is how it is prior to the spaces being added.

> The default value for `display_format` will be changed from 'dense'to 'sparse' in version 0.3.0. Once that happens, you must specifydisplay_format='dense' directly.


mrossinek (approved):
Thanks @woodsp-ibm for catching this!

PR #332, opened 2021-08-18
Avoid UserWarning during unittests

mrossinek:
### Summary

The FermionicOp is emitting a UserWarning because we plan to change the
default `display_format` from 'dense' to 'sparse'. This warning will
only be emitted if the user did not actively specify the
`display_format` themselves. Hence, explicitly setting it during tests
will avoid the warnings.

### Details and comments

Please do not merge before checking the CI output for `UserWarning`s


mrossinek:
Looks like I got all of them :+1: This PR is good to go :tada:

manoelmarques (approved):

PR #339, opened 2021-08-25
CI annotations

mrossinek:

### DIFF
--- a/.github/problem_matchers/mypy.json
+++ b/.github/problem_matchers/mypy.json
@@ -1,0 +1,17 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "mypy",
+            "pattern": [
+                {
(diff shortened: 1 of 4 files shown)


mrossinek:
Another thing to note is that currently the annotations will be added for all OS variants which we are testing. Arguably we can limit these to a single one. Although this might not catch all scenarios it will probably cover most cases if we limit the annotations to the action with the oldest Python version. @manoelmarques wdyt?

manoelmarques:
The oldest python is 3.6 but it is only used for unit tests under Ubuntu. For now probably a better choice is python 3.7 as it is used in both the checks job and unit tests. When I eventually change the checks job to python 3.8 then I can change this to python 3.8 too.

mrossinek:
> In some rare occasions some error that happens only in 3.6, 3.8, 3.9 may be missed but it is better than have a duplication all the times.

Sounds good! After all: these messages will still be there they just won't be made as visible via annotations as the ones for Python 3.7

PR #341, opened 2021-08-26
To_matrix method for FermionicOp

LaurinFischer:
### Summary
This PR implements a `to_matrix()` method for `FermionicOp`. This method returns the matrix representation of the fermionic operator over the full fermionic Fock space (with no fixed particle number). 
This was suggested and discussed some while ago in #129.

### Details and comments
The basis states over which the matrix is constructed are ordered increasingly as 00..00, 00..01, ..., 11..11.
Per default, a `scipy.sparse` matrix is returned which can be turned into a numpy array a boolean flag.


releasenotes/notes/fermionic-op-to-matrix-4abd3f1b53f6e6d5.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Add :meth:`~qiskit_nature.operators.second_quantization.FermionicOp.to_matrix()`.
+    This method returns the matrix representation of the operator over the full fermionic Fock space in
+    occupation number basis.
```
mrossinek, line 6:
```suggestion
    This method returns the matrix representation of the operator over the full fermionic Fock space in
    the occupation number basis.
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -464,6 +465,74 @@ def to_list(
...
+            The matrix of the operator in the Fock basis (scipy.sparse.csc_matrix or numpy.ndarray
+            with dtype=numpy.complex128)
+        """
+
+        csc_data, csc_col, csc_row = [], [], []
+
+        # loop over all columns of the matrix
+        for col_idx in range(2 ** self.register_length):
+            initial_occupations = [
+                int(occ) for occ in np.binary_repr(col_idx, self.register_length)
```
mrossinek, line 486: If you store these as `bool` rather than `int` you can simplify the code further quite a bit. I'll add an example there

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -464,6 +465,74 @@ def to_list(
...
+                            sign *= (-1) ** sum(occupations[: label_primitive.index])
+                            occupations[label_primitive.index] = 0
+
+                        if label_primitive.is_creation:
+                            # If this mode is already occupied, the action of '+' on this state is zero
+                            if occupations[label_primitive.index] == 1:
+                                mapped_to_zero = True
+                                break
+                            sign *= (-1) ** sum(occupations[: label_primitive.index])
+                            occupations[label_primitive.index] = 1
```
mrossinek, line 516: Assuming you change `occupations` to store booleans rather than integers you can simplify the code
by quite a bit until it becomes the below:
```suggestion
                        occ = occupations[label_primitive.index]
                        if label_primitive.is_creation == occ:
                            # Applying the creation operator on an occupied state maps to zero. So
                            # does applying the annihilation operator on an unoccupied state.
                            mapped_to_zero = True
                            break
                        sign *= (-1) ** sum(occupations[: label_primitive.index])
                        occupations[label_primitive.index] = not occ
```

mrossinek (commented):
Thank you @LaurinFischer for looking into this! I have some small refactoring ideas but other than that this looks good! :slightly_smiling_face:

qiskit_nature/operators/second_quantization/fermionic_op.py
LaurinFischer: Thanks for taking a look Max, I think switching to a boolean representation makes a lot of sense. This is adapted in 
3df5528.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -464,6 +465,67 @@ def to_list(
...
+            The matrix of the operator in the Fock basis (scipy.sparse.csc_matrix or numpy.ndarray
+            with dtype=numpy.complex128)
+        """
+
+        csc_data, csc_col, csc_row = [], [], []
+
+        # loop over all columns of the matrix
+        for col_idx in range(2 ** self.register_length):
+            initial_occupations = [
+                occ == "1" for occ in np.binary_repr(col_idx, self.register_length)
```
mrossinek, line 486:
```suggestion
                occ == "1" for occ in f"{col_idx:0{self.register_length}b}"
```
This performs slightly better than the numpy-call

woodsp-ibm:
Is this something where we can add a test, with say a small molecule, where the FermionicOp is a bit more complex for the to_matrix and then for example use numpy to extract eigenvalue(s) and assert its correctness via their expected value(s)?

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -464,6 +465,67 @@ def to_list(
...
+
+        Returns:
+            The matrix of the operator in the Fock basis (scipy.sparse.csc_matrix or numpy.ndarray
+            with dtype=numpy.complex128)
+        """
+
+        csc_data, csc_col, csc_row = [], [], []
+
+        # loop over all columns of the matrix
+        for col_idx in range(2 ** self.register_length):
```
mrossinek, line 484: I suggest you extract the computation of `2 ** self.register_length` into a variable because you actually need it again twice further down.
Also, bit-shifting becomes significantly faster for large register lengths. So you can instead do something like:
```python
dimension = 1 << self.register_length
```
A quick `timeit` example for a length of 128:
```python
>>> timeit.timeit("2 ** 128")
0.28181123600006686
>>> timeit.timeit("1 << 128")
0.04366266599981827
```

test/operators/second_quantization/test_fermionic_op.py
```diff
@@ -335,6 +338,64 @@ def test_label_display_mode(self, label, pre_processing):
...
+                ("+_1 -_1 +_3 -_3", (0.6985737398458793)),
+                ("+_1 -_1 +_2 -_2", (0.6645817352647293)),
+                ("+_0 -_0", (-1.2563391028292563)),
+                ("+_0 -_0 +_3 -_3", (0.6645817352647293)),
+                ("+_0 -_0 +_2 -_2", (0.6757101625347564)),
+                ("+_0 -_0 +_1 -_1", (0.48365053378098793)),
+            ]
+            h2_matrix = FermionicOp(h2_labels, register_length=4).to_matrix()
+            evals, evecs = eigs(h2_matrix)
+            self.assertTrue(np.allclose(np.min(evals), -1.8572750 + 0j))
```
mrossinek, line 392:
```suggestion
            self.assertTrue(np.isclose(np.min(evals), -1.8572750))
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -464,6 +465,67 @@ def to_list(
...
+                            # Applying the creation operator on an occupied state maps to zero. So
+                            # does applying the annihilation operator on an unoccupied state.
+                            mapped_to_zero = True
+                            break
+                        sign *= (-1) ** sum(occupations[: label_primitive.index])
+                        occupations[label_primitive.index] = not occ
+
+                    # add data point to matrix in the correct row
+                    if not mapped_to_zero:
+                        row_idx = int("".join([str(int(occ)) for occ in occupations]), 2)
```
mrossinek, line 513: I think we can also improve this a bit. The back-and-forth conversion of `int` and `str` caught my eye a little bit.
Instead I propose the following:
```suggestion
                        row_idx = sum(int(occ) << idx for idx, occ in enumerate(occupations[::-1]))
```
This results in quite a significant speedup:
```python
>>> timeit.timeit("int(''.join([str(int(occ)) for occ in occs]), 2)", "occs = [True] * 32 + [False] * 32")
12.031811372999982
>>> timeit.timeit("sum(int(occ) << idx for idx, occ in enumerate(occs[::-1]))", "occs = [True] * 32 + [False] * 32")
7.646822761999829
```

mrossinek (commented):
I have some minor improvement suggestions but other than that, now that Steve's comment is addressed, I believe this is good to go :+1:

qiskit_nature/operators/second_quantization/fermionic_op.py
LaurinFischer: Thanks again for these nice ideas, they are adapted here b593f70

woodsp-ibm (approved):
LGTM - thanks for adding the additional testing

mrossinek (approved):
Thanks @LaurinFischer! LGTM, too :+1:

PR #344, opened 2021-08-30
Fix #343

### LINKED ISSUE #343: Update `SparsePauliOp`-related code
The scheduled CI run from last night failed (https://github.com/Qiskit/qiskit-nature/runs/3452786046?check_suite_focus=true) most likely due to merging of this PR in Terra: https://github.com/Qiskit/qiskit-terra/pull/6826

@ikkoham would you be willing to look into fixing this? If yes, please assign yourself :+1:

ikkoham:
### Summary

Fix #343.

### Details and comments

There are two reasons: first, SparsePauliOp had been constructed by the way not officially supported (i.e. `SparsePauliOp(boolean array)` is not supported and `SparsePauliOp(PauliTable(boolean array ))` is correct). The second is that I modified the bug https://github.com/Qiskit/qiskit-terra/issues/6815, but originally nature use a specification where this phase was ignored.

### DIFF
--- a/qiskit_nature/mappers/second_quantization/bravyi_kitaev_mapper.py
+++ b/qiskit_nature/mappers/second_quantization/bravyi_kitaev_mapper.py
@@ -13,7 +13,6 @@
 """The Bravyi-Kitaev Mapper."""
 
 import numpy as np
-
 from qiskit.opflow import PauliSumOp
 from qiskit.quantum_info.operators import Pauli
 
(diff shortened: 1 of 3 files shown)


mrossinek (commented):
Hi @ikkoham, thanks for looking into this! :slightly_smiling_face: 

This already looks good but given your description above I started to wonder whether the code [in this method](https://github.com/Qiskit/qiskit-nature/blob/main/qiskit_nature/mappers/second_quantization/qubit_mapper.py#L64) also needs to be updated since it is also based on boolean arrays. Could you take a look, please?

ikkoham:
@mrossinek Thanks. The Bravyi Kitaev mapper is multiplied (&) using Pauli, but the qubit_mapper is multiplied using `SparsePauliOp`, so the point seems to have already been taken care of.

mrossinek:
Sorry, my bad. I had mistaken these lines (https://github.com/Qiskit/qiskit-nature/blob/main/qiskit_nature/mappers/second_quantization/qubit_mapper.py#L91) for constructions based on boolean arrays but these are actually based on `Pauli` instances. All good then :+1:

mrossinek (approved):

PR #346, opened 2021-09-02
Remove unused imports

mrossinek:
### Summary

Somehow the [scheduled CI run](https://github.com/Qiskit/qiskit-nature/actions/runs/1192487099) from last night caught some unused imports which previously went unnoticed. This PR removes those imports.


pbark (approved):
Thanks for the immediate fix. LGTM!

mrossinek:
I did notice that on Linux the `unused-import` errors do not always get caught properly. I am not familiar with an issue on the side of `pylint` itself tracking this, however.

PR #347, opened 2021-09-02
Fix #345

### LINKED ISSUE #345: Problem molecular_data_transformed not being set with legacy logic
In looking at the following slack posting on the Qiskit workspace it seems that 

https://qiskit.slack.com/archives/C7SJ0PJ5A/p1630241280123200

```
from qiskit_nature.drivers import PySCFDriver
from qiskit_nature.converters.second_quantization.qubit_converter import QubitConverter

molecule = "H .0 .0 .0; H .0 .0 0.739"
driver = PySCFDriver(atom=molecule)
qmolecule = driver.run()

second_q_ops = problem.second_q_ops()

main_op = second_q_ops[0]

mapper = JordanWignerMapper()
converter = QubitConverter(mapper=mapper)
```
works fine to here but then 
```
num_particles = (problem.molecule_data_transformed.num_alpha,
             problem.molecule_data_transformed.num_beta)
qubit_op = converter.convert(main_op, num_particles=num_particles)
```
fails with 
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-6-a0b98f51e882> in <module>()

----> 2 num_particles = (problem.molecule_data_transformed.num_alpha,


AttributeError: 'NoneType' object has no attribute 'num_alpha'
```

This was my response in the post

> From what I can see the above looks like a bug in the code that when the logic was adjusted, and the above became the old, now deprecated way of doings things, that unless a transform (such as ActiveSpaceTransformer) is explicitly given the transformed data ends up not getting set whereas in the past it was the same as the driver result (ie. the non-transformed data) so whether you did a transform or not that was always the right result.

mrossinek:
### Summary

If not transformers are supplied, we can provide `molecular_data_transformed` even in the non-legacy case because it is simply a copy of `molecular_data`.

Closes #345

### DIFF
--- a/qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
+++ b/qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
@@ -83,6 +83,10 @@ class ElectronicStructureProblem(BaseProblem):
                 )
 
             else:
+                if not self.transformers:
+                    # if no transformers are supplied, we can still provide
+                    # `molecule_data_transformed` as a copy of `molecule_data`
+                    self._molecule_data_transformed = self._molecule_data
                 self._grouped_property_transformed = self._transform(self._grouped_property)
 
         else:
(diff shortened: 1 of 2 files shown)


pbark (approved):
LGTM!

woodsp-ibm:
Arguably this ought to have had a reno. I think it should be backported to stable since its affecting users code from 0.1.0 version.

PR #350, opened 2021-09-02
Duck-type VQE interim results

renier:
### Summary

Add a check for the type of data expected in the callback wraper in VQEProgram. This will allow us to close Qiskit-Partners/qiskit-runtime#38

### Details and comments

Anticipating changes in the Qiskit Runtime API which will start emitting final results in the WebSocket stream. When that happens, if we don't duck-type in the VQE callback wrapper receiving the interim results, it will get a ValueError exception.

### DIFF
--- a/qiskit_nature/runtime/vqe_program.py
+++ b/qiskit_nature/runtime/vqe_program.py
@@ -209,15 +209,16 @@ class VQEProgram(MinimumEigensolver):
     def _wrap_vqe_callback(self) -> Optional[Callable]:
         """Wraps and returns the given callback to match the signature of the runtime callback."""
 
-        def wrapped_callback(*args):
+        def wrapped_callback(*args) -> None:
             _, data = args  # first element is the job id
             if isinstance(data, dict):
-                return # not expected params. skip
+                return  # not expected params. skip
             iteration_count = data[0]
             params = data[1]
             mean = data[2]
             sigma = data[3]
-            return self._callback(iteration_count, params, mean, sigma)
+            self._callback(iteration_count, params, mean, sigma)
+            return
 
         # if callback is set, return wrapped callback, else return None
         if self._callback:
--- a/qiskit_nature/runtime/vqe_program.py
+++ b/qiskit_nature/runtime/vqe_program.py
@@ -211,6 +211,8 @@ class VQEProgram(MinimumEigensolver):
 
         def wrapped_callback(*args):
             _, data = args  # first element is the job id
+            if isinstance(data, dict):
+                return # not expected params. skip
             iteration_count = data[0]
             params = data[1]
             mean = data[2]
--- a/qiskit_nature/runtime/vqe_program.py
+++ b/qiskit_nature/runtime/vqe_program.py
@@ -197,16 +197,16 @@ class VQEProgram(MinimumEigensolver):
         self._store_intermediate = store
 
     @property
-    def callback(self) -> Callable:
+    def callback(self) -> Callable[[int, np.ndarray, float, float], None]:
         """Returns the callback."""
         return self._callback
 
     @callback.setter
-    def callback(self, callback: Callable) -> None:
+    def callback(self, callback: Callable[[int, np.ndarray, float, float], None]) -> None:
         """Set the callback."""
         self._callback = callback
 
-    def _wrap_vqe_callback(self) -> Optional[Callable]:
+    def _wrap_vqe_callback(self) -> Optional[Callable[[int, np.ndarray, float, float], None]]:
         """Wraps and returns the given callback to match the signature of the runtime callback."""
 
         def wrapped_callback(*args) -> None:


qiskit_nature/runtime/vqe_program.py
```diff
@@ -211,6 +211,8 @@ def _wrap_vqe_callback(self) -> Optional[Callable]:
 
         def wrapped_callback(*args):
             _, data = args  # first element is the job id
+            if isinstance(data, dict):
+                return # not expected params. skip
```
woodsp-ibm, line 215: You will see from the error that this file needs reformatting by black sine the additional code would be restyled according to black style - I imagine the inline comment needs a second space,

Also there are lint complaints about the return below apparently returning something yes this return does not so it considers it inconsistent. Since the callback, as far as I know, is not intended to return anything calling self._callback(... and the doing the return on its own afterwards will probably fix this. Arguably the method here should be typehinted to show its return i.e. -> None.

I will note the callback setter and getter above just say Callable - I believe the Callable should have the same type signature as that on the constructor so I do not know why the typehint is not the same.

As to the typing check to differentiate data in the result stream - was anything more "robust" considered where perhaps some identifier is added to allow the different data aspects to be filtered - or is this more down to how an end user/program uses the stream and up to them. To me it seems a bit limiting/hacky but the I do not know the full details.

woodsp-ibm:
Whatever gets done here in Nature the equivalent logic will no doubt be needed in the equivalent file in Qiskit Optimization here https://github.com/Qiskit/qiskit-optimization/blob/main/qiskit_optimization/runtime/vqe_program.py

qiskit_nature/runtime/vqe_program.py
renier: > ﻿﻿As to the typing check to differentiate data in the result stream - was anything more "robust" considered where perhaps some identifier is added to allow the different data aspects to be filtered - or is this more down to how an end user/program uses the stream and up to them.

It is completely up to the runtime program and how the client-side reads results from the stream. The pattern for the stream is more of a generic event stream which can include any type of object. It is up to the program to setup the right identifiers and to the program to inspect them so they can know how to handle them. In this case, the easiest differentiator I saw was that the interim results of the VQE runtime program are an array, while the final result is a dictionary. Why are they so different, I'm not sure.

Thanks for the hints. Trying them.

qiskit_nature/runtime/vqe_program.py
```diff
@@ -209,13 +209,16 @@ def callback(self, callback: Callable) -> None:
...
             _, data = args  # first element is the job id
+            if isinstance(data, dict):
+                return  # not expected params. skip
             iteration_count = data[0]
             params = data[1]
             mean = data[2]
             sigma = data[3]
-            return self._callback(iteration_count, params, mean, sigma)
+            self._callback(iteration_count, params, mean, sigma)
+            return
```
mrossinek, line 221: I don't understand why this is necessary tbh. Could it be due to the lack of the complete type-hint on `self._callback`?
Arguably, that would need to be aligned with the type hint as written in the `__init__` method (like @woodsp-ibm also pointed out).

renier: Without doing this, you get the pylint error:
```
qiskit_nature/runtime/vqe_program.py:212:8: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
```

As to the type hint, are you saying that the `_wrap_vqe_callback` definition should look like this instead?:
```diff
--- a/qiskit_nature/runtime/vqe_program.py
+++ b/qiskit_nature/runtime/vqe_program.py
@@ -206,7 +206,7 @@ class VQEProgram(MinimumEigensolver):
         """Set the callback."""
         self._callback = callback

-    def _wrap_vqe_callback(self) -> Optional[Callable]:
+    def _wrap_vqe_callback(self) -> Optional[Callable[[int, np.ndarray, float, float], None]]:
         """Wraps and returns the given callback to match the signature of the runtime callback."""
```

mrossinek: > Without doing this, you get the pylint error:

Oh I see. Because you introduced a blank `return` statement further up top.

Yes, that typehint is correct. I believe that should also resolve the above pylint error because it specifies that the callback itself also returns `None`. While you're at it, could you also please fix the typehints on the getter and setter methods of the callback?

mrossinek: We had another internal discussion. You can keep the `self._callback` and `return` statements as they are now. But fixing the type hints in accordance with that on the `__init__` argument would be great :+1:

renier: ok, fixed the type hints

mrossinek: Thanks! :)

mrossinek (approved):
LGTM, thanks! Let's wait for @woodsp-ibm to also take another look before merging.

I don't think we need a reno for this, right?

qiskit_nature/runtime/vqe_program.py
woodsp-ibm: Ok, one last question then. It more seems then that it's the runtime itself is adding this extra info in the stream at the end if I understand the original text in the PR and your comment above. Hence whatever the user wants to put in the stream, via its server side program, it has to find a way for its client to detect this 'final result', whatever that is, which is added to the stream, to differentiate it from its normal data. Here you are simply saying if the data is a dict, since presumably the vqe script/code on the server never adds a dict, that suffices. Maybe there is further info one can see that if you wanted to send dicts in the stream for your program/client that these is a simple way to differentiate these - presumably one can at least do it by keys.

I guess the data we are currently ignoring will be expected at some point - i.e. where it just skips it in the code where it detects its a dict. Hence its preferable to just skip rather than log some warning? Or should it detect if this dict is an end result and skip and otherwise log a warning? I am just wondering if for example one updated the VQE runtime, so the server side changes and adds more data to the stream yet the user is still using an older client. I am not sure how versioning is managed in this case, if at all.

renier: > Ok, one last question then. It more seems then that it's the runtime itself is adding this extra info in the stream at the end if I understand the original text in the PR and your comment above. Hence whatever the user wants to put in the stream, via its server side program, it has to find a way for its client to detect this 'final result', whatever that is, which is added to the stream, to differentiate it from its normal data. Here you are simply saying if the data is a dict, since presumably the vqe script/code on the server never adds a dict, that suffices. Maybe there is further info one can see that if you wanted to send dicts in the stream for your program/client that these is a simple way to differentiate these - presumably one can at least do it by keys.

Yes. There is even a way to specify a schema for the results (intermediary and final) to establish that contract between the consuming client program and the runtime remote program. In the case of the current VQE runtime program however, it does not have a schema for the intermediary results just the final result. So the change here is basing itself on that final result schema (https://runtime-us-east.quantum-computing.ibm.com/programs).

> I guess the data we are currently ignoring will be expected at some point - i.e. where it just skips it in the code where it detects its a dict. Hence its preferable to just skip rather than log some warning?

Yes, the client-side VQE program fetches the final result in a different way. Since a final result is stored permanently, it uses that api to get it via the ibm qiskit provider.

About server side changes, yes, that's an aspect we are slowly improving on the API side. If you own and manage that remote program yourself, then it is not so problematic. You own both sides then. When we think about sharing this with others though, those users are more vulnerable to these problems. So in the future we might just validate a signature and also copy programs instead of being a direct reference when sharing.

woodsp-ibm (approved):

PR #351, opened 2021-09-03
Rename ``VQEProgram`` to ``VQERuntimeClient``

Cryoris:
### Summary

In the context of Qiskit Runtime, there's been a lot of confusion about what is considered a "runtime program" or "runtime script". To clear this confusion up, we rename the algorithm classes that remotely call the code on the runtime servers to _clients_ and the code on the runtime can be referred to as _program_.

We chose the word client since that's what they are: an API integrated into the algorithm flow that defers the algorithm call to the cloud.

### Details and comments

The diff might show a lot below, but it was just a renaming of the `VQEProgram` to `VQERuntimeClient` and no internal methods have been touched. 

We could also just call it `VQEClient` for brevity, if that's preferred 🙂

### DIFF
--- a/qiskit_nature/runtime/__init__.py
+++ b/qiskit_nature/runtime/__init__.py
@@ -25,9 +25,11 @@ algorithms and scripts in the cloud.
 
    VQEProgram
    VQEProgramResult
+   VQERuntimeClient
+   VQERuntimeResult
 
 """
 
-from .vqe_program import VQEProgram, VQEProgramResult
+from .vqe_program import VQEProgram, VQEProgramResult, VQERuntimeClient, VQERuntimeResult
 
-__all__ = ["VQEProgram", "VQEProgramResult"]
+__all__ = ["VQEProgram", "VQEProgramResult", "VQERuntimeClient", "VQERuntimeResult"]
--- a/qiskit_nature/runtime/vqe_program.py
+++ b/qiskit_nature/runtime/vqe_program.py
@@ -14,20 +14,24 @@
 
 
 from typing import List, Callable, Optional, Any, Dict, Union
+import warnings
 import numpy as np
 
 from qiskit import QuantumCircuit
-from qiskit.exceptions import QiskitError
 from qiskit.providers import Provider
 from qiskit.providers.backend import Backend
-from qiskit.algorithms import MinimumEigensolver, MinimumEigensolverResult, VQEResult
-from qiskit.algorithms.optimizers import Optimizer, SPSA
-from qiskit.opflow import OperatorBase, PauliSumOp
-from qiskit.quantum_info import SparsePauliOp
+from qiskit.algorithms import MinimumEigensolverResult
+from qiskit.algorithms.optimizers import Optimizer
+from qiskit.opflow import OperatorBase
+
+from .vqe_runtime_client import VQERuntimeClient, VQERuntimeResult
 
 
-class VQEProgram(MinimumEigensolver):
-    """The Qiskit Nature VQE Quantum Program to call the VQE runtime as a MinimumEigensolver."""
+class VQEProgram(VQERuntimeClient):
+    """DEPRECATED. This class has been renamed to ``qiskit_nature.runtime.VQERuntimeClient``.
+
+    This renaming reflects that this class is a client for a program executed in the cloud.
+    """
 
     def __init__(
         self,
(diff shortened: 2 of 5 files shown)


qiskit_nature/runtime/vqe_program.py
```diff
@@ -41,342 +45,68 @@ def __init__(
...
-
-        # if callback is set, return wrapped callback, else return None
-        if self._callback:
-            return wrapped_callback
-        else:
-            return None
+        warnings.warn(
+            "The qiskit_nature.runtime.VQEProgram has been renamed to "
+            "qiskit_nature.runtime.VQERuntimeClient to reflect it is a client for a program "
+            "executed in the cloud. The VQEProgram is deprecated as of Qiskit Nature 0.3.0 and "
```
mrossinek, line 51: I also see you're writing as of `0.3.0`. That won't come out for a while.. Since I think #345 can be considered a significant bug, we could think about doing a bug-fix release. Would you rather include this in `0.2.1` in that case?

qiskit_nature/runtime/vqe_program.py
```diff
@@ -41,342 +45,68 @@ def __init__(
...
-        allowed_settings.remove("trust_region")
-        allowed_settings.remove("second_order")
-
-    unsupported_args = set(settings.keys()) - set(allowed_settings)
-
-    if len(unsupported_args) > 0:
-        raise ValueError(
-            f"The following settings are unsupported for the {name} optimizer: "
-            f"{unsupported_args}"
+        warnings.warn(
```
mrossinek, line 105: The same applies to here.

qiskit_nature/runtime/vqe_program.py
```diff
@@ -41,342 +45,68 @@ def __init__(
...
+            "eigenvalue",
+            "aux_operator_eigenvalues",
+            "optimal_parameters",
+            "optimal_point",
+            "optimal_value",
+            "optimizer_evals",
+            "optimizer_time",
+            "optimizer_history",
+        ]:
+            setattr(vqe_result, attr, getattr(result, attr))
```
mrossinek, line 91: Is this manual approach really necessary? Given that `VQEProgramResult` derives from the new `VQERuntimeResult`, a simple `cast` should be sufficient, right?

qiskit_nature/runtime/vqe_runtime_client.py
```diff
@@ -0,0 +1,382 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""The Qiskit Nature VQE Quantum Program."""
```
mrossinek, line 13: I think you should update this slightly too. Maybe including some of your explanations from the PR description.

test/runtime/fake_vqeruntime.py
```diff
@@ -18,15 +18,15 @@
...
+from qiskit_nature.runtime import VQERuntimeResult
 
 
 class FakeVQEJob:
     """A fake job for unit tests."""
 
     def result(self) -> Dict[str, Any]:
         """Return a VQE result."""
-        result = VQEProgramResult()
+        result = VQERuntimeResult()
```
mrossinek, line 29: You should add a test which ensures that the deprecated code still works as intended as long as it is not removed, yet

qiskit_nature/runtime/vqe_program.py
Cryoris: I'm not sure -- if we use `typing.cast` the actual type is not changed right? So if someone does a typecheck it would fail. But maybe there's a different way of casting that I'm not aware of 😄

qiskit_nature/runtime/vqe_program.py
Cryoris: Hmm, yeah it would be nice to have it out sooner than later.

qiskit_nature/runtime/vqe_runtime_client.py
Cryoris: I updated the class docstring, since I felt that's more visible than the module docstring 👍🏻

qiskit_nature/runtime/__init__.py
```diff
@@ -25,9 +25,11 @@
 
    VQEProgram
    VQEProgramResult
+   VQERuntimeClient
+   VQERuntimeResult
 
 """
 
-from .vqe_program import VQEProgram, VQEProgramResult
+from .vqe_program import VQEProgram, VQEProgramResult, VQERuntimeClient, VQERuntimeResult
```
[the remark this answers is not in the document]
Cryoris: whoops my bad, I was programming to quickly 😉

qiskit_nature/runtime/vqe_program.py
woodsp-ibm: Correct a `cast` is more like a typehint in saying that this is the type it should be considered as - it does not change what it actually is.

woodsp-ibm:
In terms of the name - if/when these clients are moved to some library that's more associated with the Runtime would you expect to keep the name there as `VQERuntimeClient` as the Runtime part would seem then more explicit by virtue of the library it was part of. Or maybe we'll just ask users to change it again after all they would need to change some import statement in their code if this is moved.

This PR should include the runtime tutorial such that it uses the new name for this there.

qiskit_nature/runtime/vqe_program.py
mrossinek: Okay let's put it another way then: we never declare the return-type to be a `VQEProgramResult`. We only require it to be a `MinimumEigensolverResult`. Thus, I don't see the need for wrapping a `VQERuntimeResult` with a `VQEProgramResult`. We can simply return the former and deprecate the latter.

qiskit_nature/runtime/vqe_program.py
```diff
@@ -41,342 +45,68 @@ def __init__(
...
-        else:
-            return None
+        warnings.warn(
+            "The qiskit_nature.runtime.VQEProgram has been renamed to "
+            "qiskit_nature.runtime.VQERuntimeClient to reflect it is a client for a program "
+            "executed in the cloud. The VQEProgram is deprecated as of Qiskit Nature 0.3.0 and "
+            "be removed no sooner than 3 months after the release date.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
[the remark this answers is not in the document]
mrossinek: @Cryoris what about this comment? (and the identical one below)

qiskit_nature/runtime/vqe_program.py
Cryoris: Even if we didn't document it as precisely we still change the return type, so to be safe I'd prefer to still return the old type 🙂

qiskit_nature/runtime/vqe_program.py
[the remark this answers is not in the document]
Cryoris: As we discussed on Slack, we can delete the code anyways and I didn't yet change to using the `deprecation` because I thought it does the same as my message 😄 but I can also change to using the decorator 🙂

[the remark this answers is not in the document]
mrossinek: Using the decorator ensures consistency so please do that. Deleting the code is indeed fine :+1:

qiskit_nature/runtime/vqe_program.py
mrossinek: Alright. May I still suggest the use of `AlgorithmResult.combine()` to handle the updating for us?

Cryoris:
I think `VQEClient` is better, you're right that the "runtime" bit is already clear from the import path 👍🏻

qiskit_nature/runtime/vqe_program.py
```diff
@@ -41,342 +47,52 @@ def __init__(
...
-        if self._callback:
-            return wrapped_callback
-        else:
-            return None
+        warn_deprecated(
+            version="0.2.1",
+            old_type=DeprecatedType.CLASS,
+            old_name="VQEProgram",
+            new_name="VQEClient",
+            additional_msg="from qiskit_optimization.runtime",
```
mrossinek, line 55:
```suggestion
            additional_msg="from qiskit_nature.runtime",
```

qiskit_nature/runtime/vqe_program.py
```diff
@@ -41,342 +47,52 @@ def __init__(
...
+            vqe_result = VQEProgramResult()
 
+        vqe_result.combine(result)
         return vqe_result
 
 
-class VQEProgramResult(VQEResult):
-    """The VQEProgram result object.
+class VQEProgramResult(VQERuntimeResult):
+    """DEPRECATED. The ``VQEProgram`` result object has been renamed to ``VQERuntimeResult``.
```
mrossinek, line 84:
```suggestion
    """DEPRECATED. The ``VQEProgramResult`` has been renamed to ``VQERuntimeResult``.
```

releasenotes/notes/rename-runtime-program-9afac6a99a6de213.yaml
```diff
@@ -0,0 +1,9 @@
+---
+deprecations:
+  - |
+    Rename the runtime "program" to runtime "client" to avoid name confusions and
+    reflect the fact that it is an interface for code executed in the cloud. The classes
+    :class:`~qiskit_optimization.runtime.VQEProgram` and
+    :class:`~qiskit_optimization.runtime.VQEProgramResult` have been renamed to
+    :class:`~qiskit_optimization.runtime.VQEClient`,
+    :class:`~qiskit_optimization.runtime.VQERuntimeResult`, respectively.
```
mrossinek, line 9:
```suggestion
    :class:`~qiskit_nature.runtime.VQEProgram` and
    :class:`~qiskit_nature.runtime.VQEProgramResult` have been renamed to
    :class:`~qiskit_nature.runtime.VQEClient`,
    :class:`~qiskit_nature.runtime.VQERuntimeResult`, respectively.
```

mrossinek (changes requested):

docs/tutorials/07_leveraging_qiskit_runtime.ipynb
```diff
@@ -24,6 +24,15 @@
     "Let's see how you can leverage the runtime on a simple chemistry example: Finding the ground state energy of the lithium hydrate (LiH) molecule at a given bond distance."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### *Important: Version requirements*\n",
+    "\n",
+    "*This notebook requires Qiskit Terra 0.18.x and Qiskit Nature 0.3.x.*"
```
woodsp-ibm, line 33: Can we not say a minimum of 0.18.0 and 0.3.0. I imagine by the time Nature 0.3.0 is out that Terra will be on the next version

qiskit_nature/runtime/vqe_program.py
```diff
@@ -41,342 +47,52 @@ def __init__(
...
-            sigma = data[3]
-            return self._callback(iteration_count, params, mean, sigma)
-
-        # if callback is set, return wrapped callback, else return None
-        if self._callback:
-            return wrapped_callback
-        else:
-            return None
+        warn_deprecated(
+            version="0.2.1",
```
woodsp-ibm, line 51:
```suggestion
            version="0.3.0",
```
I think this is deprecated as of 0.3.0.

qiskit_nature/runtime/vqe_program.py
```diff
@@ -41,342 +47,52 @@ def __init__(
...
-        allowed_settings.remove("second_order")
-
-    unsupported_args = set(settings.keys()) - set(allowed_settings)
-
-    if len(unsupported_args) > 0:
-        raise ValueError(
-            f"The following settings are unsupported for the {name} optimizer: "
-            f"{unsupported_args}"
+        warn_deprecated(
+            version="0.2.1",
```
woodsp-ibm, line 93:
```suggestion
            version="0.3.0",
```

docs/tutorials/07_leveraging_qiskit_runtime.ipynb
Cryoris: done in 99a2433

mrossinek (approved):
LGTM! Let's await another review by @woodsp-ibm or @pbark before merging

mrossinek (approved):

PR #352, opened 2021-09-06
Add missing reno from #347

mrossinek:
### Summary

Adds the reno which was missing in #347


mrossinek:
Once this is merged I will create a PR against `stable/0.2` cherry-picking #347 and this one.

pbark (approved):
LGTM!

PR #357, opened 2021-09-09
Fix mapper

ikkoham:
### Summary

I'm not 100% sure, but I think the mapper is wrong. Please double check. How it's wrong is reflected in the [test cases](https://github.com/Qiskit/qiskit-nature/pull/357/files#diff-d92db51c4ee6797461b0113b4412281634076b2ce14ac06ab6f1819bc0df2a6aR72-R95). I don't have all the tests to go through now. If this fix is correct, I'll pass them.

### Details and comments

The point here is that the & operator is left-acting, which is the opposite of the normal order of operations.

```python

mapping = JordanWignerMapper().map
plus = FermionicOp("+")
minus = FermionicOp("-")
print((mapping(plus) @ mapping(minus)).reduce())
print((mapping(plus @ minus)).reduce())
```
The outputs should be the same. but now different

### DIFF
--- a/releasenotes/notes/fix-mapper-bug-8eba9d219595b8e1.yaml
+++ b/releasenotes/notes/fix-mapper-bug-8eba9d219595b8e1.yaml
@@ -1,5 +1,5 @@
 ---
-features:
+fixes:
   - |
     Fixed an issue where the existing Mapper was incorrectly mapping the creation operator :math:`+`
     and the annihilation operator :math:`-`. It used to be that :math:`+` was mapped to
--- a/releasenotes/notes/fix-mapper-bug-8eba9d219595b8e1.yaml
+++ b/releasenotes/notes/fix-mapper-bug-8eba9d219595b8e1.yaml
@@ -1,0 +1,7 @@
+---
+features:
+  - |
+    Fixed an issue where the existing Mapper was incorrectly mapping the creation operator :math:`+`
+    and the annihilation operator :math:`-`. It used to be that :math:`+` was mapped to
+    :math:`\sigma_+` and :math:`-` was mapped to :math:`\sigma_-`, but it is correct that
+    :math:`+` is mapped to :math:`\sigma_-` and :math:`-` is mapped to :math`\sigma_+`.
--- a/qiskit_nature/mappers/second_quantization/qubit_mapper.py
+++ b/qiskit_nature/mappers/second_quantization/qubit_mapper.py
@@ -120,17 +120,23 @@ class QubitMapper(ABC):
             # save the respective Pauli string in the pauli_str list.
             for position, char in enumerate(label):
                 if char == "+":
-                    ret_op &= times_creation_op(position, pauli_table)
+                    ret_op = ret_op.compose(times_creation_op(position, pauli_table), front=True)
                 elif char == "-":
-                    ret_op &= times_annihilation_op(position, pauli_table)
+                    ret_op = ret_op.compose(
+                        times_annihilation_op(position, pauli_table), front=True
+                    )
                 elif char == "N":
                     # The occupation number operator N is given by `+-`.
-                    ret_op &= times_annihilation_op(position, pauli_table)
-                    ret_op &= times_creation_op(position, pauli_table)
+                    ret_op = ret_op.compose(times_creation_op(position, pauli_table), front=True)
+                    ret_op = ret_op.compose(
+                        times_annihilation_op(position, pauli_table), front=True
+                    )
                 elif char == "E":
                     # The `emptiness number` operator E is given by `-+` = (I - N).
-                    ret_op &= times_creation_op(position, pauli_table)
-                    ret_op &= times_annihilation_op(position, pauli_table)
+                    ret_op = ret_op.compose(
+                        times_annihilation_op(position, pauli_table), front=True
+                    )
+                    ret_op = ret_op.compose(times_creation_op(position, pauli_table), front=True)
                 elif char == "I":
                     continue
 
--- a/qiskit_nature/mappers/second_quantization/qubit_mapper.py
+++ b/qiskit_nature/mappers/second_quantization/qubit_mapper.py
@@ -87,16 +87,16 @@ class QubitMapper(ABC):
             # 0. Some utilities
 
         def times_creation_op(position, pauli_table):
-            # The creation operator is given by 0.5*(X + 1j*Y)
+            # The creation operator is given by 0.5*(X - 1j*Y)
             real_part = SparsePauliOp(pauli_table[position][0], coeffs=[0.5])
-            imag_part = SparsePauliOp(pauli_table[position][1], coeffs=[0.5j])
+            imag_part = SparsePauliOp(pauli_table[position][1], coeffs=[-0.5j])
 
             return real_part + imag_part
 
         def times_annihilation_op(position, pauli_table):
-            # The annihilation operator is given by 0.5*(X - 1j*Y)
+            # The annihilation operator is given by 0.5*(X + 1j*Y)
             real_part = SparsePauliOp(pauli_table[position][0], coeffs=[0.5])
-            imag_part = SparsePauliOp(pauli_table[position][1], coeffs=[-0.5j])
+            imag_part = SparsePauliOp(pauli_table[position][1], coeffs=[0.5j])
 
             return real_part + imag_part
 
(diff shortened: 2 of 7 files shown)


mrossinek:
This PR raised a very important discussion which I am trying to summarize below.

The setting:
When the mappings were migrated from Aqua to Nature, we validated the code against the state of Aqua.
Right now, the following example from Aqua produces the accompanied output:
```python
import numpy as np
from qiskit.chemistry import FermionicOperator

h1 = np.asarray([[0, 1], [0, 0]])
h2 = np.zeros((2, 2, 2, 2))

aqua_op = FermionicOperator(h1, h2)
aqua_op_q = aqua_op.mapping("jordan_wigner")
print("Aqua Qubit Operator")
for (coeff, pauli) in aqua_op_q.paulis:
    print(coeff, pauli)

# Aqua Qubit Operator
# -0.25j XY
# (0.25+0j) YY
# (0.25+0j) XX
# 0.25j YX
```
The corresponding Nature snippet *prior* to this PR produces:
```python
from qiskit_nature.properties.second_quantization.electronic import ElectronicEnergy
from qiskit_nature.properties.second_quantization.electronic.bases import ElectronicBasis

mapper = JordanWignerMapper()

prop = ElectronicEnergy.from_raw_integrals(ElectronicBasis.SO, h1, h2)
op = prop.second_q_ops()[0]
qop = mapper.map(op)
print("Nature Qubit Operator")
print(qop)

# Nature Qubit Operator
# 0.25 * XX
# + -0.25j * YX
# + 0.25j * XY
# + 0.25 * YY
```
Already, we can see a problem because the signs of the `XY` and `YX` terms are flipped.

This PR was then triggered by an observation made on the public Qiskit Slack workspace regarding the following unexpected behavior in Nature:
```python
print("map(+ @ -)")
print(mapper.map(plus @ minus).to_matrix().real)
print("map(+) @ map(-)")
print((mapper.map(plus) @ mapper.map(minus)).to_matrix().real)

# map(+ @ -)
# [[0. 0.]
#  [0. 1.]]
# map(+) @ map(-)
# [[1. 0.]
#  [0. 0.]]
```
Clearly, something is odd here.

The internal discussion came to the observation that we are currently not mapping the fermionic `+` and `-` operators correctly.
Since we want to represent a fermionic `|1>` mode with a qubit `|1>` mode, we must map a fermionic `+` to a qubit `\sigma_-` operator.
```
|0> = [1, 0]^T
|1> = [0, 1]^T
vaccum = [0, 0]^T

Fermionic space:
+|0> = |1>
+|1> = vaccum
-|0> = vaccum
-|1> = |0>

Qubit space:

\sigma_+ = [[0, 1], [0, 0]]
\sigma_- = [[0, 0], [1, 0]]

\sigma_+ @ |0> = vacuum
\sigma_+ @ |1> = |0>
\sigma_- @ |0> = |1>
\sigma_- @ |1> = vaccum
```
Applying this change results in:
```
Nature Qubit Operator
-0.25 * XX
+ -0.25j * YX
+ 0.25j * XY
- 0.25 * YY

map(+ @ -)
[[1. 0.]
 [0. 0.]]
map(+) @ map(-)
[[0. 0.]
 [0. 1.]]
```
Clearly, something is still going wrong.
This leads to the second important change of this PR which is changing the order in which the single operations contained within `N` and `E` are applied.
This leads to:
```
Nature Qubit Operator
-0.25 * XX
+ -0.25j * YX
+ 0.25j * XY
- 0.25 * YY

map(+ @ -)
[[0. 0.]
 [0. 1.]]
map(+) @ map(-)
[[0. 0.]
 [0. 1.]]
```
Now, the second example works as expected.

However, we are still seeing a sign flip of our Pauli coefficients compared to the original Aqua source.
This time, all signs are flipped (and not just those of the `{XY, YX}` pair).
This sign flip explains the remaining unittest failures.
The question remains: is the new behavior *after* this PR indeed the correct one?

@ikkoham @pbark @woodsp-ibm

woodsp-ibm:
Yes, when the mappers were done I checked conversions that they matched to Aqua and I do not recall any sign differences - I am sure I would not have considered the mapping identical and proceeded with things were that the case.

In terms of referencing Aqua I can say this. The Aqua code was done on logic, which was originally in Terra, that I think came about from the original Nature paper. E.g this which is the fermionic operator/mapping logic https://github.com/Qiskit/qiskit-terra/blob/76d1e66347b383aab5a9a6fc6b26ef29d7f9b4fb/qiskit/tools/apps/fermion.py So it would have been done based on this and was refactored etc to become part of Aqua.

mrossinek:
### TL;DR
This PR now ensures that `assert map(+ @ -) == map(+) @ map(-)` for any index combination of `+` and `-`.
In the following I also explain at length how I validated this as well as validated the results against the original Aqua code.

---


Alright, @pbark and I have investigated this in more detail yesterday and I think we have identified the problem to be a wrong `compose` operation order in the `QubitMapper.mode_based_mapping` function. I have just pushed two commits to this branch: one which fixes the order and the other which fixes the only two unittests which fail because they hard-coded a wrong expected phase.

Thanks @woodsp-ibm for the pointer to the very original implementation of the mappings. We have validated that it produces the same output as the `main` branch of Aqua does right now. Furthermore, I have done two validations:

#### `assert map(+ @ -) == map(+) @ map(-)`
This is the issue which originally brought light to this bug and lead to this PR. I have written the following short snippet which validates the above assertion for a number of possible index combinations for the `+` and `-` operators:
```python

mapping = JordanWignerMapper().map

n = 10

for i in range(n):
    plus = FermionicOp(f"+_{i}", register_length=n)
    for j in range(n):
        minus = FermionicOp(f"-_{j}", register_length=n)
        print(plus @ minus, "\t", mapping(plus @ minus) == (mapping(plus) @ mapping(minus)).reduce())
```
On the `main` branch this fails for **all** possible combinations. Before my previous two commits this would only pass if `i == j`. After fixing the compose operation order this now passes for all combinations of `i` and `j` (which is what we expect).

#### Validation against Aqua
I have also validated the mapper results of this PR against the original qubit operators produced by Aqua. I did not test every single unittest but instead focused on the following: LiH computed with the `PySCFDriver` using `RHF` in the Jordan-Wigner and parity mapping (with and without two-qubit-reduction).
I have attached the produced qubit operators from Aqua, Nature's `main` branch and this PR for all of these cases in the zip file below. Interestingly enough, all of these operators are identical in terms of Pauli strings, phases and coefficients up to numerical precision (Note, the two-qubit reduction in Nature can introduce negligible complex phases which cause the output to "appear" different). I found that the results can be very nicely compared using e.g. `vimdiff lih_rhf_jw_*.txt`.
The fact that all outputs are identical is good news for us (because this means that Nature was working supposedly correctly for electronic structure problems) and it also explains why this issue went unnoticed for so long. As to why the outputs are unchanged I believe that the symmetry we often observe in electronic structure problems cancels out the inconsistencies which we have uncovered as part of this PR.

[mapper_fix_validation.zip](https://github.com/Qiskit/qiskit-nature/files/7215562/mapper_fix_validation.zip)

ikkoham:
@mrossinek Thank you very much. I spent a lot of time trying to figure out what was causing the test to fail.

I would like to add the same unit test to BK and Parity as I added to the JW conversion. If that's correct, I think this PR is mergeable.

mrossinek:
Thanks for the additional unittests, @ikkoham!
Do you think we should also add some tests for the assertion `map(+ @ -) == map(+) @ map(-)`? After all, this was how the bug surfaced in the first place...

ikkoham:
Thanks.I think it's okay to have it, but we already have the results of `map(N)` and `map(+), map(-)` guaranteed, so we already have a finer test as a unit test.

mrossinek:
Alright! Then I suggest that @woodsp-ibm and @pbark review this PR because @ikkoham and I were quite involved with the changes.

woodsp-ibm:
Only comment I have is that it ought to have a reno release note I think.

releasenotes/notes/fix-mapper-bug-8eba9d219595b8e1.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
```
woodsp-ibm, line 2:
```suggestion
fixes:
```
I think this should be 'fixes' right as it address a bug and is not a new feature

ikkoham: Thanks. My bad.

pbark:
I had a further look to this and now LGTM. Thanks a lot @ikkoham and @mrossinek .

PR #380, opened 2021-09-29
Solver Callbacks

### LINKED ISSUE #375: Add support for `callback` for minimum_eigensolver_factories

`callback` is used for getting interim results while running iterative algorithms like VQE. This is especially useful when running on hardware (with Qiskit Runtime) because the program can take hours to run.

`callback` is supported by `VQE` in qiskit terra but not the minimum eigensolver factories such as `VQEUCCFactory` and `VQEUVCFactory` in qiskit terra.

mrossinek:
### Summary

Fixes #375 

### Details and comments

- exposes `callback` on the Factory classes
- adds `kwargs` to allow arbitrary other keyword arguments to be passed in to the constructed VQE
- fixes the property getter and setter methods to actually have an effect

### DIFF
--- a/qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
+++ b/qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
@@ -81,6 +81,9 @@ class VQEUCCFactory(MinimumEigensolverFactory):
                 ansatz, the evaluated mean and the evaluated standard deviation.`
             kwargs: any additional keyword arguments will be passed on to the VQE.
         """
+        self._vqe = VQE(
+            **kwargs,
+        )
         self.quantum_instance = quantum_instance
         self.optimizer = optimizer
         self.initial_point = initial_point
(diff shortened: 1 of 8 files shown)


qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -81,75 +92,75 @@ def __init__(
...
-        self._expectation = expectation
+        if expectation is None:
+            return
+        self._vqe.expectation = expectation
 
     @property
     def include_custom(self) -> bool:
         """Getter of the ``include_custom`` setting for the ``expectation`` setting."""
-        return self._include_custom
+        return self._vqe._include_custom
```
woodsp-ibm, line 158: This is now accessing a private member `_include_custom` of VQE. We should figure something out here - maybe either delay instantiation of VQE until its needed in solve - or perhaps add public setter/getters to VQE. We should avoid using members of VQE that are marked as private.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -173,6 +184,18 @@ def initial_state(self, initial_state: Optional[QuantumCircuit]) -> None:
         the :class:`~.HartreeFock`."""
         self._initial_state = initial_state
 
+    @property
+    def callback(self) -> Optional[Callable[[int, np.ndarray, float, float], None]]:
+        """Returns the callback."""
+        return self._vqe._callback
```
woodsp-ibm, line 190: Similar comment around avoiding using private members.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -173,6 +184,18 @@ def initial_state(self, initial_state: Optional[QuantumCircuit]) -> None:
...
+
+    @callback.setter
+    def callback(self, callback: Optional[Callable[[int, np.ndarray, float, float], None]]) -> None:
+        """Sets the callback."""
+        if callback is None:
+            return
+        self._vqe._callback = callback
```
woodsp-ibm, line 197: Similar comment again for None - if you don't want the prior callback to be done then passing in None should "reset" it so no callback is called. Prviate member access too as note in getter above.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -72,84 +74,85 @@ def __init__(
...
     def gradient(self) -> Optional[Union[GradientBase, Callable]]:
         """Getter of the gradient function"""
-        return self._gradient
+        return self._vqe.gradient
 
     @gradient.setter
     def gradient(self, gradient: Optional[Union[GradientBase, Callable]]) -> None:
         """Setter of the gradient function"""
-        self._gradient = gradient
+        self._vqe.gradient = gradient
```
woodsp-ibm, line 135: Similar comments in this class to the other factory so I am not going to repeat them. I commented here since this turns out to be different (like the expectation here too) and has no None test.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -81,75 +92,75 @@ def __init__(
...
         """Getter of the gradient function"""
-        return self._gradient
+        return self._vqe.gradient
 
     @gradient.setter
     def gradient(self, gradient: Optional[Union[GradientBase, Callable]]) -> None:
         """Setter of the gradient function"""
-        self._gradient = gradient
+        if gradient is None:
+            return
```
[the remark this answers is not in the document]
mrossinek: Actually the setters of VQE for gradient and optimizer do not revert to default behavior when passing in `None`. Instead, the set the internal object to `None` which leads to errors further down the line. I agree that this is not optimal but this needs to be fixed on Terra's side (see next comment below)

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
mrossinek: Yes, I am aware of this problem. I have been meaning to open an issue on Terra to address the addition of missing property setters and to resolve the problem I mentioned above. I have simply not gotten around to opening the issue.
Because this fix is needed for the upcoming Quantum challenge I would like to merge this PR as is and address this once these shortcomings are resolved in Terra.

I can add a TODO comment in all relevant places if you like.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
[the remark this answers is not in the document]
woodsp-ibm: Either way passing in None here is in essence ignored which maybe is not expected either and arguably just as wrong. Why not pass it on, that way when VQE is corrected this will work as expected.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
woodsp-ibm: By creating VQE in solve (where these setters here just stored the vars like was there before) and using the instance locally there this would workaround things without resorting to accessing private methods and would be independent of any changes being done in Terra (unless I'm missing something). From an external perspective there would be no difference if in the future it did things as you show here directly on some internal cached instance though there maybe some performance benefit around reusing a cached instance on the same/similar problem.  So I can see there is room for improvement over what was there - indeed none of the current setters really working is less than ideal! Of course this code is not so tough to create your own factory. I imagine VQE will change as the expectation work gets further along - maybe at that point this can be revisited. Like the user passes in an ansatz I am wondering why we did not let them pass in a VQE instance too - that way they could set whatever they wanted on it and there would be no need to  expose a set of setter/getters that match VQE. Anyway if you want to go this way we need to remember to change these - maybe in additional to some todos you can raise a new issue here to fix it up and link it to the above Terra one as any changes will be dependent on that.

pbark (dismissed):

test/algorithms/ground_state_solvers/minimum_eigensolver_factories/test_vqe_uvcc_factory.py
```diff
@@ -48,13 +48,7 @@ def setUp(self):
     def test_setters_getters(self):
         """Test Getter/Setter"""
 
-        with self.subTest("Quantum Instance"):
-            self.assertEqual(self._vqe_uvcc_factory.quantum_instance, self.quantum_instance)
-            self._vqe_uvcc_factory.quantum_instance = None
-            self.assertEqual(self._vqe_uvcc_factory.quantum_instance, None)
-
```
woodsp-ibm, line 55: Was there a reason to remove this getter/setter test completely - the getter and setter still exist right. I can see that setting it to None is arguably an invalid test since the typehint does not include Optional. But a different QuanumInstance could have been created here and then set and checked via the getter similar to what is done for optimizer.

woodsp-ibm (approved):
Just to note the QuantumInstance getter/setter unit test works as the code stands, but given current VQE code if/when you change this over in the future to setting this directly into VQE it will fail since - QI setter in the VQE code is not an Optional parameter - but unlike the code here which simply stores the value and doesn't look at it at all VQE dereferences what it passed and so will fail given None.

HuangJunye:
@mrossinek Thanks a lot for fixing this issue!

PR #393, opened 2021-10-13
Fix typo in the `ElectronicEnergy._orbital_energies` variable name

mrossinek:
### Summary

Fix typo in the `ElectronicEnergy._orbital_energies` variable name

### Details and comments

Why did this not get caught by any of our linters? :thinking:


woodsp-ibm (approved):
Looks fine - in practice the code works fine even with that typo as it in effect added an extra variable that was never used. It was of course supposed to define the instance var used to store the orbital energy that the setter/getter used - which should ideally be defined upfront which this change makes sure it is.

Since it really changes nothing in respect of an external users use of the code via the public API or documentation there is no value in my mind in backporting this.

PR #411, opened 2021-11-01
Remove deprecated EvolvedOperatorAnsatz

woodsp-ibm:
EvolvedOperatorAnsatz was migrated to core Qiskit in release 0.2 and the instance here deprecated with a warning being raised if used to change to the migrated versions as a direct replacement.

A recent change Qiskit/qiskit-terra#6975 caused the unit test of the now deprecated version of the code remaining here to no longer work. As such, and since it would have been due to be removed shortly ahead of the next release, I have simply removed it now rather than attempt to fix things for this deprecated module.


manoelmarques (dismissed):

releasenotes/notes/remove_deprecated_evopansatz-36c3c872419a4329.yaml
```diff
@@ -0,0 +1,8 @@
+---
+
+upgrade:
+  - |
+    The `EvolvedOperatorAnsatz`, from the circuit.library in Nature, which was migrated in
+    as earlier release to core Qiskit i.e. Terra, has now been removed. You should change
```
mrossinek, line 6:
```suggestion
    an earlier release to core Qiskit i.e. Terra, has now been removed. You should change
```

mrossinek (changes requested):
Just a minor typo.

mrossinek (dismissed):

mrossinek (approved):

PR #413, opened 2021-11-03
Add getter and setter for the Molecule's degree_of_freedom attribute

### LINKED ISSUE #399: `Molecule` has no setter for `degrees_of_freedom`

The `Molecule` class provides property getter and setter methods for all of its attributes **except** for `degrees_of_freedom`. I don't see a good reason why this setter should not exist and suggest that we add it.

tnemoz:
### Summary
Fixes #399 by adding a getter and a setter for the `degrees_of_freedom` attribute of the `Molecule` class.


qiskit_nature/drivers/molecule.py
```diff
@@ -412,3 +412,13 @@ def perturbations(self) -> Optional[List[float]]:
     def perturbations(self, value: Optional[List[float]]) -> None:
         """Set perturbations"""
         self._perturbations = value
+
+    @property
+    def degrees_of_freedom(self) -> Optional[List[Callable]]:
+        """Get degrees_of_freedom"""
```
mrossinek, line 418: Maybe you can improve these docstrings slightly. E.g.
```
Get the degrees of freedom.
```
Same for the setter

Other than that this LGTM. Thanks! :+1:

mrossinek (approved):
Thanks for the quick contribution!

Let's await one more review from @woodsp-ibm or @pbark before merging :+1:

woodsp-ibm:
The new setter ends up not being covered by unit tests. The getter gets tested since the code was changed to fetch the value via the getter rather than the private instance variable. Maybe the simplest change to ensure the setter is used to set the same variable that the getter is reading, is to change the constructor so that instead of 
```
self._degrees_of_freedom = degrees_of_freedom
```
we first create the instance variable there that the getter/setter will use but then initialize it via the setter i.e.
```
self._degrees_of_freedom = None
self.degrees_of_freedom = degrees_of_freedom
```
Sometimes we have an explicit test for the getter/setter so that they are covered. Here they will be covered implicitly via the constructor and whatever test ends up going through the code that uses the getter, which I guess is enough unless others would prefer something explicit. My only issue with the explicit test is a simple set followed by get that all it does is check that the object can store and retrieve the value - the intent of setting it is of course that future use of the object will use that new value - that aspect is often not catered to by the test.

woodsp-ibm (approved):
Thanks for making the change!

PR #414, opened 2021-11-04
Unpin PySCF to allow latest 2.0.0

woodsp-ibm:
PySCF was released recently and wheels for Linux and Mac were added incrementally and during this time PySCF was pinned as otherwise attempts were made to compile it from the source tar.gz package.

I also unpinned h5py since the newer PySCF was fixed in regards of the deprecation issue (as far as I recall seeing before).

I am not sure if this will pass or not since before I had pinned it I had seen notebook kernel crash and other crashes. But this PR will let us view across CI what the outcome is and see how we might proceed in regards of this new version. One possibility might be to allow but pin CI to 1.7.6 if we have some failures but otherwise allow it for users as maybe it works in general. Lets see what happens with CI etc...

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,5 +4,5 @@ numpy>=1.17
 psutil>=5
 scikit-learn>=0.20.0
 setuptools>=40.1.0
-h5py<3.3
+h5py
 dataclasses; python_version < '3.7'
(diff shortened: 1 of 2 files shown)


woodsp-ibm:
2.0.1 has been released - no mac wheels yet so I expect that to fail in CI for now.

manoelmarques (approved):

mrossinek (approved):

pbark (approved):

PR #420, opened 2021-11-09
Support generalized fermionic excitations

mrossinek:
### Summary

This adds a new keyword argument to the fermionic excitation generator
method and the UCC class (and subclasses) which permits the use of
generalized excitations. Such excitation sets effectively ignore the
number of particles and instead permit all possible excitations between
all spin orbitals of a kind.

### Details and comments

A few weeks ago I was presented with a suggestion to support generalized fermionic excitations as proposed in this PR.
I never got around to opening an issue for this, hence this PR comes a little bit ouf of the blue..

### DIFF
--- a/qiskit_nature/circuit/library/ansatzes/ucc.py
+++ b/qiskit_nature/circuit/library/ansatzes/ucc.py
@@ -123,6 +123,7 @@ class UCC(EvolvedOperatorAnsatz):
         alpha_spin: bool = True,
         beta_spin: bool = True,
         max_spin_excitation: Optional[int] = None,
+        generalized: bool = False,
         reps: int = 1,
         initial_state: Optional[QuantumCircuit] = None,
     ):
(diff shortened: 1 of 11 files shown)


mrossinek:
Upon follow-up request I also decided to add a `spin_flip` option which can be used to permit spin-flipped excitations.

qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -123,6 +123,8 @@ def __init__(
         alpha_spin: bool = True,
         beta_spin: bool = True,
         max_spin_excitation: Optional[int] = None,
+        generalized: bool = False,
+        spin_flip: bool = False,
```
woodsp-ibm, line 127: I am just wondering about this as a name `spin_flip`. Generalized seems more like setting the strategy allowed when building things out. `spin_flip` sounds more like something it will do, as in change all the spins. How does `preserve_spin` seem (defaults True) - in that excitations will preserve whatever spins the particles have and will not change them i.e. it will not excite from alpha to beta or vice versa.

woodsp-ibm (approved):

PR #421, opened 2021-11-10
Remove Terra 0.19 deprecated messages

manoelmarques:


mrossinek (approved):
Thanks @manoelmarques!

woodsp-ibm (commented):
I believe this will make Nature dependent on Terra >= 0.19 so the requirements.txt needs updating accordingly

PR #422, opened 2021-11-10
Move warning about skipping 2-qubit reduction

### LINKED ISSUE #404: `QubitConverter` warns about skipping 2-qubit reduction even if not requested
### Information

- **Python version**: 3.9.7
- **Operating system**: Linux

The `QubitConverter` warns about skipping the 2-qubit reduction, even in cases where it has neither been requested nor is available:
```
The original qubit operator only contains 2 qubits! Skipping the requested two-qubit reduction!
```

```python
from qiskit_nature.transformers.second_quantization.electronic import ActiveSpaceTransformer

driver = PySCFDriver()
transformer = ActiveSpaceTransformer(2, 1)
problem = ElectronicStructureProblem(driver, [transformer])

hamiltonian = problem.second_q_ops()[0]

qubit_op = converter.convert(hamiltonian)
```

This is a simple (albeit unusual) reproduction test. I actually noticed this when using the `DirectMapper` with a `VibrationalStructureProblem` where 2 qubits can occur much quicker than in the `ElectronicStructureProblem` shown above (1 MO really does not do anything)

The warning should only be raised when the 2-qubit reduction has actually been requested.

### Suggested solutions

Move https://github.com/Qiskit/qiskit-nature/blob/745aa52/qiskit_nature/converters/second_quantization/qubit_converter.py#L319-L325 into the inner `if` statement just below.

jschuhmac:
### Summary
The `QubitConverter` now warns about skipping the 2-qubit reduction only if the reduction is requested, if it is supported by the mapper and if the number of qubits is less than or equal to 2. 
Previously, the warning was shown in all cases where the number of qubits is less than or equal to 2.
I also updated the corresponding test.

### Details and comments
Fixes #404

### DIFF
--- a/qiskit_nature/converters/second_quantization/qubit_converter.py
+++ b/qiskit_nature/converters/second_quantization/qubit_converter.py
@@ -316,16 +316,16 @@ class QubitConverter:
     ) -> PauliSumOp:
         reduced_op = qubit_op
 
-        if qubit_op.num_qubits <= 2:
-            logger.warning(
-                "The original qubit operator only contains %s qubits! Skipping the requested "
-                "two-qubit reduction!",
-                qubit_op.num_qubits,
-            )
-            return reduced_op
-
         if num_particles is not None:
             if self._two_qubit_reduction and self._mapper.allows_two_qubit_reduction:
+                if qubit_op.num_qubits <= 2:
+                    logger.warning(
+                        "The original qubit operator only contains %s qubits! Skipping the requested "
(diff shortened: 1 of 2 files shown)


woodsp-ibm (approved):
I don't think this needs a reno since its pretty minor and most people will probably have never seen this since message.

mrossinek (approved):

manoelmarques:
@Mergifyio refresh

PR #427, opened 2021-11-11
Gaussian drivers for python 3.9

manoelmarques:


mrossinek:
Since Gaussian is not available in the CI, could somebody, who has access to a License, verify that this works as expected? At least on one platform?

manoelmarques:
For reference, I tested on the following 64 bits Virtual Machines under Anaconda environment python 3.9. Each driver has the same dependencies as the previous py3.8. It would be good to have an independent test to be sure: 

1. Windows 10 Pro (latest updates)
2. Ubuntu 21.10
3. MacOS Big Sur 11.6

mrossinek (approved):
I wasn't aware you had the possibility to test these. This is all good then :+1:

manoelmarques:
I have Gaussian installed in all 3 VMs. Lets see if @woodsp-ibm has anything else to say.

woodsp-ibm (approved):

PR #365, opened 2021-09-16
Add lattice and Fermi-Hubbard model

k-tamuraphys:
### Summary
Implement lattice and Fermi-Hubbard model

### Details and comments
Gist URL of the jupyter notebook for the lattice module: 
https://gist.github.com/k-tamuraphys/6da54c468fffca3b4d18d9837de81aaf

### To Do

- [x] improve adjacency_matrix 
- [x] add the defaults position of the draw method
- [x] add unit tests
- [x] add release note

image
Added by @ikkoham

PR #430, opened 2021-11-16
Add Lattice tutorial

k-tamuraphys:
### Summary
Add a tutorial for Lattice module.
This demonstrates how to use the Lattice module in #365.


docs/tutorials/10_Lattice.ipynb
```diff
@@ -0,0 +1,943 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Lattice"
```
woodsp-ibm, line 7: Is there some more descriptive title (few more words rather than just one) that can improve on this. At a minimum I think it should be 'Lattices' plural with a s. Lattice singular does not feel right.

Now I see this tutorial shows to the point of creating second q ops. I will make a separate comment based on this and looking at the code, but I think we should show its actual usage with an algorithm to compute something so people have an example of how it would be used to do that.

woodsp-ibm:
In regards of Lattice & Models being under 'problems'. I see the Model there can generate a second_q_ops but it does not extend BaseProblem, that defines this, whereby it could be given as  problem instance to say a ground state solver. Now BaseProblem is most likely not something that is applicable here since that seems to bring in drivers hence maybe some refactoring around that could be done. Maybe having this in Problem form and passable to the algorithms is something for later.  This is the additional comment I said I would make in seeing that the Lattice tutorial stops at the point of generating a second q ops and does not use it with any quantum algorithm to show some result.

mrossinek:
Are there major differences between this tutorial and the one originally included in PR #365?

If not, then this is probably good to go once we merge that PR. I will review this after we have merged the above which will reduce the cluttered `Files` tab of this PR as it stands currently :+1:

k-tamuraphys:
> Are there major differences between this tutorial and the one originally included in PR #365?
> 
> If not, then this is probably good to go once we merge that PR. I will review this after we have merged the above which will reduce the cluttered `Files` tab of this PR as it stands currently 👍

No, this is basically the same as what was previously included in PR #365., except for a few minor modifications.

woodsp-ibm:
So once #365 is merged is the intent to remove all other files other than the tutorial. I guess there have been changes to these in #365 such that the copies here may now be different, such as renaming lattice folder to lattices etc.

ikkoham:
Please merge this PR after #365.

mrossinek (approved):

ikkoham:
Thank you. My Jupyter's black setting was set to 80 lines :cry:.

docs/tutorials/10_Lattice_Framework.ipynb
```diff
@@ -0,0 +1,959 @@
...
+   "source": [
+    "# Lattice Framework"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Introduction\n",
+    "In quantum physics including condensed matter physics, high energy physics, we often study models on lattices. \n",
```
pbark, line 15:
```suggestion
    "In quantum physics (including condensed matter physics and high energy physics) we often study models on lattices. \n",
```

docs/tutorials/10_Lattice_Framework.ipynb
```diff
@@ -0,0 +1,959 @@
...
+    "The operator $n_{i, \\sigma}$ is the number operator, which is defined by $n_{i, \\sigma} = c_{i, \\sigma}^\\dagger c_{i, \\sigma}$. \n",
+    "The matrix $t_{i, j}$ is a Hermitian matrix called hopping matrix.\n",
+    "The parameter $U$ represents the strength of the interaction."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can make the Hamiltonian for a given lattice using `FermiHubbardModel` class.\n",
```
pbark, line 754:
```suggestion
    "We can generate the corresponding Hamiltonian of a given lattice using `FermiHubbardModel` class.\n",
```

docs/tutorials/10_Lattice_Framework.ipynb
```diff
@@ -0,0 +1,959 @@
...
+    "Here, we make the Hamiltonian with uniform hopping and interaction parameters on a two-dimensional lattice."
```
pbark, line 755:
```suggestion
    "Here, we construct the Hamiltonian with uniform hopping and interaction parameters on a two-dimensional lattice."
```

docs/tutorials/10_Lattice_Framework.ipynb
```diff
@@ -0,0 +1,959 @@
...
+    "print(ham)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "`Lattice` has weights on its edges, so we can define a general hopping matrix using a Lattice instance.\n",
+    "Here, we consider the Fermi-Hubbard model on a general lattice on which non-uniform hopping parameters are given.\n",
+    "In this case, the weights of the lattice are regarded as the hopping matrix."
```
pbark, line 829:
```suggestion
    "In this case, the weights of the lattice are regarded as the hopping matrix. After generating the Hamiltonian (`second_q_ops`) we can use a qubit converter to generate the qubit operators and/or use any of the available algorithms to solver the corresponding lattice problem."
```

pbark (changes requested):

pbark:
@ikkoham @k-tamuraphys it LGTM. I did some minor stylistic changes in the text around the tutorial. If you agree with these we can merge it.

ikkoham:
Updated

pbark (approved):

mrossinek (approved):

PR #433, opened 2021-11-17
Checks for generated excitations in UCC ansatz

### LINKED ISSUE #323: Check generated excitations in UCC Ansatz before usage

In [this comment](https://github.com/Qiskit/qiskit-nature/issues/307#issuecomment-894053897) I proposed to add some checks on the generated excitations as used in the `UCC` Ansatz, [here](https://github.com/Qiskit/qiskit-nature/blob/main/qiskit_nature/circuit/library/ansatzes/ucc.py#L293).
This idea was motivated after seeing [this](https://quantumcomputing.stackexchange.com/questions/20749/define-customized-excitation-for-the-unitary-coupled-cluster-ucc-algorithm-in) stackexchange question. Since #322 only fixes the documentation aspects of #307 I am extracting this improvement to a separate issue.

Possible checks which could be done are:
- excitations are a list of pairs
- each pair consists of two tuples of identical length
- all indices within a pair are unique

A potential location for these checks is the `UCC.excitations` property setter method.
The `__init__` method should then make use of that setter to trigger those checks, too.

jschuhmac:
### Summary
Adds checks for the generated excitations in the UCC ansatz. 
The checks are only executed when the excitations were generated by a function (assuming that all other cases are handled elsewhere; if `self.excitation` is of type `str`, `int` or `List[int]`).
If one of the checks fails, a `QiskitNatureError` is raised.

### Details and comments
The expected format of the excitations is `List[Tuple[Tuple[int, ...], Tuple[int, ...]]]`.
The following checks are executed:
- the list of excitations consists of pairs of tuples
- each pair of excitation indices has the same length
- the indices within each excitation pair are unique

I also added a test to check that the errors are raised in the respective scenario.

Closes #323


qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -354,6 +358,43 @@ def _get_excitation_generators(self) -> List[Callable]:
...
+        Args:
+            excitations: the list of excitations
+
+        Raises:
+            QiskitNatureError: if format of excitations is invalid
+        """
+        logger.debug("Checking excitation list...")
+
+        if any(len(exc) != 2 for exc in excitations):
+            raise QiskitNatureError(
```
woodsp-ibm, line 378: What might be helpful if this message gets raised is to have the excitation list that is in error so the user can compare. We normally try and have the failing value in an exception message. Maybe its too much to identify the failure within that list somehow, but maybe the list is long hence including it all may may rather a long message!

jschuhmac: I've updated the error message. The checks are now executed for each excitation individually. If one of the checks fails the excitation is included in the error message.

mrossinek (commented):
Thanks, @jschuhmac! This already lgtm but I have one question about this choice:
> The checks are only executed when the excitations were generated by a function (assuming that all other cases are handled elsewhere; ...)

I assume you mean by "all other cases are handled elsewhere", that the default code from Qiskit Nature which get's triggered in common events handles everything properly. I agree that this is (hopefully :sweat_smile:) the case but does it hurt to run this check even in those cases? :thinking: 
These checks should not be a real performance penalty here so I would argue to simply run them in all cases..

jschuhmac:
@mrossinek Makes sense, I will update the code so that the checks are executed in all cases.

mrossinek (approved):
Thanks @jschuhmac!

woodsp-ibm (approved):
Thanks for doing this and updating the error messages to include relevant detail of the failure as detected

PR #442, opened 2021-11-26
Fix latest pylint errors

manoelmarques:

### DIFF
--- a/qiskit_nature/operators/second_quantization/vibrational_op.py
+++ b/qiskit_nature/operators/second_quantization/vibrational_op.py
@@ -150,7 +150,7 @@ class VibrationalOp(SecondQuantizedOp):
             # Sparse label
             dense_labels = self._convert_to_dense_labels(data, num_modals)
 
-            ops: List[VibrationalOp] = []
+            ops: List["VibrationalOp"] = []
             for dense_label, coeff in dense_labels:
                 new_op = reduce(
                     lambda a, b: a @ b,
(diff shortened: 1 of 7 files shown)


qiskit_nature/operators/second_quantization/vibrational_op.py
```diff
@@ -150,7 +150,7 @@ def __init__(
             # Sparse label
             dense_labels = self._convert_to_dense_labels(data, num_modals)
 
-            ops: List[VibrationalOp] = []
+            ops: List["VibrationalOp"] = []
```
mrossinek, line 153: Huh.. I'm surprised to see that this supposedly worked before.. There is also the option to do
```python
```
What do you think about that?

mrossinek (approved):
LGTM! I have a minor comment about a potential alternative solution to one aspect but since it's not required I will approve this as is, already :+1:

qiskit_nature/operators/second_quantization/vibrational_op.py
manoelmarques: I think the double quotes makes sense as it is what we already use everywhere in typehints with the class being declared inside itself. This fix works with both the old pylint and the new one. It is just that the old didn't report it.

manoelmarques: For the next release after Terra 0.19 we will drop support for python 3.6. The we can use `from __future__ import annotations` and remove the strings from  the type annotations

mrossinek: Right that is only available in python 3.7+.. I hadn't thought of that :sweat_smile:

Issue #123, opened 2021-04-12
Refactor `AdaptVQE` to leverage opflow gradients


mrossinek:
### What is the expected enhancement?

The adaptive VQE was originally implemented in Qiskit Aqua as the outcome of the Qiskit Hackathon Europe 2019, [link](https://github.com/Qiskit/qiskit-aqua/pull/685). Since then, it has already been refactored multiple times during the migration to the `GroundStateEigensolver` interface and later transition to Qiskit Nature.

As it stands today, it still performs the gradient computation manually, see [here](https://github.com/Qiskit/qiskit-nature/blob/main/qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py#L78), instead of leveraging Qiskit Terra's [gradient functionality](https://github.com/Qiskit/qiskit-terra/pull/5181) released as part of Terra 0.17.0

The goal of this issue, is to refactor the implementation of `AdaptVQE` to make use of the native gradient functionality included in Qiskit Terra.

fs1132429:
👍

PR #445, opened 2021-12-01
Gradient based AdaptVQE

fs1132429:
### Summary
Implements #123 
https://github.com/qiskit-advocate/qamp-fall-21/issues/5

### Details and comments
As part of the Qiskit Mentorship Program 2021, I have implemented the Qiskit Terra Gradient functionality into AdaptVQE algorithm.


qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -47,6 +48,7 @@ def __init__(
         threshold: float = 1e-5,
         delta: float = 1,
```
mrossinek, line 49: The argument `delta` is now no longer in use. Thus, we should deprecate it. You can do that by adding the following above the `def __init__` above:
```python
@deprecate_arguments("0.4.0", {"delta": "gradient"})
```
This will redirect users of the `delta` argument to the new `gradient` argument. We can think about adding some more information later.
You will also need to add this import at the top of the file:
```python
from qiskit_nature.deprecation import deprecate_arguments
```

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -56,22 +58,35 @@ def __init__(
...
 
+        if gradient is None:
+            gradient = Gradient(grad_method="param_shift")
+
         super().__init__(qubit_converter, solver)
 
         self._threshold = threshold
         self._delta = delta
         self._max_iterations = max_iterations
+        self._gradient = gradient
```
mrossinek, line 74: Just a comment: Terra's `VQE` class already holds a `gradient` object. In the future we could think about using the same one here. However, I am not sure if these can really be intended for the same use cases. Maybe @Cryoris knows a little more about this.

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -89,22 +104,24 @@ def _compute_gradients(
 
         Returns:
             List of pairs consisting of gradient and excitation operator.
+            parameters = ParameterVector("θ", self._num_parameters)
+            times = ParameterVector("t", self.reps * len(self.operators))
```
mrossinek, line 108:
```suggestion
```
I believe these are left over code snippets which need to be removed.

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -89,22 +104,24 @@ def _compute_gradients(
...
         """
         res = []
-        # compute gradients for all excitation in operator pool
```
mrossinek, line 94: We could keep this comment.

mrossinek: The same actually holds for the comments removed on the following lines

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -89,22 +104,24 @@ def _compute_gradients(
...
-            vqe.ansatz = self._ansatz
-            # evaluate energies
-            parameter_sets = theta + [-self._delta] + theta + [self._delta]
-            energy_evaluation = vqe.get_energy_evaluation(self._main_operator)
-            energy_results = energy_evaluation(np.asarray(parameter_sets))
-            # compute gradient
-            gradient = (energy_results[0] - energy_results[1]) / (2 * self._delta)
-            res.append((np.abs(gradient), exc))
-
+            if self.gradient._grad_method.analytic:
```
mrossinek, line 113:
```suggestion
            if self.gradient.grad_method.analytic:
```
`grad_method` is also available as a property so we should prefer accessing it via this public attribute.

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -89,22 +104,24 @@ def _compute_gradients(
...
+                vqe.ansatz = self._ansatz
+            else:
+                vqe.ansatz = self._ansatz.decompose()
+            param_sets = vqe._ansatz_params
+            op = vqe.construct_expectation(theta, self._main_operator)
+            state_grad = self.gradient.convert(operator=op, params=param_sets)
+            # Assign the parameters and evaluate the gradient
+            value_dict = {param_sets[-1]: 0.0}
+            state_grad_result = state_grad.assign_parameters(value_dict).eval()
+            print("State gradient computed with parameter shift", state_grad_result)
```
mrossinek, line 123:
```suggestion
            logger.info("State gradient computed with parameter shift", state_grad_result)
```
Instead of always printing this, we should use the `logger` instead such that a user can configure whether or not they would like to see this kind of output.

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -26,17 +26,19 @@
 from qiskit.utils import QuantumInstance
 from qiskit.algorithms import VQE
 from qiskit.algorithms.optimizers import L_BFGS_B
+from qiskit.opflow.gradients import Gradient
 from qiskit_nature.algorithms import AdaptVQE, VQEUCCFactory
 from qiskit_nature.circuit.library import HartreeFock, UCC
 from qiskit_nature.drivers import UnitsType
-from qiskit_nature.drivers.second_quantization import PySCFDriver
+from qiskit_nature.drivers.second_quantization import PySCFDriver, HDF5Driver
```
mrossinek, line 33:
```suggestion
from qiskit_nature.drivers.second_quantization import PySCFDriver
```

mrossinek (changes requested):
Thanks @fs1132429! I have some suggestions below but overall this already looks quite good :+1:

fs1132429:
@mrossinek I have made the suggested changes 👍

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -40,13 +43,15 @@
 class AdaptVQE(GroundStateEigensolver):
     """A ground state calculation employing the AdaptVQE algorithm."""
 
+    @deprecate_arguments("0.4.0", {"delta": "gradient"})
```
mrossinek, line 46: Maybe we can also add a little bit more explanation using the `additional_msg` argument. Kind of like this but with an actual explanation:
```suggestion
    @deprecate_arguments("0.4.0", {"delta": "gradient"}, additional_msg="Here goes my explanation.")
```

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -91,20 +109,23 @@ def _compute_gradients(
...
         for exc in self._excitation_pool:
             # add next excitation to ansatz
             self._ansatz.operators = self._excitation_list + [exc]
-            # set the current ansatz
-            vqe.ansatz = self._ansatz
-            # evaluate energies
-            parameter_sets = theta + [-self._delta] + theta + [self._delta]
-            energy_evaluation = vqe.get_energy_evaluation(self._main_operator)
-            energy_results = energy_evaluation(np.asarray(parameter_sets))
+            if self.gradient.grad_method.analytic:
```
mrossinek, line 115: As suggested by @Cryoris in the Qiskit Demo Day where this work was presented, we should dig a little deeper into why this is necessary. Maybe this requires a fix on the Terra level instead of this conditional check, here.

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -68,6 +70,43 @@ def test_default(self):
...
+        self.assertAlmostEqual(res.electronic_energies[0], self.expected, places=6)
+
+    def test_fin_diff(self):
+        """test for fin_diff method"""
+        solver = VQEUCCFactory(QuantumInstance(BasicAer.get_backend("statevector_simulator")))
+        grad = Gradient(grad_method="fin_diff", epsilon=1.0)
+        calc = AdaptVQE(self.qubit_converter, solver, gradient=grad)
+        res = calc.solve(self.problem)
+        self.assertAlmostEqual(res.electronic_energies[0], self.expected, places=6)
+
```
mrossinek, line 88: I suggest that you also add unittests for the other gradient types which we started to test based on Julien's feedback. Even if they fail for now, once we debug that we should have tests to guard against future incompatibilities.

mrossinek (commented):
I have some more minor comments. The focus for now should be testing the other gradient types as discussed offline.

You will also need to write a release note for this PR. The way you can do that is by doing the following in a terminal:
```
pip install reno
reno new gradient-based-adapt-vqe
# this will return the path of a new file in `releasenotes/notes/...
# you can now edit this file to reflect the changes done in this PR (I suggest to use the `feature` section and to delete all others)
```

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -42,13 +46,19 @@
 class AdaptVQE(GroundStateEigensolver):
     """A ground state calculation employing the AdaptVQE algorithm."""
 
+    @deprecate_arguments(
+        "0.4.0",
+        {"delta": "gradient"},
+        additional_msg="Instead of Delta=1.0 you have to construct gradient. For example:: Gradient(grad_method='lin_comb', epsilon=1.0)",
```
mrossinek, line 52:
```suggestion
        additional_msg="Instead of `delta=1.0` you have to construct a gradient, like so `gradient=Gradient(grad_method='fin_diff', epsilon=1.0)`.",
```

mrossinek: After applying this you should run `make black` to fix the line length.

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -93,20 +116,21 @@ def _compute_gradients(
...
         for exc in self._excitation_pool:
             # add next excitation to ansatz
             self._ansatz.operators = self._excitation_list + [exc]
             # set the current ansatz
-            vqe.ansatz = self._ansatz
-            # evaluate energies
-            parameter_sets = theta + [-self._delta] + theta + [self._delta]
-            energy_evaluation = vqe.get_energy_evaluation(self._main_operator)
-            energy_results = energy_evaluation(np.asarray(parameter_sets))
+            vqe.ansatz = self._ansatz.decompose()
```
mrossinek, line 123: Please add a comment explaining why we are always decomposing the ansatz now

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -68,6 +70,61 @@ def test_default(self):
...
+        transformer = ActiveSpaceTransformer(
+            num_electrons=2,
+            num_molecular_orbitals=3,
+        )
+        problem1 = ElectronicStructureProblem(driver1, [transformer])
+        self.expected = -1.85727503
+        solver = VQEUCCFactory(QuantumInstance(BasicAer.get_backend("statevector_simulator")))
+        calc = AdaptVQE(self.qubit_converter, solver)
+        res = calc.solve(problem1)
+        self.assertAlmostEqual(res.electronic_energies[0], self.expected, places=6)
```
mrossinek, line 126: This is running far too long for the normal CI. Please mark it with `@slow_test` and add `from qiskit.test import slow_test`

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -68,6 +72,62 @@ def test_default(self):
...
+
+    def test_natural_gradients(self):
+        """test for fin_diff method"""
+        solver = VQEUCCFactory(QuantumInstance(BasicAer.get_backend("statevector_simulator")))
+        grad = NaturalGradient(
+            grad_method="fin_diff", qfi_method="lin_comb_full", regularization="ridge"
+        )
+        calc = AdaptVQE(self.qubit_converter, solver, gradient=grad)
+        res = calc.solve(self.problem)
+        self.assertAlmostEqual(res.electronic_energies[0], self.expected, places=6)
```
mrossinek, line 107: Alright I fixed up some lint issues to get a better feel of what is still missing here. Afaict, the last problem is this unittest which still does not work as expected.
@fs1132429 could it be that the gradient is still not chosen optimally?

I think we should do several things:
- keep this unittest but improve the behavior of a scenario such as this one, such that the code does not simply break but raises a meaningful warning about the gradient not being chosen well (we attempted that but it appears as though this is either a different edge case or our previous attempt does not work as expected; see the latest CI result [here](https://github.com/Qiskit/qiskit-nature/runs/4773154402?check_suite_focus=true))
- add another natural gradient test with a well chosen gradient which works as expected :+1: 

We should also update the documentation of this class to reflect that the performance of the `AdaptVQE` can depend significantly on the chosen gradient (i.e. the gradient method, the QFI solver, the epsilon, etc.)

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -23,7 +23,10 @@
 from qiskit.algorithms import VQE
 from qiskit.circuit import QuantumCircuit
 from qiskit.opflow import OperatorBase, PauliSumOp
+from qiskit.opflow.gradients import GradientBase, Gradient
 from qiskit.utils.validation import validate_min
+from sqlalchemy import Float
```
mrossinek, line 28: This is not needed. Please remove.

You can simply use `float` below.

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -58,28 +72,44 @@ def __init__(
...
+        """Sets the gradient."""
+        self._gradient = grad
+
     def returns_groundstate(self) -> bool:
         return True
 
     def _compute_gradients(
         self,
-        theta: List[float],
+        theta: List[Float],
```
mrossinek, line 112:
```suggestion
        theta: List[float],
```
As per comment above

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -93,20 +123,23 @@ def _compute_gradients(
...
-            vqe.ansatz = self._ansatz
-            # evaluate energies
-            parameter_sets = theta + [-self._delta] + theta + [self._delta]
-            energy_evaluation = vqe.get_energy_evaluation(self._main_operator)
-            energy_results = energy_evaluation(np.asarray(parameter_sets))
+            # the ansatz needs to be decomposed for the gradient to work
+            vqe.ansatz = self._ansatz.decompose()
+            param_sets = sorted(vqe.ansatz.parameters,key=lambda p: p.name)
+            #zip will only iterate the length of the shorter list
+            theta1=dict(zip(theta,vqe.ansatz.parameters))
```
mrossinek, line 133:
```suggestion
            theta1=dict(zip(vqe.ansatz.parameters, theta))
```
The arguments here need to be flipped because you want to map from parameter names to their optimal values (stored in theta).
This resolves the error you were showing to me on Slack

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -254,6 +288,11 @@ def solve(
                 )
                 threshold_satisfied = True
                 break
+            if iteration == 1 and np.abs(max_grad[0]) < self._threshold:
+                raise QiskitNatureError(
+                    "Gradient choice is not suited as it leads to all non-zero gradients. "
+                    "Try a different gradient method."
+                )
```
mrossinek, line 295: This is an attempt at catching bad gradient configurations. However, as it stands now it will never occur because line 284 will be executed beforehand.
You will need to move this section above that `if` statement or you can integrate the two like so:
```python
            if np.abs(max_grad[0]) < self._threshold:
                if iteration == 1:
                    raise QiskitNatureError(
                        "Gradient choice is not suited as it leads to all non-zero gradients. "
                        "Try a different gradient method."
                    )
                logger.info(
                    "Adaptive VQE terminated successfully with a final maximum gradient: %s",
                    str(np.abs(max_grad[0])),
                )
                threshold_satisfied = True
                break
```

Once you do this, the `test_natural_gradient` test case will no longer run into an awkward Python-internal error but instead will raise this warning.
You then have two options for that test case:
- figure out a natural gradient which works
- rewrite the test using `assertRaises` as done for example here: https://github.com/Qiskit/qiskit-nature/blob/main/test/converters/second_quantization/test_qubit_converter.py#L264

mrossinek (changes requested):

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -58,22 +75,38 @@ def __init__(
...
+        """Returns the gradient."""
+        return self._gradient
+
+    @gradient.setter
+    def gradient(self, grad: Optional[GradientBase]):
+        """Sets the gradient."""
+        self._gradient = grad
+
     def returns_groundstate(self) -> bool:
         return True
```
dlasecki, line 78: Let's add docs here.

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -68,6 +73,62 @@ def test_default(self):
...
+            charge=0,
+            spin=0,
+            basis="sto3g",
+        )
+        transformer = ActiveSpaceTransformer(
+            num_electrons=2,
+            num_molecular_orbitals=3,
+        )
+        problem1 = ElectronicStructureProblem(driver1, [transformer])
+        self.expected = -1.85727503
```
dlasecki, line 126: Is it redundant? I can see it being in the set up.

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -58,22 +75,38 @@ def __init__(
...
         self._main_operator: PauliSumOp = None
         self._ansatz: QuantumCircuit = None
 
+    @property
+    def gradient(self) -> Optional[GradientBase]:
+        """Returns the gradient."""
+        return self._gradient
+
+    @gradient.setter
+    def gradient(self, grad: Optional[GradientBase]):
```
dlasecki, line 106: Return typehint missing (None).

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -58,23 +75,40 @@ def __init__(
...
         self._main_operator: PauliSumOp = None
         self._ansatz: QuantumCircuit = None
 
+    @property
+    def gradient(self) -> Optional[GradientBase]:
+        """Returns the gradient."""
+        return self._gradient
+
+    @gradient.setter
+    def gradient(self, grad: Optional[GradientBase] = None):
```
mrossinek, line 106:
```suggestion
    def gradient(self, grad: Optional[GradientBase] = None)  -> None:
```
@dlasecki had actually meant the lack of return type. But the inclusion of `= None` was also correct here :+1:

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -60,6 +65,18 @@ def setUp(self):
         self.expected = -1.85727503
 
         self.qubit_converter = QubitConverter(ParityMapper())
+
```
mrossinek, line 68:
```suggestion

```
You cannot have trailing whitespace, otherwise the linter will complain.
I am not sure if this is the only reason why the style checker is also failing. I suggest that you run `make black` locally to ensure that is fixed.
You can also run `make lint` locally to check that the linter passes.

mrossinek (changes requested):

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -61,13 +66,67 @@ def setUp(self):
...
+    def test_natural_gradients_invalid(self):
+        """test that an exception is thrown when an invalid gradient method is used"""
+        solver = VQEUCCFactory(QuantumInstance(BasicAer.get_backend("statevector_simulator")))
+        grad = NaturalGradient(
+            grad_method="fin_diff", qfi_method="lin_comb_full", regularization="ridge"
+        )
+        calc = AdaptVQE(self.qubit_converter, solver, gradient=grad)
+        with self.assertRaises(QiskitNatureError):
+            _ = calc.solve(self.problem)
```
mrossinek, line 120: Just a comment for another future reviewer: we're aware that this is not optimal but currently we have not yet figured out a working scenario with `NaturalGradient`. This is planned to happen though when we migrate `AdaptVQE` to Terra as part of https://github.com/qiskit-advocate/qamp-spring-22/issues/3

mrossinek (commented):
This is in a state where I can approve it now! I'm holding off on doing that though until either @pbark or @woodsp-ibm also had a look :+1:

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -248,6 +284,11 @@ def solve(
                         gradlog += "\t(*)"
                 logger.info(gradlog)
             if np.abs(max_grad[0]) < self._threshold:
+                if iteration == 1:
+                    raise QiskitNatureError(
+                        "Gradient choice is not suited as it leads to all non-zero gradients. "
+                        "Try a different gradient method."
+                    )
```
woodsp-ibm, line 291: I think this error message here should says it `leads to alls-zero gradients` right rather than non-zero. That would then match too with the text in Raises docstring above.

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -40,15 +42,28 @@
...
+    """
+
+    @deprecate_arguments(
+        "0.4.0",
+        {"delta": "gradient"},
+        additional_msg=(
+            "Instead of `delta=1.0` you have to construct a gradient, like so "
+            "`gradient=Gradient(grad_method='fin_diff', epsilon=1.0)`."
+        ),
+    )
```
woodsp-ibm, line 58: Normally when deprecating something the code should continue to work as it did. E.g if I had code that used/set delta it ought to work here albeit we see some deprecation message saying that in the future only gradient will be supported. Hence I would expect to see paths in the code for if delta is used (that will be removed at some point) and paths for Gradient. As far as I can see delta is set on the instance but otherwise ignored; a default gradient is created if your code uses it that  way and predates the gradient being added, the code then just works on this gradient.

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -135,6 +139,7 @@ def _compute_gradients(
             # zip will only iterate the length of the shorter list
             theta1 = dict(zip(vqe.ansatz.parameters, theta))
             op = vqe.construct_expectation(theta1, self._main_operator)
+            print("GRADIENT::", self.gradient)
```
woodsp-ibm, line 142: Left over from debugging I imagine

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -77,15 +78,18 @@ def __init__(
...
         validate_min("threshold", threshold, 1e-15)
-        validate_min("delta", delta, 1e-5)
+
+        if isinstance(gradient, float):
+            # this scenario can only occur while using the deprecate_arguments wrapper which will
+            # move any argument supplied to delta into gradient.
+            gradient = Gradient(grad_method="fin_diff", epsilon=gradient)
 
         if gradient is None:
             gradient = Gradient(grad_method="param_shift")
```
woodsp-ibm, line 88: I must admit I don't quite understand how gradient could ever be None if delta is copied over it - in the future when delta is gone, but now...

How do I supply a gradient explicitly now - maybe the copy is done only if gradient is not explicitly provided, or delta provided explicitly - though if its the latter case that would be problematic if defaulted (I should look at the deprecate arguments decorator code I guess!).

mrossinek: Yes, the copy is only done by the deprecation decorator if and only if the user provided a `delta` explicitly. Thus, `gradient is None` can still occur when the user has provided neither a custom `delta` nor a custom `gradient`.

This scenario is also covered by the unittests (the `test_default` case).

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -40,15 +42,29 @@
...
+        ),
+    )
+    # pylint: disable=unused-argument
     def __init__(
         self,
         qubit_converter: QubitConverter,
         solver: MinimumEigensolverFactory,
         threshold: float = 1e-5,
-        delta: float = 1,
+        delta: float = 1,  # delta is copied into gradient by the deprecate_arguments wrapper
```
dlasecki, line 65:
```suggestion
        delta: float = 1.0,  # delta is copied into gradient by the deprecate_arguments wrapper
```

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
woodsp-ibm: Let me ask more specifically. Is the default behavior when delta is not supplied explicitly, the same as it was before. Ie if I ran it before with default delta do I get the same still result - it would seem in this case it would drop with gradient is None path right, and hence it seems is should/could be different as I see the find_diff path being equivalent to the old way.

mrossinek: The previous default behavior remains unchanged: when a delta is supplied, a `fin_diff` gradient with the given delta will be used (as in the old days). However, the new default behavior uses param_shift instead which we found to be more performant.

woodsp-ibm: So basically anyone who never supplied delta, but used the provided default, will see a difference then. So if some one wants to reproduce something and was implicitly using defaulted delta=1 to get the same outcome they would have to change their code to explicitly provide delta=1.

What I was suggesting before, as a transition step it could use fin_diff, based on delta, with gradient defaulting to None. To use gradient one would explicitly provide it and we could suggest to use param shift which could become default once delta is removed.

If you feel how it is better I imagine the algo did not get wide usage before so perhaps any impact to the default being different than before won't have any real world impact. If someone is trying to say reproduce something from before maybe docs should say that to get the same prior default behavior you can supply delta=1 explicitly.

mrossinek: The end result is not affected significantly so I do not think that this change in default has any real world impact. We can state in the docs how to reproduce the original default case. @fs1132429 could you please add a sentence to the docs explaining this?

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -58,23 +74,44 @@ def __init__(
...
         validate_min("threshold", threshold, 1e-15)
-        validate_min("delta", delta, 1e-5)
+
+        if isinstance(gradient, float):
+            # this scenario can only occur while using the deprecate_arguments wrapper which will
+            # move any argument supplied to delta into gradient.
+            gradient = Gradient(grad_method="fin_diff", epsilon=gradient)
+
+        if gradient is None:
+            #To reproduce the original default case supply delta=1 explicitly.
```
mrossinek, line 88: I really meant that you should add this information in the documentation string above. That way, it will be visible in the online documentation which will not be the case with a simple comment like this one. You could also be a bit more explicit by saying:
```
To reproduce the default behavior of AdaptVQE prior to Qiskit Nature 0.4 you should supply `delta=1` explicitly.
This will use a finite difference scheme for the gradient evaluation whereas after version 0.4 a parameter shift gradient will be used.
[Maybe insert a link to the Qiskit Terra Gradient documentation here].
```

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -40,15 +42,34 @@
...
-
+    """A ground state calculation employing the AdaptVQE algorithm.
+
+    The performance of AdaptVQE can significantly depend on the choice of gradient method, QFI
+    solver (if applicable) and the epsilon value.
+
+    To reproduce the default behavior of AdaptVQE prior to Qiskit Nature 0.4 you should supply
+    `delta=1` explicitly. This will use a finite difference scheme for the gradient evaluation
+    whereas after version 0.4 a parameter shift gradient will be used.
+    [https://qiskit.org/documentation/tutorials/operators/02_gradients_framework.html]
```
mrossinek, line 53: Let's be a bit more explicit here:
```suggestion
    For more information refer to the gradient framework of Qiskit Terra:
    https://qiskit.org/documentation/tutorials/operators/02_gradients_framework.html
```

mrossinek (approved):
This looks good to me!

Thanks @fs1132429 for the hard work on this!

@pbark @woodsp-ibm @dlasecki I invite you to take another look before merging this :+1:

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -97,16 +139,19 @@ def _compute_gradients(
...
             # add next excitation to ansatz
             self._ansatz.operators = self._excitation_list + [exc]
-            # set the current ansatz
-            vqe.ansatz = self._ansatz
-            # evaluate energies
-            parameter_sets = theta + [-self._delta] + theta + [self._delta]
-            energy_evaluation = vqe.get_energy_evaluation(self._main_operator)
-            energy_results = energy_evaluation(np.asarray(parameter_sets))
+            # the ansatz needs to be decomposed for the gradient to work
+            vqe.ansatz = self._ansatz.decompose()
```
mrossinek, line 143:
```suggestion
            vqe.ansatz = self._ansatz
```
This has actually since been fixed in Terra so the decomposition is no longer needed.

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -23,6 +23,7 @@
 from qiskit.algorithms import VQE
 from qiskit.circuit import QuantumCircuit
 from qiskit.opflow import OperatorBase, PauliSumOp
```
mrossinek, line 25:
```suggestion
from qiskit.opflow import OperatorBase, PauliSumOp, CircuitSampler
```
See below

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -97,16 +139,19 @@ def _compute_gradients(
         for exc in self._excitation_pool:
```
mrossinek, line 139:
```suggestion
        sampler = CircuitSampler(vqe.quantum_instance)
        for exc in self._excitation_pool:
```
See below

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -97,16 +139,19 @@ def _compute_gradients(
...
+            theta1 = dict(zip(vqe.ansatz.parameters, theta))
+            op = vqe.construct_expectation(theta1, self._main_operator)
             # compute gradient
-            gradient = (energy_results[0] - energy_results[1]) / (2 * self._delta)
-            res.append((np.abs(gradient), exc))
-
+            state_grad = self.gradient.convert(operator=op, params=param_sets)
+            # Assign the parameters and evaluate the gradient
+            value_dict = {param_sets[-1]: 0.0}
+            state_grad_result = state_grad.assign_parameters(value_dict).eval()
```
mrossinek, line 152:
```suggestion
            state_grad_result = sampler.convert(state_grad, params=value_dict).eval()
```
This change is necessary in order to respect the VQE backend. If we do not use a sampler (like we did before) we would always use statevector evaluations.

mrossinek (changes requested):
I had some more discussion with @Cryoris and we noticed these improvements :+1:

mrossinek (approved):
@woodsp-ibm @pbark @dlasecki Do you think we can get this merged this week?

releasenotes/notes/gradient-based-adapt-vqe-aa4e89f44c5aa69e.yaml
```diff
@@ -0,0 +1,11 @@
+---
+features:
```
woodsp-ibm, line 2: Perhaps we should note the default behavior change in the release note so as to alert anyone that used a default delta value that the behavior is different. Maybe it suffices to just highlight this that here and link/refer to the note in the class for more info.

pbark (approved):
I think this LGTM.

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -94,19 +136,25 @@ def _compute_gradients(
         """
         res = []
         # compute gradients for all excitation in operator pool
+        sampler = CircuitSampler(vqe.quantum_instance)
```
Cryoris, line 139: Do you expect that sometimes the same ansatz will be evaluated here? If yes, it could be useful to store one single sampler in the class since it will cache the transpiled circuits

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -58,23 +80,43 @@ def __init__(
...
 
+    @property
+    def gradient(self) -> Optional[GradientBase]:
+        """Returns the gradient."""
+        return self._gradient
+
+    @gradient.setter
+    def gradient(self, grad: Optional[GradientBase] = None) -> None:
+        """Sets the gradient."""
+        self._gradient = grad
```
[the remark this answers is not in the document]
fs1132429: Yes if Gradient is `None`, it sets it to `param_shift` method.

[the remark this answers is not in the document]
Cryoris: It does in the initializer, but here in the setter it doesn't, does it? I mean in a use case like
```python
vqe = AdaptVQE(...)
vqe.gradient = None
# is vqe.gradient now None or Gradient("param_shift")?
```

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
mrossinek: The ansatz will change every time so I do not think it makes a difference

mrossinek: I ended up extracting the sampler anyways :+1:

qiskit_nature/algorithms/ground_state_solvers/adapt_vqe.py
```diff
@@ -58,23 +81,43 @@ def __init__(
...
 
         self._excitation_pool: List[OperatorBase] = []
         self._excitation_list: List[OperatorBase] = []
 
         self._main_operator: PauliSumOp = None
         self._ansatz: QuantumCircuit = None
+        self._sampler: CircuitSampler = None
+
+    @property
+    def gradient(self) -> Optional[GradientBase]:
```
Cryoris, line 108: Now this cannot be `None` anymore, right? 🙂 
```suggestion
    def gradient(self) -> GradientBase:
```

releasenotes/notes/gradient-based-adapt-vqe-aa4e89f44c5aa69e.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - |
+    Added Qiskit Terra's Gradient Functionality to calculate gradients in AdaptVQE. You can choose
+    the gradient method and input it while calling ``AdaptVQE``.
+    for example::
+
+      from qiskit.opflow.gradients import Gradient
+      grad = Gradient(grad_method="lin_comb", epsilon=1.0)
+      calc = AdaptVQE(self.qubit_converter, solver, gradient=grad)
```
Cryoris, line 10: Would it be nice if this is a working example since this is what users will see in the releasenotes? 🙂

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -69,6 +74,72 @@ def test_default(self):
         res = calc.solve(self.problem)
         self.assertAlmostEqual(res.electronic_energies[0], self.expected, places=6)
 
+    def test_param_shift(self):
```
Cryoris, line 77: We could group most of these tests into one using `data` and `ddt` and avoid boilerplate code 😄 
```python
from ddt import ddt, data

@ddt
class TestAdaptVQE(...):

    @data("param_shift", "fin_diff", "lin_comb")
    def test_gradient(self, grad_method):
        grad = Gradient(grad_method)
        ...
```
(we'd have to add a bit more to also pass `epsilon=1` for finite differences)

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -69,6 +74,72 @@ def test_default(self):
...
+        solver = VQEUCCFactory(QuantumInstance(BasicAer.get_backend("statevector_simulator")))
+        grad = Gradient(grad_method="fin_diff", epsilon=1.0)
+        calc = AdaptVQE(self.qubit_converter, solver, gradient=grad)
+        res = calc.solve(self.problem)
+        self.assertAlmostEqual(res.electronic_energies[0], self.expected, places=6)
+
+    def test_lin_comb(self):
+        """test for fin_diff method"""
+        solver = VQEUCCFactory(QuantumInstance(BasicAer.get_backend("statevector_simulator")))
+        grad = Gradient(grad_method="lin_comb", epsilon=1.0)
```
Cryoris, line 96: Is `epsilon` supposed to be specified here?
```suggestion
        grad = Gradient(grad_method="lin_comb")
```

manoelmarques:
Pyscf is not available on windows or python 3.10. The tests that use pycsf should add the decorator:
```
```
So that they just get skipped and don’t fail.
You can search other unit tests that use pyscf to see how it is applied.

mrossinek (dismissed):

releasenotes/notes/gradient-based-adapt-vqe-aa4e89f44c5aa69e.yaml
```diff
@@ -0,0 +1,51 @@
...
+      self.driver = PySCFDriver(
+            atom="H .0 .0 .0; H .0 .0 0.735", unit=UnitsType.ANGSTROM, basis="sto3g"
+            )
+      self.problem = ElectronicStructureProblem(self.driver)
+      self.qubit_converter = QubitConverter(ParityMapper())
+      solver = VQEUCCFactory(QuantumInstance(BasicAer.get_backend("statevector_simulator")))
+      grad = Gradient(grad_method="lin_comb", epsilon=1.0)
+      calc = AdaptVQE(self.qubit_converter, solver, gradient=grad)
+      res = calc.solve(self.problem)
+deprecates:
```
woodsp-ibm, line 26:
```suggestion
deprecations:
```
This section is not correctly named is not appearing in release notes. I am not sure if thats why the section looks a bit odd or not. It maybe a prior note needs sorting too, not sure.

image

woodsp-ibm (changes requested):

woodsp-ibm (approved):
I think the import in the sample could have done it from qiskit.algorithms.AdaptVQE, where we expose it, rather than directly from the module itself  i.e. `from qiskit_nature.algorithms.ground_state_solvers.adapt_vqe import AdaptVQE` but I guess it works as the latter

mrossinek (approved):
Thank you, @fs1132429, for your hard work on this project!

pbark (approved):
LGTM! Thanks for the great work everyone!

PR #447, opened 2021-12-02
Prepare 0.3.0 Release

manoelmarques:


mrossinek (changes requested):
We need to await the merging of #365 and #430 before releasing 0.3.0

manoelmarques:
Yes, I created those PRs in all 4 repos to be prepared when Terra releases but they will be merged only when the repos are ready.

docs/conf.py
```diff
@@ -110,7 +110,7 @@
 templates_path = ["_templates"]
 html_css_files = ["style.css", "custom.css", "gallery.css"]
 
-nbsphinx_timeout = 2400
+nbsphinx_timeout = 360
```
mrossinek, line 113: Can we change this back?

manoelmarques: Once the performance bug was fixed in Terra I put back the old values in all repositories. This helps us identify another bug like that one.

mrossinek (approved):

Issue #436, opened 2021-11-22
BKSF mapper fails to map H2 operator from driver


woodsp-ibm:
### Environment

- **Qiskit Nature version**: `main` branch (did not try release)
- **Python version**: 3.6
- **Operating system**: Linux

### What is happening?

In #432 I was asked had I tried with BKSF as I had added Parity and BK mappers into the [test_excited_states_solvers.py](https://github.com/Qiskit/qiskit-nature/blob/main/test/algorithms/excited_state_solvers/test_excited_states_solvers.py) test case when fixing the referenced issue and observing issues around two qubit reduction and symmetries. I did not try BKSF, you can see my comment in the referenced PR, but in trying it fails in the ground state computation. Now the bksf test file includes an pre-defined operator that is commented as H2 - in this case its built from the driver and here it fails with the following exception.

```
<snip>
line 480, in _add_edges_for_term
    _type = _interaction_type(n_number, n_raise, n_lower)
  File "...qiskit_nature/mappers/second_quantization/bksf.py", line 421, in _interaction_type
    raise ValueError("unexpected number of number operators: ", n_number)
ValueError: ('unexpected number of number operators: ', 0)
```

### How can we reproduce the issue?

Based on the 
```
from qiskit import BasicAer
from qiskit.utils import algorithm_globals, QuantumInstance
from qiskit_nature.drivers import UnitsType
from qiskit_nature.drivers.second_quantization import PySCFDriver
from qiskit_nature.mappers.second_quantization import BravyiKitaevSuperFastMapper, BravyiKitaevMapper
from qiskit_nature.converters.second_quantization import QubitConverter
from qiskit_nature.problems.second_quantization import ElectronicStructureProblem
from qiskit_nature.algorithms import  GroundStateEigensolver, VQEUCCFactory
 


algorithm_globals.random_seed = 8
driver = PySCFDriver(
    atom="H .0 .0 .0; H .0 .0 0.75",
    unit=UnitsType.ANGSTROM,
    charge=0,
    spin=0,
    basis="sto3g",
)
es_problem = ElectronicStructureProblem(driver)
quantum_instance = QuantumInstance(
    BasicAer.get_backend("statevector_simulator"),
    seed_transpiler=90,
    seed_simulator=12,
)


#converter = QubitConverter(BravyiKitaevMapper())
converter = QubitConverter(BravyiKitaevSuperFastMapper())
solver = VQEUCCFactory(quantum_instance)
[...]

PR #451, opened 2021-12-06
Handle corner cases in BKSF mapper

jlapeyre:
This PR does the following to handle corner cases instead of raising an error that is not as specific as it could be:

## Changes

1. Skip FermionicOp terms with coefficient equal to zero when building a Pauli operator.
   This might be a bit more efficient if any such terms sneak through. In particular,
   an identity term with zero coefficient will be skipped.

2. Before, if there were no contributing terms, say the coefficient of each Fermi term is zero,
   then the `SparsePauliOp` would be None. With this PR, we instead construct a `SparsePauliOp`
   representing zero. We do this by constructing an identity operator with the correct number of
   qubits and setting the coefficient to zero.

3.  Fix a typo in a doc string.

## Discussion

This PR fixes some of the issue in #436, but is not enough for the broken example given to complete without throwing an error.

The VQE problem calls the BKSF mapper several times, including once with a Hamiltonian proportional to the identity operator, with coefficient zero.  If the rest of the Fermionic Hamiltonian were correct, this PR allows the mapper to give the correct result. However, later in the optimization, another non-particle-preserving Hamiltonian is sent. BKSF cannot handle this. The mapper throws an informative error, saying that the number of raising and lowering operators is not equal. It looks like something is rather badly broken in the process that generates the Fermionic Hamiltonian. But, it is so glaring, that I suspect I might be missing something.

Another question is whether BKSF or other mappers should handle terms proportional to the identity, eg.  `1.3 * IIII`. This *could* be done, eg, by adding to this PR as another corner case. But, if I understand correctly, each mapper, BKSF, J-W, etc. would handle such a term by simply adding the equivalent Pauli operator as a special case. That is each one would would reproduce the same, special, code path. However, it it seems our policy is to handle this constant energy term *after* the entire optimization procedure is done, just by adding the energy to the final result. I think this is clearly the best way. Having each mapper support such a term anyway would add, for no good reason, complexity that needs to be maintained. However, I think having each mapper throw an error at whatever point a function down the call-stack gets illegal input is the wrong behavior. The mappers should probably detect an identity term and throw an error saying that this is illegal, that is the API of the mappers should explicitly disallow it. I could put something like that in this PR. But, perhaps we should open another issue asking that this change is made for all mappers and including a checklist of mappers.


mrossinek (commented):
Alright this already looks like it's in good shape. Could you please:
- add the check to `HartreeFock` which I suggested here: https://github.com/Qiskit/qiskit-nature/issues/436#issuecomment-990799628
- add a reno

Other than that I don't think there is anything missing here.

qiskit_nature/mappers/second_quantization/bksf.py
```diff
@@ -178,9 +182,14 @@ def _operator_coefficient(term: Tuple) -> float:
     return term[1]
 
 
-def _pauli_id(n_qubits: int) -> SparsePauliOp:
+def _pauli_id(n_qubits: int, coeff=None) -> SparsePauliOp:
     """Return the identity for `SparsePauliOp` on `n_qubits` qubits."""
-    return SparsePauliOp(Pauli((np.zeros(n_qubits, dtype=bool), np.zeros(n_qubits, dtype=bool))))
+    _coeff = coeff
```
dlasecki, line 187: Why do we need this assignment? Can we operate directly on `coeff`?

jlapeyre: Yes. I don't know why that was there.

qiskit_nature/circuit/library/initial_states/hartree_fock.py
```diff
@@ -39,7 +40,14 @@ def __init__(
             num_particles: The number of particles as a tuple storing the number of alpha- and
                            beta-spin electrons in the first and second number, respectively.
             qubit_converter: a QubitConverter instance.
+
+        Raises:
+            ValueError: If qubit_converter contains BravyiKitaevSuperFastMapper.
         """
+        if isinstance(qubit_converter.mapper, BravyiKitaevSuperFastMapper):
+            raise ValueError(
```
mrossinek, line 48: Not sure about the `ValueError`. I would rather have it be a `QiskitNatureError`

qiskit_nature/circuit/library/initial_states/hartree_fock.py
```diff
@@ -39,7 +40,14 @@ def __init__(
...
+                "Unsupported mapper in qubit_converter: ", type(qubit_converter.mapper), "."
```
mrossinek, line 49: Accordingly the phrasing could be a little bit updated. We could for example point to #436 where one can read up on why this mapper is not supported.

qiskit_nature/circuit/library/initial_states/hartree_fock.py
jlapeyre: Ok. I just copied other errors in qiskit-nature that used `ValueError` for an "unsupported" value.

jlapeyre: I also see `TypeError` for the same thing in qiskit-nature:
```
qiskit_nature/transformers/second_quantization/electronic/active_space_transformer.py
427:            raise TypeError(f"{type(prop)} is an unsupported Property-type for this Transformer!"
```

mrossinek: Huh interesting. We should maybe try to unify that as part of a non-backportable PR..
@woodsp-ibm @manoelmarques @pbark Do you have any thoughts on this?

jlapeyre: The python docs suggest this is the one: https://docs.python.org/3/library/exceptions.html#TypeError
But, I have often seem software raise errors that name the project, eg `QiskitNatureError`. I am currently searching for discussions on why to choose one or the other.

woodsp-ibm: I think generally its better to raise one of the standard exceptions when it applies, as would be the case for type or value checking if its invalid as its more specific rather than using the nature custom one. For me the QiskitNatureError is more a general catch-all for others that don't fit - though arguably one could use the generic RuntimeError and we need not have a custom one. I think these more generic custom ones were to be consistent with Terra that defined/used its own custom one; it does have more specific subclasses of that for certain error categorization.

jlapeyre: Here is a guide that discusses some designs with references to real-world projects. There are a couple of strategies that do both at the same time: tell what library the exception occurred in and what kind of error.
https://julien.danjou.info/python-exceptions-guide/

One strategy is to follow what is already in qiskit-nature by subclassing `QiskitNatureError`:
https://github.com/Qiskit/qiskit-nature/blob/d6a4db83942cd55c3e23cfbf98497bbc41779594/qiskit_nature/exceptions.py#L24-L27

It might be `UnsupportedTypeError`. But, the python docs say that `TypeError` means exactly that a type is unsupported. So perhaps just `TypeError(QiskitNatureError)`.

jlapeyre: > I think generally its better to raise one of the standard exceptions when it applies

The arguments in the links above say that the reason to have something like `class TypeError(QiskitNatureError)` is so that the consumer can see which library the error occurred in (And maybe facilitate the consumer's handling of the errors).  Even, if this is a reasonable approach, I wonder the extent to which qiskit-nature is a library, i.e. called from higher level code, or an application, i.e. with classes instantiated and methods called in a jupyter notebook. The idea is that an error handling hierarchy is useful for a library that may be used in conjunction with many others in an application. Also, how important is it to get correct right now? If it changed would we have to deprecate it? I mean, someone could write code that catches this error. If we change the error, that code would break.

But, if it is just between `QiskitNatureError` and `TypeError`, then as @woodsp-ibm says, the latter is clearly better. In fact, as I read about it, I don't find anyone recommending a design that has just one error class for a whole library. (I looked in my favorite/only Python book, "Fluent Python", but it doesn't address exceptions at all)

woodsp-ibm: I'd agree that is the error is distinct where some application code can reasonably act on the error differently then it makes sense - I am just not sure a subclass of type error would ever do that. A consumer can see where the error in the stack trace, and hopefully its clear enough from that - given our naming I think it is.

jlapeyre: > consumer can see where the error in the stack trace

Yes, that's important; it's why I don't understand the rationale entirely. The only thing I can think of is the same as what the author of the post above says:
> It then makes it easy to use except ShoeError when doing anything with that piece of code related to shoes. For example, Django does not do that for some of its exceptions, making it hard to catch "any exception raised by Django".

So I have a function that calls into qiskit and might go through nature and other libraries or structures. When catching errors, I might want to begin by catching "any exception raised by qiskit-nature". But, I don't have, and the author did not give, a concrete example.

jlapeyre: Subclassing seems like added complexity unless we can articulate a clear use case. The Python docs themselves talk about a different reason for custom exceptions. The exception is really not described by any of the standard ones and/or it is useful to include some additional attributes for debugging. But, in this case, we have precisely a `TypeError`.

jlapeyre: Another example, JAX: They raise both builtin and custom errors (although some builtin errors are reworked by decorators). They subclass `TypeError`. So instead of `TypeError(QiskitNatureError)`, they do `QiskitNatureTypeError(TypeError)`.  Then they further subclass these. Why they have to be `TypeError` is not clear. But, they do clearly have a reason. They signal specific types of errors, and include information, such a a link to a webpage.  So JAX seems to raise builtin when they can, and custom otherwise.

We might want to consider having a more user-friendly "frictionless" exception setup, with webpages. Or maybe `IncompatibleDriver` classes or something like that. But, we don't have any of that in place yet.

mrossinek: From what I gather here it seems that this is a longer discussion. I see you already opened a separate issue for this (#505). Shall we use a `TypeError` (as you said is suggested by the Python docs) for the time being?

qiskit_nature/circuit/library/initial_states/hartree_fock.py
```diff
@@ -46,8 +46,9 @@ def __init__(
         """
         if isinstance(qubit_converter.mapper, BravyiKitaevSuperFastMapper):
             raise TypeError(
-                "Unsupported mapper in qubit_converter: ", type(qubit_converter.mapper),
-                ". See issue 436."
+                "Unsupported mapper in qubit_converter: ",
+                type(qubit_converter.mapper),
+                ". See issue 436.",
```
mrossinek, line 51: I suggest changing this to the entire URL. With this number not everyone may know where to find this.
A URL is also very convenient to open directly from the terminal.

mrossinek (dismissed):
LGTM, thanks @jlapeyre!

Let's get one more review on this before merging

manoelmarques (dismissed):

mrossinek (approved):

qiskit_nature/circuit/library/initial_states/hartree_fock.py
```diff
@@ -39,7 +40,16 @@ def __init__(
...
             qubit_converter: a QubitConverter instance.
+
+        Raises:
+            TypeError: If qubit_converter contains BravyiKitaevSuperFastMapper.
         """
+        if isinstance(qubit_converter.mapper, BravyiKitaevSuperFastMapper):
+            raise TypeError(
+                "Unsupported mapper in qubit_converter: ",
+                type(qubit_converter.mapper),
+                ". See https://github.com/Qiskit/qiskit-nature/issues/436",
```
mrossinek, line 51:
```suggestion
                ". See https://github.com/Qiskit/qiskit-nature/issues/537",
```

qiskit_nature/circuit/library/initial_states/hartree_fock.py
```diff
@@ -39,7 +40,16 @@ def __init__(
             num_particles: The number of particles as a tuple storing the number of alpha- and
                            beta-spin electrons in the first and second number, respectively.
             qubit_converter: a QubitConverter instance.
+
+        Raises:
+            TypeError: If qubit_converter contains BravyiKitaevSuperFastMapper.
```
mrossinek, line 45:
```suggestion
            TypeError: If qubit_converter contains BravyiKitaevSuperFastMapper. See
                https://github.com/Qiskit/qiskit-nature/issues/537 for more information.
```

mrossinek (changes requested):

woodsp-ibm (approved):

mrossinek (approved):

PR #455, opened 2021-12-08
Include the option for imaginary parts in the UCC ansatz 

tanvipg:
### Summary
A feature added to the ucc.py file to include imaginary contributions to the ucc wave function along with the real parts. To allow for complex values in the Hamiltonian, changes to the 'so_matrix' initiation in the 'two_body_electronic_integrals.py' were added as well. 

### Details and comments
1. Added dtype = 'complex' in so_matrix creation in 'two_body_electronic_integrals.py'
2. Added a boolean flag, 'include_imaginary', in the 'ucc.py' file initialization to include the option of creating the imaginary part of the ucc wavefunction with twice the number of free parameters as described below

Screen Shot 2021-12-07 at 6 29 49 PM
3. Modified the '_build_fermionic_excitation_ops()' in 'ucc.py' file to incorporate the additions described above

### DIFF
--- a/qiskit_nature/circuit/library/ansatzes/ucc.py
+++ b/qiskit_nature/circuit/library/ansatzes/ucc.py
@@ -515,7 +515,7 @@ class UCC(EvolvedOperatorAnsatz):
             
             # Appending the corresponding imaginary part
             if self._include_imaginary:
-                op_plus = op + op_adj
+                op_plus = -1*(op + op_adj)
                 operators.append(op_plus) 
 
         return operators
--- a/qiskit_nature/circuit/library/ansatzes/ucc.py
+++ b/qiskit_nature/circuit/library/ansatzes/ucc.py
@@ -172,9 +172,9 @@ class UCC(EvolvedOperatorAnsatz):
                 setting does _not_ influence the `excitations`. When relying on the default
                 generation method (i.e. not providing a `Callable` to `excitations`), these will
                 always be constructed with respect to a `HartreeFock` reference state.
-            include_imaginary: boolean flag which when set to 'True' expands the ansatz to include 
-                               imaginary parts using twice the number of free parameters. Additional operators 
-                               are appended by the '_build_fermionic_excitation_ops()' 
+            include_imaginary: boolean flag which when set to 'True' expands the ansatz to include
+                imaginary parts using twice the number of free parameters. Additional operators
+                are appended by the '_build_fermionic_excitation_ops()'
         """
         self._qubit_converter = qubit_converter
         self._num_particles = num_particles
@@ -505,17 +505,17 @@ class UCC(EvolvedOperatorAnsatz):
                 label[occ] = "+"
             for unocc in exc[1]:
                 label[unocc] = "-"
-            # Real part of the wave-function 
+            # Real part of the wave-function
             op = FermionicOp("".join(label), display_format="dense")
             op_adj = op.adjoint()
-            op_minus = 1j*(op - op_adj)
+            op_minus = 1j * (op - op_adj)
             # we need to account for an additional imaginary phase in the exponent hence multiplied by 1j (see also
             # `PauliTrotterEvolution.convert`)
             operators.append(op_minus)
-            
+
             # Appending the corresponding imaginary part
             if self._include_imaginary:
-                op_plus = -1*(op + op_adj)
-                operators.append(op_plus) 
+                op_plus = -1 * (op + op_adj)
+                operators.append(op_plus)
 
         return operators
--- a/qiskit_nature/circuit/library/ansatzes/ucc.py
+++ b/qiskit_nature/circuit/library/ansatzes/ucc.py
@@ -173,8 +173,7 @@ class UCC(EvolvedOperatorAnsatz):
                 generation method (i.e. not providing a `Callable` to `excitations`), these will
                 always be constructed with respect to a `HartreeFock` reference state.
             include_imaginary: boolean flag which when set to 'True' expands the ansatz to include
-                imaginary parts using twice the number of free parameters. Additional operators
-                are appended by the '_build_fermionic_excitation_ops()'
+                imaginary parts using twice the number of free parameters.
         """
         self._qubit_converter = qubit_converter
         self._num_particles = num_particles
--- a/qiskit_nature/circuit/library/ansatzes/ucc.py
+++ b/qiskit_nature/circuit/library/ansatzes/ucc.py
@@ -129,7 +129,8 @@ class UCC(EvolvedOperatorAnsatz):
         generalized: bool = False,
         preserve_spin: bool = True,
         reps: int = 1,
-        initial_state: QuantumCircuit | None = None,
+        initial_state: Optional[QuantumCircuit] = None,
+        include_imaginary: bool = False,
     ):
         """
 
@@ -171,6 +172,9 @@ class UCC(EvolvedOperatorAnsatz):
                 setting does _not_ influence the `excitations`. When relying on the default
                 generation method (i.e. not providing a `Callable` to `excitations`), these will
                 always be constructed with respect to a `HartreeFock` reference state.
+            include_imaginary: boolean flag which when set to 'True' expands the ansatz to include 
+                               imaginary parts using twice the number of free parameters. Additional operators 
+                               are appended by the '_build_fermionic_excitation_ops()' 
         """
         self._qubit_converter = qubit_converter
         self._num_particles = num_particles
@@ -181,6 +185,7 @@ class UCC(EvolvedOperatorAnsatz):
         self._max_spin_excitation = max_spin_excitation
         self._generalized = generalized
         self._preserve_spin = preserve_spin
+        self._include_imaginary = include_imaginary
 
         super().__init__(reps=reps, evolution=PauliTrotterEvolution(), initial_state=initial_state)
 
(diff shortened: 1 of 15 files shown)


qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -414,11 +419,17 @@ def _build_fermionic_excitation_ops(self, excitations: Sequence) -> List[Fermion
...
+            op_minus = 1j*(op - op_adj)
+            # we need to account for an additional imaginary phase in the exponent hence multiplied by 1j (see also
             # `PauliTrotterEvolution.convert`)
-            op *= 1j  # type: ignore
-            operators.append(op)
+            operators.append(op_minus)
+            
+            # Appending the corresponding imaginary part
+            if self._include_imaginary:
+                op_plus = op + op_adj
```
tanvipg, line 432: Suggestion: `op_plus = -1*(op+op_adj)`

While the -1 will get incorporated in parameter optimization, the negative sign is consistent with `PauliTrotterEvolution` class

pbark: fixed in 21a3456

a-matsuo:
I think unittests are required for this new feature since the feature is totally new.

qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -167,10 +168,14 @@ def __init__(
                 particles.
             preserve_spin: boolean flag whether or not to preserve the particle spins.
             reps: The number of times to repeat the evolved operators.
+
```
pbark, line 171:
```suggestion
```

pbark (changes requested):

qiskit_nature/properties/second_quantization/electronic/integrals/two_body_electronic_integrals.py
```diff
@@ -164,7 +164,7 @@ def to_spin(self) -> np.ndarray:
         if self._basis == ElectronicBasis.SO:
             return self._matrices  # type: ignore
 
-        so_matrix = np.zeros([2 * s for s in self._matrices[0].shape])
+        so_matrix = np.zeros([2 * s for s in self._matrices[0].shape],dtype = 'complex')
```
woodsp-ibm, line 167: This change to complex will end up consuming a bunch more memory (double) since commonly we have real values for these integrals I think. I am also trying to figure out how this change factors in since the UCC operators are built directly as FermionicOps and not via integrals right.

qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -171,6 +172,9 @@ def __init__(
                 setting does _not_ influence the `excitations`. When relying on the default
                 generation method (i.e. not providing a `Callable` to `excitations`), these will
                 always be constructed with respect to a `HartreeFock` reference state.
+            include_imaginary: boolean flag which when set to 'True' expands the ansatz to include
+                imaginary parts using twice the number of free parameters. Additional operators
+                are appended by the '_build_fermionic_excitation_ops()'
```
mrossinek, line 177:
```suggestion
            include_imaginary: boolean flag which when set to 'True' expands the ansatz to include
                imaginary parts using twice the number of free parameters.
```
This part is not relevant since this is non-public API.

I am also wondering whether we could call this, `include_complex_conjugate` which is a better description, imo

qiskit_nature/circuit/library/ansatzes/uccsd.py
```diff
@@ -35,6 +35,7 @@ def __init__(
         initial_state: Optional[QuantumCircuit] = None,
         generalized: bool = False,
         preserve_spin: bool = True,
+        include_imaginary: bool = False,
```
mrossinek, line 38: The same setting should also be exposed in the other subclasses of `UCC`

qiskit_nature/properties/second_quantization/electronic/integrals/two_body_electronic_integrals.py
```diff
@@ -164,7 +164,7 @@ def to_spin(self) -> np.ndarray:
         if self._basis == ElectronicBasis.SO:
             return self._matrices  # type: ignore
 
-        so_matrix = np.zeros([2 * s for s in self._matrices[0].shape])
+        so_matrix = np.zeros([2 * s for s in self._matrices[0].shape], dtype="complex")
```
mrossinek, line 167: I agree with @woodsp-ibm's original comment, that this is duplicating the memory requirements unnecessarily in most cases.
Thus, this should be a configurable setting.

We can either:
1. add a setting in `qiskit_nature.settings`
2. add an `__init__` argument to these integrals (since the `OneBody` case should probably deal with this, too) which would then also need to be propagated down to the `ElectronicEnergy`

I think, I would prefer the former option.

@pbark any thoughts on this?

test/circuit/library/ansatzes/test_ucc.py
```diff
@@ -19,14 +19,15 @@
...
-from qiskit_nature.mappers.second_quantization import JordanWignerMapper, ParityMapper
+from qiskit_nature.mappers.second_quantization import JordanWignerMapper
 from qiskit_nature.operators.second_quantization import FermionicOp
 
 
 def assert_ucc_like_ansatz(test_case, ansatz, num_spin_orbitals, expected_ops):
     """Assertion utility."""
-    excitation_ops = ansatz.excitation_ops()
 
+    # excitation_ops = ansatz.excitation_ops()[::2] if include_imaginary else ansatz.excitation_ops()
```
mrossinek, line 29: Is the code that was commented here not needed anymore? If yes, the comment ought to be removed

test/circuit/library/ansatzes/test_ucc.py
```diff
@@ -133,88 +253,3 @@ def custom_excitations(num_spin_orbitals, num_particles):
...
-            ucc.excitations = "s"
-            self.assertIsNotNone(ucc.operators)
-            self.assertEqual(len(ucc.operators), 4)
-
-        with self.subTest("Change qubit converter"):
-            ucc.qubit_converter = QubitConverter(ParityMapper(), two_qubit_reduction=True)
-            # Has not been used to convert so we need to force it to do two qubit reduction
-            ucc.qubit_converter.force_match(ucc.num_particles)
-            self.assertIsNotNone(ucc.operators)
-            self.assertEqual(ucc.num_qubits, 4)
```
mrossinek, line 220: I don't think these tests were planned to be removed. Looks like a bad merge since these were only added back in March

mrossinek (changes requested):

releasenotes/notes/ucc_imaginary-8d7d9c3322899a34.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Imaginary contributions to the UCC ansatz can be added along with the real contributions by 
+    setting the boolean flag 'include_imaginary' to True. A separate parameter is used for the real 
+    and the imaginary contribution coming from a single excitation
```
woodsp-ibm, line 6: @mrossinek @declanmillar So setting imaginary here doubles up the number of parameters. How does this impact/affect the recent InitialPoint work that would set the starting point value for these?

qiskit_nature/properties/second_quantization/electronic/integrals/two_body_electronic_integrals.py
mrossinek: Thanks for the reply. Is it needed at all for this PR to work as intended though? If not, I would like to leave this change out and consider this scenario as part of #665 instead

releasenotes/notes/ucc_imaginary-8d7d9c3322899a34.yaml
declanmillar: @mrossinek, @woodsp-ibm, @tanvipg, I've opened a PR [here](https://github.com/Qiskit/qiskit-nature/pull/455) that should patch this behavior. Let me know if you spot any issues with it.

mrossinek: Thanks a lot, @declanmillar! The patch looks good, thanks for also including a unittest!
I pinged Tanvi on Slack, too, in order to hopefully get your PR merged asap :+1:

qiskit_nature/properties/second_quantization/electronic/integrals/two_body_electronic_integrals.py
mrossinek:
```suggestion
        so_matrix = np.zeros([2 * s for s in self._matrices[0].shape])
```
After another offline check with Panos, we decided to keep this change out of this PR and instead introduce this as a configurable option as part of the operator-overhaul (#665)

mrossinek (approved):
This LGTM now (assuming CI passes). I'll await another review by @pbark or @woodsp-ibm before merging :+1:

releasenotes/notes/ucc_imaginary-8d7d9c3322899a34.yaml
woodsp-ibm, line 6: I think we can make things clearer here following the first sentence. While we know parameter here means parameter of the UCC ansatz circuit it could be talking about a method parameter here as it seems a bit vague to me. How about the second sentence being replaced by something like the following which also spells out what happens when you set it  i.e. doubles the number of parameters. as that can have implications for optimization.

`When True an extra ansatz parameter is added to each excitation that controls the imaginary contribution to its evolution. This setting then doubles the total number of ansatz parameters, as compared to the default setting of False.`

mrossinek: Thanks, I updated the release note with your suggestion :+1: 

With regards to the example, this would be something for @pbark or @tanvipg to answer

mrossinek:
With complex integrals no longer being an issue in the stack (see also #1144), I have updated this PR to the latest changes.
I am not confident that the proposed inclusion of this feature in the SUCCD case will work as intended since this was applied to the code version prior to the addition of the `mirror` feature in #729. Thus, I refrained from exposing this keyword argument in that specific subclass in this PR.

woodsp-ibm (approved):

PR #457, opened 2021-12-10
fix issue 398 - BOPESSampler does not support MinimumEigensolverFactories

### LINKED ISSUE #398: `BOPESSampler` does not support `MinimumEigensolverFactories`
### Information

- **Python version**: 3.9.7
- **Operating system**: Linux

```python
  File "/home/oss/Files/Qiskit/src/qiskit-nature/main/tmp10.py", line 32, in <module>
    bs = BOPESSampler(VQEUCCFactory)
  File "/home/oss/Files/Qiskit/src/qiskit-nature/main/qiskit_nature/algorithms/pes_samplers/bopes_sampler.py", line 95, in __init__
    if isinstance(self._gss.solver, VariationalAlgorithm):  # type: ignore
AttributeError: type object 'VQEUCCFactory' has no attribute 'solver'
```

```python
from functools import partial

import numpy as np

from qiskit_nature.algorithms.ground_state_solvers.minimum_eigensolver_factories import (
    VQEUCCFactory,
)
from qiskit_nature.algorithms.pes_samplers import BOPESSampler
)

stretch = partial(Molecule.absolute_distance, atom_pair=(0, 1))

molecule = Molecule(
    geometry=[
        ["H", [0.0, 0.0, 0.0]],
        ["H", [0.0, 0.0, 1.0]],
    ],
    degrees_of_freedom=[stretch],
)
driver = ElectronicStructureMoleculeDriver(molecule, "sto3g", ElectronicStructureDriverType.PYSCF)

converter = QubitConverter(ParityMapper())

bs = BOPESSampler(VQEUCCFactory)

res = bs.sample(problem, np.linspace(0.5, 3.0, 20))
print(res)
```

The `BOPESSampler` should support `MininmumEigensolverFactory` classes.

### Suggested solutions

Add a check similar to the one in the `GroundStateEigensolver`: https://github.com/Qiskit/qiskit-nature/blob/main/qiskit_nature/algorithms/ground_state_solvers/ground_state_eigensolver.py#L103-L105

iuliazidaru:
### Summary

Fixes #398 

### Details and comments
Added support for MinimumEigensolverFactory in BOPESSampler.


qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -24,29 +24,63 @@
...
 
     def __init__(
         self,
-        gss: GroundStateSolver,
+        gss: Union[GroundStateSolver, MinimumEigensolverFactory],
         tolerance: float = 1e-3,
         bootstrap: bool = True,
         num_bootstrap: Optional[int] = None,
         extrapolator: Optional[Extrapolator] = None,
+        qubit_converter: Optional[QubitConverter] = None
```
iuliazidaru, line 79: QubitConverter is needed to initialize solver from MinimumEigenSolverFactory.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -74,6 +108,11 @@ def __init__(
         self._points_optparams: Dict[float, List[float]] = None
         self._num_bootstrap = num_bootstrap
         self._extrapolator = extrapolator
+        self._solver = None
+        if isinstance(self._gss, GroundStateSolver):
+            self._solver = self._gss.solver
```
iuliazidaru, line 113: Saved solver in a member in order to have the same code for both implementations (GroundStateSolver and MinimalEigensolverFactory). 
GroundStateSolver has the solver in the instance, MinimalEigensolverFactory produces a solver.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -92,10 +131,12 @@ def __init__(
                     "num_bootstrap must be None or an integer greater than or equal to 2"
                 )
 
-        if isinstance(self._gss.solver, VariationalAlgorithm):  # type: ignore
+        if isinstance(self._gss, GroundStateSolver) and \
```
iuliazidaru, line 134: The solver from MinimalEigensolverFactory is not initialized at this point, as the problem is needed in the argument.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -132,6 +173,14 @@ def sample(self, problem: BaseProblem, points: List[float]) -> BOPESSamplerResul
         if self._driver.molecule is None:
             raise QiskitNatureError("Driver MUST be configured with a Molecule.")
 
+        # force the initialization of the solver before starting the computation
+        if isinstance(self._gss, MinimumEigensolverFactory):
+            if self._qubit_converter is None:
+                raise QiskitNatureError("When using MinimumEigensolverFactory "
+                                        "you must specify a qubit converter")
+            prepare_problem(problem, self._qubit_converter)
+            self._solver = self._gss.get_solver(problem,self._qubit_converter)
```
iuliazidaru, line 182: Initialize solver before the tricks on initial_point. We have to apply second quantisation on the problem to obtain the solver.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -217,20 +266,28 @@ def _run_single_point(self, point: float) -> EigenstateResult:
...
-        result = self._gss.solve(self._problem)
+        if isinstance(self._gss, GroundStateSolver):
+            result = self._gss.solve(self._problem)
+        else:
+            # computation done in _gss is expanded here
+            _main_operator = prepare_problem(self._problem, self._qubit_converter)
+            self._solver = self._gss.get_solver(self._problem, self._qubit_converter)
+
+            raw_mes_result = self._solver.compute_minimum_eigenvalue(_main_operator)
+            result = self._problem.interpret(raw_mes_result)
```
iuliazidaru, line 287: The computation performed in GroundStateSolver has to be done here.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -24,29 +24,62 @@
...
         tolerance: float = 1e-3,
         bootstrap: bool = True,
         num_bootstrap: Optional[int] = None,
         extrapolator: Optional[Extrapolator] = None,
+        qubit_converter: Optional[QubitConverter] = None,
     ) -> None:
         """
         Args:
-            gss: GroundStateSolver
+            gss: GroundStateSolver or MinimumEigensolverFactory
```
mrossinek, line 82: I suggest you mention that in the latter case a `QubitConverter` is required already here.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -132,6 +173,15 @@ def sample(self, problem: BaseProblem, points: List[float]) -> BOPESSamplerResul
         if self._driver.molecule is None:
             raise QiskitNatureError("Driver MUST be configured with a Molecule.")
 
+        # force the initialization of the solver before starting the computation
+        if isinstance(self._gss, MinimumEigensolverFactory):
+            if self._qubit_converter is None:
+                raise QiskitNatureError(
+                    "When using MinimumEigensolverFactory " "you must specify a qubit converter"
```
mrossinek, line 180:
```suggestion
                    "When using MinimumEigensolverFactory you must specify a qubit converter"
```

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -132,6 +173,15 @@ def sample(self, problem: BaseProblem, points: List[float]) -> BOPESSamplerResul
...
+                )
+            prepare_problem(problem, self._qubit_converter)
```
mrossinek, line 182: `prepare_problem` returns the mapped `main_operator` but you're not using that here.

EDIT: I see that this is not necessary but maybe add a comment why.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -132,6 +173,15 @@ def sample(self, problem: BaseProblem, points: List[float]) -> BOPESSamplerResul
         if self._driver.molecule is None:
             raise QiskitNatureError("Driver MUST be configured with a Molecule.")
 
+        # force the initialization of the solver before starting the computation
```
mrossinek, line 176: Why do we need to force this initialization?

mrossinek (commented):
Thanks @iuliazidaru for your work on this! I have some comments but in general this is already looking good :+1:

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
iuliazidaru: In this case we don't need the main_operator. Prepare_problem is called because we need a problem with second quantisation applied in order to create the solver. The main_operator is used later, when the real calculations are performed. 

I must admit that this is not as clean as it should be and I did 3 implementations to get to this solution. The problem is that MinimumEigensolverFactory produces the solver based on a problem with second quantisation applied and GroundStateSolver is a wrapper which does the second quantisation and calculates the ground state on a raw problem(in the same method). Bopes sampler adds some parameters to the solver GroundStateSolver has, but in case of MinimulEigensolverFactory the creation of the solver must be delayed... So, the row above force the second quantisation on the problem in order to create a solver to gather the parameters that bopes sampler calculates.

One cleaner solution would have been to create a GroundStateSolver instance based on the MinimumEigensolverFactory and keep all the code. This solution didn't work as it introduces cyclic dependencies.

iuliazidaru:
@mrossinek Thank you for your review. I fixed the code. Sorry for the delay, I missed the notifications for the review..

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
mrossinek: I actually meant that you add a short comment in the code such that, when somebody else looks at the code in the far future, the comment helps to understand why it is neccessary.

mrossinek (commented):
I left a comment on our conversation above.

Besides that, we will also need a release note for this PR. You can add one by running
```
reno new bopessampler-eigensolver-factory-support
```
from a terminal while inside your local Qiskit Nature folder. This will output the path to a new release note file which you can edit. I suggest that you describe the changes made by this PR in one or two sentences of the `feature` section. The other sections of the auto-generated file can be removed.

mrossinek (approved):
This looks good to me now, thanks @iuliazidaru for the contribution!
In terms of performance it is a bit unfortunate that we need to run the driver twice when using a factory but I don't see a good solution for that.

Let's await another review from @pbark or @woodsp-ibm. Maybe they have some more inputs.

iuliazidaru:
@mrossinek Thank you for your review!

dlasecki (changes requested):

mrossinek (changes requested):
I'm retracting my previous approval as explained here: https://github.com/Qiskit/qiskit-nature/issues/398#issuecomment-1013157196

mrossinek:
Closing this in favor of #510

PR #458, opened 2021-12-10
Bump black to the latest version

manoelmarques:
### Summary
Updates black with `black[jupyter]==21.12b0` following Qiskit/qiskit-terra#7363 and applies black to docs too.


mrossinek (approved):
Thanks @manoelmarques!

Issue #459, opened 2021-12-10
Preparation of Slater determinants and fermionic Gaussian states


kevinsung:
### What should we add?

Slater determinants and fermionic Gaussian states are important classes of initial states for quantum simulations which can be efficiently prepared on a quantum computer [1]. Qiskit Nature should include routines to generate circuits to prepare these states. Efficient algorithms to prepare these states are described in [arxiv:1711.05395](https://arxiv.org/abs/1711.05395), and these are already implemented in [OpenFermion](https://github.com/quantumlib/OpenFermion). I can contribute these routines to Qiskit Nature. Note that these algorithms assume the Jordan-Wigner Transformation is used to map fermions to qubits.

### Implementation details
Qiskit Nature already has routines in the `qiskit_nature.circuit.library` module to prepare initial states, such as the `HartreeFock` class which prepares computational basis states. The new functionality should be implemented by adding new classes `SlaterDeterminant` and `FermionicGaussianState` to this module. My current plan for these classes is to have them called like this:
```python
SlaterDeterminant(transformation_matrix)

FermionicGaussianState(transformation_matrix, occupied_orbitals)
```

In both cases, each row of `transformation_matrix` contains the coefficients of a transformed creation operator in terms of the original creation (and possibly annihilation) operators.

For `SlaterDeterminant`, the input `transformation_matrix` is an `η x N` matrix where N is the number of orbitals (qubits) and η is the number of particles.

For `FermionicGaussianState`, the input `transformation_matrix` is is either `N x N` or `N x 2N`, the latter case corresponding to when both creation and annihilation operators are mixed together, rather than only creation operators. The argument `occupied orbitals` is a list of integers indicating the indices of the transformed orbitals that should be occupied.

A common use case for these routines is preparing eigenstates of quadratic Hamiltonians. In addition to these routines, we will want to add functionality for representing quadratic Hamltonians and computing the transformation matrices from them, but that can be done in a separate issue.

---
[1] https://arxiv.org/abs/cond-mat/0012334

PR #462, opened 2021-12-15
Add RYXXY gate for YX - XY interaction

kevinsung:
### Summary
This gate is used in the algorithms for preparing Slater determinants and fermionic Gaussian states discussed at #459 .

### Details and comments
I created a `gates` module under `circuit.library` as a place to put definitions of gates needed in quantum simulation algorithms.


kevinsung:
The spell check is failing. It's hard for me to debug because it fails for me locally on main too; see #463 .

mrossinek (commented):
Thanks @kevinsung for your contribution! I am not an expert on this topic so I suggest that someone else verifies the definition of this gate. I also don't really have any comments on the code, all is looking good!
For the debugging of the spelling/lint errors, you can also switch to the `Files changed` tab of this PR. There you should see auto-generated comments from Github which look like this:
screenshot_1639726576

The spelling ones are not perfect because the linter does not provide the exact line number for these errors. But there is another lint warning which is positioned correctly.

kevinsung:
@mrossinek Who do you think should verify the definition of the gate? This is not a textbook gate so I'm not sure what they would verify it against.

kevinsung:
`Ryxxy(θ) = (T^-1 ⊗ T) iSwap^(2·θ/π) (T ⊗ T^-1)` so if Qiskit's iSwapGate supported exponentiation then I could use that instead.

woodsp-ibm:
Maybe @Cryoris can comment on the implementation since he has a bunch of experience with gates, and also if there is any consideration needed in regards of the Qiskit Runtime and using this gate.

kevinsung:
I just noticed that the documentation for this gate does not show up when I build locally; what do I need to do for that?

kevinsung:
Never mind, I figured it out.

Cryoris:
The gate structure looks good to me, but I have a few questions:

1. General 2 qubit gates can be decomposed with 3 CX, but this uses 4 CX. Is there a more efficient decomposition we could use?
2. If we'd like to make this decomposition generally available for compilation we could implement it in Terra. This would include e.g. translating circuit _into_ `exp(i t (XY - YX))` gates, or generally using this exact decomposition instead of Trotterization if users want to evolve the operator `XY - YX`.
3. Our naming scheme currently is `R{pauli rotations on qubit}` so the name `RYXXY` would mean a 4 qubit rotation. Could we give this another name that fits the naming (possibly difficult), or something chemistry specific?

kevinsung:
@Cryoris thanks for the comments.
>General 2 qubit gates can be decomposed with 3 CX, but this uses 4 CX. Is there a more efficient decomposition we could use?

I will try to come up with a better decomposition. Does that need to block this PR though? What gates are allowed in the decomposition? Is that documented somewhere?
>If we'd like to make this decomposition generally available for compilation we could implement it in Terra. This would include e.g. translating circuit _into_ `exp(i t (XY - YX))` gates, or generally using this exact decomposition instead of Trotterization if users want to evolve the operator `XY - YX`.

This is a generally useful gate, so it makes sense to add it to Terra instead of here, if that's what you mean. I'm happy to move this PR to Terra. As I mentioned in a comment above, this gate is closely related to iSwap so an alternative way to implement this gate is to implement exponentiation for iSwap (a useful feature in itself) and then use that to implement this gate.

>Our naming scheme currently is `R{pauli rotations on qubit}` so the name `RYXXY` would mean a 4 qubit rotation. Could we give this another name that fits the naming (possibly difficult), or something chemistry specific?

I've changed the name to `YXMinusXYInteractionGate`.

Cryoris:
Thanks for the changes @kevinsung! 

> However, I'd still like to know in general what gates are allowed in the decomposition.

The goal of the decomposition is to tell the transpiler how to implement your new gate. Therefore you can generally use any gate that Qiskit knows already. However, you have to keep in mind that we often don't want the simplest definition but the most efficient one. In that case it usually makes sense to use standard gates (like H RZ CX ...) that are easy to decompose for hardware.

Cryoris:
We can discuss the incorporation into Terra in Qiskit/qiskit-terra#7470.

qiskit_nature/circuit/library/gates/yx_minus_xy_interaction.py
```diff
@@ -0,0 +1,146 @@
...
+                \end{pmatrix}
+    """
+
+    def __init__(self, theta: ParameterValueType, label: Optional[str] = None):
+        """Create new YXMinusXYInteractionGate gate."""
+        super().__init__("ryxxy", 2, [theta], label=label)
+
+    def __array__(self, dtype=None):
+        """Return a numpy array for the YXMinusXYInteractionGate gate."""
+        theta = float(self.params[0])
```
dlasecki, line 112: Not familiar with the topic but is it guaranteed not to be a complex number?

test/circuit/library/gates/test_yx_minus_xy_interaction.py
```diff
@@ -0,0 +1,66 @@
...
+
+from qiskit import QuantumCircuit
+from qiskit.quantum_info import Operator
+from qiskit_nature.circuit.library import YXMinusXYInteractionGate
+
+
+class TestYXMinusXYInteractionGate(QiskitNatureTestCase):
+    """Tests for YXMinusXYInteractionGate gate"""
+
+    def test_matrix(self):
```
dlasecki, line 26: How about using ddt?

qiskit_nature/circuit/library/gates/yx_minus_xy_interaction.py
kevinsung: Yes thanks, fixed.

test/circuit/library/gates/test_yx_minus_xy_interaction.py
kevinsung: Done.

kevinsung:
I learned that `XYGate` has been added to Qiskit, so this gate is no longer needed.

PR #467, opened 2021-12-17
Initialize ``Lattice``s from ``networkx.Graph``s

### LINKED ISSUE #465: Initialize ``Lattice``s with ``networkx.Graph``s
### What should we add?

It would be nice to allow the initialization on ``Lattice``s not only with retworkx graphs but also with networks graphs. Networkx, to date, still has a lot more features and pre-defined graphs that users have access to, and is probably _the_ standard graph library in Python.

### LINKED ISSUE #466: Type check edge weights in ``Lattice`` constructions?

Right now, a ``Lattice`` can be constructed from a graph with any value for the edge weights, e.g.
```python
>>> from qiskit_nature.problems.second_quantization.lattice import Lattice
>>> import retworkx as rx
>>> graph = rx.PyGraph(multigraph=False)
>>> graph.add_nodes_from(range(2))
>>> graph.add_edges_from([(0, 1, "dilution fridge")])
>>> Lattice(graph)
```
In the above case, the construction of a second quantized op or a qubit op based on the lattice will obviously fail. But there might be more subtle cases.

It would be nice to type check the weights and ensure that they are numeric (or in some supported format). Or is there a reason to support arbitrary types as weights?

Cryoris:
### Summary

Fixes #465: Support the initialization of lattices from networkx graphs. 
Fixes #466: Allow only numerical values for edge weights.

### Details and comments

Now we can for instance construct lattices as
```python
import networkx as nx
from qiskit_nature.problems.second_quantization.lattice import Lattice
# 3-regular random graph on 6 nodes
graph = nx.generators.random_graphs.random_regular_graph(3, n=6)
lattice = Lattice(graph)
```

### DIFF
--- a/qiskit_nature/problems/second_quantization/lattice/lattices/lattice.py
+++ b/qiskit_nature/problems/second_quantization/lattice/lattices/lattice.py
@@ -15,11 +15,10 @@
 from copy import deepcopy
 from dataclasses import asdict, dataclass
 from typing import Callable, List, Optional, Sequence, Tuple, Union
+import numbers
 
 import numpy as np
 import networkx as nx
-import numbers
-
 from retworkx import NodeIndices, PyGraph, WeightedEdgeList, adjacency_matrix, networkx_converter
 from retworkx.visualization import mpl_draw
 
--- a/qiskit_nature/problems/second_quantization/lattice/lattices/lattice.py
+++ b/qiskit_nature/problems/second_quantization/lattice/lattices/lattice.py
@@ -18,7 +18,11 @@ from typing import Callable, List, Optional, Sequence, Tuple, Union
 import numbers
 
 import numpy as np
-import networkx as nx
+try:
+    import networkx as nx
+    HAS_NX = True
+except ImportError:
+    HAS_NX = False
 from retworkx import NodeIndices, PyGraph, WeightedEdgeList, adjacency_matrix, networkx_converter
 from retworkx.visualization import mpl_draw
 
@@ -121,7 +125,7 @@ class Lattice:
             ValueError: If the input graph is a multigraph.
             ValueError: If the graph edges are non-numeric.
         """
-        if isinstance(graph, nx.Graph):
+        if HAS_NX and isinstance(graph, nx.Graph):
             graph = networkx_converter(graph)
 
         if graph.multigraph:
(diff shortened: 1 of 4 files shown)


qiskit_nature/problems/second_quantization/lattice/lattices/lattice.py
```diff
@@ -107,23 +110,34 @@ class LatticeDrawStyle:
...
-            for start, end, weight in weighted_edges:
-                graph.update_edge(start, end, weight)
+
+        # validate the edge weights
+        for edge_index, weight in enumerate(graph.edges()):
+            if weight is None or weight == {}:
+                # None or {} is updated to be 1
+                graph.update_edge_by_index(edge_index, 1)
+            elif not isinstance(weight, numbers.Number):
+                raise ValueError(f"Unsupported weight {weight} on edge with index {edge_index}.")
```
mtreinish, line 139: This won't work if the graph has edges removed. In retworkx the edge indices (or node indices) aren't guaranteed to be contiguous as removals will leave holes in the index list (see the description here: https://qiskit.org/documentation/retworkx/stubs/retworkx.PyGraph.html#retworkx.PyGraph ). So if someone passes in a graph with a removed node this will be updating the wrong edge. Instead I would change this to:

```suggestion
        for edge_index, weight in graph.edge_index_map().items():
            if weight is None or weight == {}:
                # None or {} is updated to be 1
                graph.update_edge_by_index(edge_index, 1)
            elif not isinstance(weight, numbers.Number):
                raise ValueError(f"Unsupported weight {weight} on edge with index {edge_index}.")
```

Cryoris: Thanks! I've also added a test to cover this case.

releasenotes/notes/lattice-from-networkx-20c7c8119af77f36.yaml
```diff
@@ -0,0 +1,17 @@
+---
+features:
```
mrossinek, line 2: I'm wondering whether we should split out the edge type validation fix into another PR which could become part of a `0.3.1` bugfix release. If it get's included in here we will need to wait for the `0.4.0` feature release.

Cryoris: It's probably not very urgent since it might be expected that putting some random values on the edges breaks somewhere down the stack. But of course it's nice if it breaks early and with a clear warning, I can separate it out if you like!

mrossinek: As you see fit. I'm just seeing now that the actual comment I wanted to write got somehow lost :thinking: 
Cause I wanted to say that this fix should be listed separately in the reno :sweat_smile:

Cryoris:
Closing in favor of splitting this PR in two: the edge type check #471 and the networkx support #472.

PR #478, opened 2021-12-21
Improve problem attribute access errors

mrossinek:
### Summary

A question came up on Slack (https://qiskit.slack.com/archives/CB6C24TPB/p1640100013370800) why accessing `num_particles` and `num_spin_orbitals` was not possible. After a quick offline discussion with @woodsp-ibm, we decided to at least add some useful error messages and improve the documentation to prevent a scenario like this.

### Details and comments

I am not sure whether a basic fix as this one needs a release note...?

### DIFF
--- a/qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
+++ b/qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
@@ -139,7 +139,7 @@ class ElectronicStructureProblem(BaseProblem):
         """Generates the hopping operators and their commutativity information for the specified set
         of excitations.
 
-        You cannot call this method before calling `second_q_ops()`.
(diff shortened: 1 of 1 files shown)


qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
```diff
@@ -124,6 +139,8 @@ def hopping_qeom_ops(
         """Generates the hopping operators and their commutativity information for the specified set
         of excitations.
 
+        You cannot call this method before calling `second_q_ops()`.
```
pbark, line 142: I think it would be better if we turn it to 
```suggestion
        This method should can be used after calling `second_q_ops()`.
```

pbark (approved):

pbark (approved):
LGTM!

PR #481, opened 2021-12-22
Added support for Bohr unit type in molecule

### LINKED ISSUE #477: Support `UnitsType.BOHR` in `Molecule`

Currently the `Molecule` class requires coordinates in Angstrom and is not capable of handling other unit types by itself.
In fact, it does not even take a `units` argument to its constructor based on which one could guard against unit-related errors such as the one fixed by #476.

Thus, the following things can be improved:
- add a `unit` (or `units` ?) argument to `Molecule.__init__`
- interpret said argument and always store in Angstrom internally (to make things consistent)
- take the original `unit` argument into account in `Molecule.__str__` and print the coordinates in the format used during construction to avoid confusion of the user

Ryand1234:
fixes #477 
### Summary
Added support for Bohr unit type in Molecule Class


qiskit_nature/drivers/molecule.py
```diff
@@ -39,6 +39,7 @@ def __init__(
         geometry: List[Tuple[str, List[float]]],
         multiplicity: int = 1,
         charge: int = 0,
+        unit: str = "Angstrom",
```
mrossinek, line 42: I suggest you use the `UnitsType` enum which is being imported anyways.

mrossinek (changes requested):

qiskit_nature/drivers/molecule.py
Ryand1234: I tried using UnitsType.ANGSTROM but its not resolving to Angstrom
```
print(UnitsType.ANGSTROM)
UnitsType.ANGSTROM
```
I am getting this.

Ryand1234:
@mrossinek can you tell me which test cases are failing. I don't understand what exactly is the error

mrossinek:
You need to run `make black` because the formatter is failing.
I'm not sure about the tutorial case. Will check in the new year if it continues to fail.

woodsp-ibm:
When we use the Molecule, say for an ElectronicStructureMoleculeDriver, when the Molecule is converted/mapped/used to create an instance of the actual chemistry driver, then the units property is used during the conversion e.g. in PySCF in this method https://github.com/Qiskit/qiskit-nature/blob/e39594b8ede4533af7d22b2b121f29d624e569ed/qiskit_nature/drivers/second_quantization/pyscfd/pyscfdriver.py#L316 (each driver PyQuante, PSI4 etc has the same method - different implementation of course but uses the Molecule to create what the driver needs). Hence the driver will now see the units, via the property units and self._unit, and will see Bohr, and with geometry being converted to Angstrom the driver result will not be as expected. 

Also the Molecule's geometry property returns the geometry so that will be confusing seeing something different passed back. And we have the perturbations on Molecule that allow it to be stretched etc that needs consideration.

My thought is the the units could be passed/set just as they are now but the geometry left in these units. This way everything about the Molecule is consistent - it would all be either in Bohr or Angstrom with the print in the same units (no conversion there either) so everything is consistent. Pertubations would be in the same units when its specifying distance. I think this is in fact less of a change. The underlying drivers should just work if the Molecule returns Bohr instead of Angstrom via its units property - since this is passed over the real driver but was fixed to just the one value today. To ensure it works the unit test for the Molecule and the drivers based on it should be augmented to include geometries that are in Bohr (maybe choose the exact same molecule as its tests today, but switch to Bohr).

mrossinek:
I gave this some more thought and I agree with @woodsp-ibm.

In short: _no conversion of the coordinates should happen within in `Molecule`_

A driver therefore needs to check the unit return by `Molecule.unit` and apply a conversion themselves.
@Ryand1234 can you please change this PR accordingly?

qiskit_nature/drivers/molecule.py
```diff
@@ -62,7 +64,15 @@ def __init__(
         """
         Molecule._check_consistency(geometry, masses)
 
+        # if unit is Bohr, convert geometry to unit angstrom
+        if unit == UnitsType.BOHR:
+            n_geometry: List[Tuple[str, List[float]]] = []
+            for atom, xyz in geometry:
+                ang_xyz = [e * BOHR for e in xyz]
+                n_geometry.append((atom, ang_xyz))
+            geometry = n_geometry
```
mrossinek, line 73: As mentioned in my previous comment we should refrain from doing this conversion inside of `Molecule`. I do not believe that any of the code here in this class explicitly relies on the units being `Angstrom`.
Instead, when a driver gets passed a `Molecule`, it should deal with whatever unit the `Molecule` is in, by itself.

We could think about providing a small utility method on the class, e.g. `Molecule.angstrom_to_bohr` and vice versa for convenience. But I am not sure that it is really necessary for a computation as simple as this.

mrossinek (changes requested):
As mentioned before, if you change the PR accordingly, you will need to double check that all drivers can handle this as expected. Ideally you could add a unittest to all Molecule-based driver tests which uses a Bohr-based `Molecule` instance.

qiskit_nature/drivers/molecule.py
woodsp-ibm: And related to this would be the `__str__`  method where I think if you print the Molecule you would expect to see the same values you created it with. Maybe the method could/should display the units too in that case for completeness.

mrossinek: Ah yes I just picked the first occurrence to add this comment. Thanks for pointing this out :+1:

mrossinek:
Hi @Ryand1234, are you still interested in finishing this PR?

mrossinek:
@woodsp-ibm I did some slight refactoring as we had discussed. Please take a look what you think of it now

qiskit_nature/drivers/molecule.py
```diff
@@ -84,6 +90,40 @@ def __str__(self) -> str:
                 string += [f"\t\t{atom}\t{mass}"]
         return "\n".join(string)
 
+    @staticmethod
+    def convert_units(molecule: Molecule, unit: UnitsType) -> Molecule:
```
woodsp-ibm, line 94: I am wondering about the need/use of this method. This alters the geometry but if there are degrees of freedom setup with the molecule then these remain how they were. Would it be better/simpler, if this conversion is really needed/wanted to just return the base geometry in the supplied units and let the caller build out a new Molecule. I wonder in practice how much this would be used - for other driver inputs we let the user specify either way and its up to them to do the conversion themselves. This could be left that way so the user supplies the molecule in either bohr or angstrom and if they want to convert from one set of units to the other we let them do it themselves.

mrossinek: Hm yes I did not think about the degrees of freedom.. I guess the mathematics are pretty straight forward anyways..
Would you rather we remove this entire conversion capability then or would you still provide it but instead of returning `Molecule` it just returns the new geometry?

woodsp-ibm: My personal choice would be to remove it entirely. A user can work in whatever units they want and if they have something that's needs conversion it easy enough for them to do it - they have to do it for input to other drivers. If one really wanted to supply conversion functions one could do that in with the constants - not that I am proposing to do that since again its straightforward to do.

qiskit_nature/drivers/molecule.py
```diff
@@ -39,6 +41,7 @@ def __init__(
         geometry: List[Tuple[str, List[float]]],
         multiplicity: int = 1,
         charge: int = 0,
+        unit: UnitsType = UnitsType.ANGSTROM,
```
woodsp-ibm, line 44: It seems we are missing the `unit` in the docstring args

test/drivers/second_quantization/test_driver_molecule.py
```diff
@@ -55,6 +57,7 @@ def test_construct(self):
                 degrees_of_freedom=[stretch],
                 masses=[0.7, 0.8],
             )
+            self.assertEqual(mol.units, UnitsType.ANGSTROM)
```
woodsp-ibm, line 60: It seems that while Bohr was added there is nothing that tests this, say explicitly creating a molecule in Bohr units that this. Maybe one of these could be altered to Bohr and perhaps we could pass Angstom in explicitly too in one of these so it does not rely on default?.

In looking at the logic here we have the property `units` as per mol.units yet the constructor arg is `unit` - maybe since the latter has just been added `units` would be better for the constructor param to keep things consistent.

pbark (approved):

woodsp-ibm (approved):

Issue #460, opened 2021-12-13
Data structures for quadratic and other Hamiltonians


kevinsung:
### What should we add?

A quadratic Hamiltonian is composed of terms that are quadratic in the ladder operators:
![image](https://user-images.githubusercontent.com/4296166/145867381-4f14e16d-e312-4446-82ad-0eb1a860452d.png)
Quadratic Hamiltonians arise in mean-field models and their eigenstates are an important class of states that can be efficiently prepared on a quantum computer. The most convenient way to represent quadratic Hamiltonians is by storing the coefficients `M_pq` and `Δ_pq` as matrices because they need to be processed with linear algebra routines in order to get the eigenstates of the Hamiltonian. Therefore, the `FermionicOp` class is not a convenient way to represent them. This observation applies more generally; for example, the most convenient way to represent a molecular Hamiltonian is often to store the one-body and two-body tensors as numpy arrays. In OpenFermion, these numpy array representations are encompassed by the PolynomialTensor class, and the implementation in Qiskit Nature could use that as a model.

As a first step, I would like to add a `QuadraticHamiltonian` class that stores the matrices `M_pq` and `Δ_pq`. This class probably belongs in `qiskit_nature.operators`, but it's not clear whether it belongs in `qiskit_nature.operators.second_quantization` because it will not inherit from `SecondQuantizedOp`.

kevinsung:
During offline discussion with @mrossinek, we considered adding a class analogous to PolynomialTensor from OpenFermion that both QuadraticHamiltonian and the electronic integral classes would inherit from. Upon further investigation, I have found that while this is possible, some tradeoffs would need to be made.

PolynomialTensor stores the coefficients of creation and annihilation operators in a dictionary of tensors. Let's call the dictionary `tensors`. Then `tensors[(1, 0)]` would be an `N x N` matrix storing the coefficients of `+_i -_j` and `tensors[(1, 1, 0, 0)]` would be an `N x N x N x N` tensor storing the coefficients of `+_i +_j -_k -_l`. These are enough to represent the one- and two- body integrals. For QuadraticHamiltonian, we would also have `tensors[(1, 1)]` which stores the coefficients of `+_i +_j`. Note that the concept of "spin" is absent here, which is consistent with FermionicOp, and reasonable because these are supposed to be representations of general operators.

[...]

PR #482, opened 2021-12-22
Add QuadraticHamiltonian class

kevinsung:
### Summary
Part of #460. Adds a data structure to represent quadratic Hamiltonians.

### Details and comments
I have implemented a method `_fermionic_op` to convert a QuadraticHamiltonian to a FermionicOp. For now, I have kept that method hidden because I'm not sure what general approach we should take to converting between operator types, e.g., whether we want to have instance methods or whether we want to have a functional style like `qiskit_nature.fermionic_op(quadratic_hamiltonian)`


woodsp-ibm:
> The CI failure is unrelated to this PR.

Was just fixed so I updated your branch - should pass now

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,306 @@
...
+        self,
+        hermitian_part: Optional[np.ndarray] = None,
+        antisymmetric_part: Optional[np.ndarray] = None,
+        constant: float = 0.0,
+    ) -> None:
+        r"""Initialize a QuadraticHamiltonian.
+
+        Args:
+            hermitian_part: The matrix :math:`M` containing the
+                coefficients of the terms that conserve particle number.
```
mrossinek, line 47: I think this is line-wrapped too early.
Potentially more below.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,306 @@
...
+            ValueError: Either a Hermitian part or antisymmetric part must be specified.
+        """
+        if hermitian_part is not None:
+            self._n_orbitals, _ = hermitian_part.shape
+        elif antisymmetric_part is not None:
+            self._n_orbitals, _ = antisymmetric_part.shape
+        else:
+            raise ValueError("Either a Hermitian part or antisymmetric part must be specified.")
+
+        # TODO maybe validate matrix properties
```
mrossinek, line 62: Please address this

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,306 @@
...
+
+        Raises:
+            ValueError: Either a Hermitian part or antisymmetric part must be specified.
+        """
+        if hermitian_part is not None:
+            self._n_orbitals, _ = hermitian_part.shape
+        elif antisymmetric_part is not None:
+            self._n_orbitals, _ = antisymmetric_part.shape
+        else:
+            raise ValueError("Either a Hermitian part or antisymmetric part must be specified.")
```
mrossinek, line 60: Would it make sense to provide a keyword argument like `num_orbitals` to handle the creation of a zero-operator?

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,306 @@
...
+    def antisymmetric_part(self) -> np.ndarray:
+        """The matrix of coefficients of terms that do not conserve particle number."""
+        return self._antisymmetric_part
+
+    @property
+    def constant(self) -> float:
+        """The constant."""
+        return self._constant
+
+    def _fermionic_op(self) -> FermionicOp:
```
mrossinek, line 87: Why is this a private method?

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,306 @@
...
+        """Convert to FermionicOp."""
+        terms = []
+        # TODO I shouldn't have to use string labels
```
mrossinek, line 90: You can refactor this now that #506 is merged.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,306 @@
...
+        # diagonal terms
+        for i in range(self._n_orbitals):
+            terms.append((f"+_{i} -_{i}", self.hermitian_part[i, i]))
+        # off-diagonal terms
+        for i in range(self._n_orbitals):
+            for j in range(i + 1, self._n_orbitals):
```
mrossinek, line 96:
```suggestion
        for i in range(self._n_orbitals):
            # diagonal terms
            terms.append((f"+_{i} -_{i}", self.hermitian_part[i, i]))
            # off-diagonal terms
            for j in range(i + 1, self._n_orbitals):
```

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,306 @@
...
+                terms.append((f"+_{i} -_{j}", self.hermitian_part[i, j]))
+                terms.append((f"+_{j} -_{i}", self.hermitian_part[j, i]))
+                terms.append((f"+_{i} +_{j}", self.antisymmetric_part[i, j]))
+                terms.append((f"-_{j} -_{i}", self.antisymmetric_part[i, j].conj()))
+        # constant
+        terms.append(("", self.constant))
```
mrossinek, line 102: How about you initialize `terms = [("", self.constant)]` before entering the loop above?

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,306 @@
...
+            [
+                [self.antisymmetric_part, self.hermitian_part],
+                [-self.hermitian_part.conj(), -self.antisymmetric_part.conj()],
+            ]
+        )
+        eye = np.eye(self._n_orbitals, dtype=complex)
+        majorana_basis = np.block([[eye, eye], [1j * eye, -1j * eye]]) / np.sqrt(2)
+        matrix = -1j * majorana_basis.conj() @ original @ majorana_basis.T.conj()
+        constant = 0.5 * np.trace(self.hermitian_part) + self.constant
+        # imaginary parts should be zero
```
mrossinek, line 139: Do you want to assert this and raise an error if not?

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,306 @@
...
+                \vdots \\
+                a_N
+           \end{pmatrix},
+
+        where now :math:`W` is an :math:`N \times 2N` matrix with orthonormal rows
+        (which satisfies additional constraints).
+
+        Returns:
+            - The matrix :math:`W`, which is either an :math:`N \times N` or
+              an :math:`N \times 2N` matrix
```
mrossinek, line 202: Maybe repeat that the latter is the case the `\Delta != 0`

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,306 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021, 2022.
```
mrossinek, line 3: I believe that, because of the way how our copyright checker works, this is going to fail once merged to `main` because on that branch a single squashed commit dated to 2022 will be the only commit history related to this file.

@manoelmarques is there anything reasonable we can do about that?

mrossinek (commented):
As outlined [here](https://github.com/Qiskit/qiskit-nature/issues/460#issuecomment-1039226596) we can proceed with this PR.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: I wrapped it earlier; let me know if it's still not okay.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: Done.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: Done.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: Done.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: It's not necessary because it's guaranteed if the input matrices are valid, and if the user turned validation off, then this shouldn't be checked.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: Done.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: Done.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: Done, I named it `num_modes`.

kevinsung:
@mrossinek I have addressed all your comments besides the one related to the copyright checker. If needed, I can manually squash my commits into a single one with a 2022 timestamp.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
woodsp-ibm: I did not believe we checked the validity as such of the earliest date the code was copyrighted - unless that's the only date in there since its the same as the last date it was updated, due to code getting moved around across repos/refactored etc its kinda difficult to do. The time the code was last updated, the latter (or sole) date, we did check to make sure it was current with the changes.

mrossinek:
> I have addressed all your comments besides the one related to the copyright checker. If needed, I can manually squash my commits into a single one with a 2022 timestamp.

Sounds like the copyright should be fine as only the last year is being checked. Please fix CI and then we can progress with merging this :+1:

kevinsung:
@mrossinek I'm still issues running checks locally (see https://github.com/Qiskit/qiskit-nature/issues/567) but I've fixed the issues raised by CI.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,349 @@
...
+                terms.append(([("+", j), ("-", i)], self.hermitian_part[j, i]))
+                terms.append(([("+", i), ("+", j)], self.antisymmetric_part[i, j]))
+                terms.append(([("-", j), ("-", i)], self.antisymmetric_part[i, j].conjugate()))
+        return FermionicOp(terms, register_length=self._num_modes)
+
+    def conserves_particle_number(self) -> bool:
+        """Whether the Hamiltonian conserves particle number."""
+        return np.allclose(self.antisymmetric_part, 0.0)
+
+    def majorana_form(self) -> tuple[np.ndarray, float]:
```
mrossinek, line 152: In order to be able to use a type hint such as this one in Python versions <3.9 you have to add the following import:
```python
```
That is what is causing the CI to break.

kevinsung: Done.

kevinsung:
@woodsp-ibm I've added a release note.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,351 @@
...
+    @property
+    def hermitian_part(self) -> np.ndarray:
+        """The matrix of coefficients of terms that conserve particle number."""
+        return self._hermitian_part
+
+    @property
+    def antisymmetric_part(self) -> np.ndarray:
+        """The matrix of coefficients of terms that do not conserve particle number."""
+        return self._antisymmetric_part
+
```
woodsp-ibm, line 127: So this has a number of getters but no setters to change this instance later?

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,351 @@
...
+    def constant(self) -> float:
+        """The constant."""
+        return self._constant
+
+    @property
+    def num_modes(self) -> float:
+        """The number of modes this operator acts on."""
+        return self._num_modes
+
+    def to_fermionic_op(self) -> FermionicOp:
```
mrossinek, line 138: I am wondering if we should rename this to `to_second_q_op` which is inline with the namings in the `properties` module.
If we (in the future) design a common base class for some matrix container that will likely be the name of this function, avoiding the deprecation of this one.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,351 @@
...
+        """Convert to FermionicOp."""
+        terms = [([], self.constant)]  # type: list[tuple[list[tuple[str, int]], complex]]
```
mrossinek, line 140:
```suggestion
        terms: list[tuple[list[tuple[str, int]], complex]] = [([], self.constant)]
```

mrossinek (commented):
Just a minor comment, other than that this lgtm :+1:

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: I think the current name is much better because it precisely describes what the method does. `to_second_q_op` is confusing because QuadraticHamiltonian is already a "second-quantized op" (though not a SecondQuantizedOp). From the perspective of a user trying to discover this method, I think they are more likely to guess the current name.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: Done.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,351 @@
...
+        self,
+        hermitian_part: Optional[np.ndarray] = None,
+        antisymmetric_part: Optional[np.ndarray] = None,
+        constant: float = 0.0,
+        num_modes: Optional[int] = None,
+        validate: bool = True,
+        rtol: float = 1e-5,
+        atol: float = 1e-8,
+    ) -> None:
+        r"""Initialize a QuadraticHamiltonian.
```
woodsp-ibm, line 57: This text can/should be removed `Initialize a QuadraticHamiltonian.` We had other places where things like X constructor is put as text. Since this is Python this is what the init method its not necessary to state the obvious. As the class docstring is merged with the docstring here when its comes out in the docs we can go straight to args I think

image

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,351 @@
...
+
+        Args:
+            hermitian_part: The matrix :math:`M` containing the coefficients of the terms
+                that conserve particle number.
+            antisymmetric_part: The matrix :math:`\Delta` containing the coefficients of
+                the terms that do not conserve particle number.
+            constant: An additive constant term.
+            num_modes: Number of fermionic modes. This should be consistent with hermitian_part
+                and antisymmetric_part if they are specified.
+            validate: Whether to validate the inputs.
```
woodsp-ibm, line 67: What is the rationale of giving a user the choice here (performance over hermitian and antisymmetric checks?)? I imagine things will go bad if the inputs do not conform to certain expectations, though I see these checks are only done when both matrices are specified. Things like the hermitian and anti-symmetric tests are not done on individual inputs. Nor is num_modes checked, if given, if only a single matrix is supplied.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,351 @@
...
+                "Either Hermitian part, antisymmetric part, or number of modes must be specified."
+            )
+
+        if validate:
+            if (
+                hermitian_part is not None
+                and antisymmetric_part is not None
+                and hermitian_part.shape != antisymmetric_part.shape
+            ):
+                raise ValueError("Hermitian part and antisymmetric part must have same shape.")
```
woodsp-ibm, line 97: In these error messages it is helpful for a user and debugging etc for them to be formatted with the actual failing values.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: Yes, the user should be allowed to turn off validation for performance reasons.

>Things like the hermitian and anti-symmetric tests are not done on individual inputs. Nor is num_modes checked, if given, if only a single matrix is supplied.

This is incorrect. The checks are, in fact, performed in those situations, and I have included tests for that behavior.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: Done.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
kevinsung: Done.

kevinsung:
@woodsp-ibm I have addressed your comments. Please take another look.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
woodsp-ibm: My bad, I read the first if, under validate, of the and of the two and somehow misread the other tests as being nested under that. I see now that they are at the same level,

woodsp-ibm:
I think you missed the **f** from the string they now need to be like `f"my value {x}"` for the vars to be formatted in.

Also although I put the comment in one place for the error message I was meaning it could apply to a number of the error messages in the various checks to show the failing values where that's applicable.

woodsp-ibm (approved):

mrossinek (approved):

PR #483, opened 2021-12-22
Add Slater determinant and fermionic Gaussian state initial states

### LINKED ISSUE #459: Preparation of Slater determinants and fermionic Gaussian states

kevinsung:
### Summary
Fixes #459 . Comes after #462 and #482.

### Details and comments
I made the following changes to the pylint rules:
- Increased allowed function name length to 50 in order to accommodate the name `_prepare_fermionic_gaussian_state_jordan_wigner`
- Added `p` to the list of allowed variables since it is a common name for an index or number.

Let me know if this is not acceptable.


woodsp-ibm:
The pylint rules are common throughout qiskit and hence we should try and find variable names etc that fit within the current rules rather than changing the rules.

kevinsung:
Very well. I reverted the pylint rules, and instead disabled the checks for the violating lines by adding comments.

woodsp-ibm:
> Very well. I reverted the pylint rules, and instead disabled the checks for the violating lines by adding comments.

Ok, it would be more acceptable though to choose names etc within the rules rather than disabling them to skirt the rules. Having to disable an exceptional case where pylint is having trouble but things are as needed does happen, but disabling the rule inline should not be the norm here.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_states.py
```diff
@@ -0,0 +1,361 @@
...
+
+from qiskit import QuantumRegister, QuantumCircuit
+from qiskit.circuit import Gate, Qubit
+from qiskit.circuit.library import RZGate, XGate, XYGate
+
+from qiskit_nature.converters.second_quantization import QubitConverter
+from qiskit_nature.mappers.second_quantization import JordanWignerMapper
+
+
+class SlaterDeterminant(QuantumCircuit):
```
mrossinek, line 27: I would split this out into its own file

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_states.py
```diff
@@ -0,0 +1,361 @@
...
+        m, n = transformation_matrix.shape
+        if m > n:
+            raise ValueError("transformation_matrix must have orthonormal rows.")
+        # TODO maybe actually check if the rows are orthonormal
+
+        if qubit_converter is None:
+            qubit_converter = QubitConverter(JordanWignerMapper())
+
+        register = QuantumRegister(n, "q")
+        # TODO maybe use a shorter name
```
mrossinek, line 81: What do you mean here?

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_states.py
```diff
@@ -0,0 +1,361 @@
...
+            ValueError: transformation_matrix must have orthonormal rows.
+            NotImplementedError: Currently, only the Jordan-Wigner Transform is supported.
+        """
+        if not len(transformation_matrix.shape) == 2:
+            raise ValueError("transformation_matrix must be a 2-dimensional array.")
+
+        m, n = transformation_matrix.shape
+        if m > n:
+            raise ValueError("transformation_matrix must have orthonormal rows.")
+        # TODO maybe actually check if the rows are orthonormal
```
mrossinek, line 75: I would expect this if you want to throw the above error.

mrossinek (commented):
Just some minor comments. I assume you will refactor #482 according to our offline discussion so I am holding off with reviewing that PR

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_states.py
kevinsung: Splitting the two classes into separate files would result in a lot of code duplication because many of the private helper functions would need to be copied to both files. What would be the benefit?

mrossinek: So far we have always been splitting each class into its own file. I only noticed the common module-level functions later in my review. We don't have a lot of module level functions in Qiskit Nature and if we do we sometimes create a `utils` module. See for example here: https://github.com/Qiskit/qiskit-nature/tree/main/qiskit_nature/circuit/library/ansatzes/utils

@woodsp-ibm @pbark what do you think?

kevinsung: It might make sense to factor out and expose some of the more generally useful private functions as utilities, but I don't think that's appropriate for the highly specific helper functions like `_prepare_slater_determinant_jordan_wigner`. Is there a benefit to enforcing that every class go in its own file?

woodsp-ibm: For me the two circuits would seem reasonable separate modules if only that when browsing the code base the fact that there are these 2 different circuits that can be used stands out from just looking at the files and their names (hopefully!). The utility methods can be with the class that needs them - for the common ones, which seem to be matrix ones  they could simply be put in another file. I am not sure it would need a whole new utils folder here, and so far Nature does not have a root level utils where perhaps more general matrix manipulation functions might have been situated. Perhaps arranging the structure this way then if another similar circuit was created in the future it would facilitate that better and the whoever supplied it would not be tempted to just add it to the same file.

kevinsung: > For me the two circuits would seem reasonable separate modules if only that when browsing the code base the fact that there are these 2 different circuits that can be used stands out from just looking at the files and their names (hopefully!).

It's much more reliable in general to discover different classes by browsing the parsed documentation rather than making assumptions about how code is distributed between source files.

> The utility methods can be with the class that needs them - for the common ones, which seem to be matrix ones they could simply be put in another file.

I agree. The (uncommon) utility methods here belong with the class that needs them. Since in this case they are needed by both classes, both classes belong in the same file.

kevinsung: Very well. I guess I can put the helper functions in a private file without exposing that file to the public API.

kevinsung: Done.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_states.py
kevinsung: I changed the error message.

qiskit_nature/circuit/library/__init__.py
```diff
@@ -14,7 +14,7 @@
 Circuit Library (:mod:`qiskit_nature.circuit.library`)
 ======================================================
 
-A collection of circuits used as building blocks or inputs for algorithms.
+A collection of gates and circuits used as building blocks or inputs for algorithms.
```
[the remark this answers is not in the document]
kevinsung: Done.

qiskit_nature/operators/second_quantization/quadratic_hamiltonian.py
```diff
@@ -0,0 +1,351 @@
+# This code is part of Qiskit.
```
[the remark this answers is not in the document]
kevinsung: Yup.

releasenotes/notes/quad-ham-7c98cc11f700bdc5.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Adds :class:`qiskit_nature.operators.second_quantization.QuadraticHamiltonian`.
+    This class is used to represent quadratic fermionic Hamiltonians
+    and includes methods to efficiently diagonalize them.
```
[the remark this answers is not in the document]
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/_helper.py
```diff
@@ -0,0 +1,197 @@
+# This code is part of Qiskit.
```
mrossinek, line 1: We usually create a `utils` module in which we place helper functions. See e.g. `qiskit_nature/circuits/library/ansatzes/utils`

qiskit_nature/circuit/library/initial_states/_helper.py
```diff
@@ -0,0 +1,197 @@
...
+from typing import Iterable, List, Sequence, Tuple, Union
+
+import numpy as np
+
+from qiskit import QuantumRegister
+from qiskit.circuit import Gate, Qubit
+from qiskit.circuit.library import RZGate, XGate, XXPlusYYGate
+
+
+def _prepare_slater_determinant_jordan_wigner(  # pylint: disable=invalid-name
```
mrossinek, line 24: We have moved away from hard-coded fermion-to-qubit transformations everywhere in the stack. There are rare exceptions where a certain mapping is not allowed but in general it should not be the case that only one single case is allowed.

If there is no way of implementing a general method which works for at least JW and parity, please explain why.

qiskit_nature/circuit/library/initial_states/_helper.py
```diff
@@ -0,0 +1,197 @@
...
+    register: QuantumRegister, transformation_matrix: np.ndarray
+) -> Iterable[Tuple[Gate, Tuple[Qubit, ...]]]:
+    """Prepare a Slater determinant under the Jordan-Wigner Transform."""
```
mrossinek, line 27: I know that this is a private utility method but since you started a docstring please complete it with function argument and return type documentation.

qiskit_nature/circuit/library/initial_states/_helper.py
```diff
@@ -0,0 +1,197 @@
...
+                )
+                # update matrix
+                current_matrix = _apply_matrix_to_slices(
+                    current_matrix, givens_matrix, [(Ellipsis, j), (Ellipsis, j - 1)]
+                )
+
+    yield from reversed(decomposition)
+
+
+def _prepare_fermionic_gaussian_state_jordan_wigner(  # pylint: disable=invalid-name
```
mrossinek, line 72: There is quite some code overlap with the previous function. I suggest you refactor these slightly as to avoid this duplication.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -0,0 +1,121 @@
...
+            W_1 W_2^T + W_2 W_1^T = 0
+
+        Currently, only the Jordan-Wigner Transformation is supported.
+
+        Reference: arXiv:1711.05395
+
+        Args:
+            transformation_matrix: The matrix :math:`W` that specifies the coefficients of the
+                new creation operators in terms of the original creation and annihilation operators.
+                This matrix must satisfy special constraints; see the docstring of this function.
```
mrossinek, line 89: This is the docstring of this function. Please clarify where you are referring to

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -0,0 +1,121 @@
...
+            qubit_converter: a QubitConverter instance.
+            circuit_kwargs: Keyword arguments to pass to the QuantumCircuit initializer.
+
+        Raises:
+            ValueError: transformation_matrix must be a 2-dimensional array.
+            ValueError: transformation_matrix must have shape (n_orbitals, 2 * n_orbitals).
+            NotImplementedError: Currently, only the Jordan-Wigner Transform is supported.
```
mrossinek, line 96: You should link to the `JordanWignerMapper` here

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Currently, only the Jordan-Wigner Transformation is supported.
+
+        Reference: arXiv:1711.05395
+
+        Args:
+            transformation_matrix: The matrix :math:`W` that specifies the coefficients of the
+                new creation operators in terms of the original creation and annihilation operators.
+                This matrix must satisfy special constraints; see the docstring of this function.
+            qubit_converter: a QubitConverter instance.
```
mrossinek, line 90: `occupied_orbitals` is missing from the documentation. This should also explain why the it will fall back to zero occupied orbitals and what effect this has

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -0,0 +1,121 @@
...
+        super().__init__(register, **circuit_kwargs)
+
+        if isinstance(qubit_converter.mapper, JordanWignerMapper):
+            operations = _prepare_fermionic_gaussian_state_jordan_wigner(
+                register, transformation_matrix, occupied_orbitals
+            )
+            for gate, qubits in operations:
+                self.append(gate, qubits)
+        else:
+            raise NotImplementedError("Currently, only the Jordan-Wigner Transform is supported.")
```
mrossinek, line 121: This should be more explicit in telling the user how to fix this (i.e. provide a `QubitConverter(JordanWignerMapper())`)

qiskit_nature/circuit/library/initial_states/slater_determinants.py
```diff
@@ -0,0 +1,88 @@
...
+            ValueError: transformation_matrix must have orthonormal rows.
+            NotImplementedError: Currently, only the Jordan-Wigner Transform is supported.
+        """
+        if not len(transformation_matrix.shape) == 2:
+            raise ValueError("transformation_matrix must be a 2-dimensional array.")
+
+        m, n = transformation_matrix.shape
+        if m > n:
+            raise ValueError("transformation_matrix must be square.")
+        # TODO maybe actually check if the rows are orthonormal
```
mrossinek, line 75: Yes, please do this.

test/circuit/library/initial_states/test_fermionic_gaussian_states.py
```diff
@@ -0,0 +1,116 @@
...
+from qiskit_nature.mappers.second_quantization import JordanWignerMapper
+from qiskit_nature.operators.second_quantization.quadratic_hamiltonian import QuadraticHamiltonian
+
+
+def _random_antisymmetric(dim: int):
+    mat = np.random.randn(dim, dim) + 1j * np.random.randn(dim, dim)
+    return mat - mat.T
+
+
+class TestSlaterDeterminant(QiskitNatureTestCase):
```
mrossinek, line 30: This file is called `test_fermionic_gaussian_states` yet it appears to focus on testing the `SlaterDeterminant` class (as the name suggests). Please split the tests out into two properly named files with distinct sets of tests as this helps with the code+test navigation.

mrossinek (changes requested):

qiskit_nature/circuit/library/initial_states/_helper.py
kevinsung: The situation here seems different because there are no public functions to expose. I don't think it makes sense to create a public `utils` with no public members. Please advise.

qiskit_nature/circuit/library/initial_states/_helper.py
kevinsung: >There are rare exceptions where a certain mapping is not allowed but in general it should not be the case that only one single case is allowed.

Many algorithms in the literature (such as this one) are designed specifically for the Jordan-Wigner transform, so this does not seem like a tenable policy. It may be possible to adapt them to the parity transform, but I think the benefit hardly justifies the effort, since the parity transform has no advantages over the Jordan-Wigner transform. It seems unlikely that these algorithms could be adapted to the Bravyi-Kitaev transform.

qiskit_nature/circuit/library/initial_states/_helper.py
mrossinek: Simply creating `utils` will not make it part of the public API as long as we do not include it in the sphinx instructions for the documentation.

qiskit_nature/circuit/library/initial_states/_helper.py
mrossinek: Well the parity transform has the benefit of allowing a trivial reduction of two qubits.

I agree that the BK transform may be out-of-scope which, as I said above, is also the case elsewhere.

qiskit_nature/circuit/library/initial_states/_helper.py
kevinsung: In that case it would be part of the public API, just not part of the documentation. Those are separate things, right? I.e. if I create a `utils` module it you could still import it, it would show up in tab auto-complete. etc.

qiskit_nature/circuit/library/initial_states/_helper.py
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/slater_determinants.py
kevinsung: Done.

test/circuit/library/initial_states/test_fermionic_gaussian_states.py
kevinsung: Actually both classes were tested together to reduce code duplication. But I've split them into separate files now.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/_helper.py
kevinsung: Could you be specific? Which lines are duplicated? How should they be refactored?

mrossinek: `_prepare_slater_determinant_jordan_wigner` and `_prepare_fermionic_gaussian_state_jordan_wigner` share a lot of common code. This should be improved

mrossinek: Sorry, I was fooled a bit by the very similar looking structure of the code but you're right, extracting methods would overcomplicate the logic flow to a point where it would be harder to follow than this.

qiskit_nature/circuit/library/initial_states/_helper.py
mrossinek: Please just make it consistent with the rest of Qiskit Nature

qiskit_nature/circuit/library/initial_states/_helper.py
mrossinek: Okay, we can leave it as is. However, please add unittest to ensure that non-JW mappers are caught gracefully :+1:

qiskit_nature/circuit/library/initial_states/_helper.py
kevinsung: Ok. Just to clarify, you would like me to create a `utils` directory and put `_helper.py` in there. Then, you would like me to add an `__init__.py` to the `utils` directory to initialize it as an empty module.

mrossinek: Yes, create a `utils` module with an `__init__` but maybe give the file a more descriptive name than `_helper`. You can also make the file name not start with an underscore at that point

kevinsung: Done.

qiskit_nature/circuit/library/initial_states/_helper.py
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -0,0 +1,159 @@
...
+        super().__init__(register, **circuit_kwargs)
+
+        if isinstance(qubit_converter.mapper, JordanWignerMapper):
+            operations = _prepare_fermionic_gaussian_state_jordan_wigner(
+                register, transformation_matrix, occupied_orbitals
+            )
+            for gate, qubits in operations:
+                self.append(gate, qubits)
+        else:
+            raise ValueError(
```
mrossinek, line 154:
```suggestion
            raise NotImplementedError(
```
I actually preferred the `NotImplementedError` which you used before. This is also what you still write in the docstring

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -0,0 +1,159 @@
...
+        raise ValueError(
+            "transformation_matrix does not describe a valid transformation "
+            "of fermionic ladder operators. A valid matrix should have the block form "
+            "[W1 W2] where W1 @ W1.T.conj() + W2 @ W2.T.conj() = I and "
+            "W1 @ W2.T + W2 @ W1.T = 0."
+        )
+
+
+class FermionicGaussianState(QuantumCircuit):
+    """A circuit that prepares a fermionic Gaussian state."""
```
mrossinek, line 51: I think you should add some documentation for this and the `SlaterDeterminant` class below, in terms of how these are supposed to be used.
You should also explicitly refer to the `QuadraticHamiltonian` which I assume these are meant to be worked with compared to other ansatze we have in Qiskit Nature like `UCC` which explicitly works with the `HartreeFock` initial state.

qiskit_nature/circuit/library/initial_states/slater_determinants.py
```diff
@@ -0,0 +1,104 @@
...
+        _, n = transformation_matrix.shape
+        register = QuantumRegister(n)
+        super().__init__(register, **circuit_kwargs)
+
+        if isinstance(qubit_converter.mapper, JordanWignerMapper):
+            operations = _prepare_slater_determinant_jordan_wigner(register, transformation_matrix)
+            for gate, qubits in operations:
+                self.append(gate, qubits)
+        else:
+            raise ValueError(
```
mrossinek, line 99:
```suggestion
            raise NotImplementedError(
```

test/circuit/library/initial_states/test_fermionic_gaussian_states.py
```diff
@@ -0,0 +1,101 @@
...
+    def test_circuit_kwargs(self):
+        """Test that circuit keyword arguments are actually passed through."""
+        circuit = FermionicGaussianState(
+            np.array([[0.5, 0.5, 0.5, -0.5], [0.5, 0.5, -0.5, 0.5]]), name="efgh"
+        )
+        assert circuit.name == "efgh"
+
+    def test_unsupported_mapper(self):
+        """Test passing unsupported mapper fails gracefully."""
+        with self.assertRaisesRegex(ValueError, "supported"):
```
mrossinek, line 97:
```suggestion
        with self.assertRaisesRegex(NotImplementedError, "supported"):
```

test/circuit/library/initial_states/test_slater_determinants.py
```diff
@@ -0,0 +1,81 @@
...
+            _ = SlaterDeterminant(np.ones((2, 3)))
+
+    def test_circuit_kwargs(self):
+        """Test that circuit keyword arguments are actually passed through."""
+        circuit = SlaterDeterminant(np.eye(2), name="abcd")
+        assert circuit.name == "abcd"
+
+    def test_unsupported_mapper(self):
+        """Test passing unsupported mapper fails gracefully."""
+        with self.assertRaisesRegex(ValueError, "supported"):
```
mrossinek, line 80:
```suggestion
        with self.assertRaisesRegex(NotImplementedError, "supported"):
```

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
kevinsung: Done.

test/circuit/library/initial_states/test_fermionic_gaussian_states.py
kevinsung: Done.

kevinsung:
@mrossinek I've addressed your comments.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -0,0 +1,165 @@
...
+from qiskit_nature.mappers.second_quantization import JordanWignerMapper
+
+from .utils.givens_rotations import _prepare_fermionic_gaussian_state_jordan_wigner
+
+
+def _validate_transformation_matrix(
+    mat: np.ndarray, rtol: float = 1e-5, atol: float = 1e-8
+) -> None:
+    if not len(mat.shape) == 2:
+        raise ValueError("transformation_matrix must be a 2-dimensional array.")
```
mrossinek, line 29: I think all of these errors could be improved if you provide the current value. I.e. something like the below:
```suggestion
        raise ValueError(f"transformation_matrix must be a 2-dimensional array, not {len(mat.shape)}-dimensional.")
```

mrossinek: Similarly for other errors further down

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -0,0 +1,165 @@
...
+        rtol: float = 1e-5,
+        atol: float = 1e-8,
+        **circuit_kwargs,
+    ) -> None:
+        r"""
+        Args:
+            transformation_matrix: The matrix :math:`W` that specifies the coefficients of the
+                new creation operators in terms of the original creation and annihilation operators.
+                This matrix must satisfy special constraints; see the main body of the docstring
+                of this function.
```
mrossinek, line 122:
```suggestion
                of this class.
```

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -0,0 +1,165 @@
...
+            atol: Absolute numerical tolerance for input validation.
+            circuit_kwargs: Keyword arguments to pass to the QuantumCircuit initializer.
+
+        Raises:
+            ValueError: transformation_matrix must be a 2-dimensional array.
+            ValueError: transformation_matrix must have shape (n_orbitals, 2 * n_orbitals).
+            NotImplementedError: Currently, only the Jordan-Wigner Transform is supported.
+                Please use
+                :class:`qiskit_nature.mappers.second_quantization.JordanWignerMapper`
+                to construct the qubit mapper.
```
mrossinek, line 139: Maybe we can be more explicit here and refer to the actual argument `qubit_converter`.

qiskit_nature/circuit/library/initial_states/slater_determinants.py
```diff
@@ -0,0 +1,104 @@
...
+
+    .. math::
+        b^\dagger_1 \cdots b^\dagger_{N_f} \lvert \text{vac} \rangle
+
+    where
+
+    .. math::
+        b^\dagger_j = \sum_{k = 1}^N Q_{jk} a^\dagger_k,
+
+    - :math:`Q` is an :math:`N_f \times N` matrix with orthonormal rows
```
mrossinek, line 43: Compared to the explanation of `W` in the `FermionicGaussianState`, the explanation of `Q` is lacking. Maybe you can again add a section which explains how this matrix can be obtained further down in this docstring.

qiskit_nature/circuit/library/initial_states/slater_determinants.py
```diff
@@ -0,0 +1,104 @@
...
+            atol: Absolute numerical tolerance for input validation.
+            circuit_kwargs: Keyword arguments to pass to the QuantumCircuit initializer.
+
+        Raises:
+            ValueError: transformation_matrix must be a 2-dimensional array.
+            ValueError: transformation_matrix must have orthonormal rows.
+            NotImplementedError: Currently, only the Jordan-Wigner Transform is supported.
+                Please use
+                :class:`qiskit_nature.mappers.second_quantization.JordanWignerMapper`
+                to construct the qubit mapper.
```
mrossinek, line 79: Same as above

qiskit_nature/circuit/library/initial_states/slater_determinants.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        if validate:
+            if not len(transformation_matrix.shape) == 2:
+                raise ValueError("transformation_matrix must be a 2-dimensional array.")
+            if not _rows_are_orthonormal(transformation_matrix, rtol=rtol, atol=atol):
+                raise ValueError("transformation_matrix must have orthonormal rows.")
```
mrossinek, line 85: It's a bit inconsistent compared to the other class where you have a single function to do all validation compared to the inline validations here.
I suggest you also extract these into a function.

Also, the error messages can be improved inline with my comment on the other class.

mrossinek (commented):
A few more comments. I suggest that @pbark and @woodsp-ibm also take a look now.

Besides all of these comments, I think it is important that you also add some information tying this PR and the `QuadraticHamiltonian` together in more detail. Either you can write a tutorial outlining how (and why) this would be used or you at least flesh out the release note a bit with an actual code example.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/slater_determinants.py
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/slater_determinants.py
kevinsung: I didn't mention QuadraticHamiltonian here because I think the most common way to obtain `Q` will be from electronic structure codes. I will add that to the tutorial.

kevinsung: On further thought, I don't think `Q` will be obtained from electronic structure codes because they output the optimized molecular orbitals directly and the initial guess is not important. I've updated the docstring to mention how to obtain the matrix using the QuadraticHamiltonian class.

kevinsung:
@mrossinek I've addressed your comments and added a tutorial notebook.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "\n",
+    "circuit = FermionicGaussianState(\n",
+    "    transformation_matrix, occupied_orbitals=occupied_orbitals\n",
+    ")\n",
+    "circuit.draw(\"mpl\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
```
mrossinek, line 231: Also this cell appears quite detached from the section of the text which
discusses the mapper. This could be reshuffled slightly to improve things.

You could even consider printing the qubit operator rather than just asserting
its content.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "    b^\\dagger_N \\\\\n",
+    "    \\end{pmatrix}\n",
+    "    = W\n",
+    "    \\begin{pmatrix}\n",
+    "    a^\\dagger_1 \\\\\n",
+    "    \\vdots \\\\\n",
+    "    a^\\dagger_N \\\\\n",
+    "    \\end{pmatrix},\n",
+    "$$\n",
+    "where now $W$ is an $N \\times N$ matrix. Furthermore, the $\\{\\varepsilon_j\\}$ are allowed to be negative."
```
mrossinek, line 270: If you name the `\epsilon_j` where you introduce them, referring to them by name
here will be easier to understand. Right now, the reader just sees a reference
to a set of variables which do not appear in the equation immediately above.

mrossinek (commented):
Some comments on the tutorial but other than that this LGTM 👍

Please also make sure that the CI passes (styling of the tutorial + 1 spelling exception).

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "$$\n",
+    "where $M$ is a Hermitian matrix ($M^\\dagger = M$) and $\\Delta$ is an antisymmetric matrix ($\\Delta^T = -\\Delta$), and the $\\{a_j\\}$ are fermionic annihilation operators which satisfy the anticommutation relations\n",
+    "$$\n",
+    "    \\begin{align}\n",
+    "        a_j a_k + a_k a_j &= 0 \\\\\n",
+    "        a_j a^\\dagger_k + a^\\dagger_k a_j &= \\delta_{pq}\n",
+    "    \\end{align}.\n",
+    "$$\n",
+    "\n",
+    "Quadratic Hamiltonians are an important class of Hamiltonians that are classically tractable. Their eigenstates are called fermionic Gaussian states, and they can be efficiently be prepared on a quantum computer.\n",
```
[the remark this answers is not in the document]
kevinsung: Done.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "\n",
+    "Of course, the `FermionicOp` class can be used to represent any quadratic Hamiltonian. The reason to have a class specifically for quadratic Hamiltonians is that they support special numerical routines that involve performing linear algebra on the matrices $M$ and $\\Delta$. The internal representation format of `FermionicOp` is not suitable for these routines.\n",
```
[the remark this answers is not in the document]
kevinsung: This paragraph answers the question of why one would choose one class over another. This is a very natural question and important to clarify for users. Besides, there certainly are users who consider internal design aspects when choosing what classes to use. So I really think this paragraph should stay. If you disagree, you can go ahead and remove it before merging.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "\n",
+    "A quadratic Hamiltonian is initialized using `QuadraticHamiltonian`."
```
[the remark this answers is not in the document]
kevinsung: Done.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "    \\vdots \\\\\n",
+    "    a^\\dagger_N \\\\\n",
+    "    a_1 \\\\\n",
+    "    \\vdots \\\\\n",
+    "    a_N\n",
+    "    \\end{pmatrix},\n",
+    "$$\n",
+    "where $W$ is an $N \\times 2N$ matrix. Given a basis of eigenvectors of the Hamiltonian, each eigenvector is labeled by a subset of $\\{0, \\ldots, N - 1\\}$, which we call the occupied orbitals. The corresponding eigenvalue is the sum of the corresponding values of $\\varepsilon_j$, plus the constant.\n",
+    "\n",
+    "The transformation matrix $W$ is used to construct a circuit to prepare an eigenvector of the Hamiltonian. The circuit is constructed using the `FermionicGaussianState` class. Currently, only the Jordan-Wigner Transform is supported. The circuit for the Jordan-Wigner Transform has linear depth and uses only linear qubit connectivity. The algorithm is from [Phys. Rev. Applied 9, 044036](https://journals.aps.org/prapplied/abstract/10.1103/PhysRevApplied.9.044036)."
```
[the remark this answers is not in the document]
kevinsung: Done.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
kevinsung: I've added a text section introducing this cell.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
kevinsung: Done.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,454 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e7d2a25e-5f5c-44ae-9e19-ff980704ed3a",
+   "metadata": {},
+   "source": [
+    "## Slater determinants\n",
+    "\n",
+    "When the antisymmetric part $\\Delta = 0$, then the Hamiltonian conserves particle number. In this case, the basis change only needs to mix creation operators, not annihilation operators:\n",
```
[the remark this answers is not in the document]
kevinsung: Done.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "    \\end{pmatrix},\n",
+    "$$\n",
+    "where $W$ is an $N \\times 2N$ matrix. Given a basis of eigenvectors of the Hamiltonian, each eigenvector is labeled by a subset of $\\{0, \\ldots, N - 1\\}$, which we call the occupied orbitals. The corresponding eigenvalue is the sum of the corresponding values of $\\varepsilon_j$, plus the constant.\n",
+    "\n",
+    "The transformation matrix $W$ is used to construct a circuit to prepare an eigenvector of the Hamiltonian. The circuit is constructed using the `FermionicGaussianState` class. Currently, only the Jordan-Wigner Transform is supported. The circuit for the Jordan-Wigner Transform has linear depth and uses only linear qubit connectivity. The algorithm is from [Phys. Rev. Applied 9, 044036](https://journals.aps.org/prapplied/abstract/10.1103/PhysRevApplied.9.044036)."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
```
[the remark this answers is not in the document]
kevinsung: Done.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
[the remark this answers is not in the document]
mrossinek: I see your point. However, in that case this is valuable information which is missing from the documentation itself. You should add it there, too, if you think it has a place here.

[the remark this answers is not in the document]
kevinsung: I've added a note for this to the QuadraticHamiltonian docstring.

kevinsung:
The tutorial check is failing because the server can't find the kernel I used. What should I do?
```
Notebook error:
NoSuchKernel in tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb:
No such kernel named qiskit-nature
```

kevinsung:
Shouldn't the CI server be overriding the kernelspec?

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "%qiskit_version_table\n",
+    "%qiskit_copyright"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "qiskit-nature",
+   "language": "python",
+   "name": "qiskit-nature"
```
mrossinek, line 450:
```suggestion
   "display_name": "Python 3",
   "language": "python",
   "name": "python3"
```
This should fix the kernel issue. @manoelmarques might have more insight on how to deal with this in a more general setting when he is back

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -0,0 +1,171 @@
...
+
+    The matrix :math:`W` is commonly obtained by calling the
+    :meth:`~.QuadraticHamiltonian.diagonalizing_bogoliubov_transform`
+    method of the :class:`~.QuadraticHamiltonian` class.
+    This matrix is used to create circuits that prepare eigenstates of the
+    quadratic Hamiltonian.
+
+    Currently, only the Jordan-Wigner Transformation is supported.
+
+    Reference: arXiv:1711.05395
```
woodsp-ibm, line 110: It might be nice if the reference were an actual link in the resulting documentation for the user

qiskit_nature/circuit/library/initial_states/slater_determinants.py
```diff
@@ -0,0 +1,120 @@
...
+    The matrix :math:`W` can be obtained by calling the
+    :meth:`~.QuadraticHamiltonian.diagonalizing_bogoliubov_transform`
+    method of the :class:`~.QuadraticHamiltonian` class when the
+    quadratic Hamiltonian conserves particle number.
+    This matrix is used to create circuits that prepare eigenstates of the
+    quadratic Hamiltonian.
+
+    Currently, only the Jordan-Wigner Transformation is supported.
+
+    Reference: arXiv:1711.05395
```
woodsp-ibm, line 69: Similar comment in regards a link

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    ") = hamiltonian.diagonalizing_bogoliubov_transform()\n",
+    "\n",
+    "print(\"Shape of matrix W:\")\n",
+    "print(transformation_matrix.shape)\n",
+    "print()\n",
+    "print(\"Orbital energies:\")\n",
+    "print(orbital_energies)\n",
+    "print()\n",
+    "print(\"Transformed constant:\")\n",
+    "print(transformed_constant)"
```
woodsp-ibm, line 184: This is a choice I guess to have the output over multiple lines.

Would it be easier on the eyes though and to read in the output keeping the text and value on one line by doing things more like this?
```
print(f"Shape of matrix W: {transformation_matrix.shape}")
print(f"Orbital energies: {orbital_energies}")
print(f"Transformed constant: {transformed_constant}")
```
Comment would apply to other places where printing is similar.

qiskit_nature/circuit/library/initial_states/slater_determinants.py
kevinsung: Done.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
kevinsung: Done.

releasenotes/notes/slater-53415163fff84313.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    Adds functionality to prepare Slater determinants and fermionic Gaussian states
+    using the Givens rotation strategy. This functionality is accessed via the
+    following classes:
```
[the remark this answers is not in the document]
kevinsung: Done.

kevinsung:
test failures seem unrelated

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,440 @@
...
+    "\n",
+    "A quadratic Hamiltonian is a Hamiltonian of the form\n",
+    "$$\n",
+    "    H =\n",
+    "    \\sum_{j, k} M_{jk} a^\\dagger_j a_k\n",
+    "    + \\frac12 \\sum_{j, k}\n",
+    "        (\\Delta_{jk} a^\\dagger_j a^\\dagger_k + \\Delta^*_{jk} a_k a_j)\n",
+    "    + \\text{constant}\n",
+    "$$\n",
+    "where $M$ is a Hermitian matrix ($M^\\dagger = M$) and $\\Delta$ is an antisymmetric matrix ($\\Delta^T = -\\Delta$), and the $\\{a_j\\}$ are fermionic annihilation operators which satisfy the anticommutation relations\n",
```
mrossinek, line 18:
```
and the $\\{a_j\\}$ are fermionic annihilation operators
```
I suggest to be a bit more explicit here and also include the creation operators (the same way you do it in the documentation).

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,440 @@
...
+   "id": "93b2ca42-9947-45da-ad1e-7c11735dc5a5",
+   "metadata": {},
+   "source": [
+    "## Diagonalization and state preparation\n",
+    "A quadratic Hamiltonian can always be rewritten in the form\n",
+    "$$\n",
+    "    H = \\sum_{j} \\varepsilon_j b^\\dagger_j b_j + \\text{constant}\n",
+    "$$\n",
+    "where $\\varepsilon_0 \\leq \\varepsilon_1 \\leq \\cdots \\leq \\varepsilon_N$ are non-negative real numbers called orbitals energies and\n",
+    "the $\\{b_j^\\dagger\\}$ are a new set of fermionic creation operators\n",
```
mrossinek, line 125: And analogously here, you should also include the `b_j` annihilation operators.

mrossinek (commented):
Other than this final nitpicky comment, I think this is good to go! Thanks a lot, @kevinsung!

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
kevinsung: Done.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
kevinsung: Sorry, I'm not sure how you would like it to be worded. The current wording seems perfect to me. If you insist on changing it, please make an explicit suggestion. You can also commit it yourself before merging.

kevinsung:
@mrossinek I addressed one of your comments and wasn't sure how to address the other. Please take a look.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -0,0 +1,441 @@
...
+    "$$\n",
+    "where $M$ is a Hermitian matrix ($M^\\dagger = M$) and $\\Delta$ is an antisymmetric matrix ($\\Delta^T = -\\Delta$), and the $\\{a_j\\}$ are fermionic annihilation operators which satisfy the anticommutation relations\n",
+    "$$\n",
+    "    \\begin{align}\n",
+    "        a_j a_k + a_k a_j &= 0 \\\\\n",
+    "        a_j a^\\dagger_k + a^\\dagger_k a_j &= \\delta_{pq}\n",
+    "    \\end{align}.\n",
+    "$$\n",
+    "The hermitian conjugates $\\{a_j^\\dagger\\}$ are called fermionic creation operators.\n",
+    "\n",
```
mrossinek, line 26:
```suggestion
```

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
mrossinek:
```suggestion
    "the $\\{b_j, b_j^\\dagger\\}$ are a new set of fermionic annihilation and creation operators\n",
```

kevinsung:
@mrossinek I disagree with your suggestions because they make it seem like the creation and annihilation operators are independent of each other. I think the set notation {a_j} should only include one or the other, not both.

kevinsung:
Specifically, the notation {a_j, a_j^\dagger} makes it appear that a_j^\dagger is an independent symbol rather than simply the hermitian conjugate of a_j.

mrossinek:
In that case, please make it consistent whether you define the creation or annihilation set because right now you define `a_j` but later you define `b_j^\dagger`

kevinsung:
@mrossinek Done.

mrossinek (commented):
I'd say this lgtm now. @woodsp-ibm please also take a look :+1:

woodsp-ibm (approved):

PR #490, opened 2022-01-02
Fix mypy errors created with latest numpy

manoelmarques:
### Summary
numpy 1.22.0 released for python > 3.7 added mypy errors.
Errors fixed in this PR

### DIFF
--- a/qiskit_nature/algorithms/excited_states_solvers/qeom.py
+++ b/qiskit_nature/algorithms/excited_states_solvers/qeom.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2020, 2021.
+# (C) Copyright IBM 2020, 2022.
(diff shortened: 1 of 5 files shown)


woodsp-ibm:
As far as I can see the latest numpy requires Python 3.8 or above so will never support 3.7 https://github.com/numpy/numpy/blob/984a1ac8c911fa0130a6801a897642dd850796e6/setup.py#L36-L39 Are we therefore obliged to keep this as a solution now until such time that we no longer support Python 3.7 - from the EOL date for 3.7 it seems that is most likely mid next year (2023),

manoelmarques:
@woodsp-ibm I just changed the fix to use `cast` instead of `ignore` so it works in both old a new numpy without the need to change `mypy.ini`

mrossinek (approved):
Thanks @manoelmarques!

PR #503, opened 2022-01-11
fix: update BlueprintCircuit derivatives

mrossinek:
### Summary

This updates our circuit library Ansatzes deriving from `BlueprintCircuit` in accordance with https://github.com/Qiskit/qiskit-terra/pull/7216


pbark (approved):

pbark:
thanks @mrossinek . This LGTM

woodsp-ibm:
@mrossinek This was a change in the dev version of Terra i.e. 0.20.0 such that Nature is now dependent on that i.e. the _is_built was added. As such we ought to update the requirements.txt to bump the dependent version - need to watch out for this when we adjust code to latest Terra to also update the dependent version if needed..

PR #506, opened 2022-01-13
Support "Initialize FermionicOp with tuple as label #493"

### LINKED ISSUE #493: Initialize FermionicOp with tuple as label

The only way to initialize a FermionicOp is by providing a string label (i.e. `"II+"` or `"+_2"`, etc.). This is awkward when writing code that generates FermionicOps programmatically, where it is much more natural to use a tuple `(index, action)` instead, (i.e. `(2, 1)` instead of `"+_2"`).

ikkoham:
### Summary

Support the initialization of FermionicOp with tuple as label.
FermionicOp can be initialized using a tuple `(action, index)` like:

```python
         
FermionicOp(
    [([("-", 2 + 0j), ("+", 10)], 2 + 0j), ([("+", 3 + 0j)], 34)],
    register_length=4,
    display_format="sparse",
)
```

Close #493.


releasenotes/notes/initialize-ferop-with-tuple-as-label-0dc3f1dbd754ca29.yaml
```diff
@@ -0,0 +1,14 @@
+---
+features:
+  - |
+    Support the initialization of FermionicOp with tuple as label.
+    :class:`~qiskit_nature.operators.second_quantization.FermionicOp` can be initialized using
+    a tuple `(action, index)` like::
```
mrossinek, line 6:
```suggestion
    a tuple of integers, `(action, index)`, like below:
```

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -141,6 +167,17 @@ class FermionicOp(SecondQuantizedOp):
 
     are possible labels.
 
+    As a programmatic approach, a list of tuples as labels can be input.
+    FermionicOp can be initialized using a list of tuples `(action, index)`.
+    The action is integer and `0` is an annihilation and `1` is a creation.
+    The following labels are equivalent to those in the sparse case above:
```
mrossinek, line 173:
```suggestion
    As a programmatic approach, you can input a list of tuples as the labels,
    too. Thus, a FermionicOp can be initialized using a list of tuples of the
    form, `(action, index)`. `action` is an integer where `0` denotes an
    annihilation and `1` a creation operation. The `index` is also an integer
    denoting the fermionic mode.
    The following labels are equivalent to the example given for sparse labels
    above:
```

mrossinek (commented):
Thanks, @ikkoham! I just have some minor rephrasing suggestions.

ikkoham:
@mrossinek Thank you!

releasenotes/notes/initialize-ferop-with-tuple-as-label-0dc3f1dbd754ca29.yaml
```diff
@@ -0,0 +1,14 @@
...
+    :class:`~qiskit_nature.operators.second_quantization.FermionicOp` can be initialized using
+    a tuple of integers, `(action, index)`, like below:
+
+      from qiskit_nature.operators.second_quantization import FermionicOp
+
+      FermionicOp(
+          [([(0, 2), (1, 10)], 2), ([(0, 3)], 34)],
+          register_length=4,
+          display_format="sparse",
+      )
```
mrossinek, line 14:
```suggestion
      )
      
```
I'm assuming that this blank line is what is needed to make the docs generation happy? :thinking:

mrossinek: @ikkoham I guess it didn't help.. Maybe attempting to fix this locally is the best approach then..

ikkoham: I think :: is correct (not :).
https://github.com/Qiskit/qiskit-nature/pull/506/commits/0f388758729129b40d9523cdb79309bff08837a5

mrossinek: Yep looks good. Now we just need to fix the main CI again :smile:

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -141,6 +167,20 @@ class FermionicOp(SecondQuantizedOp):
 
     are possible labels.
 
+    As a programmatic approach, you can input a list of tuples as the labels,
+    too. Thus, a FermionicOp can be initialized using a list of tuples of the
+    form, `(action, index)`. `action` is an integer where `0` denotes an
+    annihilation and `1` a creation operation. The `index` is also an integer
```
mrossinek, line 173: I have started to wonder whether it's really that beneficial to use an integer for the `action` here.
Personally, I would find `"+"` and `"-"` a little bit less confusing and it surely is not too much of a hassle to write.

What do you think?

mrossinek: That's a good question actually. I don't see why we should not support those. Especially `I`. Because `N` and `E` can be expressed in terms of `+` and `-`, but in the current implementation we cannot construct an identity operation with a certain coefficient via this interface, right?

kevinsung: The identity operation should be constructed by proving an empty tuple:
```python
FermionicOp(((), 2.0))
```

ikkoham: OK. I don't support N and E.
I'll support `FermionicOp(([()], 2.0))`. 

This is for programming, so I want to only support integer.

kevinsung: @ikkoham The syntax should be `FermionicOp(([], 2.0))`, not `FermionicOp(([()], 2.0))`. You can see this because the string syntax is `FermionicOp(("", 2.0))` and `""` corresponds to `[]`, not `[()]`. There is no need to handle an empty tuple. In my comment above, I should have used an empty list instead of an empty tuple to be more clear.

ikkoham: @kevinsung right. empty list is correct.

mrossinek (commented):
I think this looks good now! Am I understanding things correctly that you also removed some unnecessary complexity (e.g. the `_FermionicLabelPrimitive`)?

ikkoham:
Yes, for `_FermionicLabelPrimitive `, it was a temporary private class until the internal data structure is finalized. I removed it as it was finalized in this PR.

releasenotes/notes/initialize-ferop-with-tuple-as-label-0dc3f1dbd754ca29.yaml
```diff
@@ -0,0 +1,15 @@
...
+features:
+  - |
+    Support the initialization of FermionicOp with tuple as label.
+    :class:`~qiskit_nature.operators.second_quantization.FermionicOp` can be initialized using
+    a tuple of integers, `(action, index)`, like below::
+
+      from qiskit_nature.operators.second_quantization import FermionicOp
+
+      FermionicOp(
+          [([(0, 2), (1, 10)], 2), ([(0, 3)], 34)],
```
ikkoham, line 11:
```suggestion
          [([("-", 2), ("+", 10)], 2 + 0j), ([("-", 3)], 34 + 0j)],
```

mrossinek (dismissed):
LGTM! Thanks @ikkoham!

@pbark @woodsp-ibm what do you think?

pbark (changes requested):
This LGTM, except the small proposed change.

releasenotes/notes/initialize-ferop-with-tuple-as-label-0dc3f1dbd754ca29.yaml
```diff
@@ -0,0 +1,15 @@
...
+features:
+  - |
+    Support the initialization of FermionicOp with tuple as label.
+    :class:`~qiskit_nature.operators.second_quantization.FermionicOp` can be initialized using
+    a tuple of integers, `(action, index)`, like below::
+
+      from qiskit_nature.operators.second_quantization import FermionicOp
+
+      FermionicOp(
+          [([("-", 2), ("+", 10)], 2 + 0j), ([("-", 3)], 34 + 0j)],
```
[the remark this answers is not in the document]
ikkoham: Thank you for finding this.

mrossinek (approved):

Issue #384, opened 2021-10-04
VQE Factory method improvements


mrossinek:
### What is the expected enhancement?

#380 introduces dependencies on private variables of the `VQE` class. This is a temporary workaround until https://github.com/Qiskit/qiskit-terra/issues/7093 gets resolved. This issue is used to keep track of revoking this workaround.

Another possibility would be to construct `self._vqe` freshly every time `get_solver` gets called. But we decided against that.

woodsp-ibm:
As the above Terra PR has been merged it would be possible to do this now so I removed the on hold label.

However I am wondering whether a better approach would be to let the user pass in a VQE instance, that the factory then uses, and have a getter/setter for it like the ansatz. Since the intent seems to be to re-use the vqe/uccsd instances. By passing in a VQE instance the user is free to set it however they like and via the vqe getter can update as they choose. This way the factory does not have to delegate all the setter/getters which has the disadvantage of getting out of sync with VQE if/when it alters its parameters.

mrossinek:
I like this idea of passing in a VQE instance. If we do this, we should arguably deprecate the other getter/setter methods and `__init__` arguments in favor the properties exposed on the `VQE` directly. What do you think?

woodsp-ibm:
Yes, given a getter/setter for VQE the instance is always then available via the factory and it can be changed directly; exposing aspects of VQE, whether via the constructor or properties, I would agree should then be deprecated/removed. I would expect VQE to be changing as these expectation objects are introduced using an directly would allievate trying to keep it in sync. The one things is that the factory is responsible for the ansatz so we might have to note that and when getting the solver ensure that the anstaz is the factory UCC instance maybe just by setting it there such that should the user attempt to do it directly via VQE that would be ignored/overridden by the factory

woodsp-ibm:
One other thought - maybe we should thing about allowing an instance of the VQEClient runtime here too as an alternative to the local running VQE. While UCCSD maybe problematic today on an actual device the runtime can still be used with the remote simulator right.

iuliazidaru:
@mrossinek Thank you for thinking about this issue and providing so many comments here.
Please let me check if I understood correctly what this task is about:
MinimumEigenSolverFatory implementation should create an instance of solver and expose it via minimum_eigensolver method?

If I understand correctly the approach, based on the experience working on #510 I would have some comments.
[...]

PR #510, opened 2022-01-17
BOPESSampler does not support MinimumEigensolverFactories

### LINKED ISSUE #398: `BOPESSampler` does not support `MinimumEigensolverFactories`

iuliazidaru:
### Summary
Fixes #398

### Details and comments

Factory provided in GroundStateEigensolver.
solver = GroundStateEigensolver(converter, VQEUCCFactory(StatevectorSimulator()))


qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -113,6 +113,13 @@ def sample(self, problem: BaseProblem, points: List[float]) -> BOPESSamplerResul
         """
         self._problem = problem
         self._driver = problem.driver
+        # We have to force the creation of the solver so that we work on the same solver
+        # instance before and after _gss.solve
+        self._gss.prepare_solve(problem, None)
```
mrossinek, line 118: This is essentially the same approach that you took in #457. For this fix (now that we have determined in more detail what the actual issue of #398 is) I do not think you need to change the `GroundStateEigensolver` at all. Instead, this fix should be self-contained within the `BopesSampler` class (see next comment).

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -113,6 +113,13 @@ def sample(self, problem: BaseProblem, points: List[float]) -> BOPESSamplerResul
...
+        if isinstance(self._gss.solver, VariationalAlgorithm):  # type: ignore
```
mrossinek, line 119: There are various of these `isinstance` checks in the `BopesSampler` code. This is what the culprit of the problem appears to be, because this check will fail when `VQEUCCFactory` is being used. Thus, I suggest you try to extend this check instead, similar to the following:
```suggestion
        if isinstance(self._gss.solver, (VariationalAlgorithm, VQEUCCFactory)):  # type: ignore
```
If you do this everywhere, that should also fix the problem.

I am deliberately suggesting to use `VQEUCCFactory` here, because `MinimumEigensolverFactory` would also contain the `NumPyMinimumEigensolverFactory` (which we do not want).
@pbark does the `BopesSampler` also support `VibrationalStructureProblems`? If so, we will also need to support the `VQEUVCCFactory`.

mrossinek (changes requested):

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
iuliazidaru: @mrossinek  In BopesSampler we are manipulating internal structure of members of GroundStateEigensolver. This breaks the encapsulation and produces this hard to extend design. Also, the method solve in GroundStateEigensolver processes the problem, creates solver instance and performs the calculation. This breaks single responsibility principle. When SOLID principles are violated, extension is hard and produces even worst design. I don't think that the solution is to put in this class if code everywhere for internal structure of other classes. We have to allow GroundStateEigensolver implementations to vary and BOPESSampler must still work.

mrossinek: I see what you're saying. However, in that case you should also change the `_run_single_point` method to remove the overhead of `prepare_solve` being called twice and instead call `compute_minimum_eigenvalue` and `problem.interpret` directly.

iuliazidaru: @mrossinek You're right. I'll check.

mrossinek (commented):
Alright, I had another offline discussion with @woodsp-ibm.

We now agree that this is a good solution for the time being and that you (@iuliazidaru) are right that we should not rely on instance checking for `VQEUCCFactory`.

However, we feel that `prepare_problem` could be better named `get_qubit_operators`. I would also prefer to return `main_operator, aux_ops` in this order.

In the future, we would like to tackle [this TODO](https://github.com/Qiskit/qiskit-nature/blob/main/qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py#L219) from the `VQEUCCFactory`. The related issue in Terra has been resolved for a while and we roughly outlined the idea in #384.
I will write a comment there in a moment to explain the idea in a bit more detail and how this connects to the BopesSampler

iuliazidaru:
@mrossinek Thank you for your comments. Without having too much experience in this project, I think that VQEUCCFactory has a good design and I wrote a comment in #384.
VQEUCCFactory needs a problem with second quantisation applied to create the solver instance. GroundStateSolver creates this instance just before compute_minimum_eigenvalue and we cannot do the initial calculation in BopesSampler before that. After solver creation, it is not recreated as it is saved in GroundStateSolver.


IMO BOPESSampler design has to be improved. Basically, the business is based on an instance which is managed by another instance(GroundStateSolver). Based on OpenClosed Principle, a class has to be open to extension and closed to modification. By changing initial state for the solver in GroundStateSolver we are modifying the instance. The main problem with this task is that  we want to modify the state of that instance and the instance is not created yet. You can see that the class is full of "# type: ignore" comments and those are the points which have to be changed. 
I didn't try a refactoring solution at the moment, but I think that a cleaner approach would be to provide extension points in GroundStateSolver so that the initial state is passed before _solver.compute_minimum_eigenvalue. Maybe just provide an optional initial_state to solve method?

mrossinek:
> VQEUCCFactory needs a problem with second quantisation applied to create the solver instance

Technically, this is more a `UCC` factory because this (+ the `HartreeFock` initial state) is the only piece which needs to be created at runtime by the factory. The `VQE` can (since some changes in Terra have been made) be constructed earlier and the relevant attributes/properties can be set on the instance as needed (see the discussion in #384).

About the `BOPESSampler` design: I agree, that it is not optimal. The cleanest would be:
- for each point during the sampling:
    - solve the problem using a new driver as to not modify the instances
However, since the solvers support re-usability, we should leverage that.

Currently, afaict _all_ `# type: ignore` comments result from the fact the the `solver` property is defined on `GroundStateEigensolver` instead of `GroundStateSolver`. I do not know why this is the case though. If we were to define the `solver` property on the base class, the `BOPESSampler` would indeed only access information which is supposed to be available.

I agree, that we could make the `GroundStateSolver` a bit more flexibel. However, I would only do that by splitting the current rather long functions into more modular pieces, visible in the public API.
Adding an optional `initial_state` to the `solve` method does not seem like a good idea.

Maybe, what we could do instead, is to add a utility method on the `GroundStateSolver` which returns whether or not the internal `MinimumEigensovler` is a `VariationalAlgorithm`. The `BOPESSampler` can then use this cleaner interface for the check and handle the bootstrapping based on that.

qiskit_nature/algorithms/ground_state_solvers/ground_state_eigensolver.py
```diff
@@ -87,11 +87,21 @@ def solve(
...
+        raw_mes_result = self._solver.compute_minimum_eigenvalue(main_operator, aux_ops)  # type: ignore
+
+        result = problem.interpret(raw_mes_result)
+        return result
+
+    def get_qubit_operators(
+        self,
+        problem: BaseProblem,
+        aux_operators: Optional[ListOrDictType[Union[SecondQuantizedOp, PauliSumOp]]] = None,
+    ) -> Tuple[PauliSumOp, Optional[ListOrDictType[Union[SecondQuantizedOp, PauliSumOp]]]]:
```
mrossinek, line 100:
```suggestion
    ) -> Tuple[PauliSumOp, Optional[ListOrDictType[PauliSumOp]]]:
```
At the end of function, you should only have `PauliSumOp`s left.

qiskit_nature/algorithms/ground_state_solvers/ground_state_eigensolver.py
```diff
@@ -87,11 +87,21 @@ def solve(
             An interpreted :class:`~.EigenstateResult`. For more information see also
             :meth:`~.BaseProblem.interpret`.
         """
+        main_operator, aux_ops = self.get_qubit_operators(problem, aux_operators)
+        raw_mes_result = self._solver.compute_minimum_eigenvalue(main_operator, aux_ops)  # type: ignore
+
+        result = problem.interpret(raw_mes_result)
+        return result
+
+    def get_qubit_operators(
```
mrossinek, line 96: This being a public function it needs to have a docstring.

mrossinek: EDIT: I see its being inherited from the base class.

qiskit_nature/algorithms/ground_state_solvers/ground_state_solver.py
```diff
@@ -58,6 +58,22 @@ def solve(
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def get_qubit_operators(
+        self,
+        problem: BaseProblem,
+        aux_operators: Optional[ListOrDictType[Union[SecondQuantizedOp, PauliSumOp]]] = None,
+    ) -> Tuple[PauliSumOp, Optional[ListOrDictType[Union[SecondQuantizedOp, PauliSumOp]]]]:
+        """Prepare problem and ground_state_solver state for solve method.
```
mrossinek, line 67: Needs to be updated to reflect that this is constructing the qubit operators by:
- getting the second quantized operators from the `problem` (potentially running a driver in doing so [can be computationally expensive])
- using a `QubitConverter` to map + reduce the operators to qubit ops

qiskit_nature/algorithms/ground_state_solvers/ground_state_solver.py
```diff
@@ -58,6 +58,22 @@ def solve(
...
+
+        Args:
+            problem: a class encoding a problem to be solved.
+            aux_operators: Additional auxiliary operators to evaluate.
+
+        Returns:
```
mrossinek, line 73: Needs to be documented

qiskit_nature/algorithms/ground_state_solvers/ground_state_solver.py
```diff
@@ -58,6 +58,22 @@ def solve(
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def get_qubit_operators(
+        self,
+        problem: BaseProblem,
+        aux_operators: Optional[ListOrDictType[Union[SecondQuantizedOp, PauliSumOp]]] = None,
+    ) -> Tuple[PauliSumOp, Optional[ListOrDictType[Union[SecondQuantizedOp, PauliSumOp]]]]:
```
mrossinek, line 66:
```suggestion
    ) -> Tuple[PauliSumOp, Optional[ListOrDictType[PauliSumOp]]]:
```
same as above

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
mrossinek: I wonder if we should already add the `is_variational` utility to the `GroundStateSolver` which I proposed in #384 :thinking: 
What do you think?

iuliazidaru:
@mrossinek I like this is_variational method and I'd like to implement it. However, I'm not sure that the place to implement these changes is this PR. Maybe you can add a new issue for BOPESSampler refactoring which incorporates also the changes in #384 ?
Also, we have to discuss what should return is_variational method in the case of MinimalEigensolverFactory.

mrossinek (commented):
I think this is shaping up quite nicely. There will still be some changes coming when dealing with #514 but the general structure looks good now.

@woodsp-ibm what do you think?

mrossinek (approved):
LGTM! But let's await another review as mentioned above.

test/algorithms/pes_samplers/test_bopes_sampler.py
```diff
@@ -155,6 +158,41 @@ def test_vqe_bootstrap(self):
         np.testing.assert_almost_equal(result.points, ref_points)
         np.testing.assert_almost_equal(result.energies, ref_energies)
 
+    @requires_extra_library
```
mrossinek, line 161: We just merged a new approach for dealing with optional dependencies after Terra introduced lazy loading. Can you please update the test accordingly? You should find plenty of migration examples in the PR #535

mrossinek (changes requested):

mrossinek:
@iuliazidaru I see you opened the follow-up PR :slightly_smiling_face: Is there anything you need help with here for migrating the unittest to the usage of `optionals` instead of `requires_extra_library`?

iuliazidaru:
@mrossinek Sorry for the delay, I had a couple of issues in progress ... I changed the tests now.

mrossinek (approved):
LGTM, again! :+1:

@woodsp-ibm @pbark please also take a look at this one

mrossinek (approved):
LGTM, again! :+1:

@woodsp-ibm @pbark please also take a look at this one

woodsp-ibm:
1. I think since this was addressing a feature enhancement we should have a reno release note for this change.

2. If the following a true statement? The BopesSampler now supports a GroundStateSolver (GSS) built using a Factory... but unlike when using VQE directly in the GSS, if a Factory based on VQE is used, the solver therein is not treated as variational and hence the interpolation that is other possible when using VQE directly with the GSS, via the initial point setting/getting, is not done when using a Factory supplied VQE. As far as I can see from the code it is but want to check.
Now the interpolation can be quite a performance improvement overall in terms of total evaluations needed over a run. So if the above is indeed the case at some point the initial point interpolation ought to be leveraged. Maybe this is simpler once the factory refactoring is done? and hence might be better left until then?

mrossinek:
@woodsp-ibm 

1. That's a very good point, thanks! @iuliazidaru could you please [add a release note](https://github.com/Qiskit/qiskit-nature/blob/main/CONTRIBUTING.md#adding-a-new-release-note)?
2. Sorry I am not fully catching what you are saying/asking here. We did discuss some further refactorings in the conversation of this PR. Iulia has already working on another PR which will add an `is_variational` check to the factories which will allow us to further improve things in the `BOPESSampler` (#570). Is that what you are referring to?

woodsp-ibm:
> Sorry I am not fully catching what you are saying/asking here. We did discuss some further refactorings in the conversation of this PR. Iulia has already working on another PR which will add an is_variational check to the factories which will allow us to further improve things

Ok, there was discussion, sorry it was quite a bit to go through and was unsure of the outcome given conversations above. If the factory directly can pass back the algorithm, via some common algorithm getter, and that is valid at any time so it can updated/viewed by nature code and/or and end user, then of course the variational check can be done directly on the algorithm (as is done with VQE today). Whether its a separate is_variational method or whatever it seems that BopesSampler will be improved in respect of initial point behavior when a variational algo, such as VQE, is used via a factory after the factory refactoring as perhaps seemed reasonable.

iuliazidaru:
@mrossinek I added the releasenote. Please check if anything is left. Thank you!

woodsp-ibm (approved):
Looks good to me. (I understand that it leveraging VQE initial point for interpolation, when VQE is supplied via such a factory, is to come in some subseqent PR).

iuliazidaru:
@woodsp-ibm @mrossinek Can we have this PR merged?

woodsp-ibm (approved):

pbark (approved):
LGTM!

Issue #511, opened 2022-01-18
BKSF mapper fails on LiH example


nonhermitian:
### Environment

- **Qiskit Nature version**: 0.3.0
- **Python version**: 
- **Operating system**: 

### What is happening?

BKSF mapper fails on LiH example due to mult missing from `SparsePauliOp`:

```bash

~/mambaforge/envs/qiskit/lib/python3.8/site-packages/qiskit_nature/converters/second_quantization/qubit_converter.py in _map(self, second_q_op)
    339 
    340     def _map(self, second_q_op: SecondQuantizedOp) -> PauliSumOp:
--> 341         return self._mapper.map(second_q_op)
    342 
    343     def _two_qubit_reduce(

~/mambaforge/envs/qiskit/lib/python3.8/site-packages/qiskit_nature/mappers/second_quantization/bksf.py in map(self, second_q_op)
     47 
     48         edge_list = _bksf_edge_list_fermionic_op(second_q_op)
---> 49         sparse_pauli = _convert_operator(second_q_op, edge_list)
     50 
     51         ## Simplify and sort the result

~/mambaforge/envs/qiskit/lib/python3.8/site-packages/qiskit_nature/mappers/second_quantization/bksf.py in _convert_operator(ferm_op, edge_list)
    123                 # Note that h2_pqrs is not divided by 2 here, as in the aqua code
    124                 sparse_pauli = _add_sparse_pauli(
--> 125                     sparse_pauli, _number_excitation(edge_list, p, q, r, s, h2_pqrs)
    126                 )
    127             else:

~/mambaforge/envs/qiskit/lib/python3.8/site-packages/qiskit_nature/mappers/second_quantization/bksf.py in _number_excitation(edge_list, p, q, r, s, h2_pqrs)
    343         a_pr = _edge_operator_aij(edge_list, p, r)
    344         a_pr = -a_pr if r < p else a_pr
--> 345         qubit_op = (a_pr * b_r + b_p * a_pr) * (id_op - b_q)
    346         final_coeff = 1j * 0.25
    347     else:

TypeError: unsupported operand type(s) for *: 'SparsePauliOp' and 'SparsePauliOp'

```

### How can we reproduce the issue?

Just follow along here: 

https://github.com/IBM-Quantum-Technical-Enablement/Fraunhofer-Dec-13-2021/blob/main/applications/qiskit-nature.ipynb

and replace JW mapper with BKSF (BK and Parity mappers work).

### What should happen?

Should work

[...]

PR #512, opened 2022-01-19
Use dot rather than obsolete * for composing ops

jlapeyre:
### Summary

Fixes bug reported in #511 in BKSF mapper when processing a number-excitation term

### Details and comments

A test for a bigger molecule, LiH, that previously triggered the bug has been added.
A reno release note has been added.
~A test and a reno will be added.~


mrossinek (commented):
Thanks, @jlapeyre! Did you happen to validate the LiH hamiltonian which you obtained for BKSF with another working source? E.g. the old Aqua code?

jlapeyre:
>  Did you happen to validate the LiH hamiltonian 

@mrossinek , No, all my energy went into pleasing the linters ;). I did look into doing something more like unit testing, eg. test just the function that had the change (for handling number-excitation terms). But, the inputs are rather complicated. (I have another idea to try, though). Regarding Aqua, I recall finding a bug. For the existing H2 test, I compared to OpenFermion. This might be a good option for LiH as well.

mrossinek (approved):
I guess we can assume that the Hamiltonian is correct. This should probably go into the next bugfix release so I'm approving this.

@woodsp-ibm or @pbark Please also take a look :+1:

jlapeyre:
Merging this as is (after other reviews) seems like a good option. Maybe another PR to improve, or rethink, tests would be good. I also thought about unittesting just the internal method that has been fixed here... that could show that it does what I intended to code (or the original authors did). But not that it gives the correct "physical" answer. Unit testing this way is probably a fair amount of work and would be verbose. On the other hand comparing to openfermion would be much closer to an integration test. The utility would mostly be in the process of writing the test; if they disagree, then we would have to find which (or both) of the packages gives the wrong answer. In fact, I don't think we have anything to convert between openfermion and qiskit-nature, right? Some simple conversion tools would be useful for uncovering potential bugs not only in BKSF, but more widely.

woodsp-ibm (approved):

PR #518, opened 2022-01-26
Remove deprecated drivers

manoelmarques:


mrossinek:
> I kept `GaussianForcesDriver` as it is used in other unit tests. Those unit tests need to either migrate to the new driver or be deleted.

I think the only tests which are still using the old driver are those related to the vibrational label/op builder utilities. The only function from those facing the public-API is also deprecated since 0.2.0 (https://github.com/Qiskit/qiskit-nature/blob/main/qiskit_nature/problems/second_quantization/vibrational/builders/vibrational_op_builder.py#L75) so you can safely remove those tests as well.

Should we also remove the legacy transformer classes related to the legacy drivers?

mrossinek (approved):
Afaict this looks good to me! At least I don't find anything else which should be removed :smile: 

Lets give @pbark and @woodsp-ibm to also take a look :+1:

PR #520, opened 2022-01-26
Implement detecting and transforming index order of two body electronic integrals

### LINKED ISSUE #337: Add `TwoBodyElectronicIntegrals` notation conversion utilities

Currently the `TwoBodyElectronicIntegrals` matrices are expected to be given in chemistry notation. Utility methods to convert to and from the physics notation would be valuable additions to this class.

Analogous to this we may also be interested in utility methods to convert the `SO`-basis matrices of the one- and two-body electronic integrals between the (currently required) blocked and interleaved spin formats.

@jlapeyre: I know that you previously worked on some utility methods for symmetry detection+conversion. Would you be interested in joining efforts to implement this enhancement?

jlapeyre:
- [X] Support AO input matrix in chemists', physicists', "intermediate", or unknown index-order convention. Convert or raise an error for unsupported input.
- [x] Tests to cover changes.
- [x] Updated the documentation accordingly.
- [x] Release note.
- [X] I have read the CONTRIBUTING document.

### Summary

When constructing `TwoBodyElectronicIntegrals` validate the data and possibly correct it.

Closes #337

Note, we considered the following, but they will not be included here:
 Support spin-orbital input matrices. 
 Include changing spin-orbital index convention? Maybe a subsequent PR

### Details and comments

This PR will transform physicists' index-order convention to chemists'. It will also raise an error on detecting an unsupported format or an unrecognized format.


mrossinek:
Thanks @jlapeyre, this looks like a very good start! I only skimmed through the code just now so no in detail comments yet. Just one question: I find the choice for the filename starting with `_` a bit odd. Maybe we can find a more descriptive name? Since it's already inside of a `utils` folder I don't think we need the `_utils` at the end either.
What about `two_body_symmetry_conversion.py`? Or something along those lines?

qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals_utils/_electronic_integrals_utils.py
```diff
@@ -0,0 +1,118 @@
...
+
+"""Utility functions to detect and transform the index-ordering convention of two-body integrals"""
+
+import numpy
+
+## These are the permutation symmetries satisfied by
+## a rank-4 tensor of real two-body integrals in chemists'
+## index order.
+## HJO is Molecular Electronic Structure Theory by Helgaker, Jørgensen, Olsen
+CHEM_INDEX_PERMUTATIONS = [
```
mrossinek, line 21: I think it would be cleaner to make this an `Enum` rather than relying on the list index later on

qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals_utils/_electronic_integrals_utils.py
```diff
@@ -0,0 +1,118 @@
...
+    Convert the rank-four tensor `two_body_tensor` representing two-body integrals from chemists'
+    index order to physicists' index order: i,j,k,l -> i,k,l,j
+
+    See `phys_to_chem`, `check_two_body_symmetries`.
+
+    Note:
+    Denote `chem_to_phys` by `g` and `phys_to_chem` by `h`. The elements `g`, `h`, `I` form
+    a group with `gh = hg = I`, `g^2=h`, and `h^2=g`.
+    """
+    permuted_tensor = numpy.einsum("ijkl->iklj", two_body_tensor)
```
mrossinek, line 54: This numpy einsum exists as `TwoBodyElectronicIntegrals.EINSUM_CHEM_TO_PHYS`. I think it would make sense to refactor the `TwoBodyElectronicIntegrals` to purely make use of the utilities here instead.

qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals_utils/_electronic_integrals_utils.py
```diff
@@ -0,0 +1,118 @@
...
+    See also: `chem_to_phys`, `phys_to_chem`.
+
+    Note:
+    The first of `:chemist`, `:physicist`, and `:intermediate`, in that order, to pass the tests
+    is returned. If `two_body_tensor` has accidental symmetries, it may in fact satisfy more
+    than one set of symmetry tests. For example, if all elements have the same value, then the
+    symmetries for all three index orders are satisfied.
+    """
+    if _check_two_body_symmetries(two_body_tensor):
+        return "chemist"
```
mrossinek, line 110: I think it would also make sense to create an `Enum` for the four different types that we can encounter.

qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals_utils/two_body_symmetry_conversion.py
```diff
@@ -0,0 +1,176 @@
...
+
+def _phys_to_chem(two_body_tensor):
+    """
+    Convert the rank-four tensor `two_body_tensor` representing two-body integrals from physicists'
+    index order to chemists' index order: i,j,k,l -> i,l,j,k
+
+    See `_chem_to_phys`, `_check_two_body_symmetries`.
+    """
+    permuted_tensor = numpy.einsum("ijkl->iljk", two_body_tensor)
+    return permuted_tensor
```
kevinsung, line 75: This doesn't seem quite right. You cannot simply permute the two-body tensor; reordering terms causes some one-body terms to pop out because of the anticommutation relations. These one-body terms need to be kept track of somehow, otherwise the resulting operator is not equivalent to the original.

mrossinek: The way of doing this transformation is dating back to the early days of Qiskit Aqua: https://github.com/Qiskit/qiskit-aqua/commit/d77b4302cd781b88a010f54e0c17bd877b31a6a7

mrossinek: All I'm saying is that this is how this conversion has always been done in Nature. Maybe it was just wrong to call it a chemist<->physicist conversion. This PR was started on the assumption that this conversion is correct, but what you're saying would resolve a lot of confusion I used to have around this topic.

This PR is not going to be merged as is anyways, since it will only be integrated in the upcoming refactoring. I converted it to draft until we have more time to look into this.

TL;DR in Aqua and Nature conversion which were called chemist<->physicist conversions always _ignore_ all one-body corrections that would result from them

kevinsung: I see, thanks for clarifying. I think PolynomialTensor will make this nice because you could store the correction in the same PolynomialTensor object.

kevinsung: Ok, that makes sense. Thanks!

jlapeyre: These functions implement a relabeling of coefficients. The coefficients have four indices. Each index points to a different Factor in the product of operators in a single term. It leaves each product of operators untouched. I might have one routine that performs computations assuming one index ordering. And another routine that does exactly the same  calculation but expects a different index ordering. These relabeling functions just allow the second routine to do exactly the same thing the first routine did. I think this agrees with what Max said above.

mrossinek:
The reason that this has not progressed further, is that I am awaiting the implementation of the `PolynomialTensor` (#756) and `SparseLabelOp` (#731) after which we can tackle #666 which is going to largely refactor the existing `FermionicOp`. Once this happens, we will always store the electronic integrals in physicist order, at which point these conversion utilities will come in very handy to simplify the construction of said integral containers.

I envision these utilities to be exposed very publicly to simplify the construction of `PolynomialTensor` instances which store electronic integrals. I am potentially even considering an electronic-dedicated subclass of this tensor class, which can handle these integral conversions as well as deal with symmetry reductions/expansions to/from 8-fold symmetry in the future.

Long story short, this will get unblocked once the above issues are resolved, at which point a good location for these utilities needs to be found in the new stack :+1:

mrossinek:
I have updated this PR and moved the symmetry tools to `qiskit_nature.second_q.operators.utils` where it can live for now. I might change things a bit more when working on #666 which is next up on my todo list. When doing that and updating the electronic-structure related hamiltonians and properties to use that method, I will be making use of these tools for easier construction of the `PolynomialTensor` objects.

woodsp-ibm (approved):

PR #524, opened 2022-01-28
Alter forces test based on driver revision

woodsp-ibm:
The forces test, using Gaussian driver, was done originally using RevA.03. It seems the values produced via the driver in a later rev are not quite the same. This changes the test to detect the version/revision and select the right set of values to compare against.

This needs to be updated for RevC.01 values.


mrossinek:
Thanks @woodsp-ibm! I just updated the values for the C01 rev and also added its output alongside the A01 logfile.
I did not duplicate the expected test values inside of `test/drivers/second_quantization/gaussiand/test_driver_gaussian_log.py` because the parsing does not change.

I also synced with @paulineollitrault who agrees that the results produced by G16RevC.01 indeed make more sense.

As an FYI we found a potential bug fix mentioned in the G16 release notes (https://gaussian.com/relnotes/ on the "Bugs Fixed" tab) which could be related to this test. This bug was fixed in C01 so assuming that B01 is equivalent to A03 should be a good bet here.

mrossinek (dismissed):
I also played a part in this PR so I suggest @pbark should also review this :+1:

docs/tutorials/aux_files/CO2_freq_B3LYP_631g.log
```diff
@@ -0,0 +1,670 @@
+ **********************************************************************
```
mrossinek, line 1: This is a copy of the file `test/drivers/second_quantization/gaussiand/test_driver_gaussian_log_C01.txt`. We cannot use it from that location because it needs to be readily accessible from within the docs.

test/drivers/second_quantization/gaussiand/test_driver_gaussian_log.py
```diff
@@ -35,7 +35,7 @@ def test_log_driver(self):
             [
                 "#p B3LYP/6-31g Freq=(Anharm) Int=Ultrafine SCF=VeryTight",
                 "",
-                "CO2 geometry optimization B3LYP/cc-pVTZ",
+                "CO2 geometry optimization B3LYP/6-31g",
```
mrossinek, line 38: This was an inconsistency in the description. This has no effect on the calculation outcome.

test/problems/second_quantization/vibrational/resources/expected_labels.py
```diff
@@ -1,409 +0,0 @@
-# This code is part of Qiskit.
```
mrossinek, line 1: This was left over from the old stack and entirely unused. It is now superseded by the `expected_ops.py` file in the same `resources` folder.

mrossinek (commented):
Just some FYI comments

test/mappers/second_quantization/test_direct_mapper.py
```diff
@@ -16,39 +16,75 @@
...
+        vibration_energy = self.driver_result.get_property("VibrationalEnergy")
         vibration_energy.basis = HarmonicBasis(num_modals)
 
         vibration_op = vibration_energy.second_q_ops()[0]
 
         mapper = DirectMapper()
         qubit_op = mapper.map(vibration_op)
 
-        self.assertEqual(qubit_op, REFERENCE)
+        self.assertEqual(qubit_op, _num_modals_3_q_op)
```
mrossinek, line 87: I previously overlooked that the `DirectMapper` test also uses this log file. Since it is more fitting I also migrated the tutorial qubit_op test here.

mrossinek (dismissed):
LGTM! Although I was significantly involved in this PR so @pbark and @woodsp-ibm should probably also take a look.

pbark (approved):

manoelmarques (approved):

mrossinek (approved):

Issue #26, opened 2021-01-29
[Feature] HF/DFT Embedding Schemes


mrossinek:
### What is the expected improvement?

To add an implementation of the HF and DFT embedding schemes proposed in this paper: https://arxiv.org/abs/2009.01872.

The aforementioned paper provides an algorithm which allows embedding a quantum computing calculation into a DFT calculation. We already have a basic implementation based on the development version of Qiskit which we will open a PR for once the paper is published.

In the meantime, we can use this issue to discuss some of the implementation aspects which involve the paper. At this time, I can say that the implementation can be separated into 3 logically separate blocks. I will outline the concepts of each of these below.

#### 1. HF Embedding
The first block is a simple extension of the `QMolecule` class. It adds a new static method called `active_space_reduction` which can be used to achieve a proper HF embedding through calculating the _inactive Fock operator_ which is in turn used instead of the 1-electron integrals in all further calculations.
This allows the specification of an _active space_ to which the quantum calculation is being restricted.

Some more notes:
- I have exposed the same functionality in the `Hamiltonian` class (similarly to how `freeze_core` is exposed)
- the active space can even be selected manually (i.e. it is not limited to be a symmetric selection around the Fermi level)
- unittests are also in place

#### 2. PySCFDriver Refactoring
In order to prepare the final goal (which is the DFT Embedding) we need to refactor the `PySCFDriver` (which is for now the only supported driver). Since the `PyQuanteDriver` has common aspects with the `PySCFDriver` I have extended the refactoring to it as well, where applicable.
This refactoring includes the following aspects:
- deprecation of `HFMethodType` in favor of `pyscfd.MethodType` and `pyquantd.MethodType`
- deprecation of `hf_method` in favor of `method` (in both, `PySCFDriver` and `PyQuanteDriver`)
- DFT support in both drivers
- support for disabling the dipole integral computation in `PySCFDriver`
- support for `QMolecule` construction based on a PySCF `chkfile`

#### 3. DFT Embedding
Since this DFT Embedding poses a fully fledged application of Qiskit within the `chemistry` submodule, I have implemented this algorithm as a new class in the `qiskit.chemistry.applications` module. The implementation is currently limited to be used with the `PySCFDriver` but supports full configuration in most (I dare not say, all) other aspects. Further driver support is a WIP.
After the successful integration of the previous two paragraphs, this change should essentially only involve the addition of the new class and corresponding unittests.

[...]

PR #525, opened 2022-01-31
PyQuante Driver Update

mrossinek:
### Summary

This aligns the `PyQuanteDriver` with the `PySCFDriver`.

### Details and comments

Part of #166


qiskit_nature/constants.py
```diff
@@ -38,3 +38,19 @@
 KB_J_PER_K = 1.3806488e-23
 BOHR = 0.52917721092  # No of Angstroms in Bohr (from 2010 CODATA)
 DEBYE = 0.393430307  # No ea0 in Debye. Use to convert our dipole moment numbers to Debye
+
+PERIODIC_TABLE = [
```
woodsp-ibm, line 42: Nice, you fixed it! I noticed the other week that black had altered it and thought it was much better in the table form as it was created.

qiskit_nature/drivers/second_quantization/pyscfd/pyscfdriver.py
```diff
@@ -477,7 +477,7 @@ def _check_molecule_format(val: str) -> Union[str, List[str]]:
 
         # An xyz format has 4 parts in each atom, if not then do zmatrix convert
         # Allows dummy atoms, using symbol 'X' in zmatrix format for coord computation to xyz
-        parts = [x.strip() for x in atoms[0].split(" ")]
+        parts = [x.strip() for x in atoms[0].split()]
```
woodsp-ibm, line 480: Was lint or something complaining about explicitly providing the value that is the default. I guess its fine, but presumably either is ok.

woodsp-ibm:
As I think I mentioned with pyscf refactor we ended up going from a class that was the driver the user interacted with plus one that did the behind the scenes work to get the integrals etc, and create the return now all in a single larger file. It just feels less modular having it all now together - as well since part of the refactor's goal too is to allow custom replacement of the computation. Maybe this is how it has to be now as its harder to separate concerns into different modules as before.

qiskit_nature/constants.py
mrossinek: I actually extracted this from the `FreezeCoreTransformer`. I didn't know that this was duplicated in the `QMolecule`. Same as some other constants. Since the `QMolecule` is no loinger standalone since we added the deprecation notice to it anyways, I just refactored that too :+1:

qiskit_nature/drivers/second_quantization/pyscfd/pyscfdriver.py
mrossinek: Actually no. But I found some weeks ago that this was causing problems because I had loading the coordinates from a file where multiple spaces were used in order to make the numbers aligned and more readable. So I had been using this in my local fork for quite a while and just noticed that this was not in the original source.

mrossinek:
> It just feels less modular having it all now together - as well since part of the refactor's goal too is to allow custom replacement of the computation.

Yes you are right about this. The new approach does indeed result in larger classes overall but it has certain benefits:
- the behavior is split more logically into actual computation and result construction
- the behavior can be modified by a subclass (important especially in the context of #26) without duplicating major parts of the code

The disadvantage is indeed larger files. I guess in theory we could extract these methods into a separate module and then plug them into the class but I do not think that this is worthwhile.

qiskit_nature/constants.py
woodsp-ibm: I imagine the copy in FreezeTransformer originated from QMolecule - its a better place in constants of course!

woodsp-ibm:
> The disadvantage is indeed larger files. I guess in theory we could extract these methods into a separate module and then plug them into the class but I do not think that this is worthwhile.

Ok, but I will note that you cite this as an advantage

>the behavior is split more logically into actual computation and result construction

which is not reflected in the module design since its all in one file. Maybe splitting it does not allow the easy sub-classing of the driver as you intend. Either way I'll not say more, it obviously works as-is, but as you can tell I am not fully comfortable with it, though I think it can be approved as-is, since this is how the pyscf one is too.

woodsp-ibm (approved):

HenryMakowitz:
Please Let me know when a DFT Quantum Code is available for distribution.
Thanks.
Dr. Henry Makowitz, PhD

PR #527, opened 2022-02-01
Temporarily skip broken tests

mrossinek:
### Summary

Disables the tests which broke after https://github.com/Qiskit/qiskit-terra/pull/7551.

### Details and comments

This is to be reverted once the occurrence of close-to-zero imaginary phases arising in the `Z2Symmetry` reduction code are handled properly within Qiskit Terra.

EDIT: PR in Terra which is supposed to fix this: https://github.com/Qiskit/qiskit-terra/pull/7598


manoelmarques:
There is a decorator too `@unittest.skip` but since this is temporary, it is fine by me.

manoelmarques (approved):

Issue #48, opened 2021-02-12
Implement the ParticleHoleTransformer


mrossinek:
### What is the expected feature?

Add the `ParticleHoleTransformer`.

- [ ] actual implementation
- [ ] unittests
- [ ] documentation

mrossinek:
We attempted an initial implementation of this in #86 but closed that PR due to the reasons explained in [this gist](https://gist.github.com/mrossinek/3cc65837d2a998856f202ef341e7f455). Adding this for documentation purposes here, too.

PR #530, opened 2022-02-02
perf: remove unneeded reduction in IntegralProperty

mrossinek:
### Summary

This removes an unneeded call of `.reduce()` in the IntegralProperty.

Co-authored-by: @ikkoham


woodsp-ibm:
I can see some changes in the test, such as using len() instead of some hard coded value as an improvement but presumably net out to be equivalent. So that seems good. 

But I am wondering, for something that is "unneeded" that was removed, why the change in test cases. Maybe it changes the operator, as a result of reduce, and maybe it would better if the user did this explicitly if they want a reduced operator. Though when using the full stack I guess that's less than doable. So perhaps in regards of using the operator its therefore unneeded i.e. you get the same net outcome whether reduced or not. I guess I am struggling a bit with understanding "unneeded" since removing it makes a difference it seems in regards of having to change the reference values for the tests.

mrossinek:
You're right that "unneeded" may not be the best choice of words here. It is indeed not required because the mathematical outcome does not change and the performance gain is quite significant. (@ikkoham do you still have the numbers we discussed once?). Furthermore, I actually think that not doing this reduction is required for the Particle-Hole Transformer #48.

dlasecki (approved):

woodsp-ibm (approved):

PR #533, opened 2022-02-02
reduce memory usage of `ElectronicIntegrals`

mrossinek:
### Summary

Until now, the property framework would duplicate the alpha-spin
matrices to fill in any missing beta-spin matrices. This can cause a
very significant memory overhead when dealing with large systems.
This changes the internal workings to remove this limitation and only
store the needed data once.

### Details and comments

The effect which this PR has can best be seen by the diff of the
Property tutorial as the output is greatly reduced (because the
matrices are no longer duplicated in memory and thus neither when
printed out).


qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals.py
```diff
@@ -185,8 +228,12 @@ def __str__(self) -> str:
             string += self._render_matrix_as_sparse_list(self._matrices)
         else:
             for title, mat in zip(self._MATRIX_REPRESENTATIONS, self._matrices):
-                rendered_matrix = self._render_matrix_as_sparse_list(mat)
                 string += [f"\t{title}"]
+                if mat is None:
+                    string[-1] += " is None"
+                    string += ["\tThe corresponding primary-spin data will be used in-place."]
```
mrossinek, line 234: I specifically do not want to hard-code that pure alpha-spin data will be used, because in the example of the `TwoBodyElectronicIntegrals` the `alpha-beta` matrix will be substituted with the transposed `beta-alpha` matrix (when this is not None).

I am happy to take suggestions for a better wording here.

woodsp-ibm: In reading the output in notebook I am wondering about `in-place` which does not seem right here. Maybe `for this` so instead of it reading `The corresponding primary-spin data will be used in-place.` it will say `The corresponding primary-spin data will be used for this.` I think had it said `used in it's place.` that could also work i.e.  `The corresponding primary-spin data will be used in it's place.` Its just that `in-place` I would more read as in-situ rather that instead of, or in place of. 

I don't have any better overall suggestion in regards of primary etc though I wonder about in the string listing it as None. As I mentioned the interface by which you read the values will never return None right? I.e if you read the beta aspect and internally it was None you will be given back the alpha aspect. So instead of 
```
      "\t\t\tBeta is None\n",
      "\t\t\tThe corresponding primary-spin data will be used in-place.\n",
```
or variants of the text as discussed above, could it be more simply done as something like this
```
      "\t\t\tBeta\n",
      "\t\t\tSame values as the corresponding primary-spin data.\n",
```
which is how it would be perceived through the external public API right,

qiskit_nature/properties/second_quantization/electronic/bases/electronic_basis_transform.py
```diff
@@ -89,11 +105,15 @@ def from_hdf5(h5py_group: h5py.Group) -> ElectronicBasisTransform:
         Returns:
             A new instance of this class.
         """
+        coeff_alpha = h5py_group["Alpha coefficients"][...]
+        coeff_beta: Optional[np.ndarray] = None
+        if "Beta coefficients" in h5py_group.keys():
+            coeff_beta = h5py_group["Beta coefficients"][...]
```
dlasecki, line 111: We might consider using this cleaner approach:
```suggestion
        coeff_beta = h5py_group.get("Beta coefficients", None)
```
where I assumed `None` is a default value in case the key is not present. In fact adding `None` as a second argument is optional because it defaults to it if not provided.
Another option is `setdefault` which might be a bit more efficient.
Not sure how it fits with the ellipsis though (`[...]`)...

mrossinek: Thanks for the suggestion but due to how `h5py` works this is not possible. Precisely because of the ellipsis `[...]` which reads the data stored in `h5py_group["Beta coefficients"]` as a numpy array.

woodsp-ibm (approved):

PR #535, opened 2022-02-04
Adopt lazy imports from Terra

manoelmarques:
### Summary
Adopt the API implemented in Qiskit/qiskit-terra#7525

### DIFF
--- a/qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
+++ b/qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021.
+# (C) Copyright IBM 2021, 2022.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
@@ -81,7 +81,8 @@ class ElectronicStructureDriverType(Enum):
                 raise MissingOptionalLibraryError(
                     libname=driver_type, name="ElectronicStructureDriverType"
                 )
-            driver_class.check_installed()
+            # instantiating the object will check if the driver is installed
+            _ = driver_class()
             driver_class.check_method_supported(method)
 
         logger.debug("%s found from type %s.", driver_class.__name__, driver_type.value)
--- a/qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
+++ b/qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
@@ -78,7 +78,7 @@ class GaussianForcesDriver(VibrationalStructureDriver):
             _optionals.HAS_GAUSSIAN.require_now("GaussianForcesDriver __init__")
 
     @staticmethod
-    @_optionals.HAS_GAUSSIAN.require_in_call("GaussianForcesDriver from_molecule")
+    @_optionals.HAS_GAUSSIAN.require_in_call
     def from_molecule(
         molecule: Molecule,
         basis: str = "sto-3g",
--- a/qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
+++ b/qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2020, 2021.
+# (C) Copyright IBM 2020, 2022.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
@@ -18,10 +18,10 @@ from qiskit_nature import QiskitNatureError
 from qiskit_nature.properties.second_quantization.vibrational import (
     VibrationalStructureDriverResult,
 )
+import qiskit_nature.optionals as _optionals
 from ...units_type import UnitsType
 from ..vibrational_structure_driver import VibrationalStructureDriver
 from ...molecule import Molecule
-from .gaussian_utils import check_valid
 from .gaussian_log_driver import GaussianLogDriver
 from .gaussian_log_result import GaussianLogResult
 
@@ -75,9 +75,10 @@ class GaussianForcesDriver(VibrationalStructureDriver):
         # If running from a jcf we need Gaussian™ 16 so check if we have a
         # valid install.
         if self._logfile is None:
-            check_valid()
+            _optionals.HAS_GAUSSIAN.require_now("GaussianForcesDriver __init__")
 
     @staticmethod
+    @_optionals.HAS_GAUSSIAN.require_in_call("GaussianForcesDriver from_molecule")
     def from_molecule(
         molecule: Molecule,
         basis: str = "sto-3g",
@@ -98,7 +99,6 @@ class GaussianForcesDriver(VibrationalStructureDriver):
         """
         # Ignore kwargs parameter for this driver
         del driver_kwargs
-        GaussianForcesDriver.check_installed()
         basis = GaussianForcesDriver.to_driver_basis(basis)
 
         if molecule.units == UnitsType.ANGSTROM:
@@ -129,16 +129,6 @@ class GaussianForcesDriver(VibrationalStructureDriver):
         if basis == "sto3g":
             return "sto-3g"
         return basis
-
-    @staticmethod
-    def check_installed() -> None:
-        """
-        Checks if Gaussian is installed and available
-
-        Raises:
-            MissingOptionalLibraryError: if not installed.
-        """
-        check_valid()
 
     def run(self) -> VibrationalStructureDriverResult:
         if self._logfile is not None:
--- a/qiskit_nature/drivers/second_quantization/pyquanted/pyquantedriver.py
+++ b/qiskit_nature/drivers/second_quantization/pyquanted/pyquantedriver.py
@@ -12,7 +12,6 @@
 
 """ PyQuante Driver """
 
-import importlib
 import inspect
 import logging
 import re
@@ -21,7 +20,6 @@ from typing import Union, List, Optional, Any, Dict, Tuple
 
 import numpy as np
 
-from qiskit.exceptions import MissingOptionalLibraryError
 from qiskit.utils.validation import validate_min
 
 from qiskit_nature import QiskitNatureError
@@ -43,6 +41,7 @@ from qiskit_nature.properties.second_quantization.electronic.integrals import (
     OneBodyElectronicIntegrals,
     TwoBodyElectronicIntegrals,
 )
+import qiskit_nature.optionals as _optionals
 
 from ..electronic_structure_driver import ElectronicStructureDriver, MethodType
 from ...molecule import Molecule
@@ -51,13 +50,12 @@ from ....exceptions import UnsupportMethodError
 
 logger = logging.getLogger(__name__)
 
-try:
+if _optionals.HAS_PYQUANTE2:
+    # pylint: disable=import-error
     from pyquante2 import molecule as pyquante_molecule
     from pyquante2 import rhf, uhf, rohf, basisset, onee_integrals
     from pyquante2.geo.zmatrix import z2xyz
     from pyquante2.ints.integrals import twoe_integrals
-except ImportError:
-    logger.info("PyQuante2 is not installed. See https://github.com/rpmuller/pyquante2")
 
 
 class BasisType(Enum):
@@ -84,6 +82,7 @@ class BasisType(Enum):
         raise QiskitNatureError(f"Invalid Basis type basis {basis}.")
 
 
+@_optionals.HAS_PYQUANTE2.require_in_instance
 class PyQuanteDriver(ElectronicStructureDriver):
     """
     Qiskit Nature driver using the PyQuante2 library.
@@ -121,7 +120,6 @@ class PyQuanteDriver(ElectronicStructureDriver):
         """
         super().__init__()
         validate_min("maxiters", maxiters, 1)
-        PyQuanteDriver.check_installed()
         PyQuanteDriver.check_method_supported(method)
         if not isinstance(atoms, str) and not isinstance(atoms, list):
             raise QiskitNatureError(f"Invalid atom input for PYQUANTE Driver '{atoms}'")
@@ -228,6 +226,7 @@ class PyQuanteDriver(ElectronicStructureDriver):
         self._maxiters = maxiters
 
     @staticmethod
+    @_optionals.HAS_PYQUANTE2.require_in_call("PyQuanteDriver from_molecule")
     def from_molecule(
         molecule: Molecule,
         basis: str = "sto3g",
@@ -243,11 +242,10 @@ class PyQuanteDriver(ElectronicStructureDriver):
         Returns:
             driver
         """
-        PyQuanteDriver.check_installed()
         PyQuanteDriver.check_method_supported(method)
         kwargs = {}
         if driver_kwargs:
-            args = inspect.getfullargspec(PyQuanteDriver.__init__).args
+            args = inspect.signature(PyQuanteDriver.__init__).parameters.keys()
             for key, value in driver_kwargs.items():
                 if key not in ["self"] and key in args:
                     kwargs[key] = value
@@ -272,32 +270,6 @@ class PyQuanteDriver(ElectronicStructureDriver):
             driver acceptable basis
         """
         return BasisType.type_from_string(basis)
-
-    @staticmethod
-    def check_installed() -> None:
-        """
-        Checks if PyQuante is installed and available
-
-        Raises:
-            MissingOptionalLibraryError: if not installed.
-        """
-        try:
-            spec = importlib.util.find_spec("pyquante2")
-            if spec is not None:
-                return
-        except Exception as ex:
-            logger.debug("PyQuante2 check error %s", str(ex))
-            raise MissingOptionalLibraryError(
-                libname="PyQuante2",
-                name="PyQuanteDriver",
-                msg="See https://github.com/rpmuller/pyquante2",
-            ) from ex
-
-        raise MissingOptionalLibraryError(
-            libname="PyQuante2",
-            name="PyQuanteDriver",
-            msg="See https://github.com/rpmuller/pyquante2",
-        )
 
     @staticmethod
     def check_method_supported(method: MethodType) -> None:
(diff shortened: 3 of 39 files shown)


qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
```diff
@@ -81,7 +81,8 @@ def driver_class_from_type(
                 raise MissingOptionalLibraryError(
                     libname=driver_type, name="ElectronicStructureDriverType"
                 )
-            driver_class.check_installed()
+            # instantiating the object will check if the driver is installed
+            _ = driver_class()
```
mrossinek, line 85: This implicitly implies that:
1. all driver classes can be instantiated without any arguments
2. an instantiation of the class before we checked whether the requested method is supported

Is there a better way than this? Or are these implications fine?

mrossinek: Okay continuing to read on I think these are not really of concern. Leaving this here for other reviews to check and discuss :+1:

qiskit_nature/optionals.py
```diff
@@ -0,0 +1,90 @@
...
+    A lazy checker that a command-line tool is available. This just checks if the command is
+    not empty.
+    """
+
+    __slots__ = ("_command",)
+
+    def __init__(
+        self,
+        command: str,
+        *,
```
mrossinek, line 33: :slightly_smiling_face:

qiskit_nature/optionals.py
```diff
@@ -0,0 +1,90 @@
...
+    G16PROG,
+    name=GAUSSIAN_16_DESC,
+    msg="Please check that it is installed correctly",
+)
+
+PSI4 = "psi4"
+PSI4_DESC = "PSI4"
+PSI4PROG = which(PSI4)
+if PSI4PROG is None:
+    PSI4PROG = PSI4
```
mrossinek, line 84: Why is this needed? If `which` cannot find the command how would we be able to use it?

qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
```diff
@@ -81,7 +81,8 @@ def driver_class_from_type(
                 raise MissingOptionalLibraryError(
                     libname=driver_type, name="ElectronicStructureDriverType"
                 )
```
mrossinek, line 83: Related to the logic above which is dealing with driver importing. Can this still occur? I only noticed because I assumed that during the move to lazy imports we can get rid of the Nature-specific `MissingOptionalLibraryError` entirely..

test/drivers/second_quantization/test_molecule_driver.py
```diff
@@ -60,14 +62,16 @@ def test_invalid_kwarg(self):
...
     def test_driver(self, config):
         """test driver"""
         driver_type, hf_energy = config
         driver = ElectronicStructureMoleculeDriver(
             self._molecule, basis="sto3g", driver_type=driver_type
         )
-        driver_result = driver.run()
+        try:
+            driver_result = driver.run()
+        except MissingOptionalLibraryError as ex:
```
mrossinek, line 73: I guess this relates to my comment above: how can this still occur?
Is it because we defer the lazy loading of pyscf into another class?

mrossinek (commented):
In general this looks good to me :+1: I have some minor comments.

I did grep for `ImportError` on the repo and did find a few more occurrences though. Maybe those can be refactored, too?
```
qiskit_nature/drivers/second_quantization/gaussiand/gaussiandriver.py:257:        except ImportError as mnfe:
test/drivers/second_quantization/fcidumpd/test_driver_fcidump_dumper.py:138:        except ImportError:
test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py:320:        except ImportError as ex:
test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py:408:        except ImportError as ex:
test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py:434:        except ImportError as ex:
test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py:465:        except ImportError as ex:
test/test_readme_sample.py:41:        except ImportError as ex:
```
Grepping for `MissingOptionalLibraryError` gives the following:
```
qiskit_nature/problems/second_quantization/lattice/lattices/lattice.py:216:            MissingOptionalLibraryError: Requires matplotlib.
qiskit_nature/drivers/second_quantization/vibrational_structure_molecule_driver.py:22:from qiskit.exceptions import MissingOptionalLibraryError
qiskit_nature/drivers/second_quantization/vibrational_structure_molecule_driver.py:54:            MissingOptionalLibraryError: Driver not installed.
qiskit_nature/drivers/second_quantization/vibrational_structure_molecule_driver.py:64:                    except MissingOptionalLibraryError as ex:
qiskit_nature/drivers/second_quantization/vibrational_structure_molecule_driver.py:73:                raise MissingOptionalLibraryError(
qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py:22:from qiskit.exceptions import MissingOptionalLibraryError
qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py:59:            MissingOptionalLibraryError: Driver not installed.
qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py:72:                    except (MissingOptionalLibraryError, UnsupportMethodError) as ex:
qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py:81:                raise MissingOptionalLibraryError(
test/drivers/second_quantization/test_molecule_driver.py:20:from qiskit.exceptions import MissingOptionalLibraryError
test/drivers/second_quantization/test_molecule_driver.py:73:        except MissingOptionalLibraryError as ex:
test/drivers/second_quantization/test_molecule_driver.py:92:                except MissingOptionalLibraryError as ex:
test/drivers/second_quantization/test_molecule_driver.py:156:        except MissingOptionalLibraryError as ex:
```

qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
manoelmarques: @mrossinek We used to have a method to check if driver was installed. I removed them favoring the Terra approach of using decorators. Since there is a decorator for the __init__ and all the driver classes that are in the list of the Molecule Driver support no arguments __init__ I just test it by instantiating. Otherwise, I can reinstate the previous `check_installed` method.

mrossinek: Yep makes sense :+1: When getting into later parts of the review I started to understand that a bit better, too. I left it here for completeness sake

qiskit_nature/optionals.py
manoelmarques: @mrossinek Just because the LazySubprocessTester from Terra won't accept a None or empty command in the constructor and I still wanted to use its capabilities of actually running the command to see if it is installed.

qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
manoelmarques: Lazy imports is not about removing `MissingOptionalLibraryError` . It is just a delayed mechanism but stiil raises `MissingOptionalLibraryError` as was done before. That exception needs to be handled by other areas of code and it is raised here because the Molecule Driver has a search driver mechanism and will raise the same exception if no appropriate driver was found.

qiskit_nature/optionals.py
mrossinek: Okay, then why is the same not being done in the gaussian case a few lines further up?

qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
mrossinek: So the lazy import mechanism does not provide a utility which we could use for this use case? In that case I'm obviously fine with this :+1:

manoelmarques:
> In general this looks good to me 👍 I have some minor comments.
> 
> I did grep for `ImportError` on the repo and did find a few more occurrences though. Maybe those can be refactored, too?
> 
> ```
> qiskit_nature/drivers/second_quantization/gaussiand/gaussiandriver.py:257:        except ImportError as mnfe:
> test/drivers/second_quantization/fcidumpd/test_driver_fcidump_dumper.py:138:        except ImportError:
> test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py:320:        except ImportError as ex:
> test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py:408:        except ImportError as ex:
> test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py:434:        except ImportError as ex:
> test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py:465:        except ImportError as ex:
> test/test_readme_sample.py:41:        except ImportError as ex:
> ```
> 
> Grepping for `MissingOptionalLibraryError` gives the following:
> 
> ```
> qiskit_nature/problems/second_quantization/lattice/lattices/lattice.py:216:            MissingOptionalLibraryError: Requires matplotlib.
> qiskit_nature/drivers/second_quantization/vibrational_structure_molecule_driver.py:22:from qiskit.exceptions import MissingOptionalLibraryError
> qiskit_nature/drivers/second_quantization/vibrational_structure_molecule_driver.py:54:            MissingOptionalLibraryError: Driver not installed.
> qiskit_nature/drivers/second_quantization/vibrational_structure_molecule_driver.py:64:                    except MissingOptionalLibraryError as ex:
> qiskit_nature/drivers/second_quantization/vibrational_structure_molecule_driver.py:73:                raise MissingOptionalLibraryError(
> qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py:22:from qiskit.exceptions import MissingOptionalLibraryError
> qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py:59:            MissingOptionalLibraryError: Driver not installed.
> qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py:72:                    except (MissingOptionalLibraryError, UnsupportMethodError) as ex:
> qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py:81:                raise MissingOptionalLibraryError(
> test/drivers/second_quantization/test_molecule_driver.py:20:from qiskit.exceptions import MissingOptionalLibraryError
> test/drivers/second_quantization/test_molecule_driver.py:73:        except MissingOptionalLibraryError as ex:
> test/drivers/second_quantization/test_molecule_driver.py:92:                except MissingOptionalLibraryError as ex:
> test/drivers/second_quantization/test_molecule_driver.py:156:        except MissingOptionalLibraryError as ex:
> ```

@mrossinek I will review the ImportError to see if any can be refactored but the `MissingOptionalLibraryError` need to be checked and need to be added to the documentation as being raised. That exception didn't go away. It is still being thrown by the Lazy import framework. It just has been delayed at the point where it is used.

qiskit_nature/optionals.py
manoelmarques: If it is about the formatting, it is `black`s formatter fault  :-)

mrossinek: No it's about [PEP 3102](https://www.python.org/dev/peps/pep-3102/) :+1:

qiskit_nature/optionals.py
manoelmarques: Ahhh, because unfortunately Gaussian doesn't have a command line option like version or whatever that lets me just simply run it without side effect of actually running code. For this I created my own Lazy class derived from Terra that just checks if the command is None. Effectively just what was being done before: 
```
class NatureLazyCommandTester(LazyDependencyManager):
```

qiskit_nature/drivers/second_quantization/electronic_structure_molecule_driver.py
manoelmarques: Yes, the Molecule Driver search for an appropriate driver based on an enumeration passed is our own custom implementation. There is nothing there for it but we raise the correct exception.

qiskit_nature/optionals.py
mrossinek: Oh I was not careful. The first time I read it, it looked like you used `NatureLazyCommandTester` for Gaussian and Psi4. But I see now that the latter actually uses the Terra-provided one :+1:

test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
```diff
@@ -398,16 +394,11 @@ def test_uccsd_hf_qasm(self):
         self.assertAlmostEqual(result.total_energies[0], -1.138, places=2)
 
     @slow_test
```
mrossinek, line 396: Since these are slow tests, can we have the CI run them once for this PR to ensure they work as expected?

mrossinek (commented):
Just one question otw this LGTM :slightly_smiling_face:

test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
manoelmarques: ok, I added the `run_slow` label to this PR

qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
```diff
@@ -75,9 +75,10 @@ def __init__(
         # If running from a jcf we need Gaussian™ 16 so check if we have a
         # valid install.
         if self._logfile is None:
-            check_valid()
+            _optionals.HAS_GAUSSIAN.require_now("GaussianForcesDriver __init__")
 
     @staticmethod
+    @_optionals.HAS_GAUSSIAN.require_in_call("GaussianForcesDriver from_molecule")
```
jakelishman, line 81: If you don't supply an argument in the decorator form, it should use the qualified name of the method automatically.  If you prefer, this line is equivalent to decorating with `@_optionals.HAS_GAUSSIAN.require_in_call` (i.e. without "calling" the decorator)

qiskit_nature/drivers/second_quantization/pyquanted/pyquantedriver.py
```diff
@@ -51,13 +50,12 @@
 
 logger = logging.getLogger(__name__)
 
-try:
+if _optionals.HAS_PYQUANTE2:
```
jakelishman, line 53: I don't know if you mind, but this isn't a lazy test - the Boolean evaluation will be triggered as soon as this module is imported.

qiskit_nature/drivers/second_quantization/pyquanted/pyquantedriver.py
```diff
@@ -243,11 +242,10 @@ def from_molecule(
         Returns:
             driver
         """
-        PyQuanteDriver.check_installed()
         PyQuanteDriver.check_method_supported(method)
         kwargs = {}
         if driver_kwargs:
-            args = inspect.getfullargspec(PyQuanteDriver.__init__).args
+            args = inspect.signature(PyQuanteDriver.__init__).parameters.keys()
```
jakelishman, line 248: I put some effort into trying to make sure the wrappers would propagate through the signature and documentation (`require_in_instance` does a bit of magic to wrap the `__init__` method).  Let me know if something's broken about that.

(`inspect.signature` is the recommended way, though)

qiskit_nature/optionals.py
jakelishman: What functionality would you like in `LazySubprocessTester`?  I made it reject "empty" commands as a way of catching bugs because I couldn't think of any use-cases, but if you've got a use-case for empty commands, we could add it.

Here, I would note that calling `which` on import isn't lazy - if you wanted it for your Gaussian one, perhaps your derived class could run `shutil.which` to define its `_is_available`?  It depends how lazy you care about being, though.

jakelishman (commented):
I hope the tools are useful - let us/me know if there's anything about the existing Terra objects that might want expanding.

qiskit_nature/optionals.py
```diff
@@ -0,0 +1,90 @@
...
+from qiskit.utils import LazyDependencyManager, LazyImportTester, LazySubprocessTester
+
+
+class NatureLazyCommandTester(LazyDependencyManager):
+    """
+    A lazy checker that a command-line tool is available. This just checks if the command is
+    not empty.
+    """
+
+    __slots__ = ("_command",)
```
jakelishman, line 28: Also, you don't need to do this if you don't want, especially if you'd like to attach data to the objects.  To be honest, I probably shouldn't have done it in Terra to give you the choice of adding arbitrary data - from a Python perspective, it's a bad habit of mine that I over-use `__slots__` (it comes from me preferring stricter typing, where you can't add attributes to stuff).

qiskit_nature/drivers/second_quantization/gaussiand/gaussian_forces_driver.py
manoelmarques: Oh that is helpful. Thanks for the tip. I will change it.

qiskit_nature/drivers/second_quantization/pyquanted/pyquantedriver.py
manoelmarques: Yes, I knew. We had those at the top because they are being used in different parts of the file and even in type annotations like:
```
 self._mol: pyquante_molecule = None
 self._bfs: basisset = None
``` 

I will see if it is possible to move them inside the methods.

qiskit_nature/drivers/second_quantization/pyquanted/pyquantedriver.py
manoelmarques: Yes, the `getfullargspec` args started returning an empty array. I changed to `signature` which works.

qiskit_nature/optionals.py
manoelmarques: Yes, since I was creating a lazy class I decided to stick to the idiom there but already had someone else asking me about it too :-) I will remove it.

manoelmarques:
@jakelishman About the `LazySubprocessTester`: 

Nature has two drivers, PSI4 and Gaussian, that have their own commands. We use `which` as a way of figuring out if they are installed. In the PSI4 case, I used `LazySubprocessTester` because it is an improvement by running also the command. I don't use `LazySubprocessTester` for Gaussian though because it doesn't have any command option that would let it run without actually feeding chemistry input which gets complicated.

A subprocess tester that would be useful would be one where I construct it with a command string, but inside `_is_available` it runs `which` and if it returns None, is an error. Also it doesn't run the command automatically since there is no good way to run Gaussian. Maybe an additional flag in the `__init__` telling if it should run the command. 

I created a very simple subclass for Gaussian just to test None or not but I was thinking of doing what you suggested and put the `which` inside the `_is_available`. I will try this and you can take a look. If you think it is useful, it could be moved to Terra at some point.

mrossinek (approved):
LGTM! Thanks @manoelmarques!

PR #536, opened 2022-02-04
fix: propagate truncation_order from VibrationalStructureProblem

mrossinek:
### Summary

`VibrationalStructureProblem.truncation_order` did not have an effect.
This commit ensures that the attribute is propagated down to the
properties which may need it.

### Details and comments

Note: I reran the tutorial to ensure everything works as expected now. This is indeed the case but it still show up in the changed files mostly because the ordering of the Pauli terms in the produced `qubit_op` is different. However, this does not influence the result. You can validate that by sorting the output of the original and new tutorial and comparing them afterwards. No change will be visible.


pbark (dismissed):
Good catch. Thanks @mrossinek LGTM!

woodsp-ibm (dismissed):

pbark (approved):

woodsp-ibm (approved):

PR #543, opened 2022-02-08
Use code from README.md in test_readme_sample

manoelmarques:

### DIFF
--- a/test/test_readme_sample.py
+++ b/test/test_readme_sample.py
@@ -18,28 +18,13 @@ the issue then ensure changes are made to readme too.
 
 import unittest
 
-import sys
+import contextlib
+import io
 from pathlib import Path
 import re
 from test import QiskitNatureTestCase
 from qiskit.utils import optionals
 import qiskit_nature.optionals as _optionals
-
-
-class Writer:
-    """Custom Writer"""
-
-    def __init__(self):
-        self.encoding = None
-        self.texts = []
-
-    def write(self, text):
-        """write to array"""
-        self.texts.append(text)
-
-    def flush(self):
-        """flush"""
-        pass
 
 
 class TestReadmeSample(QiskitNatureTestCase):
@@ -57,6 +42,8 @@ class TestReadmeSample(QiskitNatureTestCase):
             self.fail(msg=f"{readme_name} not found at {readme_path}")
             return
 
+        # gets the first matched code sample
+        # assumes one code sample to test per readme
         readme_sample = None
         with open(readme_path, encoding="UTF-8") as readme_file:
             match_sample = re.search(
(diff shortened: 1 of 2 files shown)


manoelmarques:
@mtreinish We have sample code in the README.md that someone always has to remember to keep up to date. 
I created a unit test that gets the code from the README.md and runs it to make sure it works.
Since it uses python exec, I would like to know your opinion. As for malicious code someone can create a PR with malicious code anytime. It is up to us to review it. Also, first time PR creators need our approval to even run the CI.
I didn't think there was a big harm but would like to know your opinion.

test/test_readme_sample.py
```diff
@@ -18,95 +18,69 @@
...
+
+        readme_sample = None
+        with open(readme_path, encoding="UTF-8") as readme_file:
+            match_sample = re.search(
+                "```python.*```",
+                readme_file.read(),
+                flags=re.S,
+            )
+            if match_sample:
+                readme_sample = match_sample.group(0)[9:-3]
```
mrossinek, line 68: Can you add a comment what `[9:-3]` is doing? I get that it's stripping the markdown code block syntax but it took my a second :sweat_smile:

test/test_readme_sample.py
```diff
@@ -18,95 +18,69 @@
...
+            self.fail(msg=f"{readme_name} not found at {readme_path}")
+            return
+
+        readme_sample = None
+        with open(readme_path, encoding="UTF-8") as readme_file:
+            match_sample = re.search(
+                "```python.*```",
+                readme_file.read(),
+                flags=re.S,
+            )
```
mrossinek, line 66: This currently assumes that there is a single match, right? While this is fine (at least currently), maybe write a comment about this such that we know this needs updating in case this ever changes.

test/test_readme_sample.py
```diff
@@ -18,95 +18,69 @@
...
+            if match_sample:
+                readme_sample = match_sample.group(0)[9:-3]
+
+        if readme_sample is None:
+            self.skipTest(f"No sample found inside {readme_name}.")
+            return
+
+        writer = Writer()
+        sys.stdout = writer
```
mrossinek, line 75: I've started using [contextlib.redirect_stdout](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout) for this purpose. See for example https://github.com/Qiskit/qiskit-nature/blob/main/test/results/test_electronic_structure_result.py#L30-L31

This will remove the need for the `Writer` class :+1:

test/test_readme_sample.py
manoelmarques: yes, we have only one sample per readme.

mrossinek (approved):
LGTM! Thanks! :slightly_smiling_face:

woodsp-ibm (approved):
Thanks Manoel, now we don't have to keep the test case and the readme in sync by manual copy/paste as before so it truly will always be testing the readme and not potentially and out-of-date or altered copy (as was the case before this PR where the test was using the new driver but the readme still was using the old deprecated one that in fact is no longer present in main so this sample failed to run against main)

PR #544, opened 2022-02-09
Define FermionicOp.terms

### LINKED ISSUE #541: Iterate over FermionicOp terms
### What should we add?

I was trying to figure out how to iterate over the individual terms of a FermionicOp. I found the `FermionicOp.to_list` method, but that returns the term as strings like `"+_2 -_0"` so it's not useful for accessing the indices and operators programmatically. Apparently the `_data` attribute does store the terms in a useful format, but it's private and uses the private `_FermionLabelPrimitive` class. If I'm missing something, please let me know.

There should be a way to iterate over the individual terms in a programmatically useful format. I think it should be done by adding an `__iter__` method that yields the contents of the internal `_data` object, but in a format that has a publicly supported API. Perhaps
```python
def __iter__(self) -> Iterable[Tuple[List[Tuple[int, int]], complex]]
```
where the inner Tuple would be (action, index) with action = 1 for create and 0 for annihilate.

kevinsung:
### Summary
Fixes #541.

### Details and comments
This is possible now that #506 has finalized the internal data structure.


mrossinek (commented):
LGTM! @ikkoham could you please take a look, too?

ikkoham (commented):
Basically LGTM.

This is my personal opinion:
1. The `__iter__` method expects to return the original class, i.e. FermionicOp itself. How about making the method like `data_iter()`? Terra has `label_iter()` and `matrix_iter` method.
2. I suggest to use `CustomIter` class in Terra. https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/quantum_info/operators/custom_iterator.py. It has a custom `__repr__`, which is useful to specify where the iterator come from. But this is optional.

kevinsung:
>The `__iter__` method expects to return the original class, i.e. FermionicOp itself.

@ikkoham What do you mean by this? `iter(obj)` is certainly not expected to yield objects of the same type as `obj`.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -346,6 +346,10 @@ def __repr__(self) -> str:
             ")"
         )
 
+    def __iter__(self) -> Iterable[tuple[list[tuple[str, int]], complex]]:
```
dlasecki, line 349: I think we usually use `Tuple` and `List` for type hints which come from the Typing module.

kevinsung: That's true, but @ikkoham changed that in this file in https://github.com/Qiskit/qiskit-nature/pull/506 so I did it this way to be consistent. I don't know why @ikkoham changed it.

woodsp-ibm: The lowercase variants, i.e. using tuple etc is a change for 3.9 onwards https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections which avoids duplicate similar names from Typing and confusion that results from using the different ones in the different contexts. This is better I think and I imagine we will transition the whole code base over. This has `from __future__ import annotations` to allow its use on earlier versions. I believe this became possible one we dropped 3.6 as supported where before we stuck to Typing.

woodsp-ibm (commented):
Given this is an additional feature that can be used I think maybe a reno release note to lets users know. 

Also should we be considering this for the other operators here to have them consistent in this regard. Maybe it should be a base class aspect that needs to be overridden so we ensure that we have this for any new operators goings forwards too.

And perhaps this can be be added in the main docstring somewhere to show this support and the fact that it returns label, coeff tuples.

kevinsung:
>Maybe it should be a base class aspect that needs to be overridden so we ensure that we have this for any new operators goings forwards too.

This would be counter-intuitive because `Iterable` is a protocol, not an abstract base class.

>And perhaps this can be be added in the main docstring somewhere to show this support and the fact that it returns label, coeff tuples.

I've updated the class docstring to document this.

test/operators/second_quantization/test_fermionic_op.py
```diff
@@ -141,6 +141,16 @@ def test_init_from_tuple_label(self):
...
 
+    def test_iter(self):
+        """Test __iter__"""
+        op = FermionicOp([("+", 1.0), ("N", 2.0), ("-+", 3.0)], display_format="dense")
+        expected = [
+            ([("+", 0)], 1.0),
+            ([("+", 0), ("-", 0)], 2.0),
+            ([("-", 0), ("+", 1)], 3.0),
+        ]
+        self.assertEqual(list(op), expected)
```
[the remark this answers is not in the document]
woodsp-ibm: The unit tests tests the public API. In practice you would not normally call on list on it since it removes any advantage of an iterator. In this case since its all in memory anyway though it really is no different. I just think it clearer when it states it is going to test iter that you call `op.iter()` in the unit test. You may find it easier to write the test after coercing it to a list, that's fine.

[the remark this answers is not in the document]
kevinsung: The method is called `__iter__`, not `iter`. In practice, you would never call that directly. It's a hidden dunder method. Most commonly you would do
```python
for term, coeff in op:
    ...
```

kevinsung:
>the parent of FermionicOp declare it (i.e. __iter__) there, albeit abstract

In my opinion this is counterintuitive because that would be overriding the design of Python itself, where `__iter__` is a protocol method, not an abstract base class method. Besides, I don't see why other subclasses should be forced to implement the Iterable protocol.

woodsp-ibm:
> In my opinion this is counterintuitive because that would be overriding the design of Python itself, where __iter__ is a protocol method, not an abstract base class method. Besides, I don't see why other subclasses should be forced to implement the Iterable protocol.

Its not overriding the design just expecting the classes all to implement the and assuring they are doing so. There is certain function defined in the base class that allows you to use any second quantized operator the same way. E.g. to_list() is there. All I am saying is that given FermionicOp supports iterable now it seemed like a good thing and something I was suggesting we should have in the other classes there too to keep things consistent for an end user in the way they can consume it. It seemed that this could easily be common like the to_list() there.

test/operators/second_quantization/test_fermionic_op.py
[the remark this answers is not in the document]
woodsp-ibm: We have people at different levels of Python understanding - I just think making it explicit can be clearer that its obviously testing what it says - the code doing list cast in Python no doubt checks for that built-in and can use that implicitly. It maybe clear to you what happening but I just think its better to be explicit in the unit test there rather than relying on implicit behavior.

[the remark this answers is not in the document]
kevinsung: `list(iter(op))` is not idiomatic Python. We should strive to write idiomatic Python, even in the tests. If the reader doesn't know what `list` does, they can easily look it up.

kevinsung:
@woodsp-ibm I think the question of making other classes implement Iterable should be handled as a separate issue. It shouldn't block this PR.

woodsp-ibm:
> I think the question of making other classes implement Iterable should be handled as a separate issue. It shouldn't block this PR.

That's fine. Since it seemed to be useful in FermionicOp it looked like to me that this could be the case in the other second quantized operators here too, and it would make use of them consistent. Since it seems to be a simple change it could be labelled as Good First Issue for someone else to take a look at.

As I mentioned above though I think we should have a reno stating that this new feature is available for users of FermioicOp.

kevinsung:
>As I mentioned above though I think we should have a reno stating that this new feature is available for users of FermioicOp.

@woodsp-ibm Should this be kept track of somewhere? Not in this PR, right?

woodsp-ibm:
PRs that affect the public API, including adding new features, we should have a reno release note in with the PR. The final combined document made at release will note comprises the total text from the set of PRs that are in the release. reno already gathers together the different aspects of the notes, fixes, new features etc. 

If another PR comes along in the future, to add similar to the other operators, then perhaps we'll have them simply update this existing reno if it will be in the same release, if it seems like we would really want just the one piece of text covering it all.

kevinsung:
@woodsp-ibm Thanks, I have added a release note.

kevinsung:
By the way @woodsp-ibm it turns out that Iterable _is_ an abstract base class that you can inherit. So your suggestion can be easily implemented by making the parent class inherit from Iterable. I will create a separate issue to discuss this.

kevinsung:
https://github.com/Qiskit/qiskit-nature/issues/562

ikkoham:
Then how about providing a gettter for `FermionicOp._data`? Is that not enough?

> What do you mean by this? iter(obj) is certainly not expected to yield objects of the same type as obj.

Hmmm, I think my point was wrong. It certainly doesn't have to be the same. 

For example, `SparsePauliOp` in Terra works like:
```python
op = SparsePauliOp.from_list([("XXX", 1), ("YYY", 2)])
for i in op:
    print(i)
```
```
SparsePauliOp(['XXX'],
              coeffs=[1.+0.j])
SparsePauliOp(['YYY'],
              coeffs=[2.+0.j])
```
Opflow is also
```python
op = PauliSumOp.from_list([("XXX", 1), ("YYY", 2)])
for x in op:
    print(x)
    print(type(x))
```
```
1.0 * XXX
<class 'qiskit.opflow.primitive_ops.pauli_sum_op.PauliSumOp'>
2.0 * YYY
<class 'qiskit.opflow.primitive_ops.pauli_sum_op.PauliSumOp'>
```

You may feel this behavior is non-trivial. We need to discuss whether it is always necessary to make it Iterable. Personally, I feel that many users should not be aware of internal data, and the behavior of `iter` is non-trivial.

kevinsung:
>Then how about providing a gettter for FermionicOp._data? Is that not enough?

No, it is not enough. Being able to iterate through the terms is crucial functionality. Currently the internal code uses `to_list` to achieve this functionality. As I explain in #541 this is not flexible enough.

>For example, SparsePauliOp in Terra works like:

The behavior of SparsePauliOp seems highly unusual to me. I don't think we should emulate it.

ikkoham:
I feel
```python
fer_op = FermionicOp(...)
for o in fer_op:
    some processing
```
and
```python
for o in fer_op.data:
    some processing
#or 
for o in iter(fer_op.data):
    some processing
```
are almost same. What's the difference?

Whether to unify the interface or not should be discussed. I did not creat the SparsePauliOp iterator, but I think it is better in that users can use it without knowing the internal data structure. The reason why we don't have a getter of data as public is so that people don't have to be aware of the internal data structure.

dlasecki:
> I feel
> 
> ```python
> 
> fer_op = FermionicOp(...)
> 
> for o in fer_op:
> 
>     some processing
> 
> ```
> 
> and
> 
> ```python
> 
> for o in fer_op.data:
> 
>     some processing
> 
> #or 
> 
> for o in iter(fer_op.data):
> 
>     some processing
> 
> ```
> 
> are almost same. What's the difference?
> 
> 
> 
> Whether to unify the interface or not should be discussed. I did not creat the SparsePauliOp iterator, but I think it is better in that users can use it without knowing the internal data structure. The reason why we don't have a getter of data as public is so that people don't have to be aware of the internal data structure.

If there exists a "correct" or default way of iterating over a complex Python object, it is always advised to implement `__iter__`. It saves users a headache of digging into the structure of the implementation and ensures an iteration is performed exactly as the author of the code intended. It might save users from unobvious errors and usually makes code cleaner.

kevinsung:
>I did not creat the SparsePauliOp iterator, but I think it is better in that users can use it without knowing the internal data structure.

In this case, the internal data structure contains precisely the information in needed, in the format needed, for crucial functionality. The question is not whether users should be able to access that information, but how that access should be supported. @dlasecki explained nicely why `__iter__` should be defined, so I won't repeat it.

I think there is a valid concern that the data format is somewhat convoluted. Indeed, `tuple[list[tuple[str, int]], complex]` is a complicated and confusing type to unpack. In the current state of the library, this is the best we can do. Improving this would require defining a class to represent a single fermionic term (not a linear combination). Then the type could be something like `tuple[FermionicTerm, complex]`.

I would strongly disagree with having `__iter__` yield objects of the same type, like SparsePauliOp does. It is both more confusing and less useful. Iterating through an object gets things that are `in` it, which should be a simpler type. Note that the behavior of SparsePauliOp implies that it satisfies a recursive type of the form `SparsePauliOp = Iterable[SparsePauliOp]` which goes to show that it is not trivial.

ikkoham:
@dlasecki Thanks. I can certainly understand that. But the problem here, as the discussion is split, is that there is no "correct" or default way. I'm concerned that the user must always understand what the iterator is providing, which may result in lower usability.

Personally, I'm fine with either. (I won't use the iterator because it is too complicated. But I can understand some users want to use the iterator. So let's define iterator.)
But users who are familiar with Qiskit feel the same type iterator is more natural. If users need other iterators, we can define multiple iterators, depending on the purpose. e.g. SparsePauliOp has label_iter() and matrix_iter(). I don't want users to be aware of the internal data structure because it is too complicated.

This is just a preference issue.

test/operators/second_quantization/test_fermionic_op.py
```diff
@@ -141,6 +141,16 @@ def test_init_from_tuple_label(self):
         ]
         self.assertListEqual(actual.to_list(), desired)
 
+    def test_iter(self):
```
ikkoham, line 144: Could you check the type of iter(op)? If it returns `_data`, this past will be passed.

ikkoham (approved):
LGTM

mrossinek:
I'd like to try and summarize some observations here:

Personally, I find the behavior of Qiskit Terra _very_ unnatural. As @kevinsung pointed out:
> Note that the behavior of SparsePauliOp implies that it satisfies a recursive type of the form `SparsePauliOp = Iterable[SparsePauliOp]` which goes to show that it is not trivial.

That said, I do agree with @ikkoham that we should align our code with Qiskit Terra as to not confuse end-users.

Therefore, I think a better approach than this implementation is:
1. improve `to_list()` instead of adding another means of iteration (@kevinsung you even suggest in #562 to deprecate `to_list` so why not instead improve that)
2. check with the Terra team the reasons for their implementation and align ourselves accordingly.

A final note: In Qiskit Terra the iteration implementation appears to rely on `__getitem__` rather than `__iter__`. Not sure if this is relevant to anyone but I wanted to bring it up.

jakelishman:
@chriseclectic wrote the initial `SparsePauliOp` in Qiskit/qiskit-terra#3932, but from looking at that, I don't think there was any discussion of `__getitem__`.  Chris might be able to say more, but it's possible that `__getitem__` was originally meant simply to index into the operator, and gaining iteration from that was a side-effect (but that's a documented "feature" of Python).  In our case there's no need for us to define `__iter__`, though - it would be identical to the `__getitem__` version - so we're fine to rely on the implicit one.

From my perspective, it's also strange that `SparsePaulOp.__getitem__` returns `SparsePauliOp`, but I suspect that the original reason is that this is the only symplectic operator in Terra that has a `coeffs` field; we can't directly return `Pauli` because if `coeffs` doesn't contain a valid symplectic phase it'd change the operator.  Perhaps if we were starting again we'd return a 2-tuple `(Pauli, complex)`, but it's going to be very hard for us to change the API now without very unpleasant breaking changes for users, and I'd be very cautious about setting out down that path.

If you don't have that problem, it's probably better to respect the general type-system perspective that iterating or getting an item from a collection-like object should reduce the nested types or the logical dimensionality of the object by one.

jakelishman:
Just to be clear, Python itself doesn't always respect this type-system "rule":
```python
>>> my_object = "hello, world"
>>> type(my_object[0]) == type(my_object)
True
>>> single_char = "h"
>>> single_char[0] is single_char
True
```

mrossinek:
> Perhaps if we were starting again we'd return a 2-tuple `(Pauli, complex)`

Agreed. I suppose that's what `label_iter` is for (to some degree at least)

> but it's going to be very hard for us to change the API now without very unpleasant breaking changes for users,
<...>
If you don't have that problem, it's probably better to respect the general type-system perspective that iterating or getting an item from a collection-like object should reduce the nested types or the logical dimensionality of the object by one.

I agree that this probably makes more sense. Although I wonder what the better approach is in terms of aligning the implementation of `iter(...)` and `to_list()`. Having both only seems warranted in a case like Terra's where they are obviously different. However, having said that we don't really need Terra's way of defining `__getitem__`, I wonder what the cleanest approach is in terms of user expectation in the Qiskit framework.

qiskit_nature/operators/second_quantization/fermionic_op.py
jakelishman: A minor point, but the return value of `__iter__` is `Iterator`, not `Iterable` - the sole difference being that `Iterator` has the `__next__` method as well.

ikkoham:
> we can't directly return Pauli because if coeffs doesn't contain a valid symplectic phase it'd change the operator

exactly. so I suggested that PauliList return Pauli when we made PauliList.

I know many people feel this is strange (I [said](https://github.com/Qiskit/qiskit-nature/pull/544#issuecomment-1041575248) you may feel non-trivial) and previously causes some bug https://github.com/Qiskit/qiskit-terra/issues/5408. I made sure this bug is not a problem for FermionicOp.

kevinsung:
>That said, I do agree with @ikkoham that we should align our code with Qiskit Terra as to not confuse end-users.

There is already a very prominent precedent for diverging from Qiskit Terra's design choices: using big-endian instead of little endian. I think that the justification in this case for diverging is stronger (and unlikely to cause nearly as much confusion). I'm not suggesting that recursive types are always inappropriate. But they are in this case, and as @jakelishman said about `SparsePauliOp`, things should have been done differently from the beginning. I see this as an opportunity to learn from Qiskit rather than copy its flaws. I think the best course of action is to design a lower-level class analogous to `Pauli`, but which is appropriate for being used with `FermionicOp`. Perhaps the term should store its coefficient together with it, but I'm not sure that's necessary.

@ikkoham Would you be okay with `__iter__` if it yielded a type such as `tuple[FermionicTerm, complex]`?

>@kevinsung you even suggest in https://github.com/Qiskit/qiskit-nature/issues/562 to deprecate to_list so why not instead improve that

@mrossinek There are two reasons to deprecate `to_list` rather than improve it:
1. Iterable is a more flexible type than List (you can iterate without constructing the whole list in memory).
2. `list` is already a built-in function and `list(op)` is the idiomatic way to turn an object into a list.

mrossinek:
I agree with your summary. However, we should take a moment to consider whether having slicing functionality is a worthwhile addition. If yes, is this possible with an `__iter__` implementation or does this require `__getitem__`.

Once `iter(...)` works, `to_list()` can be deprecated.

kevinsung:
> However, we should take a moment to consider whether having slicing functionality is a worthwhile addition.

I don't see any use case for slicing because typically the ordering of terms in the FermionicOp is arbitrary.

>If I take a slice, will it return a list of these list[tuple[FermionicTerm, complex]]?

See above. I do not propose to add slicing, only iteration.

>I don't think so. For example, numpy.ndarray has tolist() and list(np.ndarray) is simply list[ndarray].

`numpy.ndarray.tolist` is a non-trivial function that converts a numpy array into a nested list. There is no need for such a function in our situation.

>to_list() can take the parameters (arguments), but __iter__ cannot. How do you pass the option? (or the option is unnecessary?)

No options are needed.

kevinsung:
@mrossinek @ikkoham In light of the controversy over defining `__iter__`, I have renamed the method to `terms`.

mrossinek (dismissed):
LGTM now, thanks for the compromise! :+1:

kevinsung:
@mrossinek if you merge https://github.com/Qiskit/qiskit-nature/pull/607 first then this won't be a breaking change

test/operators/second_quantization/test_fermionic_op.py
```diff
@@ -141,6 +141,16 @@ def test_init_from_tuple_label(self):
...
 
+    def test_terms(self):
+        """Test terms"""
+        op = FermionicOp([("+", 1.0), ("N", 2.0), ("-+", 3.0)], display_format="dense")
+        expected = [
+            ([("+", 0)], 1.0),
+            ([("+", 0), ("-", 0)], 2.0),
+            ([("-", 0), ("+", 1)], 3.0),
+        ]
+        self.assertEqual(list(op.terms()), expected)
```
woodsp-ibm, line 152: The expected terms here are in the  same order as the constructor. Is this an expected behavior or might they be in a different order - terms just says it gives an iterable. Would a better test here more say that the set of terms is the same as the set provided such that order is irrelevant? Or is the order part of the expected contract of the terms method?

kevinsung: Good catch, the order doesn't matter, so this test should be comparing the sets rather than the lists. However, I can't use `set` here until #607 is merged, so please merge that first.

kevinsung: Done.

mrossinek (approved):

PR #549, opened 2022-02-10
Apply `simplify` more often in `QubitMapper.mode_based_mapping`

### LINKED ISSUE #545: Jordan-Wigner transform slow on parity operator

The Jordan-Wigner transform generally has good performance. For example, it is usually faster than OpenFermion. However, I have found one case in which it performs very badly. This is the case of the parity operator. The Jordan-Wigner transform of the parity operator is just a single Pauli string of all Z's. So, in practice one wouldn't actually need to run this. However, I think it is worth investigating why it is so slow in this case and whether that can be fixed. Perhaps this performance issue affects other cases too.

In the benchmark below, Qiskit takes up to 14x as long as OpenFermion, and this factor appears to increase with the number of modes. I ran the benchmark up to 12 modes, but that took a lot of time and RAM so I edited the code below to only go up to 11. I do not plot the generation times because I checked that they are negligible (less than 0.1 seconds in all cases).

Benchmark code:
```python
import timeit
import matplotlib.pyplot as plt
import openfermion as of

def parity_op(n_modes: int) -> FermionicOp:
    op = FermionicOp.one(n_modes)
    for i in range(n_modes):
        op @= FermionicOp.one(n_modes) - 2 * FermionicOp(f"N_{i}")
    return op

def parity_op_of(n_modes: int) -> of.FermionOperator:
    op = of.FermionOperator.identity()
    for i in range(n_modes):
        op *= of.FermionOperator.identity() - 2 * of.number_operator(n_modes, i)
    return op

n_modes_range = range(7, 12)

generation_times_qiskit = []
generation_times_openfermion = []
transformation_times_qiskit = []
transformation_times_openfermion = []

for n_modes in n_modes_range:
    # generate Qiskit
    t0 = timeit.default_timer()
    parity = parity_op(n_modes)
    t1 = timeit.default_timer()
    generation_times_qiskit.append(t1 - t0)

    # generate OpenFermion
    t0 = timeit.default_timer()
    parity_of = parity_op_of(n_modes)
    t1 = timeit.default_timer()
    generation_times_openfermion.append(t1 - t0)

    # map Qiskit
    t0 = timeit.default_timer()
    parity_jw = JordanWignerMapper().map(parity)
    t1 = timeit.default_timer()
    transformation_times_qiskit.append(t1 - t0)

    # map OpenFermion
    t0 = timeit.default_timer()
    parity_jw_of = of.jordan_wigner(parity_of)
    t1 = timeit.default_timer()
    transformation_times_openfermion.append(t1 - t0)

ax = plt.subplot()
ax.plot(n_modes_range, transformation_times_qiskit, label='qiskit')
ax.plot(n_modes_range, transformation_times_openfermion, label='openfermion')
ax.set_xticks(n_modes_range)
ax.set_xlabel('# modes')
ax.set_ylabel('Time (s)')
ax.set_title('Transformation time')
ax.legend()

plt.show()
```
Output (edited to go up to 12 modes):
image

Qiskit Nature version: be91c8e9657082d8670fd16fd79d02b565fa9f9c
Qiskit Terra version: fcec842f1de9fd12120e30a1bf73bf7c52b1bf81
Python version: 3.9.10

t-imamichi:
### Summary

This workaround can resolve the issue #545.
But it may get slower for other cases because it applies `simplify` too many times.

Update:
I added a threshold of number of Pauli strings to apply `simplify`.
`simplify` pays off only if there are many Pauli strings.

Fixes #545

### DIFF
--- a/qiskit_nature/mappers/second_quantization/qubit_mapper.py
+++ b/qiskit_nature/mappers/second_quantization/qubit_mapper.py
@@ -141,7 +141,9 @@ class QubitMapper(ABC):
                     raise QiskitNatureError(
                         f"FermionicOp label included '{char}'. Allowed characters: I, N, E, +, -"
                     )
-                ret_op = ret_op.simplify()
+                if ret_op.size >= 100:
+                    # `simplify` pays off only if there are many Pauli strings
+                    ret_op = ret_op.simplify()
             ret_op_list.append(ret_op)
 
         return PauliSumOp(SparsePauliOp.sum(ret_op_list).simplify())
--- a/qiskit_nature/mappers/second_quantization/qubit_mapper.py
+++ b/qiskit_nature/mappers/second_quantization/qubit_mapper.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021.
+# (C) Copyright IBM 2021, 2022.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
@@ -141,6 +141,7 @@ class QubitMapper(ABC):
                     raise QiskitNatureError(
                         f"FermionicOp label included '{char}'. Allowed characters: I, N, E, +, -"
                     )
+                ret_op = ret_op.simplify()
             ret_op_list.append(ret_op)
 
         return PauliSumOp(SparsePauliOp.sum(ret_op_list).simplify())


t-imamichi:
I also made a PoC of performance improvement of `SparsePauliOp.simplify` for large number of qubits.
I welcome your feedback. https://github.com/Qiskit/qiskit-terra/pull/7656

woodsp-ibm:
In the original referenced issue there was a plot of performance - how does that look now with the above change?

t-imamichi:
I tried Kevin's parity operator and a random operator as follows.
```python
import random
from timeit import timeit

from qiskit_nature.mappers.second_quantization import JordanWignerMapper, BravyiKitaevMapper


def random_op(k, n):
    random.seed(123)
    op = sum(
        (
            FermionicOp("".join(random.choices("+-ENI", k=k)), display_format="dense")
            * random.random()
        )
        for _ in range(n)
    )
    return op


def parity_op(n_modes: int) -> FermionicOp:
    op = FermionicOp.one(n_modes)
    for i in range(n_modes):
        op @= FermionicOp.one(n_modes) - 2 * FermionicOp(f"N_{i}", display_format="dense")
    return op


def bench(op):
    mapper = JordanWignerMapper()
    qubit_conv = QubitConverter(mapper)
    print(f"{timeit(lambda: qubit_conv.convert(op), number=1)} sec")


print("random")
op = random_op(12, 100)
bench(op)

print("\nparity")
op = parity_op(10)
bench(op)
```

main
```
random
12.206206522 sec

parity
11.337571101000002 sec
```

this PR
```
random
0.4637313729999999 sec

parity
0.983379751 sec
```

qiskit_nature/mappers/second_quantization/qubit_mapper.py
```diff
@@ -141,6 +141,9 @@ def mode_based_mapping(
                     raise QiskitNatureError(
                         f"FermionicOp label included '{char}'. Allowed characters: I, N, E, +, -"
                     )
+                if ret_op.size >= 100:
```
dlasecki, line 144: How did you arrive at 100 being a good choice here? Did you perform experiments?

t-imamichi: I tested it with the microbenchmark https://github.com/Qiskit/qiskit-nature/pull/549#issuecomment-1039787299. But, I want actual cases hopefully as part of app-benchmarks.

dlasecki: Cool. Perhaps we could include this information as a comment here.

woodsp-ibm:
I see you put a comment in #560

> The technique #549 can perhaps make the qubit mapping even faster. So, I leave it for discussion.

How do the timings you ran above look with 560 applied as well as this change - you have the timings in 560 for that alone, does including this proposed change too make much of a difference where it is worth considering?

t-imamichi:
I quickly tried this PR + #560. The combination does not look faster. So, I will close this PR when #560 is merged.

this PR + #560
```
random
0.3882492110000001 sec

parity
0.602769689 sec
```

only #560 
```
random
0.27690272000000005 sec

parity
0.514391453 sec
```

woodsp-ibm:
Thanks for checking, indeed it does seem to slow it down after #560 adding this

PR #554, opened 2022-02-11
Update `load_from_hdf5`

mrossinek:
### Summary

The documentation of the `load_from_hdf5` suggested that one could
immediately obtain a driver result from it. However, in reality, this
would have assigned a generator method instead of the actual result.
In order to be aligned with the `save_to_hdf5` implementation, this
commit changes the `load_from_hdf5` function to return the first
constructed object and ignore any further ones. In practice this should
be sufficient because the Qiskit Nature stack is written such that only
a single object will be stored per HDF5 file.


manoelmarques (approved):

qiskit_nature/hdf5.py
```diff
@@ -147,9 +150,23 @@ def load_from_hdf5(
...
+            encountered in the file.
     """
+    ret: HDF5Storable = None
     with h5py.File(filename, "r") as file:
-        yield from _import_and_build_from_hdf5(file, skip_unreadable_data=skip_unreadable_data)
+        for obj in _import_and_build_from_hdf5(file, skip_unreadable_data=skip_unreadable_data):
+            if ret is not None:
+                msg = (
+                    "Encountered more than one HDF5Storable object in the file! "
+                    "This is not supported and only the first one will be returned."
```
woodsp-ibm, line 162: I did not see this logic around more than one object being tested. Maybe it would require creating some custom hdf5 test file for this. In fact if I look at this files coverage its just under 60% with none of the exception conditions  in the import_and_build_from_hdf5 being covered. Maybe again some custom hdf5 file resources we can cover these with for CI?

pbark (approved):
LGTM!

PR #558, opened 2022-02-14
Updated UCC docs

### LINKED ISSUE #489: What is the default reference state for the UCCSD ansatz?
Hi, this issue is mainly a question that pertains to the documentation for the UCC and UCCSD ansatzes. The unitary coupled cluster ansatz is typically defined with respect to some specified reference state. This could be the Hartree Fock state, but it does not necessarily have to be. For UCCSD, the T_1 and T_2 operators consist of a summation over an index of orbitals that are unoccupied in the reference state and an index of orbitals which are occupied in the reference state. What's not clear to me from the documentation on `UCC` and `UCCSD` is how Qiskit determines which orbitals are occupied and which are unoccupied when creating an instance of either of these two class objects, given that it's not necessary for you to explicitly provide it a reference state. 

There are other constraints that you must provide, such as restricting the Ansatz to single/single-double excitations, the number of electrons, the number of spin electrons, ect... but these by themselves do not seem to be enough to uniquely specify a reference state or which spin orbitals are unoccupied/occupied. Does Qiskit default to the generalized version of UCCSD (where all excitation operators subject to the other constraints are included) or does it implicitly default to some other reference state? The former seems strange since it would effectively override the default `generalized=False` argument. Defaulting to some reference state unspecified in the documentation (probably something reasonable such as the Hartree Fock state) seems opaque and less than ideal since it assumes that the user has this particular reference state in mind. Given that running the line of code:

`print(generate_fermionic_excitations(num_spin_orbitals=6, num_particles=(1,1), num_excitations=1))`

returns:

`[((0,), (1,)), ((0,), (2,)), ((3,), (4,)), ((3,), (5,))]`

it seems that the likely default reference state for `UCCSD` is the Hartree Fock state, but any insight or information on what is going on under the hood would be greatly appreciated.

javabster:
### Summary
Added a couple extra sentences to the documentation for clarification on the reference state used for `UCC` Ansatz

fixes #489 

### Details and comments
html formatting:

Screenshot 2022-02-14 at 4 36 26 PM
Screenshot 2022-02-14 at 4 36 40 PM


qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -36,7 +36,8 @@ class UCC(EvolvedOperatorAnsatz):
     This Ansatz is an `EvolvedOperatorAnsatz` given by :math:`e^{T - T^{\dagger}}` where
     :math:`T` is the *cluster operator*. This cluster operator generally consists of excitation
     operators which are generated by
-    :meth:`~qiskit_nature.circuit.library.ansatzes.utils.generate_fermionic_excitations`.
+    :meth:`~qiskit_nature.circuit.library.ansatzes.utils.generate_fermionic_excitations`, using
+    a `HartreeFock` reference state by default.
```
mrossinek, line 40:
```suggestion
    :meth:`~qiskit_nature.circuit.library.ansatzes.utils.generate_fermionic_excitations`.
    This method constructs the requested excitations based on a `HartreeFock` reference state by
    default.
```

qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -164,7 +165,10 @@ def __init__(
                 particles.
             preserve_spin: boolean flag whether or not to preserve the particle spins.
             reps: The number of times to repeat the evolved operators.
-            initial_state: A `QuantumCircuit` object to prepend to the circuit.
+            initial_state: A `QuantumCircuit` object to prepend to the circuit. Note that setting
+                this argument does not automatically affect how the excitations are generated. If
+                this argument is set to a state other than `HartreeFock` the `excitations` argument
+                will need to be matched accordingly.
```
mrossinek, line 171:
```suggestion
            initial_state: A `QuantumCircuit` object to prepend to the circuit. Note, that this
                setting does _not_ influence the `excitations`. When relying on the default
                generation method (i.e. not providing a `Callable` to `excitations`), these will
                always be constructed with respect to an `HartreeFock` reference state.
```

mrossinek (commented):
Some rephrasing suggestions.

I wonder whether we should extend the explanation added to the first docstring here also to the docstring of the `generate_fermionic_excitations` method.

javabster:
> Some rephrasing suggestions.
> 
> I wonder whether we should extend the explanation added to the first docstring here also to the docstring of the `generate_fermionic_excitations` method.

I ended up moving most of what was in the first docstring into the function docstring as I think it's all useful info that would be good for a regular user to be able to find 😄  lmk if there's any further rephrasing needed

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
```diff
@@ -106,6 +96,19 @@ def generate_fermionic_excitations(
...
+    be set to `False` in order to disable the inclusion of alpha-spin or beta-spin excitation,
+    respectively. `max_spin_excitation` takes an integer value which defines the maximum number of
+    excitations that can occur within the same spin. Thus, setting `max_spin_excitation=1` and
+    `num_excitations=2` yields only those double excitations which do not excite the same spin species
+    twice. `generalized` is another boolean flag which enables generalized excitations which are
+    effectively ignoring the spin orbital occupancies. Excitations will be produced based on an
+    initial `Hartree-Fock` occupation by default unless `generalised` is set to `True`. Therefore,
+    the excitations are only determined based on the number of spin orbitals and are independent from
+    the number of particles.
+
```
woodsp-ibm, line 111: An overall comment. This seems to be a bit of a duplication of text that is Args. I am wondering, if further explanation is needed (which seems to be the case here), perhaps improve the args text a bit, and also if it relates to values of args interacting maybe just having some example calls here with explanation whats happening and the corresponding output might be more helpful, easier to understand? These can be done as doctest so they would be examples that will be checked by CI which verifies the output matches that stored in the example when run.

For this text `The method must be called for each type of excitation (singles, doubles, etc.) that is to be considered in the Ansatz.` I think it might be better to just end it at considered (ie. drop in the Ansatz). In the past in Aqua we used a similar excitation list generator in places other than an ansatz where it applied (e.g. as part of the computation of mp2 perturbation coeffs). I think this here is more of a common utility even though its nested under the ansatzes.

mrossinek: Yep I agree with @woodsp-ibm here.

For some context: I had discussed with @javabster offline that we should check the contents of the top-level docstring in this file for information which should be moved/copied into this methods docstring. The reason being that the top-level docstring is not actually included in the documentation because only this specific method is exported in the `__init__` of this module. If you think that this ought to be changed, @woodsp-ibm, please let me know. I did not see a reason for exposing `get_{alpha|beta}_excitations` so I tend to say that this is not needed.

That being said, @javabster if you could try to deduplicate the information in this docstring and maybe add some examples, that would be great. You can see an example of how an example (:sweat_smile:) is added to a docstring here: https://github.com/Qiskit/qiskit-nature/blob/be91c8e9657082d8670fd16fd79d02b565fa9f9c/qiskit_nature/operators/second_quantization/fermionic_op.py#L157-L166

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
```diff
@@ -106,6 +96,12 @@ def generate_fermionic_excitations(
     """Generates all possible excitations with the given number of excitations for the specified
     number of particles distributed among the given number of spin orbitals.
 
+    The method must be called for each type of excitation (singles, doubles, etc.) that is to be
+    considered in the Ansatz. Excitations will be produced based on an initial `Hartree-Fock`
+    occupation by default unless `generalised` is set to `True`. Therefore, the excitations are
```
mrossinek, line 101:
```suggestion
    occupation by default unless `generalized` is set to `True`. Therefore, the excitations are
```
I know.. British English :smile: But in this case it's important because here we're actually referring to the argument name.

mrossinek: I'm also not sure about `Therefore` being used here. It should be stressed that the excitations are only **in**dependent from the number of particles if and only if `generalized=True`. The `Therefore` reads a bit differently imo

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
```diff
@@ -128,6 +124,56 @@ def generate_fermionic_excitations(
...
+                generate_fermionic_excitations
+            )
+
+            num_excitations = 1
+            num_spin_orbitals = 4
+            num_particles = [1,1]
+
+            excitations = generate_fermionic_excitations(
+                num_excitations, num_spin_orbitals, num_particles
+            )
```
mrossinek, line 143:
```suggestion
            excitations = generate_fermionic_excitations(1, 4, (1, 1))
```
I think for these examples we can inline the arguments.
Also note that `num_particles` should be a tuple and not a list.

Same holds below but I will not comment on each occurrence now.

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
```diff
@@ -128,6 +124,56 @@ def generate_fermionic_excitations(
         The list of excitations encoded as tuples of tuples. Each tuple in the list is a pair. The
         first tuple contains the occupied spin orbital indices whereas the second one contains the
         indices of the unoccupied spin orbitals.
+
+    Examples:
+        Generate excitations with basic inputs.
+
+        .. jupyter-execute::
```
mrossinek, line 131: Did you make sure that the execution works as expected and adds the output to the documentation?

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
```diff
@@ -128,6 +124,56 @@ def generate_fermionic_excitations(
...
+
+            excitations = generate_fermionic_excitations(
+                num_excitations, num_spin_orbitals, num_particles
+            )
+
+        Generate excitations limiting the maximum number of excitations per spin species.
+
+        .. jupyter-execute::
+
+            from qiskit_nature.circuit.library.ansatzes.utils.fermionic_excitation_generator import (
```
mrossinek, line 149: If I recall correctly the `jupyter-execute` blocks are executed sequentially and context-aware. That means that the import is already present from a previously executed blocks so you can skip the import here and further down below :+1:

mrossinek (commented):
Some more suggestions :slightly_smiling_face:

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
woodsp-ibm: In terms of jupyter-execute I know Terra is trying to limit their usage, see  Qiskit/qiskit-terra#7661

When I referred to including examples above I referred to doctest i.e. https://docs.python.org/3/library/doctest.html which we support here in CI and examples will be run and the output as stored checked. The code samples would need to produce output, for example the list of excitations produced

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
javabster: yeah the execution works fine, but it doesn't print any output by default so wasn't sure if I should add anything to the example. Should I explicitly call `print(excitations)` in the example and put the output in the doc?

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
javabster: I actually think it is a clearer introductory example to write them all out explicitly. Maybe I can leave the first basic example written out in full and then inline them for the other examples?

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
javabster: @woodsp-ibm do you have any more specific doctest examples of how this is implemented in nature? I am struggling to get this to work, the following docstring doesn't seem to render any output when i build the docs:
```
.. testcode::
    from qiskit_nature.circuit.library.ansatzes.utils.fermionic_excitation_generator import (
           generate_fermionic_excitations
     )
            
      excitations = generate_fermionic_excitations(1, 4, (1, 1), generalized=True)
      print(excitations)
        
.. testoutput::
    [((0,), (1,)), ((2,), (3,))]
```

Screenshot 2022-02-24 at 3 04 51 PM

I also tried the following which just rendered a completeley empty box:
```
.. doctest::
    >>>excitations = generate_fermionic_excitations(1, 4, (1, 1), generalized=True)
    >>>print(excitations)
    [((0,), (1,)), ((2,), (3,))]
```

woodsp-ibm: As far as I am aware you just include it in the doctring. Here is an example from finance https://github.com/Qiskit/qiskit-finance/blob/5c9107ddd733e312569f3f2ef4d5a0cdfd19da36/qiskit_finance/circuit/library/probability_distributions/lognormal.py#L62-L70 (I don't see anything in Nature). Now looking at the output that results maybe it needs to be inside some markup https://qiskit.org/documentation/finance/stubs/qiskit_finance.circuit.library.LogNormalDistribution.html#qiskit_finance.circuit.library.LogNormalDistribution as it has rolled all the output into a single line @manoelmarques any thoughts here - does it need to be inside a code python block perhaps?

mrossinek: I believe that doctest enforces a whitespace after `>>>`. See here: https://docs.python.org/3/library/doctest.html#how-it-works They don't explicitly state that but when explaining where the expected output should be formatted they explicitly write `>>> ` and `... ` including the trailing whitespace!

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
mrossinek: In that case I would inline it:
```
excitations = generate_fermionic_excitations(
    num_excitations=1,
    num_spin_orbitals=4,
    num_particles=(1, 1),
)
```

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
mrossinek: Yes I think printing them would make sense. I guess you already assumed that given the discussion below :+1:

javabster:
ok I've updated the doctest examples, thanks for the suggestions @woodsp-ibm and @mrossinek. The current format renders like this:

Screenshot 2022-02-25 at 5 47 27 PM

There doesn't seem to be a nice way to render the longer lines of code (tbh i think the jupyter execute blocks render much more readable examples generally) but it's looking better than before!

Let me know if there are any further changes needed 😄

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
```diff
@@ -128,6 +124,29 @@ def generate_fermionic_excitations(
...
+
+        >>> excitations = generate_fermionic_excitations(1, 4, (1, 1), max_spin_excitation=1)
+        >>> print(excitations)
+        [((0,), (1,)), ((2,), (3,))]
+
+        Generate generalized excitations.
+
+        >>> excitations = generate_fermionic_excitations(1, 4, (1, 1), generalized=True)
+        >>> print(excitations)
+        [((0,), (1,)), ((2,), (3,))]
```
woodsp-ibm, line 148: Although these are all different examples, showing different parameter usage, they all end up as the same list being printed. What might be nice is an example where the parameter being shown makes a difference to the outcome.

For parameters in the function calls there should be no space around = for the style used. This is the case in the latter 2 examples but not the first.

As Python will print the return of a function when invoked interactively I think the examples can avoid the print and be more like the following.
```
        >>> generate_fermionic_excitations(1, 4, (1, 1), generalized=True)
        [((0,), (1,)), ((2,), (3,))]
```

javabster: Thanks Steve - I've made the changes 😄

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
```diff
@@ -128,6 +124,26 @@ def generate_fermionic_excitations(
...
+        Generate excitations limiting the maximum number of excitations per spin species.
+
+        >>> generate_fermionic_excitations(1, 6, (1, 1), max_spin_excitation=1)
+        [((0,), (1,)), ((0,), (2,)), ((3,), (4,)), ((3,), (5,))]
+
+        Generate generalized excitations.
+
+        >>> generate_fermionic_excitations(1, 6, (1, 1), generalized=True)
+        [((0,), (1,)), ((0,), (2,)), ((1,), (2,)), ((3,), (4,)), ((3,), (5,)), ((4,), (5,))]
+
```
woodsp-ibm, line 146: Could perhaps we also make the first example here 1,6,(1,1) like the last two. Then you would see the difference the specific parameters make. As it stands the max_spin_excitation of 1 on the second example does not do anything, we would need have more excitations such as num_excitations =2. Then there would need to be more orbitals and more particles in order that there are enough particles to actually do two of the same i.e. and alpha, alpha excitation e.g if you do (2,8,(2,2)) as the parameters then adding max_spin_excitation=1 shows a difference in outcome. But of course by this time the list it generates is rather larger (One can see the difference between them if you (2,8,(2,0)) so there are only two alpha particles where adding max_spin_excitation=1 removes the only possible double particle excitation, which is the two alpha particles and so returns an empty list which is not such a great example). I am thinking perhaps the second example could just be removed - certainly as it stands if you remove max_spin_excitation=1 its the exact same thing so it really does not demonstrate what the parameter does at all. If one tries to show what it does then the output is much larger - or my edge case where either its something or nothing!

javabster: Changed the first example to 1,6,(1,1) and removed the second example ✅

qiskit_nature/circuit/library/ansatzes/utils/fermionic_excitation_generator.py
```diff
@@ -128,6 +124,21 @@ def generate_fermionic_excitations(
...
         indices of the unoccupied spin orbitals.
+
+    Examples:
+        Generate excitations with basic inputs.
+
+        >>> from qiskit_nature.circuit.library.ansatzes.utils.fermionic_excitation_generator import (\
+        generate_fermionic_excitations)
+        >>> generate_fermionic_excitations(num_excitations=1, num_spin_orbitals=6,\
+        num_particles=(1,1))
+        [((0,), (1,)), ((0,), (2,)), ((3,), (4,)), ((3,), (5,))]
```
woodsp-ibm, line 135: Thanks for updating this. I took a final look at the result - you can download the documentation artifact, i.e. the zip of the built content from CI, from the build results page if you did not know. It looks a bit off depending on page width

Here you see the leading space of the first line above (i.e. the import)
image

And in this one its ahead of the last parameter of the 2nd line in the function call.
image

I am thinking perhaps it might be better to have the content on a single line for these 2 above that are continued and disable pylint line length check. What is happening in the output is that leading space in the continuation line is kept and can make the output look odd depending on page width and how things wrap.

I don't have any better idea other than disabling pylint check and letting the line be a little longer in the docs here. I think removing the leading whitespace as things are may well cause other issues with sphinx. Maybe you have seen that already.

javabster: Ok I kept it all on one line and disabled the pylint check, screenshots of how it looks on different screen sizes. As I said earlier I don't think there is a perfect formatting solution using doctest, but I think this should be enough that users will understand at least

Screenshot 2022-03-04 at 5 20 05 PM
Screenshot 2022-03-04 at 5 21 22 PM

woodsp-ibm (approved):

pbark (approved):
Thanks @javabster !

PR #560, opened 2022-02-15
Apply simplify to creation_op and annihilation_op in qubit mapping

### LINKED ISSUE #545: Jordan-Wigner transform slow on parity operator

t-imamichi:
### Summary

I found that #545 occurs because creation operators and annihilation operators are not simplified.
I apply simplify to them. I think this PR is ready to merge.

~~The technique #549 can perhaps make the qubit mapping even faster. So, I leave it for discussion.~~
I tried the combination of #549 and #560 and found that the combination is not faster than only #560.
See https://github.com/Qiskit/qiskit-nature/pull/549#issuecomment-1040333454 for details.

Closes #549 
Fixes #545

### Details and comments

micro benchmark of Kevin's parity operators #545 and random operators

```python
import random
from timeit import timeit

from qiskit_nature.mappers.second_quantization import JordanWignerMapper, BravyiKitaevMapper


def random_op(k, n):
    random.seed(123)
    op = sum(
        (
            FermionicOp("".join(random.choices("+-ENI", k=k)), display_format="dense")
            * random.random()
        )
        for _ in range(n)
    )
    return op


def parity_op(n_modes: int) -> FermionicOp:
    op = FermionicOp.one(n_modes)
    for i in range(n_modes):
        op @= FermionicOp.one(n_modes) - 2 * FermionicOp(f"N_{i}", display_format="dense")
    return op


def bench(op):
    mapper = BravyiKitaevMapper()
    #mapper = JordanWignerMapper()
    qubit_conv = QubitConverter(mapper)
    print(f"{timeit(lambda: qubit_conv.convert(op), number=1)} sec")


print("random")
op = random_op(12, 100)
bench(op)

print("\nparity")
op = parity_op(10)
bench(op)
```

main
```
random
12.88540858 sec

parity
12.476451123999999 sec
```

this PR
```
random
0.2673552950000001 sec

parity
0.5437850960000001 sec
```


mrossinek (approved):
Thanks this is indeed much more straight forward! :+1:

t-imamichi:
I ran Kevin's script #545 with this PR.
parity

woodsp-ibm (approved):
Very nice performance change indeed! I had to do a double take since at first glance the graph looked similar to what I recalled from before - but places have been switched and the scale is very different. Good job!

PR #565, opened 2022-02-16
Create logging utility functions

### LINKED ISSUE #419: Simplify logging configuration

Qiskit Aqua used to provide some simplified means of enabling Python's internal logging for certain modules:

https://github.com/Qiskit/qiskit-aqua/blob/main/qiskit/chemistry/_logging.py#L30-L37

I believe, Qiskit Nature would benefit a lot from a similarly simple configuration setting.
#406 is adding a `qiskit_nature.settings` object which permits simple global configuration settings. Based on another setting, logging could be configured in a similar manner to how it used to be done in Aqua.

Here are some ideas as to what logging options should be possible (for now at least):
- set all `qiskit_nature` logging to a given level
  - (optionally) specify a list of pairs matching logging levels to submodules (not required for now imo)
  - (optionally) support arbitrary logger names in the above list (could be used to easily control logging of other modules like Qiskit Terra, PySCF, etc.)
- default to a `StreamHandler` which logs to `sys.stdout`
  - (optionally) support custom handler addition (`FileHandler`, `RotatingFileHandler`, etc.)
  - (optionally) support custom `Formatter`

#### More Aqua-related references:
- https://github.com/Qiskit/qiskit-aqua/blob/main/qiskit/aqua/_logging.py

manoelmarques:
### Summary
Fixes #419


qiskit_nature/logging.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(self):
+        """
+        Initialize the handler.
+        """
+        super().__init__(stream=sys.stdout)
+        self.setFormatter(logging.Formatter(fmt="%(asctime)s:%(name)s:%(levelname)s: %(message)s"))
+
+
+class QiskitNatureLogging:
+    """Qiskit Nature Logging."""
```
woodsp-ibm, line 33: Would it be useful here in the docstring to say a bit more around the intent of the class and perhaps have some common example usage for how one might typically set things to show users how to make use of this class?

qiskit_nature/__init__.py
```diff
@@ -46,6 +55,7 @@
    properties
    results
    runtime
+   settings
```
mrossinek, line 58: Why do you add this twice?

We need to be a bit careful here because we actually do `from .settings import settings` because the latter is a "singleton" instance of `QiskitNatureSettings`

qiskit_nature/logging.py
```diff
@@ -0,0 +1,140 @@
...
+        names: List[str] = []
+        for name, level in name_levels.items():
+            names.append(name)
+            logger = self._logging.getLogger(name)
+            logger.setLevel(level)
+            logger.propagate = False
+            handlers = logger.handlers
+            if add_default_handler and not any(
+                True for h in handlers if isinstance(h, DefaultStreamHandler)
+            ):
```
mrossinek, line 73: Isn't this the same?
```suggestion
            if add_default_handler and not any(
                isinstance(h, DefaultStreamHandler) for h in handlers
            ):
```

qiskit_nature/logging.py
```diff
@@ -0,0 +1,140 @@
...
+                self.add_handler(name, DefaultStreamHandler())
+
+        return self.get_levels_for_names(names)
```
mrossinek, line 76: What's you reason for doing this?

qiskit_nature/logging.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        super().__init__(stream=sys.stdout)
+        self.setFormatter(logging.Formatter(fmt="%(asctime)s:%(name)s:%(levelname)s: %(message)s"))
+
+
+class QiskitNatureLogging:
+    """Qiskit Nature Logging."""
+
+    def __init__(self) -> None:
+        self._logging = logging
```
mrossinek, line 36: What's you reason for doing this?

qiskit_nature/logging.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def log_to_file(self, names: List[str], path: str) -> None:
+        """Logs modules to file.
+
+        Args:
+            names: list of module names (qiskit_nature.drivers, qiskit.algorithms etc)
+            path: file path
+        """
+        filepath = os.path.expanduser(path)
+        handler = logging.FileHandler(filepath, mode="w")
```
mrossinek, line 137: I think append-mode should be configurable here, too.

It may also be worthwhile to provide a keyword enabling the use of a [RotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#rotatingfilehandler).

qiskit_nature/logging.py
```diff
@@ -0,0 +1,140 @@
...
+        formatter = logging.Formatter(fmt="%(asctime)s:%(name)s:%(levelname)s: %(message)s")
```
mrossinek, line 138: I see you use the same `fmt` as in the `DefaultHandler`. Should we extract this to something like `QiskitNatureLogging.LOG_FORMAT`?
I don't see a good reason for making this configurable but having a fixed constant might be better in case we change this in the future (i.e. only have the string in one place)

releasenotes/notes/add-logging-171aceb37c21a3d5.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Added logging utility functions accessible through `qiskit_nature.settings.logging`.
```
mrossinek, line 4: That's not how it currently appears in the documentation. There you expose `QiskitNatureLogging` directly on the root so we should unify this.
I'm fine with either exposing this as part of `settings` or via it's own singleton instance (e.g. `qiskit_nature.logging`).

test/test_logging.py
```diff
@@ -0,0 +1,177 @@
...
+    def setUp(self):
+        super().setUp()
+        self._test_handler = TestHandler()
+        nature_settings.dict_aux_operators = True
+        self._logging_dict = {"qiskit_nature": logging.DEBUG, "qiskit": logging.DEBUG}
+        self._old_logging_dict = nature_settings.logging.get_levels_for_names(
+            self._logging_dict.keys()
+        )
+
+        nature_settings.logging.set_levels_for_names(self._logging_dict, add_default_handler=False)
```
mrossinek, line 64: I see the default handler being disabled here but do not see it being tested anywhere with it actually set.
That should be possible using [unittest.assertLogs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs). Maybe that even alleviates the need for `TestHandler`?

test/test_logging.py
```diff
@@ -0,0 +1,177 @@
...
+            nature_settings.logging.remove_handler(name, handler=self._test_handler)
+        nature_settings.logging.set_levels_for_names(
+            self._old_logging_dict, add_default_handler=False
+        )
+
+    @unittest.skipIf(not _optionals.HAS_PYSCF, "pyscf not available.")
+    def test_logging_to_handler(self):
+        """logging test"""
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", ResourceWarning)
```
mrossinek, line 80: Where can a `ResourceWarning` be raised from? :thinking:

qiskit_nature/logging.py
manoelmarques: The append-mode is fine but for other file handlers then why not let the user add them in the`add_handler` method ?

releasenotes/notes/add-logging-171aceb37c21a3d5.yaml
manoelmarques: I understood the issue as being exposed through `qiskit_nature.logging`. The reason for exposing `QiskitNatureLogging` in the root was because I could not find a way to show the sphinx documentation for this calss otherwise. Suggestions are welcome :-)

test/test_logging.py
manoelmarques: It seems to be raised by pyscf. I should probably just have a unit test that doesn't require pyscf in the first place.

qiskit_nature/__init__.py
manoelmarques: The `settings` weren't documented anywhere. I first added to modules the thought having it in a Globals section would be better and forgot to remove it from the modules.

qiskit_nature/logging.py
mrossinek: I was actually only referring to the `return`-statement. I don't see a good reason for that..

qiskit_nature/logging.py
mrossinek: Sure that's fair enough :+1:

releasenotes/notes/add-logging-171aceb37c21a3d5.yaml
mrossinek: Okay I see :thinking: I also think `qiskit_nature.logging` makes more sense but the reno says otherwise..

qiskit_nature/logging.py
manoelmarques: I thought useful to return the effective logging level. That indicates exactly the levels for the domains entered in case for some reason the NOTSET is entered, there may still logging set for the domains and the user won't know what level:

getEffectiveLevel

Indicates the effective level for this logger. If a value other than NOTSET has been set using setLevel is returned.
Otherwise, the hierarchy is traversed towards the root until a value other than NOTSET is found, 
and that value is returned. The value returned is an integer, typically one of logging.DEBUG, logging.INFO etc.

mrossinek: Hm I see. I was just wondering whether it is a good default for the setter utility to automatically also function as a getter. If there were no public getter I would agree but since there is I would rather follow a Linux-philosophy: no news is good news :smile: 

I don't have a strong opinion on this though I just wanted to discuss your reasoning :slightly_smiling_face:

qiskit_nature/__init__.py
```diff
@@ -26,9 +26,18 @@
    :toctree: ../stubs/
    :nosignatures:
 
+   QiskitNatureLogging
```
mrossinek, line 29: I'm still not convinced whether this is the right way to document this..

As things are now, a user will modify the logging via `qiskit_nature.settings.logging`. However, the documentation may lead a user to believe that they can simply instantiate `QiskitNatureLogging()` and use that.

Maybe we should follow the same approach as the settings singleton here? I.e.
```python
# in qiskit_nature/logging.py
logging = QiskitNatureLogging()

# in qiskit_nature/__init__.py
from .logging import logging
# and then expose logging as part of Globals (just like settings)
```

qiskit_nature/logging.py
manoelmarques: I removed the return value

releasenotes/notes/add-logging-171aceb37c21a3d5.yaml
manoelmarques: I changed to qiskit_nature.logging

qiskit_nature/__init__.py
manoelmarques: Changed

test/test_logging.py
```diff
@@ -0,0 +1,150 @@
...
+                    msg=f"{name}: logging level {level} < reference level {ref_level} ",
+                )
+
+    def test_logging_to_handler(self):
+        """logging test"""
+        self._set_logging(False)
+        # ignore Qiskit TextProgressBar that prints to stderr
+        with contextlib.redirect_stderr(io.StringIO()):
+            unittest.TextTestRunner().run(
+                unittest.TestSuite([TestBosonicESCCalculation("test_numpy_mes")])
```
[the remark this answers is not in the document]
manoelmarques: Ignore the failure in external test now. Just checking for logging

test/test_logging.py
```diff
@@ -0,0 +1,146 @@
...
+        self.records.append(record)
+
+
+class TestLogging(QiskitNatureTestCase):
+    """Test logging"""
+
+    def setUp(self):
+        super().setUp()
+        self._test_handler = TestHandler()
+        nature_logging.dict_aux_operators = True
```
mrossinek, line 49: If you want to avoid the `DeprecationWarnings` currently raised if this is False you now need to update this to actually use the settings:
```suggestion
        nature_settings.dict_aux_operators = True
```
Andd add `from qiskit_nature import settings as nature_settings` at the top of the file :+1:

mrossinek (commented):
One minor comment, other than that this LGTM!

woodsp-ibm (approved):
LGTM!

PR #566, opened 2022-02-17
Revert "Temporarily skip broken tests (#527)"

Cryoris:
### Summary

Now that Qiskit/qiskit-terra#7598 is merged, we can again enable the tests relying on Z2 symmetries which were temporarily skipped in #527.


woodsp-ibm (approved):
Great to have the change/fix merged in Terra, thanks for re-enabling these again.

PR #570, opened 2022-02-22
VQE Factory method improvements

### LINKED ISSUE #384: VQE Factory method improvements

iuliazidaru:
### Summary
 Fixes #384 VQE Factory method improvements

### Details and comments

Work in progress


iuliazidaru:
I have problems with this code:
```
from qiskit_nature.circuit.library import UCC
uccs = UCC(excitations='d')
uccs.draw()
```

UCC is not usable until   qubit_converter, num_particles, num_spin_orbitals are set. 
VQE displays information about members in constructor: `logger.info(self.print_settings())`. If UCC is not yet usable, this code throws an exception.

failing test: TestGroundStateEigensolver. test_vqe_uccsd.

Any idea @mrossinek ?

mrossinek:
Where and why exactly would you want to use this code? I don't quite understand the need for it.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -91,6 +91,17 @@ def __init__(
         self.initial_state = initial_state
         self.callback = callback
         self._kwargs = kwargs
+        self._vqe = VQE(
+            ansatz=ansatz,
```
mrossinek, line 95: Is this the line of culprit which you were referring to above when you said you had problems with `UCC`?
I actually do not see a reason for setting ansatz already up here. You can simply leave it at None and only set it during `get_solver` as in the past.

Essentially the idea here should be, that all the attributes which are handled by the factory (`ansatz` and in the future `initial_point` #571), are set during `get_solver` on some internal minimum eigensolver instance. The remaining property setters should be deprecated in favor of exposing the `MinimumEigensolver` directly (I think that's what you attempted in the numpy factory above, but I do not see you exposing the `self._minimum_eigensolver` via a property getter which is required for the user to actually have access to it.

iuliazidaru: @mrossinek You are right! Thank you for the comment!

woodsp-ibm:
A couple of comments. One is that Nature has a set of deprecation decorators both for convenience and to ensure the deprecation message over the version and time when it may be removed is consistent. These decorators allow a custom message to be included as well etc. They are part of a deprecation.py file in the root here, along with some functions that are sometimes directly used. There is a unit test case for the above file, so there are examples of its use there, as well as examples you can find in the main Nature code where logic is deprecated e.g. search for deprecate_method for instance.

For the ansatz I am thinking we ought to check that any instance passed in via the constructor is a UCC instance since later the code is going to try and set properties that only exist for that - e.g. num_particles etc and would not work as expected otherwise. Perhaps if we type check it in the constructor if None we can also create the UCCSD default instance at that time too and set it into the VQE we create. Later when the solver is fetched, and we go to update the ansatz in vqe we should probably re-check the vqe ansatz is still a UCC instance, since a user can update/alter it via the minimum eigensolver getter right. If its not a UCC then raise an error.

I guess in looking at the code, the way HF initial state exists it has to be built out each time; rather than being like UCC which can be updated and it adapts. So the ansatz that can be passed in by a user is updateable but if they pass in an initial state they have to know up front the problem and how to build that to be compatible later. Since the VQE UCC factory was more geared to doing this late configuration based on problem data, you may not know up front, I wonder in general whether passing in an initial state is so useful in practice. It is what it is for now though - in looking at things again now I am commenting - maybe this is something that we can look at again in the future, I'd leave it as is here though.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -208,29 +356,21 @@ def get_solver(  # type: ignore[override]
...
-
-        return vqe
+        if self.ansatz is None:
+            self._vqe.ansatz = UCCSD()
+        self._vqe.ansatz.qubit_converter = qubit_converter
+        self._vqe.ansatz.num_particles = num_particles
+        self._vqe.ansatz.num_spin_orbitals = num_spin_orbitals
+        self._vqe.ansatz.initial_state = initial_state
+        # maintain initialization done in vqe constructor
+        logger.info(self._vqe.print_settings())
```
iuliazidaru, line 366: This line of code is in VQE constructor and does "magic" :). 
The main issue here is that the ansatz has all the members optional in constructor and is initialised whenever the caller wants to. So, if we just set an ansatz, we may have an object which is invalid. If it is fully set, still the number of qubits are not calculated properly.  This line `_vqe.print_settings()` calls` self.ansatz.draw(output="text")` which builds the circuit, i.e. has the right members set. This is not clean and relates to the comment above (we may add an invalid ansatz to VQE and it fails when logging the information...).

woodsp-ibm: Ok, I would have to look. The print_settings() in VQE is something we had in the very early days such that you could see all those aspects when running it if you wanted. Quite a bit has changed since then. But I will note that the heuristic anstazes likes TwoLocal etc are BluePrint circuits and also are in an incomplete state (trying not to say invalid here) until the number of qubits is set when the operator is finally given to VQE. I.e they are more like template - it knows how to build but it needs to be told how big - num_qubits, Until then its just an emply template if you will. Now with UCC I guess the num qubits is more indirectly set, but given None is a valid value for num particles etc it ought to support draw in this state without failing.

iuliazidaru: @woodsp-ibm, thank you for looking into the issue!
Yes, there is a side effect. If we remove that log/print_settings, some tests will fail as the number of qubits in the ansatz( 0) doesn't match the number of qubits needed in the circuit. So, the ansatz is not "built" just by setting the properties.

woodsp-ibm: Update: I have UCC working locally such that all the current unit tests pass without the side-effect from the logging call in vqe (presently I deleted the logging call in VQE though I may leave it but change to have it inside an if that checks if info level logging is enabled.). I want to reflect on the change over the weekend, but expect to do a PR early next week.

woodsp-ibm: @iuliazidaru I made a change to UCC in #601, which is now merged. You should no longer need to rely on the side-effect of that logging statement for things to work. The equivalent change still needs to be done to UVCC - see #602 - but I think things should now be fine for the VQE UCC factory.

woodsp-ibm: UVCC is also now fixed (by #615)

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/numpy_minimum_eigensolver_factory.py
```diff
@@ -45,30 +45,59 @@ def __init__(
...
-        """returns filter criterion"""
+        """DEPRECATED. Use ``minimum_eigensolver`` method and solver properties instead.
+        Returns filter criterion."""
+        warnings.warn(
+            "The `filter_criterion` getter method is deprecated as of Qiskit Nature 0.4, "
+            "and will be removed in a future release. Use ``minimum_eigensolver`` method and "
+            "solver properties instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
mrossinek, line 62: I think you should be able to find a suitable method in `qiskit_nature.deprecation` for this purpose. I suggest `deprecate_property` which you can use as a decorator below the `@property` decorator. See here for an example: https://github.com/Qiskit/qiskit-nature/blob/fd016e2dc3774dedb41f934b37e389446435e27d/qiskit_nature/problems/second_quantization/base_problem.py#L62

EDIT: The same applies below

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
mrossinek, line 366:
```suggestion
```
Since this workaround is no longer needed (Thanks to @woodsp-ibm!) this can be removed.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -232,7 +353,22 @@ def get_solver(  # type: ignore[override]
...
+        ansatz = self.ansatz
+        if ansatz is None:
+            ansatz = UVCCSD()
+        ansatz.qubit_converter = qubit_converter
+        ansatz.num_modals = num_modals
+        ansatz.initial_state = initial_state
+
+        self._vqe.ansatz = ansatz
+        # maintain initialization done in vqe constructor
+        logger.info(self._vqe.print_settings())
```
mrossinek, line 365:
```suggestion
```
The same holds here

mrossinek (commented):
Besides these minor comments I think this is good to go! :+1:

docs/tutorials/thermodynamics_utils/partition_function.py
```diff
@@ -301,9 +301,9 @@ def eval_translational(self, p, t):
         """
         m = sum(self.molecule.masses)
 
-        debrog = (const.H_J_S**2 / (2 * np.pi * m * const.KB_J_PER_K * t)) ** (1 / 2)
+        debrog = (const.H_J_S ** 2 / (2 * np.pi * m * const.KB_J_PER_K * t)) ** (1 / 2)
```
mrossinek, line 304: It looks like you ran `make black` once while having an outdated `black` installed which is causing the CI failures. You should update your installation, run `make black` and hopefully that should resolve the CI errors.

iuliazidaru: Thank you!

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -81,107 +86,167 @@ def __init__(
...
+        self._initial_state = initial_state
+
+        # Not needed
+        self._quantum_instance = quantum_instance
+        self._optimizer = optimizer
+        self._initial_point = initial_point
+        self._gradient = gradient
+        self._expectation = expectation
+        self._include_custom = include_custom
+        self._callback = callback
```
mrossinek, line 99: The way that these are "deprecated" now is actually not correct because they do not still work any more (which means they are effectively already removed instead of "just" deprecated). I am, unfortunately, not sure what the best course of action is.

The only suggestion I have is to check during `get_solver` whether any of these internal attributes are `not None` (which they all (except `include_custom`) default to anyways) and if so, set the respective attribute of the solver.The downside of this is that a user needs to ensure to leave these attributes at `None` when switching to the correct new behavior, otherwise they will have there settings overwritten.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -81,107 +85,161 @@ def __init__(
...
+        self._initial_state = initial_state
+
+        # Not needed
+        self._quantum_instance = quantum_instance
+        self._optimizer = optimizer
+        self._initial_point = initial_point
+        self._gradient = gradient
+        self._expectation = expectation
+        self._include_custom = include_custom
+        self._callback = callback
```
mrossinek, line 98: The same as the above applies here.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -81,107 +86,167 @@ def __init__(
...
+            gradient=gradient,
+            expectation=expectation,
+            include_custom=include_custom,
+            callback=callback,
+            **kwargs,
+        )
+
+    @property  # type: ignore
+    @deprecate_property(
+        "0.4", additional_msg="Use `minimum_eigensolver` and 'solver properties' instead."
```
mrossinek, line 115:
```suggestion
        "0.4", additional_msg="Use `minimum_eigensolver` and 'solver properties' instead"
```
The dot gets added by the decorator

mrossinek (changes requested):
Besides the above it would be great if you could:

1. Add some unittests which test both the old and new behavior. (when testing the old behavior filter out the deprecation messages)
2. Update the tests using the old method to instead use the new method.

You can check which tests are affected by checking the output of the CI which gathers all deprecation warnings here: https://github.com/Qiskit/qiskit-nature/runs/5871403093?check_suite_focus=true#step:13:7

mrossinek:
@iuliazidaru Since we need these changes to unblock another feature which has to be part of the next release, @MarcDrudis will finish up the remaining changes on this PR. He will open up a new PR which supersedes this one shortly.

iuliazidaru:
Sorry, @mrossinek, I was not available in latest weeks. Thank you, @MarcDrudis, for continuing the work.

PR #578, opened 2022-02-24
Deprecate all LegacyDriver related code

### LINKED ISSUE #561: Deprecate all `LegacyDriver` related code

The legacy driver results, `QMolecule` and `WatsonHamiltonian`, have been deprecated in Qiskit Nature 0.2.0.
We currently still have various places where we use them via their public API in order to provide migration/conversion utilities. Most notably in the `HDF5Driver` (as pointed out [here](https://github.com/Qiskit/qiskit-nature/pull/550#discussion_r805149589)) and in the property framework via the `.from_legacy_driver_result` conversion methods.

In order to be able to actually remove these legacy driver results, we must first deprecate all methods which use them in their public API. Ideally, we add these deprecations as part of 0.4.0
Once that is done, we should be able to fully remove these classes from the public API in 0.5.0 (we can still provide a conversion utility in the HDF5Driver for example, it would just be a private method instead of being part of the public, user-facing API).

manoelmarques:
### Summary
Fixes #561

Unit test hdf5 files were converted with the exception of the legacy needed to test legacy logic.


woodsp-ibm:
In terms of the 'convert` on the hdf5 driver, that supports existing hdf5 files for users that might have them, this has only just been added so it seems a little odd to come out with a new release with this already in a deprecated state. In the end I guess it depends on how long we want to keep the code around that can convert existing files that users might have i.e. past versions of hdf5. Certainly if keeping this longer term, any conversion code, along with QMolecule, needs to go from the public interfaces to where its private code just used by hdf5 driver in support of convert - I think this is what is alluded to in the referenced issue as deprecating them and making private conversion. As it stands, if we already deprecate this now it would be removed no sooner than 0.6.0 so anyone looking to convert any files after than point would be obliged to install an older nature or lose them.

qiskit_nature/drivers/second_quantization/hdf5d/hdf5driver.py
```diff
@@ -76,6 +77,7 @@ def _get_path(self) -> pathlib.Path:
 
         return hdf5_file
 
+    @deprecate_method("0.4.0")
```
mrossinek, line 80: I agree with Steve's analysis and think that we should not yet deprecate this method. Since it does not expose `QMolecule` via the public API we should be able to simply migrate that to a private class in some future release and can then choose to deprecate this at any time in the future once we deem users have had enough time to migrate/convert their files.

qiskit_nature/problems/second_quantization/vibrational/builders/vibrational_op_builder.py
```diff
@@ -26,45 +24,6 @@
...
-        num_modals = [num_modals] * num_modes
-
-    boson_hamilt_harm_basis = HarmonicBasis(
-        watson_hamiltonian, num_modals, truncation_order
-    ).convert()
-
-    return build_vibrational_op_from_ints(boson_hamilt_harm_basis, num_modes, num_modals)
-
-
 @deprecate_function(
```
mrossinek, line 27: Is this function below (Github won't let me comment further down) still being used somewhere? If yes, why/where? If not we should be able to remove this now, too.

test/drivers/second_quantization/hdf5d/test_driver_hdf5.py
```diff
@@ -40,7 +43,10 @@ def test_convert(self):
         legacy_file_path = self.get_resource_path(
             "test_driver_hdf5_legacy.hdf5", "drivers/second_quantization/hdf5d"
         )
-
+        msg_convert_ref = (
+            "The convert method is deprecated as of version 0.4.0 and "
+            + "will be removed no sooner than 3 months after the release."
```
mrossinek, line 48: Is the `+` even necessary here? I think Python will do automatic string concatenation

test/drivers/second_quantization/hdf5d/test_driver_hdf5.py
```diff
@@ -49,17 +55,22 @@ def test_convert(self):
...
+
+                driver.run()
             finally:
                 os.unlink(tmp_file.name)
 
+        msg_mol_ref = (
+            "The HDF5Driver.run with legacy HDF5 file method is deprecated as of version 0.4.0 "
+            + "and will be removed no sooner than 3 months after the release "
+            + ". Your HDF5 file contains the legacy QMolecule object! You should "
+            + "consider converting it to the new property framework. See also HDF5Driver.convert."
```
mrossinek, line 72: Same here

mrossinek (commented):
Just some minor comments :slightly_smiling_face:

qiskit_nature/drivers/second_quantization/hdf5d/hdf5driver.py
manoelmarques: Deprecate removed

qiskit_nature/problems/second_quantization/vibrational/builders/vibrational_op_builder.py
manoelmarques: Not used anywhere. I removed the file

qiskit_nature/drivers/second_quantization/hdf5d/hdf5driver.py
```diff
@@ -116,17 +119,24 @@ def run(self) -> GroupedSecondQuantizedProperty:
...
-                    "Your HDF5 file contains the legacy QMolecule object! You should consider "
-                    "converting it to the new property framework, see HDF5Driver.convert"
+                warn_deprecated(
+                    "0.4.0",
+                    DeprecatedType.METHOD,
+                    "HDF5Driver.run with legacy HDF5 file",
+                    additional_msg=(
+                        ". Your HDF5 file contains the legacy QMolecule object! You should consider "
+                        "converting it to the new property framework. See also HDF5Driver.convert"
+                    ),
```
woodsp-ibm, line 129: I guess we switched from a warning/alert to consider changing to more of an impetus to change with the deprecation.

woodsp-ibm (approved):
It all looks ok to me. I put a comment - but I think its fine and we can how things look etc when the deprecated conversion methods are dealt with down the road.

qiskit_nature/drivers/second_quantization/hdf5d/hdf5driver.py
manoelmarques: I changed from logger, which the user would have to set to warning level to see it, to a deprecation that will be seen once, because the user should ideally convert the hdf5 file to the new format.

Issue #585, opened 2022-03-05
`ElectronicStructureProblem` might generate a redundant operator


t-imamichi:
### What is the expected enhancement?

`ElectronicStructureProblem` of main branch generates a redundant operator than that by stable though they are identical from matrix perspective. Because the operator is larger, qubit mapping takes more time with main branch than stable.
I think we need to apply `reduce` somewhere.

```python
from qiskit_nature.drivers import Molecule
from qiskit_nature.drivers.second_quantization import ElectronicStructureDriverType, \
    ElectronicStructureMoleculeDriver
from qiskit_nature.problems.second_quantization import ElectronicStructureProblem

h2_geometry = [('H', [0, 0, 0]), ('H', [0, 0, 1])]

molecule = Molecule(geometry=h2_geometry, charge=0, multiplicity=1)
driver = ElectronicStructureMoleculeDriver(molecule, basis="sto3g",
                                           driver_type=ElectronicStructureDriverType.PYSCF)
es_problem = ElectronicStructureProblem(driver)
second_q_op = es_problem.second_q_ops()
lst = sorted(second_q_op[0].to_list(display_format="dense"), key=lambda e: e[0])
print(f'size {len(lst)}')
for op in lst:
    print(op)
print("\nmatrix")
print(second_q_op[0].to_matrix())
```

main
```
size 28
('+-+-', (0.09839529174273506+0j))
('+-+-', (0.09839529174273506+0j))
('+--+', (-0.0983952917427351+0j))
('+--+', (-0.09839529174273509+0j))
('-++-', (-0.09839529174273509+0j))
('-++-', (-0.0983952917427351+0j))
('-+-+', (0.09839529174273512-0j))
('-+-+', (0.09839529174273512-0j))
('IIIN', (-0.589121003706083+0j))
('IINI', (-1.1108441798837272+0j))
('IINN', (-0.0983952917427351+0j))
('IINN', (0.31085338155985665+0j))
('IINN', (0.31085338155985653+0j))
('IINN', (-0.09839529174273509+0j))
('INII', (-0.589121003706083+0j))
('ININ', (0.326535373471287+0j))
('ININ', (0.326535373471287+0j))
('INNI', (0.31085338155985653+0j))
('INNI', (0.31085338155985665+0j))
('NIII', (-1.1108441798837272+0j))
('NIIN', (0.31085338155985665+0j))
('NIIN', (0.31085338155985653+0j))
('NINI', (0.3132012497647589+0j))
[...]

PR #586, opened 2022-03-07
Performance improvement of `FermionicOp.reduce`

t-imamichi:
### Summary

This PR introduces the following two changes.
- Vectorize `FerimionicOp.reduce`
- Add `normal_order` option to `FerimionicOp.reduce` (default is True)

```python
import random
from timeit import timeit


def random_op(k, n):
    random.seed(123)
    op = sum((FermionicOp("".join(random.choices("+-ENI", k=k)), display_format="dense") * random.random()) for _ in range(n))
    return op

def bench_reduce(op):
    print(f"normal order: {timeit(lambda: op.reduce(), number=1)} sec")
    try:
        print(f"no normal order: {timeit(lambda: op.reduce(normal_order=False), number=1)} sec")
    except TypeError:
        pass

op = random_op(5, 10000)
bench_reduce(op)
```

main
```
normal order: 2.4304106099998535 sec
```

this PR
```
normal order: 2.271573748999799 sec
no normal order: 0.05692009299946221 sec
```

### Details and comments

The motivation to apply `reduce` before qubit mapping is #585. The output operator can be redundant.
So, I want to simplify the operators. But, I don't think qubit operator requires the normal order.

I added the `normal_order` option because applying `normal_order` may increase the number of operators and `to_normal_order` can be slow.
If users apply `reduce` to an operator before qubit mapping, it takes more time than without `reduce` due to the larger number of operators.
I attach an example. The size of operator grows from 1000 to 18950 by applying `reduce`.

```python
import random
from timeit import timeit

def random_op(k, n):
    random.seed(123)
    op = sum((FermionicOp("".join(random.choices("+-E", k=k)), display_format="dense") * random.random()) for _ in range(n))
    return op

qubit_converter = QubitConverter(mapper=JordanWignerMapper())

def bench_convert(op):
    print(f"w/o reduce: {timeit(lambda: qubit_converter.convert(op), number=1)} sec")
    print(f"w/ reduce of normal order: {timeit(lambda: qubit_converter.convert(op.reduce()), number=1)} sec")
    try:
        print(f"w/ reduce no normal order: {timeit(lambda: qubit_converter.convert(op.reduce(normal_order=False)), number=1)} sec")
    except TypeError:
        pass

op = random_op(10, 1000)

print("qubit convert")
bench_convert(op)

print("\noperator sizes")
print(f"original op: {len(op)}")
print(f"w/ reduce of normal order: {len(op.reduce())}")
try:
    print(f"w/ reduce no normal order: {len(op.reduce(normal_order=False))}")
except TypeError:
    pass
```

this PR
```
qubit convert
w/o reduce: 1.7863747950004836 sec
w/ reduce of normal order: 17.12350188399978 sec
w/ reduce no normal order: 1.741173968000112 sec

operator sizes
original op: 1000
w/ reduce of normal order: 18950
w/ reduce no normal order: 989
```


t-imamichi:
I have not written the docstring of the new option `reduce(normal_order)`. If the option is allowed, I will write it.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -538,24 +538,25 @@ def adjoint(self) -> FermionicOp:
             display_format=self.display_format,
         )
 
-    def reduce(self, atol: Optional[float] = None, rtol: Optional[float] = None) -> FermionicOp:
+    def reduce(
```
pbark, line 541: Why we do not have a docstring here?

kevinsung:
This PR is made obsolete by https://github.com/Qiskit/qiskit-nature/pull/607.

t-imamichi:
I will port this optimization to the upcoming `simplify` methods #607. 

https://github.com/Qiskit/qiskit-nature/pull/607#issuecomment-1076457920

PR #587, opened 2022-03-08
Fix lists in electronic structure tutorial

garrison:
### Summary

This is in the spirit of https://github.com/Qiskit/qiskit-optimization/pull/278.  The incorrectly displayed lists begin in [this section](https://qiskit.org/documentation/nature/tutorials/01_electronic_structure.html#The-Hartree-Fock-initial-state).

### Details and comments

I also grepped and looked through the other tutorials, and I was unable to find any other list that displays incorrectly.


woodsp-ibm (approved):
Looks fine to me in terms of fixing the formatting for the bullet lists. 

Now in looking at the latter  list

image

The text that is in quotes I think relates to the string by which these used to be selected a long time ago. Maybe nowdays it should look more like

- Jordan-Wigner (user....
- Parity (The...
- Bravyi-Kitaev (Annals...

so as simply to get rid of the strings which I think one would wonder what they mean; the jordan wigner item had mapping in it too which seems to be redundant since its a list of mappings!

Anyway since this is a different improvement than your specific fix I approved this as I did not know if you want to do this or this is better left to a separate issue for someone to address,

woodsp-ibm (approved):
Thanks for doing the other change too.

pbark (approved):
Thanks @garrison !

PR #588, opened 2022-03-08
Make VQEClient extend VariationalAlgorithm

### LINKED ISSUE #426: Allow broader use of VQEClient in Nature
While VQEClient is a MimimumEigenSolver and can be used in those contexts there are other places where we should be able to allow this to be used/leveraged but have not so far.

1. In mininum eigen solver factories that are building for example VQE+UCCSD that could be built out with VQEClient see #384 in which I commented in this regard.

2. For BopesSampler. With the extrapolation its looking for a VariationalAlgorithm interface that VQE implements but VQEClient does not. We should consider having it implement that interface to allow its use with BopesSampler.

javabster:
### Summary

fixes #426 

This PR addresses the BopesSampler part of 426. `VQEClient` now extends `VariationalAlgorithm`. Any suggestions on further testing this change would be appreciated

### DIFF
--- a/qiskit_nature/runtime/vqe_client.py
+++ b/qiskit_nature/runtime/vqe_client.py
@@ -20,13 +20,18 @@ from qiskit import QuantumCircuit
 from qiskit.exceptions import QiskitError
 from qiskit.providers import Provider
 from qiskit.providers.backend import Backend
-from qiskit.algorithms import MinimumEigensolver, MinimumEigensolverResult, VQEResult
+from qiskit.algorithms import (
+    MinimumEigensolver,
+    MinimumEigensolverResult,
+    VQEResult,
+    VariationalAlgorithm,
+)
 from qiskit.algorithms.optimizers import Optimizer, SPSA
 from qiskit.opflow import OperatorBase, PauliSumOp
 from qiskit.quantum_info import SparsePauliOp
 
 
-class VQEClient(MinimumEigensolver):
+class VQEClient(VariationalAlgorithm, MinimumEigensolver):
     """The Qiskit Nature VQE Runtime Client.
 
     This class is a client to call the VQE program in Qiskit Runtime."""


woodsp-ibm:
> Any suggestions on further testing this change would be appreciated
As VQEClient already implemented the interface as defined by VariationalAlgorithm there is no new function added at all so no need, I think, for any additional unit testing. Simply it now means that other algorithms that check if the minimum eigensolver is a VariationalAlgorithm instance will now work this way too now based on the same instance checking they already do. 

Can you confirm you from a system level perspective that it all works as expected. Its so simple it should all just work, but checking with say the tutorial that has the BopesSampler and replacing VQE etc by VQEClient, that it indeed all works fine.

One thing we ought to have is a reno for this change as users will now be able to use VQEClient with the BopesSampler and leverage bootstrapping/interpolation which can be quite savings in total evaluations as that tutorial shows.

releasenotes/notes/vqeclient-extends-variationalalg-eb5b3f951405fd16.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    :class:`~qiskit_nature.runtime.vqe_client` now extends 
+    VariationalAlgorithm, meaning users can now implement 
+    :class:`~qiskit_nature.algorithms.pes_samplers.bopes_sampler`
```
woodsp-ibm, line 6:
```suggestion
    :class:`~qiskit_nature.algorithms.pes_samplers.BOPESSampler` 
```

woodsp-ibm (commented):
The references to class need to link to the class itself not the module containing the class. If you go to the CI results, pick Details for any of the CI jobs, and select Summary then scroll down and you will find a `documentation` artifact. You can download this and unzip it and either click on release_notes html directly or go to it via the index.html. You will see this note in among the new ones and will see that they do not have any navigation. I think the suggested changes here should correct that

releasenotes/notes/vqeclient-extends-variationalalg-eb5b3f951405fd16.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
+  - |
+    :class:`~qiskit_nature.runtime.vqe_client` now extends 
+    VariationalAlgorithm, meaning users can now implement 
+    :class:`~qiskit_nature.algorithms.pes_samplers.bopes_sampler` 
+    using :class:`~qiskit_nature.runtime.vqe_client` instead of
```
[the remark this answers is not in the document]
woodsp-ibm: VQEClient was usable before with BOPESSampler, this change fixes things such that the bootstapping, which works with VQE, now works similarly with VQEClient. I think this is the essence of this change/fix right. Users are not implementing BopesSampler as such more it fixes things for VQEClients use in a GroundStateSolver used in conjunction with it.

releasenotes/notes/vqeclient-extends-variationalalg-eb5b3f951405fd16.yaml
```diff
@@ -1,8 +1,9 @@
...
 features:
   - |
-    :class:`~qiskit_nature.runtime.vqe_client` now extends 
+    :class:`~qiskit_nature.runtime.VQEClient` now extends 
     VariationalAlgorithm, meaning users can now implement 
-    :class:`~qiskit_nature.algorithms.pes_samplers.bopes_sampler` 
-    using :class:`~qiskit_nature.runtime.vqe_client` instead of
-    local VQE
+    :class:`~qiskit_nature.algorithms.pes_samplers.BOPESSampler` 
+    using bootstapping, which works similarly now with VQEClient
```
woodsp-ibm, line 7:
```suggestion
    using bootstrapping, which works similarly now with VQEClient
```

Issue #571, opened 2022-02-22
VQE initial point when using HF+UCC


woodsp-ibm:
### What is the expected enhancement?

When creating an initial_state + variable anstaz combined state in order to have HF+UCCSD produce the HF energy as the first eval the ansatz, in this case UCCSD needs to be appropriately parameterized. All zeros in this case. Now there still remains in the EvolvedOperatorAnstaz a preferred init point method where the idea was to inform VQE (or any other algo that might need it) of this parameterization such that it could be used for cases where the user did not explicitly supply an initial point and using a random point was not desired. See https://github.com/Qiskit/qiskit-terra/blob/a6cb1a37f52fd98f90d0685d1fdcba447a5345b7/qiskit/circuit/library/evolved_operator_ansatz.py#L139-L153

This mechanism had problems in applying it in general across the ansatzes, and this remains as the last vestige of that logic that still kinda works. So give VQE the UCCSD+HF and lo and behold it the first eval is the HF energy... but gotcha first transpile the HF+UCCSD and give it the transpiled circuit and it does not. The difference is that the result of transpilation is no longer the UCC sub-class of QuantumCircuit and its does not have the special method. This method is looked for by VQE in its path where the initial point is not given explicitly. If this method exists then it gets the special (preferred) start point, otherwise its random and this is the case here.

As the above linked code shows it has long been desired for something better/different. One thought is to document the initial point of all zeros in UCC and require the user, when directly using VQE + HF+UCC to explicitly set it. By doing this it does not matter whether you use a transpiled version or original it would all work - no magic implicit stuff going on in background - though VQE does state this is what it does https://github.com/Qiskit/qiskit-terra/blob/a6cb1a37f52fd98f90d0685d1fdcba447a5345b7/qiskit/algorithms/minimum_eigen_solvers/vqe.py#L103-L105 Of course knowing which ansatzes do/don't have that capability is more of a challenge. 

To keep some of 'the magic' what could be done is that the VQE + UCC factories could take care of the initial point on behalf of the user so at it gets set as we do things today. Drop down though to using VQE and UCC directly though you would be charged with the initial point yourself just like any other say heuristic anstazes we have where they have no preferred point.

It is of course possible to deprecate the current behavior warning that if we set all zeros automatically this way that it will be required that the user do it explicitly in the future.

Perhaps there are other ways. The above describes though the area that this ought to address and outlines some initial thought.

Cryoris:
[...]

PR #590, opened 2022-03-09
Migrate MP2Info

### LINKED ISSUE #89: Migrate MP2Info

I already mentioned `qiskit_nature.mp2info` [here](https://github.com/Qiskit/qiskit-nature/pull/66#issuecomment-803277819) because it is still coupled to the old `UCCSD` implementation. Do we need this for the release next week?

@pbark @woodsp-ibm

declanmillar:
### Summary

Closes #89. Migrates the class that computes Møller-Plesset 2nd order corrections from the Aqua module to Nature. The double-excitation coefficients are intended to be used as initial point values for the corresponding operators in VQE when using a UCC ansatz. This should introduce an initial point that is closer to the ground state point, leading to fewer evaluations. However, I have not done any experiments to show that this is true in practice. These will be documented in #633.

The original Aqua implementation is [here](https://github.com/Qiskit/qiskit-aqua/blob/main/qiskit/chemistry/mp2info.py), with tests [here](https://github.com/Qiskit/qiskit-aqua/blob/main/test/chemistry/test_mp2info.py).

### Details and comments

`MP2InitialPoint` inherits from the abstract base class (interface) `InitialPoint`, which takes a driver result and a UCC Ansatz. The `MP2InitialPoint` requires the number of spin orbitals, the electronic energy, and the excitation lists from the second quantized properties of the driver result.

If using a `VQEUCCFactory`, one can pass an `MP2InitialPoint` instance via the `initial_point` keyword argument, in which case the driver result and ansatz will be used to compute the initial point automatically. This will then be passed to the VQE. Outside of the factory, one can do this manually (see examples below).

The eventual intention is to retire `preferred_init_points` from `VQE` in Terra, so the implementation avoids this property. A default `HFInitialPoint` will be introduced in #632.

### Example scripts

A simple complete script to use an MP2 initial point via the factory is:

```python
from qiskit import Aer

from qiskit_nature.algorithms import VQEUCCFactory
from qiskit_nature.algorithms.initial_points.mp2_initial_point import MP2InitialPoint
)
from qiskit_nature.algorithms import GroundStateEigensolver

from qiskit_nature.transformers.second_quantization.electronic import FreezeCoreTransformer

molecule = Molecule(
    geometry=[["H", [0.0, 0.0, 0.0]], ["H", [0.0, 0.0, 0.735]]], charge=0, multiplicity=1
)

driver = ElectronicStructureMoleculeDriver(
    molecule, basis="sto3g", driver_type=ElectronicStructureDriverType.PYSCF
)

freeze_core = FreezeCoreTransformer()

problem = ElectronicStructureProblem(driver, transformers=[freeze_core])

qubit_converter = QubitConverter(JordanWignerMapper())

initial_point = MP2InitialPoint()

quantum_instance = QuantumInstance(backend=Aer.get_backend("aer_simulator_statevector"))
vqe_ucc_factory = VQEUCCFactory(quantum_instance, initial_point=initial_point)

vqe_solver = GroundStateEigensolver(qubit_converter, vqe_ucc_factory)
res = vqe_solver.solve(problem)

print(res)
```

Or to use it outside a factory:

```python
from qiskit import Aer
from qiskit.algorithms.optimizers import L_BFGS_B
from qiskit.algorithms import VQE

from qiskit_nature.circuit.library import UCCSD
from qiskit_nature.algorithms import MP2InitialPoint
from qiskit_nature.settings import settings
)
from qiskit_nature.algorithms.initial_points import mp2_initial_point


settings.dict_aux_operators = True

molecule = Molecule(
    geometry=[["H", [0.0, 0.0, 0.0]], ["H", [0.0, 0.0, 0.735]]], charge=0, multiplicity=1
)

driver = ElectronicStructureMoleculeDriver(
    molecule, basis="sto3g", driver_type=ElectronicStructureDriverType.PYSCF
)

# generate the second-quantized operators
second_q_ops = problem.second_q_ops()
main_op = second_q_ops["ElectronicEnergy"]

driver_result = problem.grouped_property_transformed

particle_number = driver_result.get_property("ParticleNumber")
electronic_energy = driver_result.get_property("ElectronicEnergy")

num_particles = (particle_number.num_alpha, particle_number.num_beta)
num_spin_orbitals = particle_number.num_spin_orbitals

# setup the classical optimizer for VQE
optimizer = L_BFGS_B()

# setup the qubit converter

# map to qubit operators
qubit_op = converter.convert(main_op, num_particles=num_particles)

# setup the ansatz for VQE
ansatz = UCCSD(
    qubit_converter=converter,
    num_particles=num_particles,
)

mp2_initial_point = MP2InitialPoint()
mp2_initial_point.grouped_property = driver_result
mp2_initial_point.ansatz = ansatz
initial_point = mp2_initial_point.to_numpy_array()

# set the backend for the quantum computation
backend = Aer.get_backend("aer_simulator_statevector")

# setup and run VQE
algorithm = VQE(ansatz, optimizer=optimizer, quantum_instance=backend, initial_point=initial_point)

result = algorithm.compute_minimum_eigenvalue(qubit_op)
print(result.eigenvalue.real)

electronic_structure_result = problem.interpret(result)
print(electronic_structure_result)
```


test/circuit/library/ansatzes/test_ucc.py
```diff
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021.
+# (C) Copyright IBM 2021, 2022.
```
woodsp-ibm, line 3: All these files that were changed before and are now in essence unchanged can you put the date back to 2021 so they get dropped completely from the PR and we are not simply pushing some date change on these

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -81,15 +92,15 @@ def __init__(
...
-        self.callback = callback
+        self._quantum_instance = quantum_instance
+        self._optimizer = optimizer
+        self._initial_point = initial_point
+        self._gradient = gradient
+        self._expectation = expectation
+        self._include_custom = include_custom
+        self._ansatz = ansatz
+        self._initial_state = initial_state
+        self._callback = callback
```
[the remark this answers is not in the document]
woodsp-ibm: See #570 for changes that are intended to come for the factory. The goal is to allow direct manipulation, of the underlying min eigen solver, e.g. vqe, in the factory and not have to mirror all its setters/getters here

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -216,6 +230,19 @@ def get_solver(  # type: ignore[override]
         ansatz.num_spin_orbitals = num_spin_orbitals
         ansatz.initial_state = initial_state
 
+        initial_point = self.initial_point
+        if initial_point is not None and not isinstance(initial_point, np.ndarray):
+            # UCC ansatz must be built early to compute excitation list.
+            ansatz._build()
```
declanmillar, line 236: Currently, one needs to call `_build()` as the existing UCC code only computes its excitation list when it builds the excitation operators. Should UCC be refactorized to try to separate the excitation_list computation and the rest of the build?

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -233,7 +233,7 @@ def get_solver(  # type: ignore[override]
         initial_point = self.initial_point
         if initial_point is not None and not isinstance(initial_point, np.ndarray):
             # UCC ansatz must be built early to compute excitation list.
-            ansatz._build()
+            ansatz.data
```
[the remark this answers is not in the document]
declanmillar: That makes sense. I was planning to separate the excitation_list computation from the operators or trigger the computation of the operators from the excitation_list getter. In the end, I just raised an error if the excitation_list is None in the getter, as I figured it's less invasive.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -40,7 +44,7 @@ def __init__(
         self,
         quantum_instance: QuantumInstance,
         optimizer: Optional[Optimizer] = None,
-        initial_point: Optional[np.ndarray] = None,
+        initial_point: Optional[Union[np.ndarray, str]] = None,
```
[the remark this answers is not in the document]
declanmillar: @woodsp-ibm raised a similar concern in Slack. I think I'm going to go with option 2.2 above (i.e., the empty instance), as Steve also suggested that would be preferable.

qiskit_nature/algorithms/point_generators/mp2_point_generator.py
```diff
@@ -0,0 +1,216 @@
...
+
+    The coefficients and energy deltas are computed when the object is
+    initialized. The `initial_point` property may then be used with VQE.
+
+    Only initial point values that correspond to double-excitations will be
+    non-zero.
+    """
+
+    # TODO: Add benchmark demonstrating that using MP2 initial point results in
+    # fewer evaluations to reach the ground state.
```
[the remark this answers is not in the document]
declanmillar: I was planning to raise a separate issue for this, however, on @woodsp-ibm's suggestion, I'm actually going to roll it into this one. I'm not sure how (if at all) I would make available these demonstrative results... maybe I could make a related tutorial and kill two birds with one stone.

pbark:
@declanmillar I have one last point that we may want to address. Can we rename this to `initial_points` instead of `point_generators`? And the corresponding file `mp2_initial_points.py `. I think that in this way we will be better and more descriptive. The rest LGTM.

pbark (dismissed):

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -15,21 +15,23 @@
...
 
 import numpy as np
+
 from qiskit.algorithms import MinimumEigensolver, VQE
 from qiskit.algorithms.optimizers import Optimizer
 from qiskit.circuit import QuantumCircuit
 from qiskit.opflow import ExpectationBase
 from qiskit.opflow.gradients import GradientBase
 from qiskit.utils import QuantumInstance
+from qiskit_nature.algorithms.initial_points.initial_point import InitialPoint
```
mrossinek, line 25: This looks like a duplicated import of the one a few lines down

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -40,7 +42,7 @@ def __init__(
         self,
         quantum_instance: QuantumInstance,
         optimizer: Optional[Optimizer] = None,
-        initial_point: Optional[np.ndarray] = None,
+        initial_point: Optional[Union[np.ndarray, InitialPoint]] = None,
```
mrossinek, line 45: I am wondering if we should add a `HFInitialPoint` class which simply returns an all zero array of the correct length here.
That way we could have this as a default. This would be a very minimal class obviously but I think it could be a bit clearer because users may not necessarily know that the all-zero case is simply the HF initial point.

Just putting this out there. Also tagging @woodsp-ibm and @pbark

releasenotes/notes/migrate-mp2-info-b1a3ee7320c46053.yaml
```diff
@@ -0,0 +1,22 @@
+features:
+  - |
+    Adds a class to compute the Moller-Plesset 2nd Order corrections. The
+    corresponding double-excitation coefficients are intended to be used as
+    an initial point for the VQE parameters (cluster operator coefficients)
+    when using a UCC ansatz. This should introduce an initial point that is
+    closer to the ground state point, leading to fewer overall evaluations for VQE.
+
+    To instantiate `~qiskit_nature.algorithms.MP2PointGenerator`, it requires the
```
mrossinek, line 9:
```suggestion
    To instantiate `~qiskit_nature.algorithms.MP2InitialPoint`, it requires the
```

releasenotes/notes/migrate-mp2-info-b1a3ee7320c46053.yaml
```diff
@@ -0,0 +1,22 @@
...
+    number of spin orbitals, the electronic energy, and the excitation lists
+    from the second-quantized properties returned by the driver.
+
+    If using a :class:`~qiskit_nature.algorithms.VQEUCCfactory`, one can set
+    `initial_point="MP2"` and :class:`~qiskit_nature.algorithms.MP2PointGenerator`
```
mrossinek, line 14:
```suggestion
    `initial_point="MP2"` and :class:`~qiskit_nature.algorithms.MP2InitialPoint`
```

mrossinek: Actually this needs some more rephrasing because we dropped the string support on `initial_point`.

releasenotes/notes/migrate-mp2-info-b1a3ee7320c46053.yaml
```diff
@@ -0,0 +1,22 @@
...
+    will be built within the factory using the same excitations as
+    `~qiskit_nature.circuit.library.UCC` automatically. This initial point will
+    then be passed to VQE.
+
+    Outside of the factory, one can pass the initial point for VQE by manually
+    getting the excitation list from UCC and instantiating
+    :class:`~qiskit_nature.algorithms.MP2PointGenerator` with it and the relevant
```
mrossinek, line 21:
```suggestion
    :class:`~qiskit_nature.algorithms.MP2InitialPoint` with it and the relevant
```

test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
```diff
@@ -539,6 +540,31 @@ def test_print_result(self):
...
+
+        solver = VQEUCCFactory(QuantumInstance(BasicAer.get_backend("statevector_simulator")))
+        calc = GroundStateEigensolver(self.qubit_converter, solver)
+        res = calc.solve(self.electronic_structure_problem)
+
+        self.assertIsNone(solver.initial_point)
+        self.assertAlmostEqual(res.total_energies[0], self.reference_energy, places=6)
+
+    def test_vqe_ucc_factory_with_mp2(self):
+        """Test when using MP2PointGenerator to generate the initial point."""
```
mrossinek, line 554:
```suggestion
        """Test when using MP2InitialPoint to generate the initial point."""
```

test/algorithms/initial_points/test_initial_point.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Test the point generator base class."""
```
mrossinek, line 13:
```suggestion
"""Test the initial point generator base class."""
```

test/algorithms/initial_points/test_mp2_initial_point.py
```diff
@@ -0,0 +1,126 @@
...
+    def setUp(self):
+        super().setUp()
+        settings.dict_aux_operators = True
+        self.mock_ansatz = Mock()
+        self.mock_ansatz.operators = None
+
+    def test_mp2_missing_orbital_energies(
+        self,
+    ):
+        """Test MP2 InitialPoint raises errors for bad input."""
```
mrossinek, line 63:
```suggestion
        """Test MP2InitialPoint raises errors for bad input."""
```

test/algorithms/initial_points/test_mp2_initial_point.py
```diff
@@ -0,0 +1,126 @@
...
+        atom1,
+        atom2,
+        distance,
+        initial_point,
+        energy_delta,
+        energy_deltas,
+        energy,
+        excitations,
+    ):
+        """Test MP2 InitialPoint with several real molecules."""
```
mrossinek, line 88:
```suggestion
        """Test MP2InitialPoint with several real molecules."""
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
pbark: I agree with @mrossinek that this would be a good addition. Idk if this should be done here or in separate PR. @declanmillar if you can add this functionality also would be great.

declanmillar: Yep, I think that sounds like a usefully illustrative addition. I'm travelling over the next few days, but I will chip away to resolve these comments when stuck on trains and planes.

qiskit_nature/algorithms/point_generators/mp2_point_generator.py
```diff
@@ -0,0 +1,216 @@
...
+
+    @property
+    def energy_delta(self) -> float:
+        """The MP2 delta energy correction for the molecule."""
+        return sum(self.energy_deltas)
+
+    @property
+    def energy(self) -> float:
+        """The absolute MP2 energy for the molecule."""
+        return self._reference_energy + self.energy_delta
```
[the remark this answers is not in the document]
woodsp-ibm: I am not sure I follow with the need - its not energy which is extracted like the transformers do right which alters the value we will find and needs to be added back at the end, To me its more like the HF energy as in its a value that is computed along the way. Although we set the MP2 informed initial point, I don't think this ends up evaluating as MP2 energy from VQE in the same way the HF starting point results in HF energy.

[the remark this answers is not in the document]
mrossinek: Ah yes you are right! :thinking: 
So this is more an extra piece of information rather than a partial energy which the user needs to take into account in the end, right?

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
woodsp-ibm: Having an HFInitialPoint sounds fine - it would go some way to sorting out #571 , in regards of removing the preferred initial point logic from EvolvedOperatorAnsatz. Something is needed for UVCC Factory though to set initial point to 0 once that is removed as a preferred initial point that VQE gets from the EvolvedOperatorAnsatz

qiskit_nature/algorithms/point_generators/mp2_point_generator.py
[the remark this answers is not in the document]
woodsp-ibm: Yes, just information if you want it. So I would not say they "need to take this into account" any more than the HF ref energy that is available if its wanted.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
declanmillar: I decided to open a separate PR for this (#632) for more clarity.

mrossinek (dismissed):
LGTM, thanks @declanmillar!

qiskit_nature/algorithms/point_generators/mp2_point_generator.py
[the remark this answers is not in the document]
declanmillar: Actually, to make this addition easier to review, I decided to raise a separate issue (#633) and will open a new PR.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -113,12 +116,12 @@ def optimizer(self, optimizer: Optional[Optimizer]) -> None:
         self._optimizer = optimizer
 
     @property
-    def initial_point(self) -> Optional[np.ndarray]:
+    def initial_point(self) -> Optional[Union[np.ndarray, InitialPoint]]:
```
woodsp-ibm, line 119: In #570 these setters are being deprecated with the underlying algorithm being exposed and the setting of whatever properties it has being direct (otherwise this mirrored the same parameters and needs maintenance etc to keep it in sync with the algorithm that seems burdensome and unnecessary). With this being kept here, to allow these specific points to be used, I guess we will need to document how things work together in that regard.

qiskit_nature/algorithms/initial_points/initial_point.py
```diff
@@ -0,0 +1,44 @@
...
+
+from abc import ABC, abstractmethod
+
+import numpy as np
+
+from qiskit_nature.properties.second_quantization import GroupedSecondQuantizedProperty
+from qiskit_nature.circuit.library import UCC
+
+
+class InitialPoint(ABC):
```
woodsp-ibm, line 23: So I am wondering if/how this might extend to UVCC+VQE. This seems to be generic in its name yet I guess it ends up more specific to UCC? I am thinking of #571 and how that relates to UVCC+VQE factory if/when EvolveOpAnsatz drops the preferred point logic. As in does this InitialPoint class concept extend to there too; should any naming etc here perhaps accommodate for that. Just putting this out there for feedback from others too.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def _compute_corrections(
+        self,
+    ) -> dict[str, tuple[float, float]]:
+        """Compute the MP2 coefficients and energy corrections for each double excitation.
+
+        Non-double excitations will have zero coefficient and energy delta.
+
+        Returns:
+            Dictionary with MP2 coefficients and energy_deltas for each excitation
```
dlasecki, line 195:
```suggestion
            Dictionary with MP2 coefficients and energy_deltas for each excitation.
```

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,255 @@
...
+            + self._orbital_energies[a]
+            - self._orbital_energies[i]
+            - self._orbital_energies[j]
+        )
+        coeff = -num / denom
+        coeff = coeff if abs(coeff) > self._threshold else 0
+        e_delta = coeff * tiajb
+        e_delta = e_delta if abs(e_delta) > self._threshold else 0
+
+        return (coeff, e_delta)
```
dlasecki, line 255:
```suggestion
        return coeff, e_delta
```

qiskit_nature/circuit/library/ansatzes/ucc.py
```diff
@@ -237,6 +241,15 @@ def excitations(self, exc: Union[str, int, List[int], Callable]) -> None:
         self._invalidate()
         self._excitations = exc
 
+    @property
+    def excitation_list(self) -> List[Tuple[Tuple[int, ...], Tuple[int, ...]]]:
+        """The excitation list that UCC is using, in the required format."""
+        if self._excitation_list is None:
+            raise QiskitNatureError(
```
dlasecki, line 248: Raised error not documented.

qiskit_nature/circuit/library/ansatzes/uccsd.py
```diff
@@ -49,7 +49,7 @@ def __init__(
                 the occupation of the spin orbitals. As such, the set of generalized excitations is
                 only determined from the number of spin orbitals and independent from the number of
                 particles.
-        preserve_spin: boolean flag whether or not to preserve the particle spins.
+            preserve_spin: boolean flag whether or not to preserve the particle spins.
```
dlasecki, line 52:
```suggestion
            preserve_spin: Boolean flag whether or not to preserve the particle spins.
```

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,255 @@
...
+        denom = (
+            self._orbital_energies[b]
+            + self._orbital_energies[a]
+            - self._orbital_energies[i]
+            - self._orbital_energies[j]
+        )
+        coeff = -num / denom
+        coeff = coeff if abs(coeff) > self._threshold else 0
+        e_delta = coeff * tiajb
+        e_delta = e_delta if abs(e_delta) > self._threshold else 0
```
dlasecki, line 253: Can we use better names, e.g. `correction_coeff`, `energy_delta` etc. How about `denom`? Can we assign it some meaning?

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,255 @@
...
+            coeff = -(2 * T[i,a,j,b] - T[i,b,j,a)/(E[b] + E[a] - E[i] - E[j])
+        where E is the orbital energy.
+        and an energy delta given by
+            e_delta = coeff * T[i,a,j,b]
+
+        All the computations are done using the molecule orbitals but the indices used
+        in the excitation information passed in and out are in the block spin orbital
+        numbering as normally used by the nature module:
+          - alpha runs from 0 to num_orbitals - 1
+          - beta runs from num_orbitals to num_orbitals * 2 - 1
```
dlasecki, line 225:
```suggestion
        numbering as normally used by the nature module:
          - alpha runs from 0 to num_orbitals - 1,
          - beta runs from num_orbitals to num_orbitals * 2 - 1.
```

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,255 @@
...
+    def _compute_correction(
+        self, excitation: tuple[tuple[int, ...], tuple[int, ...]]
+    ) -> tuple[float, float]:
+        """Compute the MP2 coefficient and energy corrections given a double excitation.
+
+        Each double excitation given by [i,a,j,b] has a coefficient computed using
+            coeff = -(2 * T[i,a,j,b] - T[i,b,j,a)/(E[b] + E[a] - E[i] - E[j])
+        where E is the orbital energy.
+        and an energy delta given by
+            e_delta = coeff * T[i,a,j,b]
```
dlasecki, line 219:
```suggestion
        and an energy delta given by
            e_delta = coeff * T[i,a,j,b].
```

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,255 @@
...
+
+        return corrections
+
+    def _compute_correction(
+        self, excitation: tuple[tuple[int, ...], tuple[int, ...]]
+    ) -> tuple[float, float]:
+        """Compute the MP2 coefficient and energy corrections given a double excitation.
+
+        Each double excitation given by [i,a,j,b] has a coefficient computed using
+            coeff = -(2 * T[i,a,j,b] - T[i,b,j,a)/(E[b] + E[a] - E[i] - E[j])
```
dlasecki, line 216:
```suggestion
            coeff = -(2 * T[i,a,j,b] - T[i,b,j,a])/(E[b] + E[a] - E[i] - E[j])
```
Can we perhaps use math mode?

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,255 @@
...
+        """The absolute MP2 energy for the molecule."""
+        return self._reference_energy + self.energy_delta
+
+    def get_initial_point(
+        self, driver_result: GroupedSecondQuantizedProperty, ansatz: UCC
+    ) -> np.ndarray:
+        """
+        Args:
+            driver_result: the second quantization properties from the driver.
+            ansatz: the UCC ansatz.
```
dlasecki, line 177:
```suggestion
            driver_result: The second quantization properties from the driver.
            ansatz: The UCC ansatz.
```

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,255 @@
...
+    @threshold.setter
+    def threshold(self, thr: float) -> None:
+        """The energy threshold for MP2 computation.
+
+        Computed initial point and energy deltas will be set to
+        zero if their value is below this threshold.
+        """
+        if thr < 0.0:
+            raise ValueError("The energy threshold cannot be negative.")
+        self._threshold = thr
```
dlasecki, line 149:
```suggestion
    def threshold(self, threshold: float) -> None:
        """The energy threshold for MP2 computation.

        Computed initial point and energy deltas will be set to
        zero if their value is below this threshold.
        """
        if threshold < 0.0:
            raise ValueError("The energy threshold cannot be negative.")
        self._threshold = threshold
```

dlasecki (changes requested):

qiskit_nature/circuit/library/ansatzes/ucc.py
declanmillar: Thanks for spotting this. Now raised.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
declanmillar: I usually only use setters if I want to do some validation/modification/extraction. As I'm not doing that here, should I just get ahead of the curve and take out this setter before it's merged? I think I only included it to mirror the existing setters anyway.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
declanmillar: Sure, I've now made these variable names more explicit.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
declanmillar: I don't think the getter/setter is necessary for initial_point. So, I've just exposed it directly to match upcoming changes in #570.

qiskit_nature/algorithms/initial_points/initial_point.py
declanmillar: @woodsp-ibm, should I open a separate issue to review this, or document it in #571 or #632 to get more eyes on it? I suppose, in principle, any MinimumEigensolverFactory using any evolved ansatz circuit could be provided with some InitialPoint that provides a suitable ndarray on request, but, I have currently just set it up for UCC. I suppose the options, as you've already laid out, are:

1. Open up InitialPoint to be used with any MinimumEigensolverFactory instance, and EvolvedOperator ansatz  (I don't know how useful this will be in practice.
2. Create a parent class for both UCC+VQE and UVCC+VQE, such that they inherit this functionality, and refactor InitialPoint to expect either a UCC or UVCC ansatz (or a new interface).
3. Rename InitialPoint to UCCInitialPoint or similar.

mrossinek: @declanmillar Sorry, I meant to already reply here. I think it makes sense to keep the UCC and UVCC stacks aligned. I already talked to @paulineollitrault and we discussed that the "HF-equivalent" in the vibrational stack should be called `VSCF` (like the in in the initial state library).

woodsp-ibm: In terms of support UVCC+VQE, I think at present it would be limited to having an equivalent to the HF initial point as Max describes above. Allowing other points to be developed, though we would not be including any (beyond what you are doing with UCC and MP2). For UVCC this would be another all zeros point. (This will then allow the removal of the preferred initial point all zeros from the EvolvedOpAnstaz base class, and VQE's use of that, relying instead on explicit initial point setting as per what you have done so far.) As to how to accommodate this type of initial point, maybe in the 632 PR as you suggest, which may end up renaming/refactoring some of these classes a little to broaden things out. The reason these factories exist is that the ansatz/initial state need configuring with information from the problem, so its not needed for any ansatz per-se. UVCC+VQE is already "similar" but I am not sure that whether a common parent, or perhaps mixin, is needed. Maybe things will become more evident, either way, when its worked out further. I would be wary though of trying to generalize things too much at this point - I was really thinking more of consistency around UCC and UVCC factories in regards of the initial point handling at this stage. Things can always be refactored later in the future if more generality is needed if perhaps we then have more cases to consider etc.

declanmillar: I've added a note to consider this as part of PR #632.

declanmillar:
@pbark, I'm pretty sure I addressed your changes, but I incorporated them manually with some changes I was working on at the time, so GitHub didn't catch on. Just mentioning it as I don't seem to be able to clear the "changes requested" blocker.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,261 @@
...
+          - ::math::`\\beta` runs from ``num_orbitals`` to ``num_orbitals * 2 - 1``.
+
+        Args:
+            excitations: Sequence of excitations.
+
+        Returns:
+            Correction coefficients and energy deltas.
+        """
+        # Since spins are the same drop to molecular orbital indexing
+        num_orbitals = self._num_spin_orbitals // 2
```
mrossinek, line 238: Follow-up to earlier: since you actually want `num_mos` rather than `num_spin_orbitals` you can even more so extract this from `self._integral_matrix.shape[0]`

mrossinek (commented):
Some more comments but other than this, it LGTM :+1:

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,261 @@
...
+    def energy(self) -> float:
+        """The absolute MP2 energy for the molecule."""
+        return self._reference_energy + self.energy_delta
+
+    def get_initial_point(
+        self, driver_result: GroupedSecondQuantizedProperty, ansatz: UCC
+    ) -> np.ndarray:
+        """
+        Args:
+            driver_result: The second quantization properties from the driver.
```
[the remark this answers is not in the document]
declanmillar: @mrossinek I renamed this property to `grouped_property` in `initial_point` and `mp2_initial_point` and updated the docstrings and error handling to try to make the requirements clearer. Let me know if it's still lacking.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,261 @@
...
+    @electronic_energy.setter
+    def electronic_energy(self, elec: ElectronicEnergy) -> None:
+        """The ElectronicEnergy property.
+
+        Raises:
+            ValueError: If ElectronicEnergy is missing orbital energies.
+        """
+        orbital_energies = elec.orbital_energies
+        if orbital_energies is None:
+            raise ValueError("The orbital energies are missing from the electronic energy.")
```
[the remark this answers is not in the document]
declanmillar: @mrossinek I've refactored the `grouped_property` setter a bit. I think it should now catch cases where the user sets it via the property.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,261 @@
...
+
+    @driver_result.setter
+    def driver_result(self, result: GroupedSecondQuantizedProperty) -> None:
+        """The driver result."""
+
+        electronic_energy = result.get_property(ElectronicEnergy)
+        self.electronic_energy = electronic_energy
+
+        particle_number = result.get_property(ParticleNumber)
+        self.num_spin_orbitals = particle_number.num_spin_orbitals
```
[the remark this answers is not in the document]
declanmillar: @mrossinek that's a very nice suggestion. I've made this change. I guess at this point, we could just pass in ElectronicEnergy rather than the grouped_property, but I've kept it to grouped_property so a user can just pass in a driver result and with an eye to other initial points and might need different properties.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,261 @@
...
+        Returns:
+            Dictionary with MP2 coefficients and energy_deltas for each excitation.
+        """
+        corrections = {}
+        for excitation in self._excitations:
+            if len(excitation[0]) == 2:
+                # MP2 needs double excitations.
+                coeff, e_delta = self._compute_correction(excitation)
+            else:
+                # Leave single excitations unchanged.
```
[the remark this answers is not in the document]
declanmillar: I've updated these comments to, hopefully, be a bit clearer.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
declanmillar: I've made this change and commented on it above.

qiskit_nature/algorithms/initial_points/initial_point.py
```diff
@@ -0,0 +1,66 @@
...
+    def initial_point(self) -> np.ndarray:
+        """The initial point."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_initial_point(
+        self, grouped_property: GroupedSecondQuantizedProperty | None, ansatz: UCC | None
+    ) -> np.ndarray:
+        """Computes the initial point."""
+        raise NotImplementedError
```
mrossinek, line 46: I am not sure if I overlooked this earlier or if this got updated slightly. But from a user perspective seeing both `initial_point` and `get_initial_point` could be a bit confusing, especially since the `initial_point` will fail if `get_initial_point` has not been called earlier.

I would do one of the following:
- rename `get_initial_point -> compute_initial_point` and add a guard inside of `initial_point` to call this if `self._corrections` is None (talking MP2 specific here)
- remove `initial_point`

If you do either of these, the documentation also in MP2 will need updating

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,302 @@
...
+
+        Raises:
+            QiskitNatureError: If the ``ElectronicEnergy`` is missing or the two-body molecular
+            orbital matrix, the orbital energies, and/or the Hartree-Fock reference energy are
+            missing from ``ElectronicEnergy``.
+        """
+        try:
+            electronic_energy: ElectronicEnergy = grouped_property.get_property(ElectronicEnergy)
+        except TypeError as invalid_grouped_property:
+            raise QiskitNatureError from invalid_grouped_property
```
mrossinek, line 74: I do not think this will ever happen because `get_property` will safely return `None` even if `ElectronicEnergy` does not exist.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,302 @@
...
+
+        try:
+            self._integral_matrix: np.ndarray = electronic_energy.get_electronic_integral(
+                ElectronicBasis.MO, 2
+            ).get_matrix()
+            # Infer the number of molecular orbitals from the MO matrix.
+            self._num_molecular_orbitals: int = self._integral_matrix.shape[0]
+        except TypeError as missing_two_body_mo_matrix:
```
mrossinek, line 82: Here, the error which will be raised is actually an `AttributeError` because the missing integral will again be `None` resulting in `get_matrix()` being called on a `NoneType`.

I suggest you also add some unittests which specifically trigger these exceptions

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,302 @@
...
+        if self._orbital_energies is None:
+            raise QiskitNatureError(
+                f"Orbital energies not found in grouped property: {grouped_property}"
+            )
+
+        self._reference_energy: float = electronic_energy.reference_energy
+        if self._reference_energy is None:
+            raise QiskitNatureError(
+                f"Hartree-Fock reference energy not found in grouped property: {grouped_property}"
+            )
```
mrossinek, line 95: Is this energy strictly needed or is it purely optional in order to provide the correct total energy? Maybe you can log a warning rather than raise one and set the internal object to `float("NaN")`. (Note: this will cause `.energy` to also return `NaN`, but `.energy_delta` is still available so I'd say this is fine)

mrossinek (commented):
Just some minor comments left, then this will be good :+1:

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
declanmillar: So, previously I had `.energy` just return `.delta` if reference_energy was None. (i.e. just make the reference_energy 0). But then I thought the reference energy would always be zero. I can reinstate that logic (with a warning).

qiskit_nature/algorithms/initial_points/initial_point.py
declanmillar: So, I sort of did both. I didn't really like `InitialPoint.get_initial_point` or `InitialPoint.initial_point`. As it's a bit confusing and verbose. It's a leftover from when the class was briefly called `MP2InformedStart`.

Now if the user tries to call a computed property such as `x` (the actual initial point array) before calling `.compute`, it'll try to call `.compute` with the `grouped_property` and `ansatz` that it already has. If it can't be computed, it'll just raise a warning and return None. I realize that `x` is also a controversial property name 😎, but I felt justified due to the way `x` is used in the [OptimizationResult](https://qiskit.org/documentation/_modules/qiskit/algorithms/optimizers/optimizer.html#OptimizerResult). Still happy to change it, regardless.

declanmillar: NB: Refactoring again to use .to_numpy_array(), and more clearly separate computation and conversion.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
declanmillar: I've overhauled these checks and added unit tests specific to each error raised.

qiskit_nature/algorithms/initial_points/initial_point.py
declanmillar: I've now rewritten the API to separate the computation and conversion etc. I've also limited the class properties to those that I expect the user to set and updated all the docs (hopefully I didn't miss any outdated references).

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,363 @@
...
+    A class for computing the second-order Møller-Plesset perturbation theory (MP2) corrections.
+    This is intended for use as an initial point for the VQE parameters in combination with a UCC
+    ansatz.
+
+    :class:`MP2InitialPoint` requires two-body molecular orbital electronic integrals and orbital
+    energies from
+    :class:`~qiskit_nature.properties.second_quantization.electronic.ElectronicEnergy`,
+    which should be passed in via the :attr:`grouped_property` attribute.
+
+    :class:`MP2InitialPoint` also requires the :attr:`excitation list` from the :attr:`~ansatz`
```
mrossinek, line 45:
```suggestion
    :class:`MP2InitialPoint` also requires the :attr:`~excitation_list` from the :attr:`~ansatz`
```
Assuming the `~` is really needed since it is also used below.

mrossinek: Follow-up: if the `~` is indeed needed, you probably also need it for the `grouped_property` in various other places of this docstring.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,363 @@
...
+    to ensure that the coefficients are mapped correctly in the initial point array. This can be
+    overwritten by setting :attr:`~excitation_list` directly.
+
+    The coefficients and energy corrections are computed using :meth:`~compute`, which requires
+    :attr:`grouped_property` and :attr:`~ansatz` to be set or passed as arguments to
+    :meth:`~compute`.
+
+    Following computation, the initial point array can be extracted via `:meth:to_numpy_array`.
```
mrossinek, line 53:
```suggestion
    Following computation, the initial point array can be extracted via :meth:`to_numpy_array`.
```

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,363 @@
...
+    Array elements with indices corresponding double excitations in the :attr:`~excitation_list`
```
mrossinek, line 54:
```suggestion
    Array elements with indices corresponding to double excitations in the :attr:`~excitation_list`
```

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,363 @@
...
+    will use the computed MP2 coefficient, while those that correspond to single, triple, or
+    higher excitations will be zero.
+
+    The array of energy corrections for each excitation can be
+    recovered using :meth:`get_energy_corrections`. The overall correction is got via
```
mrossinek, line 59:
```suggestion
    recovered using :meth:`get_energy_corrections`. The overall correction can be obtained via
```
You may need to reformat slightly to avoid the line length limit being exceeded

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,363 @@
...
+            )
+
+        orbital_energies: np.ndarray | None = electronic_energy.orbital_energies
+        if not isinstance(orbital_energies, np.ndarray):
+            raise QiskitNatureError(f"Orbital energies not in grouped property: {grouped_property}")
+
+        # Invalidate any previous computation.
+        self._corrections = None
+
+        self._integral_matrix = two_body_mo_integral.get_matrix()
```
mrossinek, line 124: This implies that the current implementation will only work for cases where the alpha and beta spin MOs are identical (i.e. came out of a restricted spin calculation). While this is okay for the time being, you should check that this is indeed the case, and, if it is not, raise an error here:
```suggestion
        self._integral_matrix = two_body_mo_integral.get_matrix()
        if not np.allclose(self._integral_matrix, two_body_mo_integral.get_matrix(2)):
            raise NotImplementedError("MP2InitialPoint only supports restricted-spin setups.")
```
The error message can be improved slightly. You should also open an issue which tracks the progress of extending this implementation to unrestricted-spin cases (should not be too hard).

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,363 @@
...
+
+    @ansatz.setter
+    def ansatz(self, ansatz: UCC) -> None:
+        """The UCC ansatz.
+
+        This is used to ensure that the :attr:`excitation_list` matches with the UCC ansatz that
+        will be used with the VQE algorithm.
+        """
+        # Operators must be built early to compute excitation list.
+        _ = ansatz.operators
```
mrossinek, line 142: Should we check the type of `ansatz` to actually be `UCC`?

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,363 @@
...
+        ansatz.
+
+        :meth:`~compute` requires two-body molecular orbital electronic integrals and orbital
+        energies from
+        :class:`~qiskit_nature.properties.second_quantization.electronic.ElectronicEnergy`,
+        which should be passed in via the :attr:`grouped_property` attribute.
+
+        :meth:`compute` also requires the :attr:`excitation list` from the :attr:`~ansatz` to
+        ensure that the coefficients are mapped correctly in the initial point array. This can be
+        overwritten by setting :attr:`~excitation_list` directly.
```
mrossinek, line 224: Some of this is duplicated from the class-level docstring. I think only one repetition on the class-level is fine so you can deduplicate things here.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,363 @@
...
+            Coefficient and energy correction for a given double excitation.
+        """
+        integral_matrix = self._integral_matrix
+        orbital_energies = self._orbital_energies
+        threshold = self._threshold
+
+        # Infer the number of molecular orbitals from the MO matrix.
+        num_molecular_orbitals: int = integral_matrix.shape[0]
+
+        i = excitation[0][0] % num_molecular_orbitals
```
mrossinek, line 317: A suggestion for the future when extending this to unrestricted spin cases:
```suggestion
        a_or_b_i, i = divmod(excitation[0][0], num_molecular_orbitals)
```
`a_or_b_i` will then be either 0 or 1 indicating which set of 2-body terms to use.

test/algorithms/initial_points/resources/test_data_mp2_point_generator.json
```diff
@@ -0,0 +1,366 @@
+[
```
mrossinek, line 1: This files name can be updated

mrossinek (commented):
Okay I have another set of comments but then we should be done.
I really like the API now :+1:

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
declanmillar: Good spot. It looks like it does format the same in the docs with and without the tilde (once the underscore is fixed). I'll remove the tildes for consistency and check that all the cross-references work.
Screen Shot 2022-04-22 at 10 56 25
Screen Shot 2022-04-22 at 10 56 32

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
declanmillar: Thanks! Added as a suggestion to #645.

mrossinek (approved):
This LGTM! Thanks a lot @declanmillar for your hard work on this!

@pbark @woodsp-ibm please also take a look before we merge :+1:

.pylintdict
```diff
@@ -270,6 +270,7 @@ moc
 mohij
 mohijkl
 mol
+møller
```
dlasecki, line 273: I wonder if this special character might have a different ordering w.r.t. "o".

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,347 @@
...
+                )
+
+        if isinstance(ansatz, UCC):
+            self.ansatz = ansatz
+        else:
+            if not isinstance(self._excitation_list, list):
+                raise QiskitNatureError(
+                    "The excitation list has not been set directly or via the ansatz. "
+                    + missing_input_error_message
+                )
```
dlasecki, line 264: Shall this become `elif`?

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -0,0 +1,347 @@
...
+        )
+
+        if isinstance(grouped_property, GroupedSecondQuantizedProperty):
+            self.grouped_property = grouped_property
+        else:
+            if not isinstance(self._grouped_property, GroupedSecondQuantizedProperty):
+                raise QiskitNatureError(
+                    "The grouped property has not been set. " + missing_input_error_message
+                )
+
```
dlasecki, line 256: Shall this become `elif`?

.pylintdict
declanmillar: Good point. So, I'm going to go with the ordering here https://en.wikipedia.org/wiki/Danish_and_Norwegian_alphabet. I did do some googling around to see if there's an agreed-upon ordering for all characters across all languages (to future proof the list) but didn't have much luck.

woodsp-ibm: To start with, and I think we have done it on occasions as well, is that this run by `sort` and I think that obeys, or looks at, what is defined for collating across multiple character sets. So while we more or less maintain this manually I might expect it to get run by sort in the future; or even regenerated since if we delete this, and assume all spellings are correct, then any thing picked will not be the in the standard dict and needs to be in such a file. I think we have done this in the past during refactor/moves etc. So it seems ok for now, and is possibly nothing to stress over too much.

mrossinek (approved):

pbark (approved):
This now also LGTM :) Thanks a lot @declanmillar !

Issue #577, opened 2022-02-24
Heisenberg model


jvscursulim:
### What should we add?

I saw that recently the Ising model was added to Qiskit Nature, this model is a special case of the Heisenberg model (https://en.wikipedia.org/wiki/Quantum_Heisenberg_model). Thus, I would like let you all know that I would like to contribute to Qiskit Nature adding this model, because I have developed a code that implements this model during the Qiskit Advocate Mentorship Program Fall 2021.

pbark:
Hi @jvscursulim ! Thanks for you  interest in contributing to our lattice models! Your contribution is very welcome. 

Please take a look [here](https://github.com/Qiskit/qiskit-nature/blob/main/CONTRIBUTING.md) or  [here](https://qiskit.org/documentation/contributing_to_qiskit.html) about contribution guidelines. 

Looking forward to your PR.

jvscursulim:
Thanks @pbark , I will read the contribution guidelines and I will create the PR as soon as possible.

PR #591, opened 2022-03-10
Heisenberg model

jvscursulim:
### Summary

This PR is related with the Issue #577 Heisenberg model. The main idea behind it, is the addition of the Heisenberg model aiming the expansion of the lattice models.

### Details and comments

I used the ising_model.py to guide the process of rewriting my original code, to avoid any conflicts with the logic behind the construction of the models through a lattice. However, I had to add some different aspects to heisenberg_model.py, once this model is more general than the Ising model, actually the Ising model is a specific case of Heisenberg model ( J_x = J_y= 0, J_z = 1 and h = 1). I tried to follow the logic of getting all information to build the model of the lattice, but this approach could lead to some problems, once we have XX, YY and ZZ terms, multiplied by constants J_x, J_y and J_z, that not necessarily must be equal. Another difference between the models is that in the Heisenberg model the anisotropic term/external magnetic field is not restricted to the x-axis. Thus, I thought it would be better if I just used the lattice to get the number of qubits and the connection between the nodes, instead of making use of multigraphs. Then, if we want to some special configuration we need to give the HeisenbergModel class two dictionaries, one to set the model constants, and other to set the direction of the external magnetic field, if the user does not pass any dictionaries to the method, a default configuration will be used, the ferromagnetic without an external magnetic field.
To finish my comments, I would like to ask an opinion of the reviewers about the construction of the code, once some of you could have a better approach to the construction of the model.


manoelmarques:
The file .pylintdict contains word exceptions used by the spell checker and shouldn't be deleted.

.pylintdict
```diff
@@ -1,509 +0,0 @@
-adjoint
```
manoelmarques, line 1: The file .pylintdict contains word exceptions for the spell checker and shouldn't be deleted.

jvscursulim: I'm sorry, I only deleted the .pylintdict because some commits that I made in another PR come together in this PR. I'll fix this.

pbark:
Hi @jvscursulim, thanks for the contribution. I had a first pass and this looks quite good already. As this is a new feature can you please add a release note ?

test/problems/second_quantization/lattice/models/test_heisenberg_model.py
```diff
@@ -0,0 +1,61 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021, 2022.
```
pbark, line 3:
```suggestion
# (C) Copyright IBM 2022.
```

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
```diff
@@ -0,0 +1,102 @@
...
+    """The Heisenberg model."""
+
+    def __init__(
+        self,
+        lattice: Lattice,
+        model_constants: Optional[dict] = None,
+        ext_magnetic_field: Optional[dict] = None,
+    ) -> None:
+        """
+        Args:
```
pbark, line 34: please add also a description of the `lattice` argument in the docstring.

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
```diff
@@ -0,0 +1,103 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"The Heisenberg model"
```
mrossinek, line 13:
```suggestion
"""The Heisenberg model."""

```

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class HeisenbergModel(LatticeModel):
+    """The Heisenberg model."""
+
+    def __init__(
+        self,
+        lattice: Lattice,
+        model_constants: Optional[dict] = None,
+        ext_magnetic_field: Optional[dict] = None,
```
mrossinek, line 31: These should not be dictionaries.
If I am understanding things correctly, we have the following parameters:
- `J_x`, `J_y`, `J_z`
- `B_x`, `B_y`, `B_z`
- `h`

How about, `J` simply gets provided as a tuple of floats of length 3.
The same can be done for `B`. That said, `B` even appears to have bool values, to simply turn on/off `h` along the given axis. Doesn't this make `h` redundant if we can simply have `B` be another tuple of floats of length 3?

Furthermore, in the `IsingModel`, we deliberately kept `lattice` as the sole input argument to `__init__`. I think we could do the same here and then supply `@classmethod` "constructors" similar to the `IsingModel.from_uniform_parameters` which would inject `J` and `B`, respectively. In that way, we could even have a method which simply provides `h` that gets expanded into the proposed tuple `B`.

mrossinek (changes requested):
Since my proposed change effectively changes the entire structure of the PR, I am going to leave the review at that for now.

Besides this comment, we also recently merge #746 which majorly restructures the code. Please update your PR accordingly be moving this class to `qiskit_nature/second_q/hamiltonians` (where the other lattice models will end up, too).

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
jvscursulim: I agree with your suggestions. I wrote the inputs as dictionaries, because I did not know how I would inject them using the lattice. But now, with your explanation, I'm starting to see a way to do this.

I will update the code according with #746 and I will try to rewrite the class in this new way.

jvscursulim:
@mrossinek I tried to create a new version of the `HeisenbergModel` class following your instructions, but I'm having some troubles to inject the coupling constants `J` (a `tuple`) and the external magnetic field `B` (a `tuple`) as you suggested. In order to solve this problem, I put `J` and `B` as input of the `second_q_ops()` method and defined some default values to them. My problem with `J` and `B` injection using the method that you suggested is that I don't know how to deal with it when the tuple `J` has different coupling constants and the tuple `B` only has one value nonzero, that means we are applying an external magnetic field in one specific direction. I hope that you can understand what I'm saying, but if you don't, the things I'm trying to say are present in the following wikipedia article: https://en.wikipedia.org/wiki/Quantum_Heisenberg_model

mrossinek:
I don't actually fully understand what the problem is in that scenario. It should be identical to what you had before, just using tuples instead of named dictionary entries..

That said, `second_q_ops()` cannot take any arguments. You must provide `J` and `B` as arguments to `__init__` and store them as instance variables on `self`

qiskit_nature/problems/second_quantization/lattice/models/__init__.py
```diff
@@ -13,6 +13,7 @@
 """Models"""
 from .fermi_hubbard_model import FermiHubbardModel
 from .ising_model import IsingModel
+from .heisenberg_model import HeisenbergModel
```
mrossinek, line 16:
```suggestion
```

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
```diff
@@ -0,0 +1,131 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""The Heisenberg model."""
```
mrossinek, line 13: Please move this file to: `qiskit_nature/second_q/properties/heisenberg_model.py`
It's actual final destination will be in the `hamiltonians` module but it will be easier to refactor alongside with the `FermiHubbardModel` and `IsingModel` once we start work on that :+1:

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
```diff
@@ -0,0 +1,131 @@
...
+        Args:
+            lattice: Lattice on which the model is defined.
+            coupling_constants (Tuple, optional): Coupling constants. Defaults to (1.0, 1.0, 1.0).
+            ext_magnetic_field (Tuple, optional): External magnetic field. Defaults to (0.0, 0.0, 0.0).
+        """
+        super().__init__(lattice)
+        self.coupling_constants = coupling_constants
+        self.ext_magnetic_field = ext_magnetic_field
+
+    def coupling_matrix(self) -> np.ndarray:
```
mrossinek, line 45: Do we need to expose the `interaction_matrix` under this new name?

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
```diff
@@ -0,0 +1,131 @@
...
+
+        Returns:
+            HeisenbergModel: The Heisenberg model generated from the given interaction
+                matrix and on-site interaction.
+        """
+        return cls(cls._generate_lattice_from_parameters(interaction_matrix))
+
+    def second_q_ops(
+        self,
+        display_format: Optional[str] = None,
```
mrossinek, line 85: You can remove this argument. It will be removed from the other classes, too :+1:

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
```diff
@@ -0,0 +1,131 @@
...
+        for node_a, node_b, _ in weighted_edge_list:
+
+            if node_a == node_b:
+                index = node_a
+                if self.ext_magnetic_field[0] != 0:
+                    hamiltonian.append((f"X_{index}", self.ext_magnetic_field[0]))
+                if self.ext_magnetic_field[1] != 0:
+                    hamiltonian.append((f"Y_{index}", self.ext_magnetic_field[1]))
+                if self.ext_magnetic_field[2] != 0:
+                    hamiltonian.append((f"Z_{index}", self.ext_magnetic_field[2]))
```
mrossinek, line 114: May I suggest smth like the following:
```python
for axis, coeff in zip("XYZ", self.ext_magnetic_field):
    if not np.isclose(coeff, 0.0):
        hamiltonian.append((f"{axis}_{index}", coeff))
```

You can do the same in the case of the `coupling_constants`

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
```diff
@@ -0,0 +1,131 @@
...
+                if self.coupling_constants[1] != 0:
+                    hamiltonian.append(
+                        (f"Y_{index_left} Y_{index_right}", self.coupling_constants[1])
+                    )
+                if self.coupling_constants[2] != 0:
+                    hamiltonian.append(
+                        (f"Z_{index_left} Z_{index_right}", self.coupling_constants[2])
+                    )
+
+        return SpinOp(hamiltonian, spin=Fraction(1, 2), register_length=register_length)
```
mrossinek, line 131: Would it make sense to make the value of `spin` configurable?
If so, you could add it as another `__init__` argument

releasenotes/notes/added-heisenberg-model-218b7ca8386e208a.yaml
```diff
@@ -0,0 +1,26 @@
+---
+features:
+  - |
+    Add: :class:`qiskit_nature.problems.second_quantization.lattice.models.HeisenbergModel
```
mrossinek, line 4: This path needs to be updated in accordance with the new location

releasenotes/notes/added-heisenberg-model-218b7ca8386e208a.yaml
```diff
@@ -0,0 +1,26 @@
...
+    implementing the Hamiltonian of the Heisenberg model.`
+    This model is used in the study of critical points and phase transitions of magnetic system. 
+    Through the choice of the model constants and the external magnetic field, 
+    we can produce many models like: XYZ, XXX, Ising model and others.
+    For example::
```
mrossinek, line 9:
```suggestion
    ..code-block:: python
```

mrossinek (commented):
Other than that, the CI fixes are most important right now :+1:

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
jvscursulim: Much better! Thanks for this code suggestion

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
jvscursulim: In this case it doesn't make sense, because this model is based on spin 1/2.

qiskit_nature/problems/second_quantization/lattice/models/heisenberg_model.py
jvscursulim: I think we don't need this, I'll remove it.

mrossinek (commented):
We are looking into adding some generalizations into the lattice models in the (near-ish) future. But this is definitely a good starting point that we should merge in its current form (i.e. current generalization level; once the final touch-ups are done to this PR :+1:)

qiskit_nature/second_q/properties/heisenberg_model.py
```diff
@@ -0,0 +1,111 @@
...
+        self,
+        lattice: Lattice,
+        coupling_constants: Tuple = (1.0, 1.0, 1.0),
+        ext_magnetic_field: Tuple = (0.0, 0.0, 0.0),
+    ) -> None:
+        """
+        Args:
+            lattice: Lattice on which the model is defined.
+            coupling_constants (Tuple, optional): Coupling constants. Defaults to (1.0, 1.0, 1.0).
+            ext_magnetic_field (Tuple, optional): External magnetic field. Defaults to (0.0, 0.0, 0.0).
```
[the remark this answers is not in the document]
jvscursulim: I'll work on that asap

test/problems/second_quantization/lattice/models/test_heisenberg_model.py
```diff
@@ -0,0 +1,61 @@
...
+
+        with self.subTest("Check the graph."):
+            self.assertTrue(
+                is_isomorphic(
+                    heisenberg_model.lattice.graph, lattice.graph, edge_matcher=lambda x, y: x == y
+                )
+            )
+
+        with self.subTest("Check the second q op representation."):
+            coupling = [("X_0 X_1", 1.0), ("Y_0 Y_1", 1.0), ("Z_0 Z_1", 1.0)]
```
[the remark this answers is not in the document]
jvscursulim: I'll work on that asap

jvscursulim:
@mrossinek I changed the docstring in order to explain better `coupling_constants` and `ext_magnetic_field`. Also, I added more tests as you requested. I added one test to see if the class are creating the right model with parameters just in x and y axes, other to see the with an external magnetic field in all directions (x, y and z) and I also added a test to check the model in a triangular lattice with an external magnetic field in y direction.

test/problems/second_quantization/lattice/models/test_heisenberg_model.py
```diff
@@ -0,0 +1,103 @@
...
+from test import QiskitNatureTestCase
+from retworkx import PyGraph, is_isomorphic
+from qiskit_nature.second_q.properties.lattices import Lattice
+from qiskit_nature.second_q.properties import HeisenbergModel, IsingModel
+
+
+class TestHeisenbergModel(QiskitNatureTestCase):
+    """TestHeisenbergModel"""
+
+    def test_init(self):
```
mrossinek, line 24: The additional tests are looking good. But I would suggest that you split this into multiple test methods. For example:
- `test_line`
- `test_triangular`
- `test_ising`
- `test_xy`
- `test_external_field`
Something along those lines.

You can also use the lattice subclasses that we have (e.g. `LineLattice`, `TriangularLattice`) instead of building out those graphs manually :+1:

test/problems/second_quantization/lattice/models/test_heisenberg_model.py
```diff
@@ -0,0 +1,157 @@
...
+        with self.subTest("Check the second q op representation."):
+            terms = [("X_0 X_1", 1.0), ("Y_0 Y_1", 1.0), ("Z_0 Z_1", 1.0)]
+
+            hamiltonian = terms
+
+            self.assertSetEqual(set(hamiltonian), set(heisenberg_model.second_q_ops().to_list()))
+
+    def test_triangular(self):
+        """Test triangular lattice."""
+        triangle_graph = PyGraph(multigraph=False)
```
mrossinek, line 46: Is there any reason not to use `qiskit_nature.second_q.properties.lattices.TriangularLattice`?

mrossinek (commented):
Other than this minor detail, I think this is good to go :+1:

mrossinek:
@woodsp-ibm if you could have a look here, too, that would be great :slightly_smiling_face:

test/problems/second_quantization/lattice/models/test_heisenberg_model.py
jvscursulim: I tried to use the `TriangularLattice`, but it seems that doesn't allow me to create just one triangle, it only create a pattern with triangles. If there is a way to create a graph with just one triangle using this class, please tell me how to do it.

releasenotes/notes/added-heisenberg-model-218b7ca8386e208a.yaml
```diff
@@ -0,0 +1,27 @@
+---
+features:
+  - |
+    Add: :class:`qiskit_nature.second_q.properties.HeisenbergModel`
+    implementing the Hamiltonian of the Heisenberg model.
+    This model is used in the study of critical points and phase transitions of magnetic system.
```
woodsp-ibm, line 6:
```suggestion
    This model is used in the study of critical points and phase transitions of magnetic systems. 
```

qiskit_nature/second_q/properties/heisenberg_model.py
```diff
@@ -0,0 +1,113 @@
...
+        lattice: Lattice,
+        coupling_constants: Tuple = (1.0, 1.0, 1.0),
+        ext_magnetic_field: Tuple = (0.0, 0.0, 0.0),
+    ) -> None:
+        """
+        Args:
+            lattice (Lattice): Lattice on which the model is defined.
+            coupling_constants (Tuple, optional): The coupling constants in each Cartesian axes.
+                                                  Defaults to (1.0, 1.0, 1.0).
+            ext_magnetic_field (Tuple, optional): Represents a magnetic field in Cartesian coordinates.
```
woodsp-ibm, line 40: The types in the args here are not needed - they duplicate the info that is  already in the typehints on the parameters which Sphinx will pick up on - so these should all be removed leaving just the parameter name: description

qiskit_nature/second_q/properties/heisenberg_model.py
```diff
@@ -0,0 +1,114 @@
...
+        self.ext_magnetic_field = ext_magnetic_field
+
+    @classmethod
+    def uniform_parameters(
+        cls, lattice: Lattice, uniform_interaction: complex, uniform_onsite_potential: complex
+    ) -> HeisenbergModel:
+        """Set a uniform interaction parameter and on-site potential over the input lattice.
+
+        Args:
+            lattice (Lattice): Lattice on which the model is defined.
```
woodsp-ibm, line 55:
```suggestion
            lattice: Lattice on which the model is defined.
```

qiskit_nature/second_q/properties/heisenberg_model.py
```diff
@@ -0,0 +1,114 @@
...
+            uniform_interaction (complex): The interaction parameter.
```
woodsp-ibm, line 56:
```suggestion
            uniform_interaction: The interaction parameter.
```

qiskit_nature/second_q/properties/heisenberg_model.py
```diff
@@ -0,0 +1,114 @@
...
+            uniform_onsite_potential (complex): The on-site potential.
```
woodsp-ibm, line 57:
```suggestion
            uniform_onsite_potential: The on-site potential.
```

qiskit_nature/second_q/properties/heisenberg_model.py
```diff
@@ -0,0 +1,114 @@
...
+            )
+        )
+
+    @classmethod
+    def from_parameters(cls, interaction_matrix: np.ndarray) -> HeisenbergModel:
+        """Return the Hamiltonian of the Heisenberg model
+        from the given interaction matrix and on-site interaction.
+
+        Args:
+            interaction_matrix (np.ndarray): A real or complex valued squared matrix.
```
woodsp-ibm, line 74:
```suggestion
            interaction_matrix: A real or complex valued squared matrix.
```

woodsp-ibm (commented):
I am sure you saw from the CI that the future import needs to be the first thing

woodsp-ibm (approved):
LGTM. Thanks Victor for completing another contribution to Qiskit Nature!

jvscursulim:
> LGTM. Thanks Victor for completing another contribution to Qiskit Nature!

Thanks Steve, I'm glad in complete another contribution to Qiskit Nature!

test/problems/second_quantization/lattice/models/test_heisenberg_model.py
mrossinek: Hm yes you appear to be right about that. I was not aware of this limitation. In that case this is good as it is now :+1:

mrossinek (approved):
Thanks @jvscursulim once more, for this great contribution to Qiskit Nature! I'm looking forward to future contributions from you! :wink:

PR #601, opened 2022-03-14
UCC operator creation on demand

woodsp-ibm:
### Summary

Change UCC such that `operators` are built on demand rather than being built only when the circuit is built. This means now, that as soon as UCC has a valid configuration, a call to get `operators` or `num_qubits` will return actual values rather than None and 0 respectively. The later behavior is problematic with VQE when used more dynamically (i.e. beyond just passing in a fully configured UCC to VQE constructor). See discussion in #570.

Operators will still be built, if they were not already, when the circuit is built, but if they were built earlier it will use these. Any built operators are invalidated if UCC config is changed as before. The config checking was improved to include None checks since UCC starts out with None for num particles and spin orbitals and config check would fail previously if it gets to the point of building and either was still None.

Why it works via the VQE constructor today happens to be a side effect of a logging statement that draws the circuit and hence builds it, see https://github.com/Qiskit/qiskit-terra/blob/c9fc6d29b4c81b06d1c154690e3d2e64749e9870/qiskit/algorithms/minimum_eigen_solvers/vqe.py#L164

The updated UCC now works whether that side-effect exists or not.

Also fixes a minor bullet list formatting in the docstring that needed a new line preceding the first item in order not to be wrapped up into the paragraph text where it did not end up as a list.

### Details and comments

I should have noted, that UVCC ought to have a similar alteration. That is not included in this PR at this point.

### DIFF
--- a/qiskit_nature/circuit/library/ansatzes/ucc.py
+++ b/qiskit_nature/circuit/library/ansatzes/ucc.py
@@ -19,7 +19,7 @@ from typing import Callable, List, Optional, Sequence, Tuple, Union
 
 from qiskit.circuit import QuantumCircuit
 from qiskit.circuit.library import EvolvedOperatorAnsatz
-from qiskit.opflow import OperatorBase, PauliTrotterEvolution
+from qiskit.opflow import PauliTrotterEvolution
 
 from qiskit_nature import QiskitNatureError
 from qiskit_nature.converters.second_quantization import QubitConverter
@@ -86,6 +86,7 @@ class UCC(EvolvedOperatorAnsatz):
 
     Keep in mind, that in all of the examples above we have not set any of the following keyword
     arguments, which must be specified before the Ansatz becomes usable:
+
     - `qubit_converter`
     - `num_particles`
     - `num_spin_orbitals`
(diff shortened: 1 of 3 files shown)


manoelmarques (approved):

pbark (approved):

Cryoris:
Could we backport this PR to 0.3? Since Qiskit/qiskit-terra#7216 merged, the UCC circuits are not correctly built but this PR fixes the issue 🙂 

The problem is that the Terra PR referenced above changes the check whether a blueprint circuit is built from `self._data is None` to using a dedicated variable `self._is_built`. The `self._data` attribute is now never `None` anymore and UCC (without this PR here) fails to be built.

mrossinek:
Sounds like a reasonable request :+1: @woodsp-ibm what do you think about backporting this PR as well as #615 ?
If we decide to do this, can we still trigger the mergify backport by adding the respective label? (@manoelmarques ?)

woodsp-ibm:
@mrossinek UCC is commonly enough used that its probably best, as was requested, to have this backported as a bug fix to stable ahead of the upcoming release. And we may as well do UVCC one too at the same time since its the same fix though its probably not as widely used.

PR #607, opened 2022-03-18
Replace reduce and to_normal_order with simplify and normal_ordered

### LINKED ISSUE #551: Normal order should reduce, not the other way around
### Environment

- **Qiskit Nature version**: 98de3b06fbc8b03e4079fbf7e256169f0196facc
- **Python version**: N/A
- **Operating system**: N/A

### What is happening?

Normal ordering should put an operator into a canonical form. This means that two equivalent operators should have the same normal order. So `FermionicOp.to_normal_order` should "reduce" the operator by merging terms with the same label. It currently does not do this.

Reducing an operator should refer to merging terms with the same label. This is a useful operation in itself. It should not perform normal ordering, which takes exponential time in general. Reducing should be efficient. Currently, reducing a FermionicOp normal orders it, which is incorrect behavior.

```python
import random

random.seed(123)
k = 50
n = 100

op = sum(
    (FermionicOp(''.join(random.choices('+-ENI', k=k)), display_format="dense") * random.random())
    for _ in range(n))
op.reduce()
```

### What should happen?

This operation should complete in at most a few seconds, probably under one second. Instead, it takes a very long time (I have not witnessed it complete, so I don't know how long).

### Any suggestions?

`FermionicOp.to_normal_order` should normal order and reduce, and `FermionicOp.reduce` should only reduce (not normal order).

kevinsung:
### Summary
Fixes #551.

### Details and comments
Deprecates `reduce` and `to_normal_order` in favor of the new methods `simplify` and `normal_ordered`.

This PR changes the type of `FermionicOp._data` from
```
list[tuple[list[tuple[str, int]], complex]]
```
to
```
list[tuple[tuple[tuple[str, int], ...], complex]]
```

The reason to use a tuple instead of a list is that tuples are hashable. This is used in the implementation of `simplify`. The old method of using `_to_dense_label_data` does not work because `_to_dense_label_data` reorders terms in order to produce dense labels. Furthermore, the new method has better asymptotic complexity (`O(n)` instead of `O(n log n)`).

@ikkoham


qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -220,7 +222,7 @@ def __init__(
             tuple[str, complex],
             list[tuple[str, complex]],
             list[tuple[str, float]],
-            list[tuple[list[tuple[str, int]], complex]],
+            list[tuple[tuple[tuple[str, int], ...], complex]],
```
ikkoham, line 225: This is API breaking change of the type hint. I have learned Python's type very recently.
If you use `collections.abc.Sequence` or `collections.abc.Iterable`, you can not break because list and tuple are subclass of them.
Sequence probably does not require any code changes, but Iterable is broader, so Iterable is better if it's consumed somewhere.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -308,26 +310,23 @@ def __init__(
         else:
             self._register_length = register_length
 
-    def _substituted_label(self, label):
-        new_label = []
+    def _substituted_label(self, label) -> Iterator[tuple[str, int]]:
```
ikkoham, line 313: Using generator is very nice!

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -16,14 +16,16 @@
 
 import re
 import warnings
+from collections import defaultdict
 from itertools import product
-from typing import Optional, Union, cast
+from typing import Iterator, Optional, Union, cast
```
ikkoham, line 21: `collections.abc.Iterator` would be better

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -559,6 +566,24 @@ def reduce(self, atol: Optional[float] = None, rtol: Optional[float] = None) ->
...
+        if rtol is None:
+            rtol = self.rtol
+
+        data = defaultdict(float)  # type: dict[tuple[tuple[str, int], ...], complex]
+        for label, coeff in self._data:
+            data[label] += coeff
+        terms = [
+            (label, coeff)
+            for label, coeff in data.items()
+            if not np.isclose(coeff, 0.0, rtol=rtol, atol=atol)
```
ikkoham, line 583: I recently noticed that `rtol` is not necessary when comparing to 0. (This should also be fixed in Terra.)

ikkoham (changes requested):
Thank you. It's a very nice change and high quality. I have a few minor comments.

qiskit_nature/operators/second_quantization/fermionic_op.py
kevinsung: Done.

qiskit_nature/operators/second_quantization/fermionic_op.py
kevinsung: Done.

qiskit_nature/operators/second_quantization/fermionic_op.py
kevinsung: Done.

t-imamichi:
This is a patch to vectorize `SpinOp.simplify`. https://github.com/kevinsung/qiskit-nature/pull/1

kevinsung:
Thanks @t-imamichi . I was planning to have that as a separate PR, but I think it's fine to include here too. But in that case we should also do VibrationalOp at the same time. The same thing can be done there, right?

t-imamichi:
You are welcome. I missed `VibrationalOp.simplify`. I will update my patch soon.

t-imamichi:
I updated `VibrationalOp.simplify`. https://github.com/kevinsung/qiskit-nature/pull/1
Btw, It's good to have unit tests of `simplify` in test_vibrational_op.py

ikkoham (approved):
LGTM! Thank you.

t-imamichi:
Alright. No problem. I wait for the merge of this PR.

test/properties/second_quantization/electronic/resources/angular_momentum_op.json
```diff
@@ -1 +1,386 @@
-[["+_0 -_1 -_4 +_5", 1], ["+_0 -_2 -_4 +_6", 1], ["+_0 -_3 -_4 +_7", 1], ["-_0 +_1 +_4 -_5", 1], ["-_0 +_2 +_4 -_6", 1], ["-_0 +_3 +_4 -_7", 1], ["+_1 -_2 -_5 +_6", 1], ["+_1 -_3 -_5 +_7", 1], ["-_1 +_2 +_5 -_6", 1], ["-_1 +_3 +_5 -_7", 1], ["+_2 -_3 -_6 +_7", 1], ["-_2 +_3 +_6 -_7", 1], ["+_7 -_7", 0.75], ["+_6 -_6", 0.75], ["+_6 -_6 +_7 -_7", 0.5], ["+_5 -_5", 0.75], ["+_5 -_5 +_7 -_7", 0.5], ["+_5 -_5 +_6 -_6", 0.5], ["+_4 -_4", 0.75], ["+_4 -_4 +_7 -_7", 0.5], ["+_4 -_4 +_6 -_6", 0.5], ["+_4 -_4 +_5 -_5", 0.5], ["+_3 -_3", 0.75], ["+_3 -_3 +_7 -_7", -1.5], ["+_3 -_3 +_6 -_6", -0.5], ["+_3 -_3 +_5 -_5", -0.5], ["+_3 -_3 +_4 -_4", -0.5], ["+_2 -_2", 0.75], ["+_2 -_2 +_7 -_7", -0.5], ["+_2 -_2 +_6 -_6", -1.5], ["+_2 -_2 +_5 -_5", -0.5], ["+_2 -_2 +_4 -_4", -0.5], ["+_2 -_2 +_3 -_3", 0.5], ["+_1 -_1", 0.75], ["+_1 -_1 +_7 -_7", -0.5], ["+_1 -_1 +_6 -_6", -0.5], ["+_1 -_1 +_5 -_5", -1.5], ["+_1 -_1 +_4 -_4", -0.5], ["+_1 -_1 +_3 -_3", 0.5], ["+_1 -_1 +_2 -_2" , 0.5], ["+_0 -_0", 0.75], ["+_0 -_0 +_7 -_7", -0.5], ["+_0 -_0 +_6 -_6", -0.5], ["+_0 -_0 +_5 -_5", -0.5], ["+_0 -_0 +_4 -_4", -1.5], ["+_0 -_0 +_3 -_3", 0.5], ["+_0 -_0 +_2 -_2", 0.5], ["+_0 -_0 +_1 -_1", 0.5]]
```
mrossinek, line 1: This change is unrelated. Please revert it

mrossinek (changes requested):
Some minor comments but in general this looks good.

You also need to add a reno which should explain also the difference between `reduce()` and `simplify()`

test/properties/second_quantization/electronic/resources/angular_momentum_op.json
kevinsung: I updated this to make the following test below pass: https://github.com/Qiskit/qiskit-nature/blob/748f96e4db4c86d90151398ef2791761b9bf5da7/test/properties/second_quantization/electronic/test_angular_momentum.py#L39-L53

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -16,14 +16,17 @@
...
 import warnings
+from collections import defaultdict
+from collections.abc import Iterable, Iterator, Sequence
 from itertools import product
 from typing import Optional, Union, cast
 
 import numpy as np
 from scipy.sparse import csc_matrix
-
+from qiskit.utils.deprecation import deprecate_function
```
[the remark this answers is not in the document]
kevinsung: Ok, but I was simply emulating what's already done in `second_quantized_op.py`.

test/properties/second_quantization/electronic/resources/angular_momentum_op.json
mrossinek: What exactly did you change then?

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -531,20 +537,28 @@ def to_matrix(self, sparse: Optional[bool] = True) -> Union[csc_matrix, np.ndarr
...
             ],
             register_length=self.register_length,
             display_format=self.display_format,
         )
 
+    @deprecate_function(
+        "The `reduce` method is deprecated as of version 0.4.0 and will be removed no "
+        "earlier than 3 months after the release date. Instead, use `simplify` "
+        "and optionally `normal_ordered`."
+    )
```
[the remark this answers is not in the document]
kevinsung: Done.

test/operators/second_quantization/test_spin_op.py
```diff
@@ -265,8 +265,13 @@ def test_adjoint(self):
 
     def test_reduce(self):
         """Test reduce"""
+        with self.assertWarns(DeprecationWarning):
+            _ = (self.heisenberg - self.heisenberg).reduce()
```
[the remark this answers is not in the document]
kevinsung: Done.

kevinsung:
@mrossinek I've added a release note and addressed or replied to all your comments.

test/properties/second_quantization/electronic/resources/angular_momentum_op.json
mrossinek: It was formatted as a single line before. While this may not be perfect, the reformatting you did here makes it impossible to identify the changes to this file. Please make it a single line again so that the change can be identified.

We can reformat the json files to non-single lines in one go in a separate PR as this will likely be easier to handle in the future.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -559,6 +576,29 @@ def reduce(self, atol: Optional[float] = None, rtol: Optional[float] = None) ->
...
+        Args:
+            atol: Absolute tolerance for checking if coefficients are zero (Default: 1e-8).
+
+        Returns:
+            The simplified operator.
+        """
+        if atol is None:
+            atol = self.atol
+
+        data = defaultdict(float)  # type: dict[tuple[tuple[str, int], ...], complex]
```
mrossinek, line 594:
```suggestion
        data: dict[tuple[tuple[str, int], ...], complex] = defaultdict(float)
```
Does this work or is mypy or black unhappy with this?

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -645,6 +689,55 @@ def to_normal_order(self) -> FermionicOp:
             return ret
         return FermionicOp(("", 0), self.register_length, "sparse")
 
+    def normal_ordered(self) -> FermionicOp:
+        """Convert to the equivalent operator with normal order.
+
+        Returns a new operator (the original operator is not modified).
+        The returned operator is in sparse label mode.
```
mrossinek, line 696: Is this really true? You are storing `self.display_format`, right?

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -645,6 +689,55 @@ def to_normal_order(self) -> FermionicOp:
...
+            operators. For example, for the case :math:`\\colon c_0 c_0^\\dagger\\colon`
+            where :math:`c_0` is an annihilation operator,
+            this method returns :math:`1 - c_0^\\dagger c_0` due to commutation relations.
+            See the reference: https://en.wikipedia.org/wiki/Normal_order#Multiple_fermions.
+
+        Returns:
+            The normal ordered operator.
+        """
+        temp_display_label = self.display_format
+        self.display_format = "dense"
```
mrossinek, line 711: How exactly do you rely on this?

mrossinek (commented):
Just a few more things I would like to clarify.

Note also my comment on the `json` change above

qiskit_nature/operators/second_quantization/fermionic_op.py
kevinsung: It's true. It looks like `self.display_format` is stored because it is changed in the function and changed back later. I actually copied this docstring and the code, verbatim, from `to_normal_order`. Do you want me to change the behavior?

qiskit_nature/operators/second_quantization/fermionic_op.py
kevinsung: Apparently this affects the return value of `self.to_list`, which is used  3 lines later. Again, I simply copied the code verbatim from the original implementation of `to_normal_order`.

qiskit_nature/operators/second_quantization/fermionic_op.py
kevinsung: It works.

test/properties/second_quantization/electronic/resources/angular_momentum_op.json
kevinsung: Done.

qiskit_nature/operators/second_quantization/fermionic_op.py
mrossinek: Ah I didn't notice that this is simple a verbatim copy. You can leave it as is for the time being then (I guess we may revisit this as part of #610).

test/properties/second_quantization/electronic/resources/angular_momentum_op.json
kevinsung: I think it's because of the change to `qiskit_nature/properties/second_quantization/electronic/angular_momentum.py` where `reduce` is replaced with `simplify`. So the operator is no longer normal-ordered.

mrossinek: Oh thanks I had absolutely overlooked that change in this file!

mrossinek (approved):
This LGTM now!

@woodsp-ibm @pbark @ikkoham Please also take a look at this :+1:

pbark (approved):

PR #620, opened 2022-03-31
[WIP]Add a LatticeModelProblem class

### LINKED ISSUE #563: Implement the Lattice Model Problem
### What should we add?

As it is done for electronic stucture and vibrational structure problems we should be able to implement a lattice model problem that support the solution of a lattice model the `second_q_ops()` and the `interpret()` methods for a lattice model Hamiltonian. In the same way an electronic structure problems is initialised with a driver and the transformers, here we should provide a lattice and a lattice model in order to construct the second quantized operators.

a-matsuo:
### Summary
Implemented a `LatticeModelProblem` class.

[] get_default_filter_criterion
[] `__str__` of `LatticeModelResult`
[] `num_occupied_modals_per_mode` of `LatticeModelResult`

### Details and comments
resolve #563

### DIFF
--- a/qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
+++ b/qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
@@ -1,0 +1,96 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021, 2022.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+"""The Lattice Model Problem class."""
+
+from typing import Union
+
+import numpy as np
+
+from qiskit.algorithms import EigensolverResult, MinimumEigensolverResult
+from qiskit_nature import ListOrDictType
+from qiskit_nature.operators.second_quantization import SecondQuantizedOp
+from qiskit_nature.results import EigenstateResult, LatticeModelResult
+from .models.lattice_model import LatticeModel
+from .. import BaseProblem
+
+
+class LatticeModelProblem(BaseProblem):
+    """Lattice Model Problem"""
+
+    def __init__(
+        self,
+        lattice_model=LatticeModel
+    ):
+        super().__init__()
+        self._lattice_model = lattice_model
+        self._main_property_name = "LatticeEnergy"
+
+    def second_q_ops(self) -> ListOrDictType[SecondQuantizedOp]:
+        """Returns the second quantized operators created based on the lattice models.
+
+        Returns:
+            A `list` or `dict` of `SecondQuantizedOp` objects.
+        """
+        return [self._lattice_model.second_q_ops()]
+
+    def interpret(
+        self,
+        raw_result: Union[EigenstateResult, EigensolverResult, MinimumEigensolverResult],
+    ) -> LatticeModelResult:
+        """Interprets an EigenstateResult in the context of this transformation.
+        Args:
+            raw_result: an eigenstate result object.
+        Returns:
+            An lattice model result.
+        """
+        eigenstate_result = None
+        if isinstance(raw_result, EigenstateResult):
+            eigenstate_result = raw_result
+        elif isinstance(raw_result, EigensolverResult):
+            eigenstate_result = EigenstateResult()
+            eigenstate_result.raw_result = raw_result
+            eigenstate_result.eigenenergies = raw_result.eigenvalues
+            eigenstate_result.eigenstates = raw_result.eigenstates
+            eigenstate_result.aux_operator_eigenvalues = raw_result.aux_operator_eigenvalues
+        elif isinstance(raw_result, MinimumEigensolverResult):
+            eigenstate_result = EigenstateResult()
+            eigenstate_result.raw_result = raw_result
+            eigenstate_result.eigenenergies = np.asarray([raw_result.eigenvalue])
+            eigenstate_result.eigenstates = [raw_result.eigenstate]
+            eigenstate_result.aux_operator_eigenvalues = [raw_result.aux_operator_eigenvalues]
+        result = LatticeModelResult()
+        result.combine(eigenstate_result)
+        # self._grouped_property_transformed.interpret(result)
+        result.computed_lattice_energies = eigenstate_result.eigenenergies
+        return result
+
+
+    # def get_default_filter_criterion(
+    #     self,
+    # ) -> Optional[Callable[[Union[List, np.ndarray], float, Optional[List[float]]], bool]]:
+    #     """Returns a default filter criterion method to filter the eigenvalues computed by the
+    #     eigen solver. For more information see also
+    #     aqua.algorithms.eigen_solvers.NumPyEigensolver.filter_criterion.
+    #     In the fermionic case the default filter ensures that the number of particles is being
+    #     preserved.
+    #     """
+
+    #     # pylint: disable=unused-argument
+    #     def filter_criterion(self, eigenstate, eigenvalue, aux_values):
+    #         # the first num_modes aux_value is the evaluated number of particles for the given mode
+    #         for mode in range(self.grouped_property_transformed.num_modes):
+    #             _key = str(mode) if isinstance(aux_values, dict) else mode
+    #             if aux_values is None or not np.isclose(aux_values[_key][0], 1):
+    #                 return False
+    #         return True
+
+    #     return partial(filter_criterion, self)
--- a/qiskit_nature/results/lattice_model_result.py
+++ b/qiskit_nature/results/lattice_model_result.py
@@ -1,0 +1,108 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2020, 2021.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""The vibrational structure result."""
+
+from typing import List, Optional
+
+import numpy as np
+
+from qiskit.algorithms import AlgorithmResult
+from .eigenstate_result import EigenstateResult
+
+
+class LatticeModelResult(EigenstateResult):
+    """The vibrational structure result."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._algorithm_result: Optional[AlgorithmResult] = None
+        self._computed_lattice_energies: Optional[np.ndarray] = None
+        self._num_occupied_modals_per_mode: Optional[List[List[float]]] = None
+
+    @property
+    def algorithm_result(self) -> Optional[AlgorithmResult]:
+        """Returns raw algorithm result"""
+        return self._algorithm_result
+
+    @algorithm_result.setter
+    def algorithm_result(self, value: AlgorithmResult) -> None:
+        """Sets raw algorithm result"""
+        self._algorithm_result = value
+
+    # TODO we need to be able to extract the statevector or the optimal parameters that can
+    # construct the circuit of the GS from here (if the algorithm supports this)
+
+    @property
+    def computed_lattice_energies(self) -> Optional[np.ndarray]:
+        """Returns computed electronic part of ground state energy"""
+        return self._computed_lattice_energies
+
+    @computed_lattice_energies.setter
+    def computed_lattice_energies(self, value: np.ndarray) -> None:
+        """Sets computed electronic part of ground state energy"""
+        self._computed_lattice_energies = value
+
+    # @property
+    # def num_occupied_modals_per_mode(self) -> Optional[List[List[float]]]:
+    #     """Returns the number of occupied modal per mode"""
+    #     return self._num_occupied_modals_per_mode
+
+    # @num_occupied_modals_per_mode.setter
+    # def num_occupied_modals_per_mode(self, value: List[List[float]]) -> None:
+    #     """Sets measured number of modes"""
+    #     self._num_occupied_modals_per_mode = value
+
+
+    def __str__(self) -> str:
+        """Printable formatted result"""
+        return 'energies'+str(self.computed_lattice_energies)
+
+
+    # def __str__(self) -> str:
+    #     """Printable formatted result"""
+    #     return "\n".join(self.formatted())
+
+    # def formatted(self) -> List[str]:
+    #     """Formatted result as a list of strings"""
+    #     lines = []
+    #     lines.append("=== GROUND STATE ===")
+    #     lines.append(" ")
+    #     lines.append(
+    #         "* Vibrational ground state energy "
+    #         f"(cm^-1): {np.round(self.computed_lattice_energies[0], 12)}"
+    #     )
+    #     if len(self.num_occupied_modals_per_mode) > 0:
+    #         lines.append("The number of occupied modals for each mode is: ")
+    #         for i, m in enumerate(self.num_occupied_modals_per_mode[0]):
+    #             lines.append(f"- Mode {i}: {np.round(m, 12)}")
+
+    #     if (
+    #         self.computed_lattice_energies is not None
+    #         and len(self.computed_lattice_energies) > 1
+    #     ):
+    #         lines.append(" ")
+    #         lines.append("=== EXCITED STATES ===")
+    #         lines.append(" ")
+
+    #         for idx, vib_energy in enumerate(self.computed_lattice_energies[1:]):
+    #             lines.append(
+    #                 f"* {(idx + 1): 3d}: Vibrational excited state energy "
+    #                 f"(cm^-1): {np.round(vib_energy, 12)}"
+    #             )
+    #             if idx < len(self.num_occupied_modals_per_mode):
+    #                 lines.append("The number of occupied modals for each mode is")
+    #                 for i, m in enumerate(self.num_occupied_modals_per_mode[idx]):
+    #                     lines.append(f"- Mode {i}: {np.round(m, 12)}")
+    #             lines.append(" ")
+
+    #     return lines
--- a/qiskit_nature/problems/second_quantization/__init__.py
+++ b/qiskit_nature/problems/second_quantization/__init__.py
@@ -24,6 +24,7 @@ Second-Quantization Problems (:mod:`qiskit_nature.problems.second_quantization`)
    BaseProblem
    ElectronicStructureProblem
    VibrationalStructureProblem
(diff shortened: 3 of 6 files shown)


a-matsuo:
It is still messy, but at least it works with `NumPyMinimumEigensolver` like the following code.
```
lmp = LatticeModelProblem(lattice_model)
np_solver = NumPyMinimumEigensolverFactory()
calc = GroundStateEigensolver(qubit_converter, np_solver)
res = calc.solve(lmp)
```

I have questions for the following method. I couldn't find the corresponding codes for them. Can I skip them? I think we need `__str__` of `LatticeModelResult` at least.
- [ ]  get_default_filter_criterion
- [ ]  `__str__` of `LatticeModelResult`
- [ ] `num_occupied_modals_per_mode` of `LatticeModelResult`
@pbark @ikkoham

ikkoham:
- I think `get_default_filter_criterion` can return `None` in `LatticeModelProblem`.
- `__str__` returns only energy, but it ie better to have header like `=== GROUND STATE ENERGY ===`. In the future, we can extend to have magnetization, or some observables...
- we don't need `num_occupied_modals_per_mode` and ElectronicStructureResult  and EigenstateResult don't have it.

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
```diff
@@ -0,0 +1,110 @@
...
+    def second_q_ops(self) -> ListOrDictType[SecondQuantizedOp]:
+        """Returns the second quantized operators created based on the lattice models.
+
+        Returns:
+            A ``list`` or ``dict`` of
+            :class:`~qiskit_nature.operators.second_quantization.SecondQuantizedOp`
+        """
+        return [self._lattice_model.second_q_ops()]
+
+    def interpret(
```
woodsp-ibm, line 48: Can we do some unit test(s) that cover the interpret and the result. As far as I can see these aspects are not covered at all.

a-matsuo:
@woodsp-ibm Thank you for reviewing. I added unittests for `interpret`. I will add some brief explanation how to use LatticeModelProblem in the tutorial.

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
```diff
@@ -0,0 +1,110 @@
...
+                :`List[int]`: a list of positive integers.
+                :`Callable`: a function which is used to generate the excitations.
+                    For more details on how to write such a function refer to one of the default
+                    methods, :meth:`generate_fermionic_excitations` or
+                    :meth:`generate_vibrational_excitations`.
+
+        Returns:
+            None
+        """
+        return None
```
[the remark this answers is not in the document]
a-matsuo: Actually, I have no idea about it (I don't even know what qeom is) . But, I changed it to return `NotImplementedError` for the future.

a-matsuo:
@mrossinek @woodsp-ibm Thank you for reviewing. I fixed the code based on the comments. I also added some explanation for the Lattice Model Problem in the tutorial.

docs/tutorials/10_lattice_models.ipynb
```diff
@@ -813,6 +813,14 @@
       "+ (1+0j) * ( -_0 +_30 )\n",
       "+ (- ...\n"
      ]
+    },
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "<ipython-input-21-3bf38577149b>:1: DeprecationWarning: The reduce function is deprecated as of version 0.4.0 and will be removed no sooner than 3 months after the release Instead, use `simplify` and optionally `normal_ordered`..\n",
+      "  ham = fhm.second_q_ops(display_format=\"sparse\").reduce()\n"
```
mrossinek, line 822: Could you change this to use `simplify()` instead?

docs/tutorials/10_lattice_models.ipynb
```diff
@@ -894,50 +902,99 @@
     "print(ham)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## LatticeModelProblem\n",
+    "We also have the Lattice Model Problem to generate the Second quantized operator from a Lattice Model. We can use the Ground State Eigensolver with it to calculate the ground state energy."
```
mrossinek, line 910:
```suggestion
    "Qiskit Nature also has a `LatticeModelProblem` class which allows the usage of the `GroundStateEigensolver` to calculate the ground state energy of a given lattice. You can use this class as follows:"
```

docs/tutorials/10_lattice_models.ipynb
```diff
@@ -894,50 +902,99 @@
...
-       "<IPython.core.display.HTML object>"
-      ]
-     },
-     "metadata": {},
-     "output_type": "display_data"
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/atsushi/github/nature/qiskit_nature/problems/second_quantization/lattice/models/fermi_hubbard_model.py:123: UserWarning: The default value for `display_format` will be changed from 'dense' to 'sparse' in version 0.3.0. Once that happens, you must specify display_format='dense' directly.\n",
+      "  return FermionicOp(ham, register_length=register_length, display_format=display_format)\n"
```
mrossinek, line 955: Could you go into the `FermiHubbardModel` and explicitly set `display_format="sparse"` in order to remove this warning?

docs/tutorials/10_lattice_models.ipynb
```diff
@@ -894,50 +902,99 @@
...
-      ]
-     },
-     "metadata": {},
-     "output_type": "display_data"
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "=== GROUND STATE ===\n",
+      " \n",
+      "* Lattice ground state energy : -2.566350190841\n"
```
mrossinek, line 947: I am not sure, if this is because of the Github plain-text view of the tutorial, but I don't think this output cell should be here, but rather should be below the next cell

docs/tutorials/10_lattice_models.ipynb
```diff
@@ -894,50 +902,99 @@
...
+     "text": [
+      "/Users/atsushi/github/nature/qiskit_nature/problems/second_quantization/lattice/models/fermi_hubbard_model.py:123: UserWarning: The default value for `display_format` will be changed from 'dense' to 'sparse' in version 0.3.0. Once that happens, you must specify display_format='dense' directly.\n",
+      "  return FermionicOp(ham, register_length=register_length, display_format=display_format)\n"
+     ]
     }
    ],
+   "source": [
+    "from qiskit.algorithms import NumPyMinimumEigensolver\n",
+    "from qiskit_nature.algorithms import GroundStateEigensolver\n",
+    "from qiskit_nature.mappers.second_quantization import JordanWignerMapper\n",
```
mrossinek, line 962: If I recall correctly, the type of lattice dictates what mapper to use, correctly? Is this being discussed anywhere in this tutorial? If not, this could be an addition for the future (but there is no direct need to add this now as part of this PR)

releasenotes/notes/lattice_model_problem-3925bfbd2267cd4c.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - |
+    Adds a new problem class,
+    :class:`qiskit_nature.problems.second_quantization.lattice.LatticeModelProblem`,
+    for lattice models. It can create second quantized operators from a lattice model.
+    We can use the Ground State Eigensolver with it to calculate the ground state energy.
+
+    .. code-block:: python
+
+        calc = GroundStateEigensolver(converter, solver)
+        res = calc.solve(LatticeModelProblem(lattice_model))
```
mrossinek, line 12: I think, if anything, this should be a fully functional snippet with all imports and variable definitions properly working. You can copy the cell from the tutorial if you want.

docs/tutorials/10_lattice_models.ipynb
a-matsuo: hmm, I have not mentioned about it in this tutorial (I didn't even know that). I will leave it for the future.

ikkoham (commented):
Totally good direction. I have a very minor comment.

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
```diff
@@ -0,0 +1,116 @@
...
+from qiskit_nature.operators.second_quantization import SecondQuantizedOp
+from qiskit_nature.results import EigenstateResult, LatticeModelResult
+from .models.lattice_model import LatticeModel
+from ..base_problem import BaseProblem
+
+
+class LatticeModelProblem(BaseProblem):
+    """Lattice Model Problem class to create second quantized operators from a lattice model."""
+
+    def __init__(self, lattice_model=LatticeModel) -> None:
```
mrossinek, line 30: I don't think I fully understand this init argument. Did you want to specify the type here but
accidentally assigned the default value as the class itself?

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
```diff
@@ -0,0 +1,116 @@
...
+    def second_q_ops(self) -> ListOrDictType[SecondQuantizedOp]:
+        """Returns the second quantized operators created based on the lattice models.
+
+        Returns:
+            A ``list`` or ``dict`` of
+            :class:`~qiskit_nature.operators.second_quantization.SecondQuantizedOp`
+        """
+        second_q_ops: ListOrDictType[SecondQuantizedOp] = self._lattice_model.second_q_ops()
+        if settings.dict_aux_operators:
+            second_q_ops = {"LatticeEnergy": second_q_ops}
```
mrossinek, line 48: Ideally, the internally used model should already return `second_q_ops` in accordance with
`settings.dict_aux_operators`. But this can also be fixed later as it requires deprecation.
Nonetheless, I suggest that you use `self._main_property_name` here in order to avoid the string
duplication.

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
```diff
@@ -0,0 +1,116 @@
...
+        self,
+        qubit_converter: QubitConverter,
+        excitations: Union[
+            str,
+            int,
+            List[int],
+            Callable[[int, Tuple[int, int]], List[Tuple[Tuple[int, ...], Tuple[int, ...]]]],
+        ] = "sd",
+    ) -> None:
+        """Generates the hopping operators and their commutativity information
```
mrossinek, line 96: More a question than a comment: do we need to duplicate this docstring or can Sphinx infer it from
the base class?

mrossinek (commented):
Besides these comments, this LGTM 👍

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
a-matsuo: You're right. I didn't notice it until you mentioned it.

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
a-matsuo: I've tested with my local environment. It looks Sphinx can infer this docstring from the base class (even I deleted this docstring in `LatticeModelProblem`, Sphinx could make a docstring from the base class for `LatticeModeProblem`'s `hopping_qeom_ops`).

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
```diff
@@ -0,0 +1,116 @@
...
+    """Lattice Model Problem class to create second quantized operators from a lattice model."""
+
+    def __init__(self, lattice_model=LatticeModel) -> None:
+        """
+        Args:
+            lattice_model: A lattice model class to create second quantized operators.
+        """
+        super().__init__()
+        self._lattice_model = lattice_model
+        self._main_property_name = "LatticeEnergy"
```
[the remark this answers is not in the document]
a-matsuo: I talked this with @ikkoham. We can pass a `main property_name` to a constructor of `BaseProblem` to set it. e.g. `super().__init__("LatticeEnergy")`. The comment is legit since `BaseProblem` has `_main_property_name` variable with an empty string. 

What do you think? @mrossinek 
I can change all problem subclasses to set their name like `super().__init__("LatticeEnergy")` if you want. Then, I will open a new PR for that change.

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
woodsp-ibm: It does inherit docstrings, we rely on this in many places so as to not have to repeat the same thing, rather just have it once. If there is some need, for some specific subclass behavior that needs explaining, then you can supply a docstring which will override any that might be inherited. But best avoided if possible.

qiskit_nature/results/lattice_model_result.py
```diff
@@ -0,0 +1,64 @@
...
+        return "\n".join(self._formatted())
+
+    def _formatted(self) -> List[str]:
+        """Formatted result as a list of strings"""
+        lines = []
+        lines.append("=== GROUND STATE ===")
+        lines.append(" ")
+        lines.append(
+            "* Lattice ground state energy " f": {np.round(self.computed_lattice_energies[0], 12)}"
+        )
```
woodsp-ibm, line 63: Given this is now in a Problem , it can presumably be used with an excited state solver, such as using NumPyEigensolver with the ExcitedStatesEigensolver (and maybe later VQD) from Terra. Shouldn't this also format up any excited state energies that may be computed as presumably this is a valid use too right?

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
a-matsuo: Cool. I added a docstring to explicitly describe a `NotimplementedError`. Also, I added an error message for the `NotimplementedError`.

qiskit_nature/results/lattice_model_result.py
a-matsuo: I copied messages for excited state energies from other problem classes.

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
```diff
@@ -115,4 +115,6 @@ def hopping_qeom_ops(
             Currently, this function is not implemented in the `LtticeProblemModel` class and
             always raises `NotImplementedError`.
         """
-        raise NotImplementedError()
+        raise NotImplementedError(
+            "Currently, it's not implemented in the `LtticeProblemModel` class."
```
woodsp-ibm, line 119:
```suggestion
            "Currently, it's not implemented in the `LatticeProblemModel` class."
```
This copied the spelling error from above in the Returns docstring. I wonder whether we could say something like this problem cannot be used with QEOM solver. @mrossinek Any thoughts, otherwise the text says little beyond whats implied by the error itself.

mrossinek: Looking at the code here as well as the qEOM code, it looks to me like this should generally be possible to implement. A quick test could be by testing the behavior of the `ExcitedStatesSolver` + `NumPyEigensolver` (not minimum) to see if that works as expected. If yes, I would handle that accordingly in the result printing and would open a separate issue to track the implementation of the `hopping_qeom_ops` method.

a-matsuo: I changed `NotImplementedError` to return None. Then, `ExcitedStatesEigensolver` + `NumPyEigensolver` returns some values (I'm not sure whether this behavior is as expected or not).
image

woodsp-ibm: > I changed NotImplementedError to return None. Then, ExcitedStatesEigensolver + NumPyEigensolver returns some values (I'm not sure whether this behavior is as expected or not).

I am not sure  I understand. Are you saying the excited states solver `ExcitedStatesEigensolver + NumPyEigensolver` fails if the `hopping_qeom_ops` raises an error - (and only if it did not, and returned None, did you get those values)? As far as I know its one called by QEOM solver and should never get invoked otherwise.

a-matsuo: Sorry for the confusion. I thought it would fail and I hadn't check it with `hopping_qeom_ops` + `NotimplementedError`. However, I checked it works, even `hopping_qeom_ops` raises `NotimplementedError`.

a-matsuo:
I changed `get_default_filter_criterion` to explicitly return None in LatticeModelProblem. Can I just leave `get_default_filter_criterion` to return None?

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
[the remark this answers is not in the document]
mrossinek: Sorry, I overlooked this comment before. Sounds like a reasonable change to me :+1: Could you do this in a separate PR?

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
```diff
@@ -0,0 +1,125 @@
...
+                :`int`: a single, positive integer denoting the excitation type (1 == `s`, etc.).
+                :`List[int]`: a list of positive integers.
+                :`Callable`: a function which is used to generate the excitations.
+                    For more details on how to write such a function refer to one of the default
+                    methods, :meth:`generate_fermionic_excitations` or
+                    :meth:`generate_vibrational_excitations`.
+
+        Returns:
+            Currently, this function is not implemented in the `LtticeProblemModel` class and
+            always raises `NotImplementedError`.
```
mrossinek, line 116: Instead of `returns` this should say `Raises`. There is also still a typo here: `Lttice -> Lattice`.

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
```diff
@@ -0,0 +1,125 @@
...
+    def hopping_qeom_ops(
+        self,
+        qubit_converter: QubitConverter,
+        excitations: Union[
+            str,
+            int,
+            List[int],
+            Callable[[int, Tuple[int, int]], List[Tuple[Tuple[int, ...], Tuple[int, ...]]]],
+        ] = "sd",
+    ) -> None:
```
mrossinek, line 95: Despite this not being implemented, I would keep the return type hint unchanged for consistency (and also to document future expectation of what this method should return)

qiskit_nature/problems/second_quantization/lattice/lattice_model_problem.py
```diff
@@ -0,0 +1,125 @@
...
+            Currently, this function is not implemented in the `LtticeProblemModel` class and
+            always raises `NotImplementedError`.
+        """
+        raise NotImplementedError(
+            "Currently, it's not implemented in the `LatticeProblemModel` class."
+        )
+
+    def get_default_filter_criterion(
+        self,
+    ) -> None:
```
mrossinek, line 124: Same here: I would try to keep the function definition aligned with the base class

mrossinek (approved):
This LGTM, thanks a lot @a-matsuo!

woodsp-ibm (approved):
LGTM too, thanks @a-matsuo

PR #621, opened 2022-03-31
Add `qiskit_nature.settings.optimize_einsum`

mrossinek:
### Summary

The new setting `qiskit_nature.settings.optimize_einsum` allows enabling
the `optimize` argument in `numpy.einsum` calls with more than
2 operands. This is known to yield significant computational efficiency
increases at the expense of higher memory consumption.
The setting defaults to `True`.

### Details and comments

For more information refer to:
https://numpy.org/doc/stable/reference/generated/numpy.einsum.html


woodsp-ibm (approved):
Do you have any idea of performance difference when True versus False - at least in terms of speed?

mrossinek:
> Do you have any idea of performance difference when True versus False - at least in terms of speed?

Here is a very simple test using a water molecule in the `6-31g*` basis and the PySCF driver:
```python
import time

from qiskit_nature.settings import settings

settings.optimize_einsum = False

start = time.time()
driver = PySCFDriver(atom="O 0 0 0.115; H 0.0 0.754 -0.459; H 0.0 -0.754 -0.459", basis="631g*")
props = driver.run()
end = time.time()
print("optimize_einsum = False", end - start)

del driver

settings.optimize_einsum = True

start = time.time()
driver = PySCFDriver(atom="O 0 0 0.115; H 0.0 0.754 -0.459; H 0.0 -0.754 -0.459", basis="631g*")
props = driver.run()
end = time.time()
print("optimize_einsum = True", end - start)
```
This results in:
```
optimize_einsum = False 96.63125371932983
optimize_einsum = True 0.8717706203460693
```

So as you can see, the speedup is __very__ significant :sweat_smile:

PR #622, opened 2022-04-01
Vectorize `SpinOp.simplify` and `VibrationalOp.simplify`

t-imamichi:
### Summary

Vectorize `SpinOp.simplify` and `VibrationalOp.simplify`

Follow-up of #607
Replacement of #586

### DIFF
--- a/qiskit_nature/operators/second_quantization/spin_op.py
+++ b/qiskit_nature/operators/second_quantization/spin_op.py
@@ -407,11 +407,10 @@ class SpinOp(SecondQuantizedOp):
             return_inverse=True,
             axis=0,
         )
-        coeff_list = np.zeros(len(self._coeffs), dtype=np.complex128)
-        for i, val in zip(indices, self._coeffs):
-            coeff_list[i] += val
-        non_zero = [i for i, v in enumerate(coeff_list) if not np.isclose(v, 0, atol=atol)]
-        if not non_zero:
+        coeff_list = np.zeros(flatten_array.shape[0], dtype=np.complex128)
+        np.add.at(coeff_list, indices, self._coeffs)
+        is_zero = np.isclose(coeff_list, 0, atol=atol)
+        if np.all(is_zero):
             return SpinOp(
                 (
                     np.zeros((3, 1, self.register_length), dtype=np.int8),
@@ -419,9 +418,10 @@ class SpinOp(SecondQuantizedOp):
                 ),
                 spin=self.spin,
             )
+        non_zero = np.logical_not(is_zero)
         new_array = (
             flatten_array[non_zero]
-            .reshape((len(non_zero), 3, self.register_length))
+            .reshape((np.count_nonzero(non_zero), 3, self.register_length))
             .transpose(1, 0, 2)
         )
         new_coeff = coeff_list[non_zero]
--- a/qiskit_nature/operators/second_quantization/vibrational_op.py
+++ b/qiskit_nature/operators/second_quantization/vibrational_op.py
@@ -291,14 +291,14 @@ class VibrationalOp(SecondQuantizedOp):
             atol = self.atol
 
         label_list, indices = np.unique(self._labels, return_inverse=True, axis=0)
-        coeff_list = np.zeros(len(self._coeffs), dtype=np.complex128)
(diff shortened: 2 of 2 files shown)


qiskit_nature/operators/second_quantization/spin_op.py
```diff
@@ -407,21 +407,21 @@ def simplify(self, atol: Optional[float] = None) -> "SpinOp":
...
             axis=0,
         )
-        coeff_list = np.zeros(len(self._coeffs), dtype=np.complex128)
-        for i, val in zip(indices, self._coeffs):
-            coeff_list[i] += val
-        non_zero = [i for i, v in enumerate(coeff_list) if not np.isclose(v, 0, atol=atol)]
-        if not non_zero:
+        coeff_list = np.zeros(flatten_array.shape[0], dtype=np.complex128)
+        np.add.at(coeff_list, indices, self._coeffs)
+        is_zero = np.isclose(coeff_list, 0, atol=atol)
```
mrossinek, line 412: What about using `np.nonzero` instead? https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html

I would believe that (since this method exists) there is some further optimization done there compared to this approach.
This would obviously imply some minor changes below, all of which could be propagated to the other file, too.

woodsp-ibm:
This is all about performance right. You had some figures in the prior PR, do you have any timings for the before/after in this case.

qiskit_nature/operators/second_quantization/spin_op.py
t-imamichi: Good catch. I forgot about `np.nonzero`. I checked the doc of `no.nonzero` and noticed that it does not take `atol`. So, as far as `simplify` takes `atol`, I think we need to use `isclose`. What do you think of `atol`?

t-imamichi:
I made a microbenchmark.

```python
import random
from timeit import timeit

from qiskit_nature.operators.second_quantization import VibrationalOp, SpinOp


def vb_random_op(k, n):
    random.seed(123)
    op = sum(
        (VibrationalOp("".join(random.choices("+-ENI", k=k)), k, 1) * random.random())
        for _ in range(n))
    return op


def spin_random_op(k, n):
    random.seed(123)
    op = sum(
        (SpinOp("".join(random.choices("IXYZ", k=k))) * random.random())
        for _ in range(n))
    return op


def bench_simplify(k, n):
    op = vb_random_op(k, n)
    print(f"vibrational op, {k} {n}: {timeit(lambda: op.simplify(), number=100)} sec")

    op = spin_random_op(k, n)
    print(f"spin op, {k} {n}: {timeit(lambda: op.simplify(), number=100)} sec")


bench_simplify(10, 1000)
```

main
```
vibrational op, 10 1000: 3.0248702310000226 sec
spin op, 10 1000: 3.114273889999822 sec
```

this PR
```
vibrational op, 10 1000: 0.16983310300020094 sec
spin op, 10 1000: 0.19599379700048303 sec
```

kevinsung:
I wonder if we should change these to a dict-based implementation, as is done in `FermionicOp`. That would have better asymptotic complexity (O(n) instead of O(n log n)). It might be slower for small operators, but if that's an issue it seems we can port it to Rust as was done at https://github.com/Qiskit/qiskit-terra/pull/7656. In any case, that should probably be done separately after merging this PR.

qiskit_nature/operators/second_quantization/spin_op.py
kevinsung: `atol` is important so we can't use `np.nonzero` here.

t-imamichi:
If `simplify` is still a bottleneck of some applications even with this PR, it would be good to think of the rust version. I hope we add more benchmarking codes to https://github.com/Qiskit/qiskit-app-benchmarks. Currently, there is only protein folding in nature.

qiskit_nature/operators/second_quantization/spin_op.py
mrossinek: I had not noticed the lack of `atol`. In that case your approach is probably best :+1:

mrossinek (approved):
Thanks @t-imamichi!

PR #623, opened 2022-04-01
[WIP] Fix the incorrect results given by QEOM with noisy simulators

### LINKED ISSUE #52: QEOM energies differ for Statevector and Qasm simulator

### LINKED ISSUE #415: QEOM evaluation of operators
In unit test module [algorithms/excited_state_solvers/test_bosonic_esc_calculation.py](https://github.com/Qiskit/qiskit-nature/blob/main/test/algorithms/excited_state_solvers/test_bosonic_esc_calculation.py) the test [test_vqe_uvccsd_factory()](https://github.com/Qiskit/qiskit-nature/blob/f73b8c182451939f3f386389291156a813cdfa53/test/algorithms/excited_state_solvers/test_bosonic_esc_calculation.py#L112-L123) uses the `aer_simulator_statevector` (a statevector simulator).

Now switching to `aer_simulator` and adding `include_custom=True` to the `VQEUVCCFactory` constructor, I think should also be an ideal result with the same outcome as the statevector. Internally `VQE` will use `AerPauliExpectation` which will use the custom Aer instruction in the expectation computation and produces an ideal result. The aforementioned test however **_fails_** under this configuration as the result ends up outside the limits allowed by the test.

In looking at QEOM, which in turn uses an `evaluate_operators` in the GroundStateSolver, I have some observations. First VQE, when using include_custom=True, returns a dictionary computed from counts
```
    'eigenstate': {'0101': 0.9960860906568267, '1001': 0.08838834764831845},
```
versus what I might have expected it to look like as comes from the statevector result
```
    'eigenstate': array([-4.90652305e-17+1.59462423e-17j,  2.15105711e-16-4.65078744e-16j,
        1.23327997e-18+2.94357950e-17j, -1.11022302e-16-4.68375339e-17j,
        1.23881957e-16-2.25705481e-16j,  9.95269314e-01+1.40225427e-16j,
        7.31772069e-05+1.06710345e-16j,  3.29597460e-17+4.93038066e-32j,
       -7.46392098e-17+5.76507565e-17j,  9.71544362e-02-5.94899346e-18j,
        6.12208684e-05-6.86833149e-17j,  1.16971296e-16-7.11236625e-17j,
        9.19962484e-17+2.94377830e-17j, -3.33066907e-16+3.64099743e-17j,
        4.90533720e-17+6.01023954e-17j, -1.21430643e-17+1.66533454e-16j]),
```
as an ideal result. Since the former is computed from counts it will not be ideal. Maybe this is fine, its not what I expected, I would have thought VQE would have given an ideal state back, like it does for statevector, buts it's logic to compute the eigenstate does not do that. The state is used in Nature in its evaluation of the matrix operators.

Either way it seems that perhaps the computation could be improved. But rather than try to fix this now - since its working albeit it could be improved, perhaps this is something that needs more change/improvements here in Nature as well as in Terra.

Maybe we should have some call on the minimum eigensolvers that given a result from an instance, if you pass that into some method say compute_expectation(operator, MinimumEigenSolverResult) it will compute an expectation (if it can do so - i.e. supports aux operator) where it can take whatever it needs from the result to compute the expectation at that same result state. Maybe we allow it to be a union for VQE to allow any parameters to be passed - then in the future it will just use the planned expectation object directly at that parameterization. This would allow the logic to go closer to the algorithms in Terra and away from the ground state solvers in Nature.

So I figured I'd raise this issue for potential discussion/consideration going forwards on improving Nature and the algorithms in Terra based on needs like this.

Anthony-Gandon:
Fixes #52
Fixes #415

### Summary

Previously, QEOM evaluated the many operators needed to create the matrices Q,M on the
groundstate resulting from a call to VQE. This worked for StateVector simulators because
this object was exactly the statefunction (type VectorStateFn). However, when using QASM or
aer_simulator this eigenstate is a dictionary of samples which was converted
into a VectorStateFn without consideration of the sign ambiguity.

This commit changes the evaluate_operators() method used in QEOM to the eval_observables()
that was added in a recent commit of qiskit-terra.
This method is called on the ansatz function evaluated at the optimal parameters given by
the solver instead of on the groundstate. This delays the call to the eval() method which
was the root of the issue.

A Qiskit-terra pull requests gives the required support for non-hermitian operators in the 
aer_pauli_expectation class.

### Details and comments

This commit modifies the evaluate_operators() method of the ground state_eigensolver to allow
input of type Circuit.
However, a recent commit to Qiskit-terra aims to unify the many such evaluation method. 
The modified QEOM function calls eval_observables() instead of evaluate_operators().
This leads in a failure to pass the first test of test_excited_states_solvers.py due to an 
error an unhandled case in eval_observables() when the solver is NumPyMinimumEigensolver().

The tests: 
- test/algorithms/excited_state_solvers/test_bosonic_esc_calculation_noisy.py
- test/algorithms/excited_state_solvers/test_excited_states_solvers_noisy.py
were added to evaluate the results of QEOM with the qasm_simulator and the 
AerPauliExpectation()

### DIFF
--- a/qiskit_nature/algorithms/excited_states_solvers/qeom.py
+++ b/qiskit_nature/algorithms/excited_states_solvers/qeom.py
@@ -21,14 +21,16 @@ import numpy as np
 from scipy import linalg
 from qiskit.tools import parallel_map
 from qiskit.tools.events import TextProgressBar
-from qiskit.utils import algorithm_globals
+from qiskit.utils import algorithm_globals, QuantumInstance
 from qiskit.algorithms import AlgorithmResult
+from qiskit.opflow.expectations.expectation_base import ExpectationBase
 from qiskit.opflow import (
     Z2Symmetries,
     commutator,
     double_commutator,
     PauliSumOp,
 )
+from qiskit.algorithms import eval_observables
 
 from qiskit_nature import ListOrDictType
 from qiskit_nature.operators.second_quantization import SecondQuantizedOp
@@ -81,6 +83,8 @@ class QEOM(ExcitedStatesSolver):
         self,
         problem: BaseProblem,
         aux_operators: Optional[ListOrDictType[SecondQuantizedOp]] = None,
+        expectation: ExpectationBase = None,
+        quantum_instance: QuantumInstance = None,
     ) -> EigenstateResult:
         """Run the excited-states calculation.
 
@@ -90,7 +94,11 @@ class QEOM(ExcitedStatesSolver):
         Args:
             problem: a class encoding a problem to be solved.
             aux_operators: Additional auxiliary operators to evaluate.
-
+            expectation: Might be different from the one used in VQE. If None, the expectation
+            from the groundstate eigensolver is imported. Please note that the include_custom
+            parameter can be used locally in the groundstate eigensolver to create an
+            AerPauliExpectation() but it cannot be retrieved in qeom.solve().
+            quantum_instance: quantum instance that can be used when the solver don't have one.
         Returns:
             An interpreted :class:`~.EigenstateResult`. For more information see also
             :meth:`~.BaseProblem.interpret`.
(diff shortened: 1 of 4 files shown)


mrossinek:
This PR was superseded by #971

PR #624, opened 2022-04-01
Add LogarithmicMapper

### LINKED ISSUE #6: Logarithmic-Mapper (spin)
Migrated from Enterprise Github.
Creator: @mrossinek

Implement and test the `LogarithmicMapper`.
This issue breaks-out of #18 and can be used to assign and discuss related topics.

- [ ] Implementation
- [ ] Unittest
- [ ] Documentation

a-matsuo:
### Summary
Implemented a LogarithmicMapper class for a new QubitMapper.
- [x] unittests
- [x] documentation

### Details and comments
resolve #6


a-matsuo:
@pbark It looks working, but I'm not sure whether the result is correct. Can we have simple test cases like the following ones in `test_linear_mapper.py`?   I checked the existing unittests for random operators, but it's hard to re-implement those tests on the current Qiskit. 

```
    spin_op1 = SpinOp([("Y_0^2", -0.432 + 1.32j)], 0.5, 1)
    ref_qubit_op1 = (-0.054 + 0.165j) * (I ^ I) + (0.054 - 0.165j) * (Z ^ Z)

    spin_op2 = SpinOp([("X_0 Z_0 I_1", -1.139 + 0.083j)], 0.5, 2)
    ref_qubit_op2 = (0.010375 + 0.142375j) * (I ^ I ^ Y ^ X) + (-0.010375 - 0.142375j) * (
        I ^ I ^ X ^ Y
    )
```

.pylintdict
```diff
@@ -150,6 +150,7 @@ fock
 formatter
 fortran
 fredkin
+gauglielmo
```
mrossinek, line 153:
```suggestion
guglielmo
```
The name in the citation is misspelled.

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+from .spin_mapper import SpinMapper
+
+
+class LogarithmicMapper(SpinMapper):
+    """A mapper for Logarithmic spin-to-qubit mapping [1].
+
+
+    [1]: Mathis, S. V., Gauglielmo, M., & Ivano, T. (2020).
+         Toward scalable simulations of Lattice Gauge Theories on quantum computer.
+         `arXiv:2005.10271 <https://arxiv.org/abs/2005.10271>`_
```
mrossinek, line 35: Besides the misspelled name, I would also suggest to use the more up-to-date reference: https://journals.aps.org/prd/abstract/10.1103/PhysRevD.102.094501

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+
+    def __init__(self, embed_padding: float = 1, embed_location: str = "upper") -> None:
+        """
+        Args:
+            embed_padding:
+                The value of the diagonal elements of the upper left  block of the embedded matrix.
+            embed_location:
+                Must be one of ['upper', 'lower']. This parameters sets whether
```
mrossinek, line 44: If this can only take a limited set of strings I suggest we extract these into an `Enum`

mrossinek, line 44:
```suggestion
                Must be one of ['upper', 'lower']. This parameter sets whether
```

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+                    operatorlist.append(identity)
+
+            # Now, we can tensor all operators in this list
+            qubit_ops_list.append(coeff * reduce(operator.xor, reversed(operatorlist)))
+
+        qubit_op = reduce(operator.add, qubit_ops_list)
+
+        return qubit_op
+
+    def _logarithmic_encoding(self, spin: Union[Fraction, float]) -> List[PauliSumOp]:
```
mrossinek, line 101: This return type is actually much stricter. Above, you are unpacking the result of this function into four variables. Inspecting the code below, it is indeed the case that this list is always of length 4. I suggest you make this evident by returning a tuple and hard-code the length of 4 into the documented return type.

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+        for idx, (_, coeff) in enumerate(second_q_op.to_list()):
+
+            operatorlist: List[PauliSumOp] = []
+
+            for n_x, n_y, n_z in zip(second_q_op.x[idx], second_q_op.y[idx], second_q_op.z[idx]):
+
+                operator_on_spin_i: List[PauliSumOp] = []
+
+                if n_x > 0:
+                    operator_on_spin_i.append(reduce(operator.matmul, [spinx] * int(n_x)))
```
mrossinek, line 78: I do not understand the need for the `reduce()` call. Can you please explain why this is necessary or update the code if it is not.

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+
+                if n_y > 0:
+                    operator_on_spin_i.append(reduce(operator.matmul, [spiny] * int(n_y)))
+
+                if n_z > 0:
+                    operator_on_spin_i.append(reduce(operator.matmul, [spinz] * int(n_z)))
+
+                if np.any([n_x, n_y, n_z]) > 0:
```
mrossinek, line 86: How about the following instead?
```suggestion
                if len(operator_on_spin_i) > 0:
```
The `> 0` is even optional in both cases.

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+                    single_operator_on_spin_i = reduce(operator.matmul, operator_on_spin_i)
+                    operatorlist.append(single_operator_on_spin_i.reduce())
```
mrossinek, line 88:
```suggestion
                    operatorlist.append(single_operator_on_spin_i)
```
Or why is the `.reduce()` strictly necessary? If it is optional in terms of mathematics, how does performance change with/without it?

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+            spin_op_encoding.append(PauliSumOp(1.0 * op))
+
+        return spin_op_encoding
+
+    def _embed_matrix(
+        self,
+        matrix: np.ndarray,
+        num_qubits: int,
+        embed_padding: float = 1,
+        embed_location: str = "upper",
```
mrossinek, line 142: This method does not use `self` and instead get's passed `embed_padding` and `embed_location` as input arguments.
I suggest you either make this a `@staticmethod` or change the latter and remove those duplicate arguments in favor of accessing them via `self`.

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+
+        dim_diff = full_dim - subs_dim
+        if dim_diff == 0:
+            full_matrix = matrix
+
+        elif dim_diff > 0:
+            if embed_location == "lower":
+                full_matrix = np.zeros((full_dim, full_dim), dtype=complex)
+                full_matrix[:dim_diff, :dim_diff] = np.eye(dim_diff) * embed_padding
+                full_matrix[dim_diff:, dim_diff:] = matrix
```
mrossinek, line 184: This seems to me like a lot of copying will take place, however, I am not 100% certain on the internals of numpy.
That said, I believe there may exist an easier implementation using `np.block` (see the first example here: https://numpy.org/doc/stable/reference/generated/numpy.block.html)

test/mappers/second_quantization/test_logarithmic_mapper.py
```diff
@@ -0,0 +1,79 @@
...
+        (spin_op6, ref_qubit_op6, 2),
+        (spin_op6, ref_qubit_op7, 2, "lower"),
+    )
+    @unpack
+    def test_mapping(self, spin_op, ref_qubit_op, embed_padding=1, embed_location="upper"):
+        """Test mapping to qubit operator"""
+        mapper = LogarithmicMapper(embed_padding, embed_location)
+        qubit_op = mapper.map(spin_op)
+        print("qubit_op", qubit_op)
+        print("ref_qubit_op", ref_qubit_op)
```
mrossinek, line 74:
```suggestion
```

mrossinek (dismissed):
Thank you @a-matsuo for the hard work on this! I have a few comments below. Please let me know here or offline if I need to clarify anything :+1:

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,199 @@
...
+        spin_matrices.append(np.eye(dspin))
+
+        # Embed the spin matrices in a larger matrix of size 2**num_qubits x 2**num_qubits
+        embed = lambda matrix: self._embed_matrix(
+            matrix,
+            num_qubits,
+            embed_padding=self._embed_padding,
+            embed_location=self._embed_location,
+        )
+        embedded_spin_matrices = list(map(embed, spin_matrices))
```
ikkoham, line 126: Is it possible to implement this without lambda functions? Since this is a method, it should be able to take parameters inside.

test/mappers/second_quantization/test_logarithmic_mapper.py
```diff
@@ -0,0 +1,79 @@
...
+        (spin_op5, ref_qubit_op5),
+        (spin_op6, ref_qubit_op6, 2),
+        (spin_op6, ref_qubit_op7, 2, "lower"),
+    )
+    @unpack
+    def test_mapping(self, spin_op, ref_qubit_op, embed_padding=1, embed_location="upper"):
+        """Test mapping to qubit operator"""
+        mapper = LogarithmicMapper(embed_padding, embed_location)
+        qubit_op = mapper.map(spin_op)
+        print("qubit_op", qubit_op)
```
ikkoham, line 73: `print` is used only for debug?

ikkoham (commented):
LGTM. Just minor comments.

woodsp-ibm:
Don't forget the release note!

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
woodsp-ibm: Could we add something in the class docstring that briefly describes the embedding in a larger matrix, I see something in the map function but when I look  at the constructor it talks about embed_padding and embed_location and there is no immediate text to put these into context here for the reader. And I wonder if simply `padding` and `location` would be easier names if the text describes these as referring to the embedded matrix.

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
mrossinek: Oh I overlooked the fact that the list * integer multiplication will chain the list. Yeh this makes sense now, thanks! :+1:

a-matsuo:
@mrossinek @ikkoham @woodsp-ibm 
Thank you for all the comments! I fixed the codes and reflected all the comments. Could you please review the code again?

releasenotes/notes/logarithmic_mapper-f655fc66d319ba08.yaml
```diff
@@ -0,0 +1,6 @@
+features:
+  - |
+    Adds :class:`qiskit_nature.mappers.second_quantization.logarithmic_mapper`.
```
woodsp-ibm, line 3:
```suggestion
    Adds :class:`qiskit_nature.mappers.second_quantization.LogarithmicMapper`.
```
This needs to refer to the class itself not the module that its in (the release note showed the term highlighted but it was not a link). I guess we want to show the full path here for the context where the class is, that's fine I think.

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,231 @@
...
+    In this local encoding transformation, each individual spin S system is represented via
+    the lowest lying 2S+1 states in a qubit system with the minimal number of qubits needed to
+    represent >= 2S+1 distinct states.
+
+    [1]: Mathis, S. V., Guglielmo, M., & Ivano, T. (2020).
+         Toward scalable simulations of lattice gauge theories on quantum computers.
+         Phys. Rev. D, 102 (9), 094501. 10.1103/PhysRevD.102.094501
+    """
+
+    def __init__(self, padding: float = 1, location: EmbedLocation = EmbedLocation.UPPER) -> None:
```
woodsp-ibm, line 48: I can't help but wonder, given that this only has two values, whether a boolean might be "better". It would not need to add an extra class/enum type that needs to be documented etc i.e.
```
embed_upper: bool = True
```
Where if False then it does lower.

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,231 @@
...
+                            np.zeros((dim_diff, subs_dim), dtype=complex),
+                            np.eye(dim_diff) * self._padding,
+                        ],
+                    ]
+                )
+
+            else:
+                raise ValueError(
+                    "location must be one of " "EmbedLocation.UPPER or EmbedLocation.LOWER"
+                )
```
woodsp-ibm, line 224: As a bool this error test would then not be necessary

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
mrossinek: I like the idea of using a boolean for this better than my original Enum suggestion, too :+1: Thanks @woodsp-ibm!

mrossinek (commented):
I agree with Steve's comments and other than that this LGTM :+1:

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
a-matsuo: I agree with @woodsp-ibm. Somehow, I never thought about using a boolean for 2 choices.

a-matsuo:
Thank you so much guys! The PR is ready again now.

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,208 @@
...
+from .spin_mapper import SpinMapper
+
+
+class LogarithmicMapper(SpinMapper):
+    """A mapper for Logarithmic spin-to-qubit mapping [1].
+    In this local encoding transformation, each individual spin S system is represented via
+    the lowest lying 2S+1 states in a qubit system with the minimal number of qubits needed to
+    represent >= 2S+1 distinct states.
+
+    [1]: Mathis, S. V., Guglielmo, M., & Ivano, T. (2020).
```
mrossinek, line 35:
```suggestion
    [1]: S. V. Mathis, G. Mazzola and I. Tavernelli (2020).
```
This makes the use of first- and surnames consistent. I am also still not sure about the `(2020)` at the end.. @pbark do you have a suggestion here?

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,208 @@
...
+        Args:
+            padding:
+                When embedding a matrix into the upper/lower diagonal block of a 2^num_qubits by
+                2^num_qubits matrix, pads the diagonal of the block matrix with
+                the value of `padding`.
+
+            embed_upper:
+                This parameter sets whether
+                the given matrix is embedded in the upper left hand corner or
+                the lower right hand corner of the larger matrix.
```
mrossinek, line 51: Please reformat this to a 100-width text block

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
```diff
@@ -0,0 +1,208 @@
...
+                I.e. using embed_upper = True returns the matrix:
+                [[ matrix,    0             ],
+                [   0   , padding * I]]
```
mrossinek, line 54: This will likely need to be wrapped in some sort of block in order for Sphinx to display this properly. Same below

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
a-matsuo: Thank you so much for reviewing the PR! I fixed everything. 

I'm not sure about the `(2020)` either. I just copied from this page, https://journals.aps.org/prd/abstract/10.1103/PhysRevD.102.094501. If we have something more appropriate, I have no problem to change it.  @mrossinek @pbark

mrossinek: I thought about this some more but forgot to update my comment here. How about we leave the names on a separate line and append the `(2020).` at the end of last line which indices the journal, volume and pages.

mrossinek: I think rather this:
```
Phys. Rev. D, 102 (9), 094501 (2020).
https://doi.org/10.1103/PhysRevD.102.094501
```

mrossinek (commented):
Other than this final formatting comment above, this LGTM :+1:

qiskit_nature/mappers/second_quantization/logarithmic_mapper.py
a-matsuo: Thx! I fixed it.

mrossinek (approved):
Let's await a review by @woodsp-ibm and/or @pbark before merging :+1:

ikkoham (approved):
LGTM

woodsp-ibm (approved):

Issue #124, opened 2021-04-12
Refactor `ooVQE`


mrossinek:
### What is the expected enhancement?

The `ooVQE` did not make it into the last release because it needs to undergo a major refactoring to fit into the framework of Qiskit Nature. The most up-to-date (but WIP) version of `ooVQE` can be found [here](https://github.com/dlasecki/qiskit-nature/blob/23e1fc77347968c1ae7bdfd3f5ce264ca705bf6a/qiskit_nature/algorithms/ground_state_solvers/orbital_optimization_vqe.py)

I am replicating a (non-exhaustive) TODO list from that file below:
- [ ] removal of `molecular_problem` from `__init__` (this will only be given to `solver()`)
- [ ] actual use of `MolecularProblem.second_q_ops()` rather than hard-coded `build_fermionic_op`
- [ ] unification with `BaseProblem.solve()`
- [ ] general refactorings of flow
- [ ] fix freeze-core support based on `FreezeCoreTransformer`
- [ ] fix dependence on `num_parameters` handling within Terra's VQE and Optimizer

ElePT:
I think that this could be a good first issue for me, but it might take more than one sprint with my current bandwidth. Would you like me to work on it?

pbark:
@ElePT that would be great. I will split in multiple issues and assign them to you. Thank you very much!

ElePT:
Changing this issue into an epic for better tracking, as it will be divided in 2.

PR #629, opened 2022-04-05
[WIP] Refactor oovqe

ElePT:
### Summary
This draft PR addresses #124 without following the current division of the epic into issues.

In particular, the algorithm involves 3 classes:

- `CustomProblem` : this class implements a small modification of the `ElectronicStructureProblem` class (maybe should be called `CustomElectronicStructureProblem`) to avoid running the driver every time `second_q_ops()` is called. This is a temporary fix while the new design of qiskit nature's `Problem` classes is finalized, as it is necessary to be able to perform the orbital rotations.
- `OrbitalRotation`: this class is an almost exact copy of the original `OrbitalRotation` class in Aqua. The name is a bit misleading, because it is not really responsible for rotating the orbitals, it is just a collection of helper functions and data containers for ooVQE. I have decided to keep it for the moment because it helps keep a cleaner `ooVQE` class, but there are high chances that there is a better way of structuring this.
-   `OrbitalOptimizationVQE`: this class extends `GroundStateEigensolver` and modifies the solver's default `compute_min_eigenvalue` and `energy_evaluation` functions to allow access to the `Problem`, and to enable the optimization of the orbital rotation parameters. After discussing it with @mrossinek, I decided to monkey-patch the `VQE` class instead of extending it, but it could be done either way. The orbital rotation logic is contained in the `rotate_orbitals()` method. 

### Details and comments

This implementation is incomplete, it is missing:
- iterative orbital optimization
- support for beta orbital rotations
- rotation logic for auxiliary operators (I am not sure if it's necessary, not sure if `transform_basis` can be applied)
- getters, setters

### DIFF
--- a/qiskit_nature/algorithms/ground_state_solvers/oovqe_algorithm.py
+++ b/qiskit_nature/algorithms/ground_state_solvers/oovqe_algorithm.py
@@ -46,6 +46,7 @@ from qiskit_nature.algorithms.ground_state_solvers.oovqe_solver import compute_m
 
 logger = logging.getLogger(__name__)
 
+
 class OrbitalOptimizationVQE(GroundStateEigensolver):
     """Solver for ooVQE"""
 
@@ -53,43 +54,20 @@ class OrbitalOptimizationVQE(GroundStateEigensolver):
         self,
         qubit_converter: QubitConverter,
         solver: Union[MinimumEigensolver, MinimumEigensolverFactory],
+        initial_point: Union[ListOrDict, np.ndarray] = None,
     ) -> None:
         super().__init__(qubit_converter, solver)
 
         # Store problem to have access during energy eval. function.
-        self.problem: CustomProblem = (
-            None  # I am using temporarily the CustomProblem class, that avoids
-        )
+        self.problem: CustomProblem = None
+        # I am using temporarily the CustomProblem class, that avoids
         # running the driver every time .second_q_ops() is called
+        self.orbital_rotation = None  # to set during solve()
+        self.num_parameters_oovqe = None  # to set during solve()
 
-        self.initial_point = None  # in the future: set by user
-        self.bounds_oo: np.array = None  # in the future: set by user
-        self.bounds: np.array = None  # ansatz + oo
-
-        self.orbital_rotation = OrbitalRotation(
-            num_qubits=self.solver.ansatz.num_qubits, qubit_converter=qubit_converter
-        )
-        self.num_parameters_oovqe = (
-            self.solver.ansatz.num_parameters + self.orbital_rotation.num_parameters
-        )
-
-        # the initial point of the full ooVQE alg.
-        if self.initial_point is None:
-            self.set_initial_point()
-        else:
-            # this will never really happen with the current code
-            # but is kept for the future
-            if len(self.initial_point) is not self.num_parameters_oovqe:
-                raise QiskitNatureError(
-                    f"Number of parameters of OOVQE ({self.num_parameters_oovqe,}) "
-                    f"does not match the length of the "
-                    f"intitial_point ({len(self.initial_point)})"
-                )
-
-        if self.bounds is None:
-            # set bounds sets both ansatz and oo bounds
-            # do we want to change the ansatz bounds here??
-            self.set_bounds(self.orbital_rotation.parameter_bound_value)
+        self.initial_point = initial_point  #
+        self.bounds_oo = None  # in the future: set by user
+        self.bounds = None  # ansatz + oo
 
     def set_initial_point(self, initial_pt_scalar: float = 1e-1) -> None:
         """Initializes the initial point for the algorithm if the user does not provide his own.
@@ -151,7 +129,7 @@ class OrbitalOptimizationVQE(GroundStateEigensolver):
                     if name in aux_ops.keys():
                         raise QiskitNatureError(
                             f"The key '{name}' is already taken by an internally constructed "
-                            "auxliliary operator! Please use a different name for your custom "
+                            "auxiliary operator! Please use a different name for your custom "
                             "operator."
                         )
                     aux_ops[name] = converted_aux_op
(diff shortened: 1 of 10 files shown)


qiskit_nature/algorithms/ground_state_solvers/oovqe_algorithm.py
```diff
@@ -0,0 +1,631 @@
...
+        ee = grouped_property_transformed.get_property('ElectronicEnergy')
+        one_body_integrals = ee.get_electronic_integral(ElectronicBasis.MO, 1)
+        two_body_integrals = ee.get_electronic_integral(ElectronicBasis.MO, 2)
+
+        # the basis transform should be applied in place, but it's not???
+        # unless I manually add the integrals, the result of second_q_ops()
+        # doesn't change.
+        # I have to look further into this.
+        ee.add_electronic_integral(one_body_integrals.transform_basis(transform))
+        ee.add_electronic_integral(two_body_integrals.transform_basis(transform))
```
ElePT, line 392: @mrossinek I am a bit confused about this, but I had to explicitly add the electronic integrals to the electronic energy grouped property after applying transform basis, or else the output from` second_q_ops()` wouldn't change... but they are supposed to happen in place, right?

ElePT: Ok, they do not happen in place. Problem solved :)

ElePT: Ok, they do not happen in place. Problem solved :)

mrossinek: Sorry for the lack of reply, I overlooked this mention notification! :see_no_evil:

qiskit_nature/algorithms/ground_state_solvers/orbital_rotation.py
```diff
@@ -0,0 +1,218 @@
...
+        self._parameters = parameters
+        if self._parameters is None:
+            self._create_parameter_list_for_orbital_rotations()
+
+        self._num_parameters = len(self._parameters)
+        self._parameter_bounds = parameter_bounds
+        if self._parameter_bounds is None:
+            self._create_parameter_bounds()
+
+        if self._qubit_converter.two_qubit_reduction is True:
```
mrossinek, line 81:
```suggestion
        if self._qubit_converter.two_qubit_reduction:
```

qiskit_nature/algorithms/ground_state_solvers/orbital_rotation.py
```diff
@@ -0,0 +1,218 @@
...
+        elif self._orbital_rotations is None and self._orbital_rotations_beta is not None:
+            raise QiskitNatureError(
+                "Only beta orbitals labels (orbital_rotations_beta) have been provided."
+                "Please also specify the alpha orbitals (orbital_rotations) "
+                "that are rotated as well. Do not specify anything to have by default "
+                "all orbitals rotated."
+            )
+
+        if self._orbital_rotations_beta is not None:
+            num_parameters = len(self._orbital_rotations + self._orbital_rotations_beta)
```
mrossinek, line 163:
```suggestion
            num_parameters = len(self._orbital_rotations) + len(self._orbital_rotations_beta)

File: qiskit_nature/algorithms/ground_state_solvers/orbital_rotation.py [hunk 4]
```diff
@@ -0,0 +1,218 @@
...
+                k_matrix_beta[exc[1]][exc[0]] = -self._parameters[i]
+        else:
+            for i, exc in enumerate(self._orbital_rotations):
+                k_matrix_alpha[exc[0]][exc[1]] = self._parameters[i]
+                k_matrix_alpha[exc[1]][exc[0]] = -self._parameters[i]
+
+            for j, exc in enumerate(self._orbital_rotations_beta):
+                k_matrix_beta[exc[0]][exc[1]] = self._parameters[j + len(self._orbital_rotations)]
+                k_matrix_beta[exc[1]][exc[0]] = -self._parameters[j + len(self._orbital_rotations)]
+
```
mrossinek, line 194: In this moment in Aqua, the freeze core logic is included. We need to figure out a way how to do the
equivalent here (i.e. handle potential transformers being passed to the `problem` instance).
For the time being, as long as your tests do not use any transformers, you can ignore this.

mrossinek (commented):
Here are a bunch of observations. I did not test any of these but maybe they lead us closer to the
source of the current issues 👍

mrossinek:
This PR has became stale over time due to the refactoring of the entire stack in version 0.5

As such, I think it will be easier to revamp this work from scratch once it becomes a priority. That said, similar results to what ooVQE can do, can now be achieved using the CASSCF integration provided by the https://github.com/qiskit-community/qiskit-nature-pyscf plugin.

YangD5014:
Could you please consider to refactor OO-VQE with qiskit-nature-pyscf?

mrossinek:
> Could you please consider to refactor OO-VQE with qiskit-nature-pyscf?

Maybe my previous comment was not clear: the Qiskit Nature PySCF plugin _already_ supports CASSCF which can achieve a similar result as the ooVQE. Thus, the addition of ooVQE does not add a lot of value right now.
As such, there is no need for further changes and you can simply use the plugin in conjunction with the CASSCF implemented by PySCF.

buttercutter:
> This PR has became stale over time due to the refactoring of the entire stack in version 0.5
> 
> As such, I think it will be easier to revamp this work from scratch once it becomes a priority. That said, similar results to what ooVQE can do, can now be achieved using the CASSCF integration provided by the https://github.com/qiskit-community/qiskit-nature-pyscf plugin.

@mrossinek When I tried to check its demo usage as well as did a search on its coding, it seems that `qiskit-nature-pyscf` does not do orbital rotation on its own ?

Feel free to correct me if wrong / miss anything.

PR #632, opened 2022-04-07
Add HF initial point

declanmillar:
### Summary

Following #590, it was decided to introduce a Hartree-Fock initial point that would act as the default argument for the VQE UCC factories (see [discussion](https://github.com/Qiskit/qiskit-nature/pull/590#discussion_r839281564)). This will go some way to addressing the concerns in #571.

This PR also introduces a VSCFInitialPoint that acts in the same way (minus the energy corrections) for UVCC.

### Details and comments

- [x] This PR introduces the HFInitialState, which inherits from the InitialPoint base class introduced in #590.
- [x] ~~The MP2InitialState now inherits from HFInitialState and all compatible methods/attributes/properties have been moved up the inheritance tree.~~
- [x] This necessitated a small change in the UCC class to ensure the excitation_list is reduced symmetrically with the operators following the application of the qubit converter.
- [x] Unit tests have been updated for full coverage of the initial point classes.
- [x] This refactoring should also consider the UVCC+VQE initial_point as raised [here](https://github.com/Qiskit/qiskit-nature/pull/590#discussion_r845288397).


qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -12,7 +12,8 @@
 
 """The minimum eigensolver factory for ground state calculation algorithms."""
 
-from typing import Optional, Union, Callable, cast
+from __future__ import annotations
+from typing import Callable, cast
```
mrossinek, line 16: I suggest you try to keep the changes to this file minimal because of the parallel effort in #646. This will make all our lives a bit easier :smile:

mrossinek: The same applies to the other factory.

qiskit_nature/algorithms/initial_points/hf_initial_point.py
```diff
@@ -0,0 +1,203 @@
...
+from .initial_point import InitialPoint
+
+logger = logging.getLogger(__name__)
+
+
+class HFInitialPoint(InitialPoint):
+    """Compute the Hartree-Fock (HF) initial point.
+
+    All excitations will have zero coefficient and energy correction as the all-zero initial point
+    corresponds to the HF initial point once the HF initial state is prepended to the circuit.
```
mrossinek, line 37: I think this phrasing needs to clarify that this only holds for the UCC ansatz. I don't think that in general this applies to any ansatz.

qiskit_nature/algorithms/initial_points/hf_initial_point.py
```diff
@@ -0,0 +1,203 @@
...
+    @property
+    def ansatz(self) -> UCC:
+        """The UCC ansatz.
+
+        This is used to ensure that the :attr:`excitation_list` matches with the UCC ansatz that
+        will be used with the VQE algorithm.
+
+        Raises:
+            QiskitNatureError: If not set using a valid
+                :class:`~qiskit_nature.circuit.library.ansatzes.ucc.UCC` instance.
```
mrossinek, line 92: I don't see this being raised here. EDIT: ah hold on. This is a docstring for the getter AND setter, right?

qiskit_nature/algorithms/initial_points/initial_point.py
```diff
@@ -30,27 +30,31 @@ class InitialPoint(ABC):
...
+    def __len__(self) -> int:
+        """The length of the computed initial point array."""
+        raise NotImplementedError
 
     @property
     @abstractmethod
-    def grouped_property(self) -> GroupedSecondQuantizedProperty:
-        """The grouped property."""
+    def ansatz(self) -> UCC | None:
+        """The UCC ansatz."""
```
mrossinek, line 43: For the VSCF-case this is actually UVCC..
I'm wondering if this needs to be restricted at all in this way :thinking:

qiskit_nature/algorithms/initial_points/hf_initial_point.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def get_energy_corrections(self) -> np.ndarray:
+        """The energy corrections for each excitation."""
+        if self._corrections is None:
+            self.compute()
+        return np.asarray([value[1] for value in self._corrections.values()])
+
+    def get_energy_correction(self) -> float:
+        """The overall energy correction."""
+        return self.get_energy_corrections().sum()
```
mrossinek, line 139: I was a bit confused to see all this logic here because all these energy terms are simply zero in the HF case, but I see that you made the MP2 point a subclass of this.
I am wondering whether it would be better to keep the HF case as simple as possible and instead have this logic only on the MP2 case. But presumably the HF point will still need these methods in place (but return zero) for API consistency, right?

What does this look like compared to VSCF? Could it make sense to extract base implementations of these methods into the `InitialPoint` super class?

qiskit_nature/algorithms/initial_points/hf_initial_point.py
```diff
@@ -0,0 +1,203 @@
...
+        return self._reference_energy + self.get_energy_correction()
+
+    def compute(
+        self,
+        ansatz: UCC | None = None,
+        grouped_property: GroupedSecondQuantizedProperty | None = None,
+    ) -> None:
+        """Compute the coefficients and energy corrections.
+
+        See :class:`HFInitialPoint` for more information.
```
mrossinek, line 157: This is basically a self-reference. I think you could simply say "See further up for more information."

qiskit_nature/algorithms/initial_points/hf_initial_point.py
mrossinek: To clarify: I am not suggesting that the generated initial point should be an all-zero vector by default. But instead, the returned energy correction could default to being `0.0`.

releasenotes/notes/add-hf-initial-point-ac29baf3091fa5b1.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+    :class:`~qiskit_nature.algorithms.initial_points.MP2InitialPoint` now
+    inherits from
+    :class:`~qiskit_nature.algorithms.initial_points.HFInitialPoint`, while
+    :class:`~qiskit_nature.algorithms.initial_points.HFInitialPoint` and
+    :class:`~qiskit_nature.initial_points.VSCFInitialPoint` both inherit
+    from the abstract base class
+    :class:`~qiskit_nature.initial_points.InitialPoint`.
+
+    These initial point are intended to be used as starting VQE parameters when
```
mrossinek, line 24:
```suggestion
    These initial points are intended to be used as starting VQE parameters when
```

qiskit_nature/algorithms/initial_points/hf_initial_point.py
declanmillar: Yeah, I think the setter docstring is ignored by Sphinx, so it acts for both.

qiskit_nature/algorithms/initial_points/hf_initial_point.py
declanmillar: For HF and VSCF one could obviously just return a NumPy all-zero array of the appropriate size. This would make their logic way more transparent. I kept the dictionary simply to keep the logic in MP2 the same, but given that HFInitialPoint is likely to be used far more often than MP2, I'm tempted to dump the dictionary... yeah, I've talked myself into doing that.

mrossinek (commented):
Wow this PR has grown a lot :smile: Overall I think it looks good but I will need to do another thorough pass..

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -218,6 +219,12 @@ def get_solver(  # type: ignore[override]
         ansatz.num_modals = num_modals
         ansatz.initial_state = initial_state
 
+        if isinstance(self.initial_point, InitialPoint):
+            self.initial_point.ansatz = ansatz
+
+            # Override the initial_point with the computed array.
+            self.initial_point = self.initial_point.to_numpy_array()
```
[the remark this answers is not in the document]
declanmillar: Yeah, that's a good call.

qiskit_nature/circuit/library/ansatzes/uvcc.py
```diff
@@ -152,21 +168,30 @@ def operators(self):  # pylint: disable=invalid-overridden-method
...
-                # behavior is suppressed.
-                self.operators = self.qubit_converter.convert_match(
-                    excitation_ops, suppress_none=True
-                )
+                # it may be that one or more excitation operators do not commute with the symmetry.
+                # The converted operators are maintained at the same index by the converter inserting
+                # None as the result if an operator did not commute.
+                operators = self.qubit_converter.convert_match(excitation_ops, suppress_none=False)
+
+                # Here we are only interested in getting the valid set of operators. To ensure that
```
woodsp-ibm, line 184: By 'valid set` this means non-None. You switched from the conversion code dropping non-commuting ones to having None instead in order that you can remove them manually since you need to ensure any initial point to matches with the given terms and this is done based on their index order in the list which dropping them as before does not allow you too track.

releasenotes/notes/add-hf-initial-point-ac29baf3091fa5b1.yaml
```diff
@@ -0,0 +1,51 @@
+---
+features:
+  - |
+    Adds Hartree-Fock (HF) and vibrational self-consistent field (VSCF) initial
+    point classes.
+    :class:`~qiskit_nature.initial_points.HFInitialPoint` and
+    :class:`~qiskit_nature.initial_points.VSCFInitialPoint`,
```
woodsp-ibm, line 7: I do not think this is where these are right - later it has correct references to qiskit_nature.algorithms.initial_points but others here are not correct.

releasenotes/notes/add-hf-initial-point-ac29baf3091fa5b1.yaml
```diff
@@ -0,0 +1,51 @@
...
+    from the abstract base class
+    :class:`~qiskit_nature.initial_points.InitialPoint`.
+
+    These initial points are intended to be used as starting VQE parameters when
+    using a UCC ansatz (cluster operator coefficients). This should introduce an
+    initial point that is closer to the ground state point, leading to fewer
+    overall evaluations for VQE.
+
+    Instances of the appropriate classes are now the default
+    :attr:`initial_point` arguments in
```
woodsp-ibm, line 30: This fails to resolve since its not on one of the classes below so its just highlighted which maybe is ok/

test/algorithms/excited_state_solvers/test_bosonic_esc_calculation.py
```diff
@@ -124,6 +124,22 @@ def test_vqe_uvccsd_factory(self):
...
+    def test_vqe_uvcc_factory_with_user_initial_point(self):
+        """Test VQEUVCCFactory when using it with a user defined initial point."""
+
+        optimizer = COBYLA(maxiter=5000)
+        quantum_instance = QuantumInstance(
+            backend=qiskit.BasicAer.get_backend("statevector_simulator"),
+            seed_simulator=algorithm_globals.random_seed,
+            seed_transpiler=algorithm_globals.random_seed,
+        )
+        solver = VQEUVCCFactory(quantum_instance, optimizer=optimizer, initial_point=[1, 2, 3])
```
woodsp-ibm, line 136: How do we know that this ` initial_point=[1, 2, 3]` does anything? The code does not fail but can we test that some behavior is influenced by that such that is actually functioning as expected. Ie. I suspect the test will pass given other values of initial point, and indeed just leaving it as default. What could be done is to set the optimizer to say a single iter and set the initial point to a prior known optimal point. Then with only one iteration if the initial point is indeed having an effect we should already be at the solution.

test/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
```diff
@@ -547,7 +547,20 @@ def test_default_initial_point(self):
...
+        self.assertAlmostEqual(res.total_energies[0], self.reference_energy, places=6)
+
+    def test_vqe_ucc_factory_with_user_initial_point(self):
+        """Test VQEUCCFactory when using it with a user defined initial point."""
+
+        initial_point = [1, 2, 3]
+        solver = VQEUCCFactory(
+            QuantumInstance(BasicAer.get_backend("statevector_simulator")),
+            initial_point=initial_point,
+        )
```
woodsp-ibm, line 560: Similar comment as per above test

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         self,
         quantum_instance: QuantumInstance,
         optimizer: Optional[Optimizer] = None,
-        initial_point: Optional[Union[np.ndarray, InitialPoint]] = None,
+        initial_point: Optional[Union[np.ndarray, InitialPoint]] = HFInitialPoint(),
```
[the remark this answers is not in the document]
declanmillar: Good point. I actually remember that now as one of the classic Python "gotchas"... I think the original goal of introducing a HFInitialPoint was to make the default arguments more explicit, but I suppose it will still be clear if we explicitly spin up an instance to replace None.

qiskit_nature/algorithms/initial_points/__init__.py
```diff
@@ -20,25 +20,43 @@
...
+The initial point interface.
 
 .. autosummary::
    :toctree: ../stubs/
    :nosignatures:
 
    InitialPoint
 
-The MP2 initial point generator.
+The HF initial point.
```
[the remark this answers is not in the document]
declanmillar: I actually would prefer users to import them explicitly from initial_points. So, I'll remove the documentation and imports from the algorithms directory.

qiskit_nature/algorithms/initial_points/hf_initial_point.py
```diff
@@ -0,0 +1,177 @@
...
+                "grouped_property and reference_energy will not be set."
+            )
+            return
+
+        electronic_energy: ElectronicEnergy | None = grouped_property.get_property(ElectronicEnergy)
+        if not isinstance(electronic_energy, ElectronicEnergy):
+            logger.warning(
+                "ElectronicEnergy not obtained from grouped_property; "
+                "grouped_property and reference_energy will not be set."
+            )
```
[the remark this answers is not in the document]
declanmillar: Thanks, Steve. That's a useful reference!

qiskit_nature/algorithms/initial_points/hf_initial_point.py
```diff
@@ -0,0 +1,177 @@
...
+        Raises:
+            QiskitNatureError: If not set using a valid
+                :class:`~qiskit_nature.circuit.library.ansatzes.ucc.UCC` instance.
+        """
+        return self._ansatz
+
+    @ansatz.setter
+    def ansatz(self, ansatz: UCC) -> None:
+        if not isinstance(ansatz, UCC):
+            raise QiskitNatureError(f"Expected a UCC ansatz, but got type: {type(ansatz)}.")
```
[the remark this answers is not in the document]
declanmillar: Yeah, I was conflicted about this. @mrossinek do you have a strong opinion? I could relax this instance checks in a number of places. It feels "unpythonic" to be this strict on the type at runtime, but it might catch the odd user input error earlier.

qiskit_nature/circuit/library/ansatzes/uvcc.py
```diff
@@ -112,29 +114,43 @@ def qubit_converter(self, conv: QubitConverter) -> None:
...
-    def excitations(self, exc: Union[str, int, List[int], Callable]) -> None:
+    def excitations(self, exc: str | int | list[int] | Callable) -> None:
         """Sets the excitations."""
         self._operators = None
         self._invalidate()
         self._excitations = exc
 
+    @property
+    def excitation_list(self) -> list[tuple[tuple[int, ...], tuple[int, ...]]]:
+        """The excitation list that UVCC is using, in the required format.
```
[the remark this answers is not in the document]
declanmillar: I meant "in the required format" as opposed to `excitations` which could be a string etc. But you're right, it's not clear at all.

qiskit_nature/circuit/library/ansatzes/uvcc.py
declanmillar: I updated the comment a bit to try to clarify what I'm doing here. Let me know if it still needs work.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -27,11 +29,20 @@
     GroupedSecondQuantizedProperty,
 )
 
-from .initial_point import InitialPoint
+from .hf_initial_point import HFInitialPoint
+
+
+@dataclass(frozen=True)
+class Correction:
```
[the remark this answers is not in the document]
declanmillar: Yeah, the user doesn't need to be exposed to this class. I've made it the Python-private `_Corrections` to reflect this. Let me know if you think it'd be better nested.

qiskit_nature/algorithms/initial_points/hf_initial_point.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._reference_energy: float = 0.0
+        self._excitation_list: list[tuple[tuple[int, ...], tuple[int, ...]]] | None = None
+        self._parameters: np.ndarray | None = None
+
+    def __len__(self) -> int:
+        """The length of the computed initial point array."""
+        return len(self.to_numpy_array())
```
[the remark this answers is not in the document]
declanmillar: In the end, I just decided to remove `__len__` as it's not really necessary. One can use `len` on the result of `to_numpy_array`. It's probably easier to maintain without it.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -27,11 +29,20 @@
...
+    """Data class for storing corrections."""
 
+    excitation: tuple[tuple[int, ...], tuple[int, ...]]
+    coefficient: float = 0.0
+    energy: float = 0.0
 
-class MP2InitialPoint(InitialPoint):
-    """Compute second-order Møller-Plesset perturbation theory (MP2) corrections.
+
+class MP2InitialPoint(HFInitialPoint):
```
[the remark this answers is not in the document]
declanmillar: Yeah, it's mostly just for the setters and getters, though I've refactored the classes a little to remove unnecessary methods and attributes from the base class and VSCF. This means that I've restricted the types expected in the factories to the appropriate subclasses. I.e. VQEUCC can only take HFInitialPoint and its subclasses (i.e. MP2InitialPoint) and VQEUVCC can only take VSCFInitialPoint and its subclasses (which don't exist yet).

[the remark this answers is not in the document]
declanmillar: Principally, the ABC InitialPoint only knows about `ansatz`, and no longer about `grouped_property` or the energy-related methods. VSCF doesn't need `grouped_property` or the energy-related methods either. HF introduces these (minus the energy corrections) and then MP2 inherits these and adds the corrections and expects more for the grouped_property.

qiskit_nature/algorithms/initial_points/hf_initial_point.py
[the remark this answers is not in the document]
declanmillar: I decided to remove the runtime type checking. I think the type hints should be sufficient for the user to debug their approach. The overhead/maintenance of type checking everything is a bad trend to set, I think. Happy to revert if there are objections.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -210,20 +213,20 @@ def get_solver(  # type: ignore[override]
         ansatz.num_spin_orbitals = num_spin_orbitals
         ansatz.initial_state = initial_state
 
-        if isinstance(self.initial_point, InitialPoint):
-            self.initial_point.grouped_property = driver_result
+        if isinstance(self.initial_point, HFInitialPoint):
```
mrossinek, line 216:
```suggestion
        if isinstance(self.initial_point, InitialPoint):

File: qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py [hunk 23]
```diff
@@ -218,13 +225,19 @@ def get_solver(  # type: ignore[override]
         ansatz.num_modals = num_modals
         ansatz.initial_state = initial_state
 
+        if isinstance(self.initial_point, VSCFInitialPoint):
```
mrossinek, line 228:
```suggestion
        if isinstance(self.initial_point, InitialPoint):

File: qiskit_nature/algorithms/initial_points/hf_initial_point.py [hunk 24]
```diff
@@ -0,0 +1,160 @@
...
+    def __init__(self) -> None:
+        super().__init__()
+        self._ansatz: UCC | None = None
+        self._excitation_list: list[tuple[tuple[int, ...], tuple[int, ...]]] | None = None
+        self._grouped_property: GroupedSecondQuantizedProperty | None = None
+        self._reference_energy: float = 0.0
+        self._parameters: np.ndarray | None = None
+
+    @property
+    def grouped_property(self) -> GroupedSecondQuantizedProperty | None:
```
mrossinek, line 54: Just to be even more explicit: in order to unify the electronic and vibrational structure cases,
I suggest that these attributes become part of the base class.

mrossinek (commented):
Okay this was another more thorough round. Besides these last comments I think this is pretty much
good to be merged 👍

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
declanmillar: Incorporated this change as part of [8c5b10f](https://github.com/Qiskit/qiskit-nature/pull/632/commits/8c5b10fc437a19c4020ce2b57d34afd5966beb3c).

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
declanmillar: Incorporated this change as part of [8c5b10f](https://github.com/Qiskit/qiskit-nature/pull/632/commits/8c5b10fc437a19c4020ce2b57d34afd5966beb3c).

qiskit_nature/algorithms/initial_points/hf_initial_point.py
declanmillar: Incorporated this change as part of [8c5b10f](https://github.com/Qiskit/qiskit-nature/pull/632/commits/8c5b10fc437a19c4020ce2b57d34afd5966beb3c).

releasenotes/notes/add-hf-initial-point-ac29baf3091fa5b1.yaml
```diff
@@ -0,0 +1,46 @@
+---
+features:
+  - |
+    Adds Hartree-Fock (HF) and vibrational self-consistent field (VSCF) initial point classes.
+    :class:`~qiskit_nature.algorithms.initial_points.hf_initial_point.HFInitialPoint` and
```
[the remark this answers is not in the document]
declanmillar: I actually wasn't sure that Sphinx respected imports or needed the raw file path. It's good to know that it does the former :).

releasenotes/notes/add-hf-initial-point-ac29baf3091fa5b1.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+    These initial points are intended to be used as starting VQE parameters when using a UCC ansatz
+    (cluster operator coefficients). This should introduce an initial point that is closer to the
+    ground state point, leading to fewer overall evaluations for VQE.
+
+    Instances of the appropriate classes are now the default :attr:`initial_point` arguments in
+    :class:`~qiskit_nature.algorithms.ground_state_solvers.minimum_eigensolver_factory.vqe_ucc_factory.VQEUCCFactory`
+    and
+    :class:`~qiskit_nature.algorithms.ground_state_solvers.minimum_eigensolver_factory.vqe_uvcc_factory.VQEUVCCFactory`.
+
```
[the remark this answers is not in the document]
declanmillar: Yeah, that makes sense, I'll amend the release note with a code snippet for using it with default stack.

pbark (approved):
I think this now looks good! Thanks!

mrossinek (approved):
LGTM, too! Thanks a lot, @declanmillar! :slightly_smiling_face:

woodsp-ibm (approved):
Looks good, nice job!

Issue #634, opened 2022-04-11
Error using VQEClient with dict-based aux_operators


cpossel:
### Environment

- **Qiskit Nature version**: 0.3.2
- **Python version**: 3.8.11
- **Operating system**: Windows 10

### What is happening?

When trying to submit a job with `VQEClient` I get the following error:
```
Traceback (most recent call last):
  File "C:\Users\poc\.conda\envs\qiskit_gaussian\lib\site-packages\qiskit_nature\runtime\vqe_client.py", line 381, in _convert_to_paulisumop
    primitive = SparsePauliOp(operator.primitive)
AttributeError: 'str' object has no attribute 'primitive'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File [...], line [...], in solve
    raw_result = self.ground_state_solver.solve(problem=self.problem)
  File "C:\Users\poc\.conda\envs\qiskit_gaussian\lib\site-packages\qiskit_nature\algorithms\ground_state_solvers\ground_state_eigensolver.py", line 144, in solve
    raw_mes_result = self._solver.compute_minimum_eigenvalue(main_operator, aux_ops)
  File "C:\Users\poc\.conda\envs\qiskit_gaussian\lib\site-packages\qiskit_nature\runtime\vqe_client.py", line 261, in compute_minimum_eigenvalue
    aux_operators = [_convert_to_paulisumop(aux_op) for aux_op in aux_operators]
  File "C:\Users\poc\.conda\envs\qiskit_gaussian\lib\site-packages\qiskit_nature\runtime\vqe_client.py", line 261, in <listcomp>
    aux_operators = [_convert_to_paulisumop(aux_op) for aux_op in aux_operators]
  File "C:\Users\poc\.conda\envs\qiskit_gaussian\lib\site-packages\qiskit_nature\runtime\vqe_client.py", line 384, in _convert_to_paulisumop
    raise ValueError(
ValueError: Invalid type of the operator <class 'str'> must be PauliSumOp, or castable to one.
```

### How can we reproduce the issue?

Getting a minimal example is a bit tricky, but it basically goes down to using the H2 ground state energy calculation example but with `VQEClient`.
I already switched from the deprecated list based `aux_operator` to dict based ones via 
```
qiskit_nature.settings.dict_aux_operators = True
```

### What should happen?

Converting to pauli sum op should terminate without error i.e., the list comprehension should be dict aware.

### Any suggestions?

[...]

PR #635, opened 2022-04-11
Iterate over dict values of aux_op.

cpossel:
### Summary

Fix bug with new dict based aux_ops via iterating over dict values instead of dict keys/list elements.

### Details and comments

For details see issue #634

### DIFF
--- a/qiskit_nature/runtime/vqe_client.py
+++ b/qiskit_nature/runtime/vqe_client.py
@@ -309,7 +309,7 @@ class VQEClient(VariationalAlgorithm, MinimumEigensolver):
             aux_op_eigenvalues = result.get("aux_operator_eigenvalues", None)
         else:
             aux_op_eigenvalues = dict(
-                zip(aux_operators, result.get("aux_operator_eigenvalues", []))
+                zip(wrapped_aux_operators, result.get("aux_operator_eigenvalues", []))
             )
             if not aux_op_eigenvalues:  # For consistency set to None for empty dict
                 aux_op_eigenvalues = None
--- a/qiskit_nature/runtime/vqe_client.py
+++ b/qiskit_nature/runtime/vqe_client.py
@@ -266,7 +266,7 @@ class VQEClient(VariationalAlgorithm, MinimumEigensolver):
 
         # try to convert the operators to a PauliSumOp, if it isn't already one
         operator = _convert_to_paulisumop(operator)
-        aux_operators = {
+        wrapped_aux_operators = {
             str(aux_op_name_or_idx): _convert_to_paulisumop(aux_op)
             for aux_op_name_or_idx, aux_op in ListOrDict(aux_operators).items()
         }
@@ -274,7 +274,7 @@ class VQEClient(VariationalAlgorithm, MinimumEigensolver):
         # combine the settings with the given operator to runtime inputs
         inputs = {
             "operator": operator,
-            "aux_operators": aux_operators,
+            "aux_operators": wrapped_aux_operators,
             "ansatz": self.ansatz,
             "optimizer": self.optimizer,
             "initial_point": self.initial_point,
@@ -305,9 +305,15 @@ class VQEClient(VariationalAlgorithm, MinimumEigensolver):
         vqe_result.cost_function_evals = result.get("cost_function_evals", None)
         vqe_result.eigenstate = result.get("eigenstate", None)
         vqe_result.eigenvalue = result.get("eigenvalue", None)
-        vqe_result.aux_operator_eigenvalues = dict(
-            zip(aux_operators, result.get("aux_operator_eigenvalues", None))
-        )
+        if isinstance(aux_operators, list):
+            aux_op_eigenvalues = result.get("aux_operator_eigenvalues", None)
+        else:
+            aux_op_eigenvalues = dict(
+                zip(aux_operators, result.get("aux_operator_eigenvalues", []))
+            )
+            if not aux_op_eigenvalues:  # For consistency set to None for empty dict
+                aux_op_eigenvalues = None
+        vqe_result.aux_operator_eigenvalues = aux_op_eigenvalues
         vqe_result.optimal_parameters = result.get("optimal_parameters", None)
         vqe_result.optimal_point = result.get("optimal_point", None)
         vqe_result.optimal_value = result.get("optimal_value", None)
--- a/qiskit_nature/runtime/vqe_client.py
+++ b/qiskit_nature/runtime/vqe_client.py
@@ -311,12 +311,9 @@ class VQEClient(VariationalAlgorithm, MinimumEigensolver):
         vqe_result.eigenstate = result.get("eigenstate", None)
         vqe_result.eigenvalue = result.get("eigenvalue", None)
         if isinstance(aux_operators, dict):
-            vqe_result.aux_operator_eigenvalues = {
-                aux_op_name: aux_op_eigenvalue
-                for aux_op_name, aux_op_eigenvalue in zip(
-                    aux_operators, result.get("aux_operator_eigenvalues", None)
-                )
-            }
+            vqe_result.aux_operator_eigenvalues = dict(
+                zip(aux_operators, result.get("aux_operator_eigenvalues", None))
+            )
         else:
             vqe_result.aux_operator_eigenvalues = result.get("aux_operator_eigenvalues", None)
         vqe_result.optimal_parameters = result.get("optimal_parameters", None)
--- a/qiskit_nature/runtime/vqe_client.py
+++ b/qiskit_nature/runtime/vqe_client.py
@@ -13,7 +13,7 @@
 """The Qiskit Nature VQE Runtime Client."""
 
 
-from typing import List, Callable, Optional, Any, Dict, Union, TypeVar
+from typing import Callable, Optional, Any, Dict, Union
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


woodsp-ibm:
Indeed VQEClient should support a dictionary or a list to conform to how the MinimumEigenSolver interface/result is now defined in Terra and the fact that Nature now defaults to dictionary of aux_ops. When a dictionary is provided of key->operators then the result needs to be key -> values, i.e. dictionary in and dictionary out with same keys but now referring to the computed values for the aux ops.

Probably the easier solution here is as you are doing which is to build a list of values, but it needs to be done only if the type is a dictionary. Then if the input was a dictionary, once the values are obtained (will be a list), the result needs to be a turned into a dictionary. I.e this line will have to be updated too that builds the result which gets passed back to the application, so that the list is provided back if it was a list input, but if dictionary then form the dictionary.
> vqe_result.aux_operator_eigenvalues = result.get("aux_operator_eigenvalues", None)

I imagine the conversion/handling could be done on the remote side instead; but doing it in the client side here will leave that unaffected. And I think looking forward, when we have Estimator primitive doing the expectation results, where these are list based, that the client is  probably the right place to handle this. @Cryoris any thought/comment here since you were more directly involved with this runtime logic,

cpossel:
I added the changes being compatible with both lists and dicts; thanks to @woodsp-ibm for the suggestions.
Instead of converting the dict to a list as suggested by @woodsp-ibm , I decided to pass a dict to vqe runtime since runtime supports both inputs. The runtime output is still a list, so the final result is restored as a dict if the initial aux_op was a dict.

qiskit_nature/runtime/vqe_client.py
```diff
@@ -30,6 +30,10 @@
 from qiskit.opflow import OperatorBase, PauliSumOp
 from qiskit.quantum_info import SparsePauliOp
 
+# Introduced new type to maintain readability.
+_T = TypeVar("_T")  # Pylint does not allow single character class names.
+ListOrDict = Union[List[Optional[_T]], Dict[str, _T]]
```
mrossinek, line 35: Rather than writing your own type variable you can simply use
```python
from qiskit_nature import ListOrDictType
```

qiskit_nature/runtime/vqe_client.py
```diff
@@ -263,7 +268,13 @@ def compute_minimum_eigenvalue(
...
         operator = _convert_to_paulisumop(operator)
         if aux_operators is not None:
-            aux_operators = [_convert_to_paulisumop(aux_op) for aux_op in aux_operators]
+            try:
+                aux_operators = {
+                    aux_op_name: _convert_to_paulisumop(aux_op)
+                    for aux_op_name, aux_op in aux_operators.items()  # type: ignore
+                }
+            except AttributeError:
+                aux_operators = [_convert_to_paulisumop(aux_op) for aux_op in aux_operators]
```
mrossinek, line 277: You could also look at the implementation in the `QubitConverter` which handles list or dict scenarios using `qiskit_nature.ListOrDict`. I do not think the code necessarily needs this here, but for consistency purposes we could stick to one way of handling this scenario.
Here is a link to the example section: https://github.com/Qiskit/qiskit-nature/blob/72d5445d59a0829ec5a65f1ba0e36be26d63e63f/qiskit_nature/converters/second_quantization/qubit_converter.py#L325-L336

qiskit_nature/runtime/vqe_program.py
```diff
@@ -28,6 +28,10 @@
 
 from .vqe_client import VQEClient, VQERuntimeResult
 
+# Introduced new type to maintain readability.
+_T = TypeVar("_T")  # Pylint does not allow single character class names.
+ListOrDict = Union[List[Optional[_T]], Dict[str, _T]]
```
mrossinek, line 33: Same as above.

mrossinek (commented):
Just some minor suggestions.

Please also add a release note for this bugfix PR. You can read [this](https://github.com/Qiskit/qiskit-nature/blob/main/CONTRIBUTING.md#release-notes) for instructions on how to do this.

qiskit_nature/runtime/vqe_client.py
cpossel: Thanks for the comment. I just blatantly copied it from minimum_eigen_solver.py in qiskit algorithms; of course I'll change it to the nicer import.

qiskit_nature/runtime/vqe_client.py
cpossel: I added a new implementation with `ListOrDict` from utils. It looks cleaner but has the disadvantage that the (deprecated) list type is now converted to a dict by the `ListOrDict` converter without a nice way to restore the initial list after retrieving `job.result()`. So, it's now a dict with strings of numbers 0,1,... as keys (from `ListOrDict`'s `enumerate()`. If that's fine with you, I would like to stick with the cleaner approach using `ListOrDict`.

qiskit_nature/runtime/vqe_client.py
```diff
@@ -299,7 +305,9 @@ def compute_minimum_eigenvalue(
         vqe_result.cost_function_evals = result.get("cost_function_evals", None)
         vqe_result.eigenstate = result.get("eigenstate", None)
         vqe_result.eigenvalue = result.get("eigenvalue", None)
-        vqe_result.aux_operator_eigenvalues = result.get("aux_operator_eigenvalues", None)
+        vqe_result.aux_operator_eigenvalues = dict(
+            zip(aux_operators, result.get("aux_operator_eigenvalues", None))
+        )
```
mrossinek, line 310: I would expect that the conversion to `dict` is depended on the type of the `aux_operators` passed into this method and the value of `qiskit_nature.settings.dict_aux_operators` (where the latter should take precedence (and overwrite) the former).

woodsp-ibm: Since this implements the MinimumEigenSolver interface from Qiskit Terra, I think ideally it ought to follow that, which means List in -> List out, and Dict in - > Dict out as far as aux ops and their result goes. While this class does sit in Nature, its really a runtime version of VQE and I think it ought to behave as the same way. Since Nature will create Dicts it should see only Dicts back even if the runtime here supports List.

qiskit_nature/runtime/vqe_client.py
woodsp-ibm: In the dev branch of Terra the ListOrDict type has been split out so it can be directly imported elsewhere. However at present we cannot use that here since we need to stay compatible with 0.20 release. For me having a copy here so it matches what is in MinimumEigenSolver in the interim seems ok. It can be replaced with with the import at some later point.

qiskit_nature/runtime/vqe_client.py
cpossel: Good points; I completely agree with the 'Type in -> Type out' approach.
Internally in VQEClient I think it does not matter that much, so let's stick to the converted Dicts (also using them as input for the runtime even if the input to VQEClient was a List).
I see your point @mrossinek about `qiskit_nature.settings.dict_aux_operators` but I think it's better to ignore this setting since the convertion to dict can't restore the correct names for the aux_operators but only fill it with numbers counting up. So, it would end up claiming to correctly convert the List to Dict but in fact doing only filling up with dummy keys. And with `qiskit_nature.settings.dict_aux_operators` set to True it should get a Dict as input either way as @woodsp-ibm pointed out -- else things would be broken already at another place.

qiskit_nature/runtime/vqe_client.py
cpossel: Do I understand it correctly that `from qiskit_nature import ListOrDictType` is not compatible with the 0.20 Terra release? So, I should change it back to the copied lines from `minimum_eigen_solver.py`?

woodsp-ibm: ListOrDictType is the same thing just with a different name https://github.com/Qiskit/qiskit-nature/blob/8367463a47fcb46770b310d2c5b2bf7a067398bf/qiskit_nature/__init__.py#L62-L64 It makes the interface look as though the type might be different, due to importing it as a different name, but in the end as you can see its the same thing.

qiskit_nature/runtime/vqe_client.py
mrossinek: Yes you have convinced me that the `Type in -> Type out` approach is the way to go here :+1: 
This probably also resolves our conversation above, right @cpossel ?

qiskit_nature/runtime/vqe_client.py
cpossel: Thanks for the explanation. Somewhat confusing with the different names `ListOrDict` and `ListOrDictType` but now I get it.

qiskit_nature/runtime/vqe_client.py
cpossel: Yes all suggestions from the two of you are included in the newest commit. I'll leave it to you to 'resolve conversation' if you are fine with the changes made.

mrossinek (approved):
This LGTM, thanks @cpossel!

@woodsp-ibm please also take another look :+1:

woodsp-ibm (approved):
The use of ListOrDictType, which is the Terra level VQE ListOrDict type, along with the ListOrDict here from Nature would need to be sorted if this is ever moved from here elsewhere. I guess its ok for now but in my mind this class was really more independent of Nature.

PR #638, opened 2022-04-13
Improve `TwoBodyElectronicIntegrals.to_second_q_op` performance

mrossinek:
### Summary

This improves the performance of the `TwoBodyElectronicIntegrals.to_second_q_op` method.
The unusable performance was observed when attempting to run the following:
```python
import time

driver = PySCFDriver(atom="O 0 0 0.115; H 0 0.754 -0.459; H 0 -0.754 -0.459", basis="cc-pVDZ")
props = driver.run()
print(props.get_property("ParticleNumber"))

start = time.time()
props.second_q_ops()
stop = time.time()
print(stop - start)
```
which is a system of size:
```
ParticleNumber:
        48 SOs
        5 alpha electrons
                orbital occupation: [1. 1. 1. 1. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
        5 beta electrons
                orbital occupation: [1. 1. 1. 1. 1. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
```
On latest `main`, this snippet takes longer to execute than I am willing to wait (>10 minutes). After this PR, it will take 4.484 seconds.

### Details and comments

I also ran the following benchmark:
```python
import time

import numpy as np

from qiskit_nature.properties.second_quantization.electronic.bases import ElectronicBasis
from qiskit_nature.properties.second_quantization.electronic.integrals import (
    TwoBodyElectronicIntegrals,
)

sizes = range(1, 50)
times = []

for size in sizes:
    print(f"{size} MOs: ", end="")
    ints = TwoBodyElectronicIntegrals(
        ElectronicBasis.MO,
        (
            np.random.random((size, size, size, size)),
            None,
            None,
            None,
        ),
    )
    local = []
    for i in range(10):
        start = time.time()
        op = ints.to_second_q_op()
        stop = time.time()
        local.append(stop - start)
    times.append(np.average(local))
    print(f"{times[-1]}s")
```
Note, that this is a worst case scenario because we have a dense, non-zero matrix. Chemistry applications are highly sparse and all zero-terms will be skipped. However, iteration still needs to happen over `#SO**4` matrix elements..

Here are the benchmark results on 72d5445d59a0829ec5a65f1ba0e36be26d63e63f
```
1 MOs: 0.0003814220428466797s
2 MOs: 0.003143501281738281s
3 MOs: 0.01747605800628662s
4 MOs: 0.10729048252105713s
5 MOs: 0.4813446283340454s
6 MOs: 2.6585819482803346s
7 MOs: 10.059082674980164s
8 MOs: 33.26701810359955s
```
And here are the results with this PR:
```
1 MOs: 0.0001928091049194336s
2 MOs: 0.00028238296508789065s
3 MOs: 0.0008608579635620117s
4 MOs: 0.005258703231811523s
5 MOs: 0.010203838348388672s
6 MOs: 0.02182309627532959s
7 MOs: 0.049395179748535155s
8 MOs: 0.07928628921508789s
9 MOs: 0.12792351245880126s
10 MOs: 0.20389955043792723s
11 MOs: 0.32534089088439944s
12 MOs: 0.5057589530944824s
13 MOs: 0.6290717124938965s
14 MOs: 0.8679617881774903s
15 MOs: 1.2400678396224976s
16 MOs: 1.8723540782928467s
17 MOs: 2.1375882148742678s
18 MOs: 2.7122227430343626s
19 MOs: 3.2011844635009767s
20 MOs: 3.8472093820571898s
21 MOs: 4.834750986099243s
22 MOs: 5.815815830230713s
23 MOs: 7.087169241905213s
24 MOs: 8.564099335670472s
25 MOs: 9.87972595691681s
```
(Note: `#SOs = 2 * #MOs`)

Note: in the future we can take this optimization further by parallelizing the implementation. However, this generation is I/O bound rather than compute bound which means that a straight-forward application of `qiskit.tools.parallel_map` which actually result in worse performance (see also https://github.com/Qiskit/qiskit-terra/issues/7741). One solution could be the use of shared memory, as attempted here: https://github.com/Qiskit/qiskit-terra/pull/7789.
For the time being, performance is greatly improved by this PR and in the near future it is unlikely that we need to generate the second quantized operators for 100+ spin orbitals. At that point storage will also become concern so we will need to support the usage of sparse matrices anyways, which will also improve the performance of this method.

### DIFF
--- a/releasenotes/notes/performance-two-body-second-q-op-08f906951005ebc1.yaml
+++ b/releasenotes/notes/performance-two-body-second-q-op-08f906951005ebc1.yaml
@@ -2,7 +2,7 @@
 features:
   - |
     Improves the performance of the
-    :meth:`qiskit_nature.properties.second_quantization.electronic.integrals.TwoBodyElectronicIntegrals.to_second_q_op`
+    :meth:`qiskit_nature.properties.second_quantization.electronic.integrals.ElectronicIntegrals.to_second_q_op`
     method significantly. Previously, generating the second quantized operators
     for a system size greater than 20 spin orbitals took on the order of minutes
     to hours (depending on the actual size). Now, even system sizes of 50 spin
--- a/releasenotes/notes/performance-two-body-second-q-op-08f906951005ebc1.yaml
+++ b/releasenotes/notes/performance-two-body-second-q-op-08f906951005ebc1.yaml
@@ -2,7 +2,7 @@
 features:
   - |
     Improves the performance of the
-    :class:`qiskit_nature.properties.second_quantization.electronic.integrals.TwoBodyElectronicIntegrals.to_second_q_op`
+    :meth:`qiskit_nature.properties.second_quantization.electronic.integrals.TwoBodyElectronicIntegrals.to_second_q_op`
     method significantly. Previously, generating the second quantized operators
     for a system size greater than 20 spin orbitals took on the order of minutes
     to hours (depending on the actual size). Now, even system sizes of 50 spin
--- a/releasenotes/notes/performance-two-body-second-q-op-08f906951005ebc1.yaml
+++ b/releasenotes/notes/performance-two-body-second-q-op-08f906951005ebc1.yaml
@@ -1,0 +1,9 @@
+---
+features:
+  - |
+    Improves the performance of the
+    :class:`qiskit_nature.properties.second_quantization.electronic.integrals.TwoBodyElectronicIntegrals.to_second_q_op`
+    method significantly. Previously, generating the second quantized operators
+    for a system size greater than 20 spin orbitals took on the order of minutes
+    to hours (depending on the actual size). Now, even system sizes of 50 spin
+    orbitals can be handled in a matter of seconds.
--- a/qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals.py
+++ b/qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals.py
@@ -15,7 +15,6 @@
 from __future__ import annotations
 
 import importlib
-import itertools
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from typing import Generator, Optional, Union
--- a/qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals.py
+++ b/qiskit_nature/properties/second_quantization/electronic/integrals/electronic_integrals.py
@@ -356,32 +356,20 @@ class ElectronicIntegrals(ABC):
         if not np.any(spin_matrix):
             return FermionicOp.zero(register_length)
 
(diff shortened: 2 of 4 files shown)


woodsp-ibm:
So basically this builds out the FermionicOp having first created all the terms rather than building one and then augmenting it term by term. Is there any aspect of the performance around how the methods used before that could considered in the future though perhaps always building it out in one go may faster than how things were done.

mrossinek:
Yes I believe that the previous way of building out the `FermionicOp` will always be less performant, simply because of the overhead involved with constructed 5 instances of `FermionicOp` per iteration, compared to a single instantiation in the new case. Obviously, this single instantiation now can take considerably longer than the previous single ones, but as you can see from the speedup achieved here, this is not a problem.

I think the old implementation was a bit more instructive/educational in terms of how the integral indices determine which creation/annihilation operator to use, but this is not worth the performance deficiencies.

releasenotes/notes/performance-two-body-second-q-op-08f906951005ebc1.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    Improves the performance of the
+    :class:`qiskit_nature.properties.second_quantization.electronic.integrals.TwoBodyElectronicIntegrals.to_second_q_op`
```
woodsp-ibm, line 5:
```suggestion
    :meth:`qiskit_nature.properties.second_quantization.electronic.integrals.TwoBodyElectronicIntegrals.to_second_q_op`
```
this is a method being referenced right

releasenotes/notes/performance-two-body-second-q-op-08f906951005ebc1.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    Improves the performance of the
+    :meth:`qiskit_nature.properties.second_quantization.electronic.integrals.TwoBodyElectronicIntegrals.to_second_q_op`
```
woodsp-ibm, line 5:
```suggestion
    :meth:`qiskit_nature.properties.second_quantization.electronic.integrals.ElectronicIntegrals.to_second_q_op`
```
The link still does not resolve in the html. I think this is because its a method of the parent and parent methods do not get duplicated docs down in the sub-classes (my recollection is that we avoided doing this due to the sheer amount of methods that we saw in cases where this was done in the past). Anyway I think changing the link to the parent should sort things out as it is documented there and I think the text works ok still like this right.

woodsp-ibm (approved):

Issue #46, opened 2021-02-12
Implement the SeniorityZeroTransformer


mrossinek:
### What is the expected feature?

Add the `SeniorityZeroTransformer`.

- [ ] actual implementation
- [ ] unittests
- [ ] documentation

PR #640, opened 2022-04-19
Seniority zero transformer

jschuhmac:
### Summary
First implementation of the SeniorityZeroTransformer (#46).
The implementation is based on to the `ActiveSpaceTransformer` and therefore the `SeniorityZeroTransformer` can be used in the same way.
Monkey patching is used to create the second quantization operators in the restricted formalism.

### Details and comments
The `SeniorityZeroTransformer` transforms the one- and two-body integrals of the (standard) unrestricted Hartree-Fock method into the restricted formalism. In the restricted formalism the electrons are only treated in pairs, halving the number of qubits required, but at the cost of losing some accuracy by neglecting the single-electron excitations.

Building the second quantized operators based on one- and two-body integrals is currently only implemented for fermions (the `second_q_ops()` method for properties in `qiskit_nature.properties.second_quantization.electronic` returns `FermionicOp`). For the moment the creation of the operators in the restricted formalism is done with monkey patching.

Next to the integrals (stored in the `ElectronicEnergy` property) also the `ParticleNumber` property is transformed. The number of particles is left unchanged, only the `second_q_ops` method is updated to return an operator that counts two particles per excitation.

Example how to use the SeniorityZeroTransformer:
```python
from qiskit_nature.drivers.second_quantization import ElectronicStructureDriverType, ElectronicStructureMoleculeDriver
from qiskit_nature.transformers.second_quantization.electronic import SeniorityZeroTransformer

molecule = Molecule(geometry=[["H", [0.0, 0.0, 0.0]], ["H", [0.0, 0.0, 0.8]]], charge=0, multiplicity=1)
driver = ElectronicStructureMoleculeDriver(molecule, basis="sto3g", driver_type=ElectronicStructureDriverType.PYSCF)

es_problem = ElectronicStructureProblem(driver, transformers=[SeniorityZeroTransformer()])
```

### Discussion
There are still some points open for discussion:
* A `VibrationalOp` is used to implement the electron pair operators, but the `VibrationalOp` is not intended to be used in such a way.
* Instead of monkey patching @mrossinek suggested to use a [Strategy](https://en.wikipedia.org/wiki/Strategy_pattern).
* We only transform the `ElectronicEnergy` and the `ParticleNumber` property. Could this also be done for the other properties, `AngularMomentum`, `Magnetization` and `ElectronicDipoleMoment`?


qiskit_nature/transformers/second_quantization/electronic/active_space_transformer.py
```diff
@@ -367,24 +367,13 @@ def _transform_property(self, prop: Property) -> Property:
...
-                except StopIteration:
-                    break
-
+            for internal_property in iter(prop):
                 try:
                     transformed_internal_property = self._transform_property(internal_property)
+                    if transformed_internal_property is not None:
+                        transformed_property.add_property(transformed_internal_property)
+                    else:
+                        transformed_property.remove_property(internal_property)
```
mrossinek, line 376: A short FYI comment from my side:
The initial implementation was trying to be clever by using `generator.send` to inject the newly transformed property back into the iterated object. However, this does not work as intended when you want to remove properties because sending `None` will not overwrite the existing iterated object.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,242 @@
...
+                except TypeError:
+                    logger.warning(
+                        "The Property %s of type %s could not be transformed!",
+                        internal_property.name,
+                        type(internal_property),
+                    )
+                    continue
+
+            if len(transformed_property._properties) == 0:
+                transformed_property = None
```
mrossinek, line 120: While this is indeed just a copy, I would leave this as such for the time being, because this code will be refactored slightly in the near future anyways :+1:

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,242 @@
...
+            transformed_property = ElectronicEnergy.from_raw_integrals(
+                ElectronicBasis.MO, hr_1, hr_2
+            )
+            transformed_property._shift = prop._shift.copy()
+            transformed_property.nuclear_repulsion_energy = prop.nuclear_repulsion_energy
+            transformed_property.reference_energy = prop.reference_energy
+
+            transformed_property.second_q_ops = partial(  # type: ignore[assignment]
+                second_q_ops_electronic_energy, transformed_property
+            )
```
mrossinek, line 137: This monkeypatching approach is currently a bit of a hack. An alternative to this approach is the use of a Strategy pattern as was already mentioned in the PR description. While this would make things slightly cleaner, it would have the downside of extracting all current `Property.second_q_ops()` implementations into a single "default" strategy.
A transformer which would affect this logic would then ship its own strategy which needs to overwrite the methods for all supported property types.

In the future, adding a new property will then cause all existing strategies to be extended by another implementation. Thus we have a NxM problem with N being the number of properties and M the number of strategies.

I am open to better suggestions of handling this :+1:

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,242 @@
...
+        else:
+            raise TypeError(f"{type(prop)} is an unsupported Property-type for this Transformer!")
+
+        return transformed_property
+
+    def _restricted_electronic_integrals(
+        self, h_1: np.ndarray, h_2: np.ndarray
+    ) -> Tuple[np.ndarray, np.ndarray]:
+        """
+        Transform PyScf integrals to the restricted formalism
```
mrossinek, line 161: Please remove the mention of `PySCF` as these integrals are irrespective of the driver

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,242 @@
...
+        arxiv: 2002.00035, Appendix A, Equations (A10),(A11),(A12)
+        https://arxiv.org/pdf/2002.00035.pdf
```
mrossinek, line 163:
```suggestion
        arxiv:2002.00035, Appendix A, Equations (A10),(A11),(A12)
        https://arxiv.org/abs/2002.00035
```

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,242 @@
...
+                        internal_property.name,
+                        type(internal_property),
+                    )
+                    continue
+
+            if len(transformed_property._properties) == 0:
+                transformed_property = None
+
+        elif isinstance(prop, ElectronicEnergy):
+            h_1 = prop.get_electronic_integral(ElectronicBasis.MO, 1).get_matrix()
```
mrossinek, line 123: I think it would be good if we raise or log a warning if the alpha and beta integrals are not identical. I.e. `if not np.allclose(....get_matrix(0), ....get_matrix(1))`

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,242 @@
...
+            h_2 = prop.get_electronic_integral(ElectronicBasis.MO, 2).get_matrix()
+
+            hr_1, hr_2 = self._restricted_electronic_integrals(h_1, h_2)
+
+            transformed_property = ElectronicEnergy.from_raw_integrals(
+                ElectronicBasis.MO, hr_1, hr_2
+            )
```
mrossinek, line 130: Technically, `hr_2` has the shape of 1-body terms but will be injected into a `TwoBodyElectronicIntegrals` object here. There is currently no better way of doing this unfortunately, because of the internal structure of the `IntegralProperty` and `ElectronicIntegrals` classes.
These will be refactored in the near future to resolve this temporary workaround.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,242 @@
...
+    hr_2 = self.get_electronic_integral(ElectronicBasis.MO, 2).get_matrix()
+
+    num_modes = hr_1.shape[0]
+
+    label_m = lambda x: "I" * x + "-" + "I" * (num_modes - x - 1)
+    label_p = lambda x: "I" * x + "+" + "I" * (num_modes - x - 1)
+
+    # Create pair creation and annihilation operators for the spatial orbitals
+    b = [VibrationalOp(label_m(i), num_modes=num_modes, num_modals=1) for i in range(num_modes)]
+    b_dag = [VibrationalOp(label_p(i), num_modes=num_modes, num_modals=1) for i in range(num_modes)]
```
mrossinek, line 206: You should be able to use sparse labels here:
```suggestion
    # Create pair creation and annihilation operators for the spatial orbitals
    b = [VibrationalOp(f"-_{i}", num_modes=num_modes, num_modals=1) for i in range(num_modes)]
    b_dag = [VibrationalOp(f"+_{i}", num_modes=num_modes, num_modals=1) for i in range(num_modes)]
```

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # Build the operators
```
mrossinek, line 208:
```suggestion
    # TODO: refactor to construct `VibrationalOp` from a list of labels rather than summing multiple instances
    # Build the operators
```
This can be improved significantly in terms of performance in a similar style to #638. However, the `VibrationalOp` is no longer on par with the `FermionicOp` so I am not sure if this is already possible now. Please leave this TODO comment in the code for the time being.

Note: you will need to reformat the file to avoid the line length limit being exceeded.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,242 @@
...
+    """Returns the second quantized particle number operator.
+    The actual return-type is determined by `qiskit_nature.settings.dict_aux_operators`.
+
+    Returns:
+        A `list` or `dict` of `VibrationalOp` objects.
+    """
+    label = lambda x: "I" * x + "N" + "I" * (self._num_spin_orbitals // 2 - x - 1)
+
+    op = VibrationalOp(
+        [(label(o), 2.0) for o in range(self._num_spin_orbitals // 2)],
```
mrossinek, line 234:
```suggestion
    op = VibrationalOp(
        [(f"N_{o}", 2.0) for o in range(self._num_spin_orbitals // 2)],
```

releasenotes/notes/seniority-zero-transformer-5d36d9658b46db8f.yaml
```diff
@@ -0,0 +1,4 @@
+features:
+  - |
+    Add SeniorityZeroTransformer to solve a chemistry problem in second quantization
+    with the restricted Hartree-Fock formalism.
```
mrossinek, line 4: Maybe we can add a short usage example here

qiskit_nature/transformers/second_quantization/electronic/__init__.py
```diff
@@ -27,13 +27,12 @@
...
 
 from .active_space_transformer import ActiveSpaceTransformer
 from .freeze_core_transformer import FreezeCoreTransformer
+from .seniority_zero_transformer import SeniorityZeroTransformer
 
-__all__ = [
-    "ActiveSpaceTransformer",
-    "FreezeCoreTransformer",
-]
+__all__ = ["ActiveSpaceTransformer", "FreezeCoreTransformer", "SeniorityZeroTransformer"]
```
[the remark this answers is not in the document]
jschuhmac: I intended to do it as you suggest, but the line was reformatted by `make black`.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
jschuhmac: I have not used sparse labels for the following reasons:
* For the `VibrationalOp` the sparse label has the notation `-_{mode_index}*{modal_index}` and I think it is not really nice to explicitly state the modal index when we do not use the modal functionality of the `VibrationalOp`.
* Using labels like `+_0*0` triggers the warning `Number of raising and lowering operators do not agree for mode 0 in label +_0*0` as the `VibrationalOp` would like to conserve the number of excitations in each mode. The warning is not triggered when I specify the label explicitly.

I will add a comment in the code to clarify why I did not use sparse notation.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
jschuhmac: Only `+` and `-` are allowed in the sparse label of a `VibrationalOp`. However, I can use the sparse labels to do the following
```python
op = VibrationalOp([(f"+_{o}*0 -_{o}*0", 2.0) for o in range(self._num_pin_orbitals // 2)], ...)
```

qiskit_nature/properties/grouped_property.py
```diff
@@ -70,6 +70,24 @@ def add_property(self, prop: Optional[T_co]) -> None:
...
+            prop: the name or type of the property to remove from the group.
+
+        Returns:
+            The removed property (or None).
+        """
+        name: str
+        if isinstance(prop, str):
+            name = prop
+        elif isinstance(prop, Property):
+            name = prop.name
```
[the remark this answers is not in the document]
jschuhmac: No, it is not needed. I put it there because e.g. for the `DipoleMoment` property `prop.name` and `DipoleMoment.__name__` are different. I will remove these two lines and directly pass the name of a property to `remove_property`,
```python
transformed_property.remove_property(internal_property.name)
```

qiskit_nature/transformers/second_quantization/electronic/__init__.py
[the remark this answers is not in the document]
mrossinek: If you put a comma after the last item in the list, black will format it the way I suggested it above :+1:

qiskit_nature/properties/grouped_property.py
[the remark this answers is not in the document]
mrossinek: Ah that's a good point yes!

qiskit_nature/transformers/second_quantization/electronic/__init__.py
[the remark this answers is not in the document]
jschuhmac: Ah, thanks. Now I have noticed the comma.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,246 @@
...
+    ElectronicBasis,
+)
+from qiskit_nature.properties.second_quantization.electronic.types import GroupedElectronicProperty
+from qiskit_nature.transformers.second_quantization import BaseTransformer
+
+logger = logging.getLogger(__name__)
+
+
+class SeniorityZeroTransformer(BaseTransformer):
+    """The Seniority-Zero restriction."""
```
woodsp-ibm, line 45: This class docstring/description of the transformer, will end up in the API reference as documentation of this. I think for many users this may not be well known and could do with a description, link(s) to paper(s) or other useful material. I see some method has a link but having things in the main docstring here will be helpful in an overall description. And perhaps examples/use cases of when this would be useful to consider using.

The second q ops creation, this now requires a use different qubit mappers right, and the fermionic ones, such as JW, should not be used. If so I think this aspect should be noted here about the change to the operators in that respect.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def __init__(self):
+        self._num_modes: int = None
+
+    def transform(
```
woodsp-ibm, line 50: Are there any ordering/usage constraints when using this with other Transformers? As it stands the problem takes a list of transformers and they are simply applied one after the other in list order.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,246 @@
...
+
+            for j in range(self._num_modes):
+                if i != j:
+                    hr_1[i, j] = h_2[i, i, j, j]
+                    hr_2[i, j] = 2 * h_2[i, j, j, i] - h_2[i, j, i, j]
+
+        return hr_1, hr_2
+
+
+def second_q_ops_electronic_energy(self: ElectronicEnergy) -> ListOrDictType[VibrationalOp]:
```
woodsp-ibm, line 190: Should this be public in this class, I think it would be better annotated as private. As a function, at first sight, having self as the first parameter is confusing, but then I see above this is monkey patched into an class instance where self is the target of the monkey patching. Since I also see that monkey patching has already been commented on, hopefully something better is figured out in general.

releasenotes/notes/seniority-zero-transformer-5d36d9658b46db8f.yaml
woodsp-ibm: Also saying it 'solves a chemistry problem` seems perhaps a bit of a mismatch. Saying it transforms a problem, and being a bit more specific about the nature of the problems its applies too, to make computation/solving a chemistry problem simpler/smaller/easier, or whatever the advantages of using it are, would be preferable I think. Then the example would follow on from there.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
mrossinek: Yes, this transformer would need to be applied after any active space or freeze core transformer. Once we get a particle hole transformer, I believe that will be incompatible to be matched with this one. @pbark can correct me on this if needed

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
mrossinek: What do you think about the strategy pattern suggestion? It is not perfect but the best idea I have had so far in order to remove the need for monkeypatching

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
woodsp-ibm: Should we be thinking about say a priority attribute such that the list could be processed in priority order too? That way the user would not have to be so concerned but can just say I want these transformations and the list order plus priority order deals with things. They would still need of course to know that when using SZ transform that they will need to alter the mapping according. But dealing with list order would be more intrinsic. At present, if they put it up front in the list, what would happen - do we check anywhere or do things just fail somehow later?

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
woodsp-ibm: The monkey patching is really in effect just dynamically creating a subclasses of ElectronicEnergy etc. Could they be created as permanent sub-classes under the `vibrational' side and just used where the transform changes it from an ElectronicEnergy in the `electronic` side to an equivalent there with just the second_q_ops overidden (that is all I see monkey patched). Its always going to alter from one property type to another right.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
mrossinek: Yes this sounds like a good idea. But I think for now we can merge this as is and implement a priority ordering in a separate PR. If you agree I can open an issue to track this :+1:

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
mrossinek: @pbark and I were discussing with @jschuhmac offline, that the naming `vibrational` is more than bad in this scenario. We are using this operator in this scenario because we want to describe _pairs of fermions_ which, together, behave like a boson. When we originally wrote the `VibrationalOp`, we voted against the name `BosonicOp` for reasons which I do not recall in all detail. However, now, I believe we need to revisit that naming choice.

All that said, I am actually wondering whether we want to make the `SeniorityZero` more of a first-class citizen by taking an approach of implementing two methods along the lines of `to_fermionic_op()` (for the current "default" behavior) and `to_vibrational_op()` (with an updated name depending on what we discuss above) in all properties directly.
Alternatively, we could even envision providing some sort of argument to `second_q_ops()` or `to_second_q_op()` (this needs to be unified, but this is another issue).
I am not sure whether this scales because I do not recall all of the details which the `ParticleHoleTransformer` will bring (i.e. does it affect only data or also the strategy) but I wanted to at least put this option up for discussion.

woodsp-ibm: I don't fully follow - as you say the transform makes the operator bosonic in nature by working with pairs, so I do not quite see how a to_fermionic_op applies. I was more thinking here that the properties were being transformed into something bosonic in nature and hence properties reflecting that so that the operator built would be bosonic. Or maybe you are thinking that properties have these methods for fermionic or bosonic and can implement one of the other as in your option to second_q_ops. Is there an option here though where some property could reasonably emit either?

My recollection of PH transform as it used to be was just a change in the data along with an energy extracted where the energy is the HF energy so that UCCSD first iteration will emit 0 once this is done.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,319 @@
...
+                "property, which is required by this transformer!"
+            )
+        electronic_energy = cast(ElectronicEnergy, electronic_energy)
+
+        one_body = electronic_energy.get_electronic_integral(ElectronicBasis.MO, 1)
+        if not np.allclose(one_body.get_matrix(0), one_body.get_matrix(1)):
+            raise QiskitNatureError(
+                "One-body integrals for alpha and beta electrons are not identical, "
+                "which is required by this transformer."
+            )
```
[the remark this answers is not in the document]
mrossinek: I am not happy with the current design of the `ElectronicIntegrals` and the `IntegralProperty` classes. Simple operations like the one you suggest here should not be as "difficult" to realize. I am gathering some feedback on the properties for a range of improvements during the next few months and will add this suggestion to the list :+1:

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,319 @@
...
+                "Only `GroupedElectronicProperty` objects can be transformed by this Transformer, "
+                f"not objects of type, {type(grouped_property)}."
+            )
+
+        electronic_energy = grouped_property.get_property(ElectronicEnergy)
+        if electronic_energy is None:
+            raise QiskitNatureError(
+                "The provided `GroupedElectronicProperty` does not contain a `ElectronicEnergy` "
+                "property, which is required by this transformer!"
+            )
```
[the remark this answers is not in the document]
mrossinek: I think erroring is the correct way to go. Especially with the seniority zero case, if we were to only log a warning/error and try to proceed without the transform being the applied, the problem would immediately be twice the size and a user may not notice and wait a significantly long time to obtain a result (if at all).

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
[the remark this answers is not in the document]
woodsp-ibm: Ok, sounds good - and maybe add a note too with your list so we remember to revisit this code here.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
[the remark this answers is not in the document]
woodsp-ibm: Yes, in this case, it seems the right thing to do as also the qubit mapper, that was presumably chosen on the assumption this would be in effect, needs to match as well.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,322 @@
...
+        elif isinstance(prop, DriverMetadata):
+            transformed_property = prop
+        elif isinstance(prop, SecondQuantizedProperty):
+            transformed_property = None
+        elif isinstance(prop, ElectronicBasisTransform):
+            transformed_property = None
+
+        else:
+            raise TypeError(f"{type(prop)} is an unsupported Property-type for this Transformer!")
```
woodsp-ibm, line 217: Since some properties are dropped, why not just warn and drop this too rather than make it an error. At least this way things would be a bit more future proof in that if additional properties are added for the problem that this will continue to work - though of course maybe it can be extended to better support such a property rather than dropping it.

qiskit_nature/transformers/second_quantization/electronic/seniority_zero_transformer.py
```diff
@@ -0,0 +1,322 @@
...
+        elif isinstance(prop, ParticleNumber):
+            transformed_property = prop
+
+            transformed_property.second_q_ops = partial(  # type: ignore[assignment]
+                _second_q_ops_particle_number, transformed_property
+            )
+        elif isinstance(prop, DriverMetadata):
+            transformed_property = prop
+        elif isinstance(prop, SecondQuantizedProperty):
+            transformed_property = None
```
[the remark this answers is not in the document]
jschuhmac: In the latest commit I have updated the docstring and the way how the unsupported properties are handled. The `_transform_property` method now returns `None` whenever it receives an unsupported property, and a warning is logged when a property is removed from the `GroupedProperty`.

mrossinek:
Over time, this PR has gotten rather stale. This is mostly because the problems we hit during its implementation led to an improved code design reflected by the last 2 releases (0.5 and 0.6). However, in order to really implement this in the best way possible, we should address https://github.com/Qiskit/qiskit-nature/issues/1143 first.

However, for the time being, the priority of such an implementation is rather low. Thus, I will close this (now stale) PR.

PR #642, opened 2022-04-20
Cache Terra/Aer wheel files in the CI

manoelmarques:
### Summary
Github Actions Cache key is based on Terra/Aer latest `main` commit id. If there is no hit, Terra and or Aer will build wheel files and store them in the caches with that commit id as key. Next time, if no new commit to `main` happened for Terra/Aer, the wheel files saved to cache will be used instead of being built.


.github/workflows/main.yml
```diff
@@ -55,9 +58,40 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: aer-cache
+          key: aer-cache-${{ matrix.os }}-${{ matrix.python-version }}-${{ env.AER_HASH }}-${{ env.CACHE_VERSION }}
+        if: ${{ !startsWith(github.ref, 'refs/heads/stable') && !startsWith(github.base_ref, 'stable/') }}
+      - uses: ./.github/actions/install-main-aer
         with:
           os: ${{ matrix.os }}
+          cache-path: aer-cache
+          cache-hit: ${{ steps.aer-cache.outputs.cache-hit }}
```
mrossinek, line 94: Is it not possible to put all of this logic (which is now duplicated 3 times) into a common place like e.g. the old `install-main-dependencies/action.yml`?

manoelmarques: The cache action needs to trigger a post action step in the main workflow in order to save the cache. When I tried initially to have inside the action, it didn't trigger it but I will try again since I changed other things after the last time I tested.

mrossinek (approved):
Thanks @manoelmarques!

I see that it worked now. Out of curiosity: what was the necessary change?

manoelmarques:
Correctly testing for the cache hit inside the action

PR #646, opened 2022-04-25
Replace #570. VQE factories.

### LINKED ISSUE #384: VQE Factory method improvements

MarcDrudis:
**Summary**
This pull request replaces and hence closes #570, which in turn was trying to fix #384.

**Details and Comments**

I have modified the getters and setters from vqe factories in order for them to access the vqe properties instead of modifying their own attributes. The reason is that if we want to use the deprecated setters they need to modify the attributes from vqe. With the current implementation we were able to set different values to vqe_ucc_factory.ATTRIBUTE and vqe_ucc_factory._vqe.ATTRIBUTE.

I have also implemented some unittests to verify that the setters and getters behave correctly.

I am not sure what would be the best course of action for the ansatz, since it is still an attribute from the factory and the vqe. I don't see a reason why we wouldn't depracate it as well, but for now its setter sets both attributes to the same value.


qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -81,108 +86,159 @@ def __init__(
...
-    @property
+        self._vqe = VQE(
+            quantum_instance=quantum_instance,
+            optimizer=optimizer,
+            initial_point=initial_point,
+            gradient=gradient,
+            expectation=expectation,
+            include_custom=include_custom,
+            callback=callback,
+            **kwargs,
```
mrossinek, line 100: Since this factory is made specifically to handle the `ansatz` and `initial_point`, I believe it would be good if we deprecate all other explicit keyword arguments of this initializer in favor of `**kwargs`. That way we do not need to keep this method as much in sync with the underlying `VQE`.

The ansatz and initial_state should still remain configurable via this initializer, but should immediately set the `self._vqe` attributes.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -208,29 +264,19 @@ def get_solver(  # type: ignore[override]
...
-            gradient=self.gradient,
-            expectation=self.expectation,
-            include_custom=self.include_custom,
-            callback=self.callback,
-            **self._kwargs,
-        )
-
-        return vqe
+        if self.ansatz is None:
+            self._vqe.ansatz = UCCSD()
```
mrossinek, line 268: As a further explanation to my comment above, consider the following case:
```python
factory = VQEUCCFactory()
factory.minimum_eigensolver.ansatz = UCCD()
vqe = factory.get_solver()
print(type(vqe.ansatz))  # UCCSD
```
The reason the above will still return `UCCSD` is because in these commented lines you are checking against `self.ansatz`, but the user may have overwritten the ansatz on the solver object directly. Obviously this defeats the purpose of this factory but by always using the solver internal attribute, such edge cases can be minimized.

mrossinek (commented):
Other than these few comments, this already LGTM!
Thank you for adding the tests and continuing @iuliazidaru's work!

Please make sure to format your files using `make black` and check that pylint and mypy are happy (`make lint` and `make mypy`, respectively).

You also need to sign the CLA (see comment above).

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/numpy_minimum_eigensolver_factory.py
```diff
@@ -87,9 +107,13 @@ def get_solver(
...
-        npme = NumPyMinimumEigensolver(filter_criterion=filter_criterion)
-        return npme
+            self._minimum_eigensolver.filter_criterion = filter_criterion
+        return self._minimum_eigensolver
 
     def supports_aux_operators(self):
         return NumPyMinimumEigensolver.supports_aux_operators()
+
+    @property
+    def minimum_eigensolver(self):
```
woodsp-ibm, line 117: I guess this should be typehinted for a return of NumPyMinimumEigensolver since we want to allow the user full access to the specific type and its properties here right . I saw the same method ion the VQE factories and was initially thinking perhaps this could be a method on the base class. But we want the user of VQE factories to be able to get to the VQE and alter things like include_custom etc., and not work with it through the common MinimumEigenSolver interface that would presumably be the only type that would make sense for such a function in the base class to say it returns.

@mrossinek With this change we are doing there is subtle behavior change in that before a factory could emit number of solvers and they were independent. Now its more one factory one solver. I think its ok, its more where we want to be in having the user have direct access to the algorithm and the user changing it directly via its properties rather than us having all the indirects here.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
MarcDrudis: Maybe I implemented the change after you checked the code but in the setter of the factory I set both the factory ansatz and the vqe ansatz to avoid this kind of error.

mrossinek: But I am using the setter of the `VQE` directly:
```python
factory.minimum_eigensolver.ansatz = UCCD()
```
This will not/cannot affect the `factory.ansatz` object.

MarcDrudis: Since self.ansatz returns self.minimum_eigensolver.ansatz, the condition will be False and that line of code won't execute. However, maybe self.ansatz should return self._ansatz instead, in which case we would have this problem.

mrossinek: No, what I meant is that you should not need `self._ansatz` at all and instead always use `self.minimum_eigensolver.ansatz`

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
MarcDrudis: So, I believe that in order to implement this we need to loose the positional arguments that can use in the current implementations. Code like VQEUCCFactory(some_quantum_instance) wouldn't work anymore and should be replaced by VQEUCCFactory(quantum_instance = some_quantum_instance). Otherwise the constructor will think of some_quantum_instance as an ansatz.
 It is a small sacrifice, but it would mean that some people would need to change their code.

mrossinek: Since all arguments of the `VQEUCCFactory` are keyword arguments, we do not give any guarantee for code to work when used via positional arguments. As such this should not be a problem. That we internally use the factories via positional arguments internally (e.g. during unittests) needs to be updated, yes

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
MarcDrudis: Oh, you mean I can just delete the attribute from the factory? Got it!

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
MarcDrudis: Perfect, I have already changed the unittests and they work fine.

MarcDrudis:
I just wanted to add a comment to say that I have changed all the unittests that were using a UVCC Factory because they were using quantum_instance positionally, and now it can only be used as a keyword argument.
This change was necessary and from now on all uses of that factory will have to take this into account.

woodsp-ibm:
> using quantum_instance positionally, and now it can only be used as a keyword argument.

Can we not keep these arguments but deprecate them so that existing code continues to work. This is what we would normally do such that a user has time to change their code over to whatever the new way would be. We never enforced the arguments to be used only as keywords.

mrossinek:
> Can we not keep these arguments but deprecate them so that existing code continues to work

We actually outright remove these arguments in this PR. You can still provide them via the factories `__init__` method but they will all be gathered inside of `kwargs`. In that way all `VQE.__init__` arguments (which are not explicitly handled by the factory, like ansatz, initial_state and initial_point) are now "automagically" in-sync with the VQE.

woodsp-ibm:
The thing is that the removal of these - even though they can still be supplied via kwargs, breaks existing code that is using them positionally - as witnessed by changes to notebooks, unit tests etc. Maybe we can make quantum_instance Optional and default it to None, same for include_custom i.e change to None from False (False is VQE's default for that anyway). Can we not find a transition path where existing code, that may have used these positionally as we ourselves did, continues to work albeit that is raises deprecation warnings when it detects them being used (i.e. non-None values). We would want to change the notebook to avoid the warnings, but the notebook (and other code users may have done) ought not be broken.

MarcDrudis:
> The thing is that the removal of these - even though they can still be supplied via kwargs, breaks existing code that is using them positionally - as witnessed by changes to notebooks, unit tests etc. Maybe we can make quantum_instance Optional and default it to None, same for include_custom i.e change to None from False (False is VQE's default for that anyway). Can we not find a transition path where existing code, that may have used these positionally as we ourselves did, continues to work albeit that is raises deprecation warnings when it detects them being used (i.e. non-None values). We would want to change the notebook to avoid the warnings, but the notebook (and other code users may have done) ought not be broken.

We could keep all of the arguments we had previously and raise deprecation warnings, or we could check for the type of ansatz (and if it is a QuantumInstance pass it to the VQE as a QuantumInstance and raise a warning).

MarcDrudis:
I tried using type checks to find an alternative that fulfills both requirements (would be usable by keyword arguments or positionally). However, one of the many issues this aproach has is that inputs such as VQEFactory(quantum_instance, ansatz=ansatz) wouldn't work (assuming that ansatz is the first positional argument in the constructor).
The only reasonable solution I can come up with is to leave the arguments of the constructor exactly as they were, but if some of them has to be deprecated, to raise a deprecation error.
However, as @mrossinek pointed out, until we finally remove it from the code, every time we make a change to VQE, we will need to modify this class as well.

woodsp-ibm:
> The only reasonable solution I can come up with is to leave the arguments of the constructor exactly as they were, but if some of them has to be deprecated, to raise a deprecation error.

Yes, in order to allow existing code to work then this should be the route we take unless its impossible and a breaking change has to be done.

> until we finally remove it from the code, every time we make a change to VQE, we will need to modify this class as well.

VQE itself should only be making changes the same way, i.e. deprecating as necessary. We can choose to only support any new function via the new way. So hopefully this does not happen in practice. But yes, the main idea of discontinuing the way things were done was to avoid this going forwards, especially as we expect changes around the algorithms as they are adapted over to use primitives.

MarcDrudis:
Do not merge the branch yet, since I still need to write the deprecation warnings for the case where we use the arguments positionally.

MarcDrudis:
The way I dealt with the constructor is that  I kept the same order of arguments but I have changed the name of the positional arguments we want to deprecate to **_deprecate. This way if the user was using them as keyword arguments, they will still be passed to vqe as **kwargs. Otherwise, if used positionally, they will raise a deprecation warning and update kwargs[**] = **_deprecated.
I believe this makes it clear for the user that he should not be using them positionally, we won't have to make any changes to the factory even if we make changes to vqe, and the old code will keep working.

MarcDrudis:
Okay, what I ended up doing is to create a decorator in deprecate.py that takes care of the particular problem we have.
If you try a script that looks something like this you should see that it works but raises deprecation warnings:
```python
from qiskit_nature.algorithms.ground_state_solvers.minimum_eigensolver_factories import VQEUCCFactory
from qiskit import Aer
import numpy as np

quantum_instance = QuantumInstance(backend=Aer.get_backend("aer_simulator_statevector"))

VQEUCCFactory(quantum_instance, None, np.array([1,2,3]), include_custom=True)
#This raises warnings for quantum_instance, optimizer (which is the second postitional argument) but not for initial_point (since it is not being deprecated), nor for include_custom (because it has been passed as a keyword argument).

```

I believe this implementation is much more clean and easy to modify.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/numpy_minimum_eigensolver_factory.py
```diff
@@ -45,28 +44,49 @@ def __init__(
         """
         self._filter_criterion = filter_criterion
```
mrossinek, line 45: Do we even still need this private attribute or could this be replaced by `self._minimum_eigensolver.filter_criterion` everywhere?

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -209,29 +277,21 @@ def get_solver(  # type: ignore[override]
...
         ansatz.num_spin_orbitals = num_spin_orbitals
         ansatz.initial_state = initial_state
+        self._vqe.ansatz = ansatz
 
         if isinstance(self.initial_point, InitialPoint):
             self.initial_point.grouped_property = driver_result
             self.initial_point.ansatz = ansatz
-
             # Override the initial_point with the computed array.
             self.initial_point = self.initial_point.to_numpy_array()
```
mrossinek, line 286: I would not overwrite `self.initial_point` here and instead simply have a local `init_point` variable (or similarly named).

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -22,33 +23,62 @@
...
+    """A factory to construct a VQE minimum eigensolver with UVCCSD ansatz wavefunction.
+
+    Note: get_solver will overwrite any value that we directly set onto the vqe for the ansatz.
+    For example:
+    .. code-block:: python
+
+        factory = VQEUCCFactory()
+        factory.minimum_eigensolver.ansatz = UCCD()
+        vqe = factory.get_solver()
+        print(type(vqe.ansatz))  # UCCSD
```
mrossinek, line 53: Needs to be updated to the vibrational case.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -22,33 +23,62 @@
...
+    )
     def __init__(
         self,
-        quantum_instance: QuantumInstance,
-        optimizer: Optional[Optimizer] = None,
-        initial_point: Optional[np.ndarray] = None,
-        gradient: Optional[Union[GradientBase, Callable]] = None,
-        expectation: Optional[ExpectationBase] = None,
-        include_custom: bool = False,
+        quantum_instance_depracated: QuantumInstance = None,
```
mrossinek, line 70: 1. Is this renaming still required with the new decorator?
2. If yes, please fix the naming: `depracated->deprecated` and also align the UCC case (where these are currently not being renamed).
3. Also if yes, you will need to disable the pylint errors here

qiskit_nature/deprecation.py
```diff
@@ -228,6 +228,59 @@ def wrapper(*args, **kwargs):
...
+    """Decorator to convert positional arguments into keyword arguments and warn upon use.
+    .. code-block:: python
+
+        @deprecate_positional_arguments("0.1",(b,d))
+        def function(a,b,c,d,**kwargs):
+            returns a+c+sum(kwargs.values())
+
+        #This two calls would be equivalent
+        function(1,2,3,4) #Would raise deprecation warning for arguments b and d.
+        function(a=1,b=2,c=3,d=4) #Would not raise any warning.
```
mrossinek, line 246:
```suggestion
    """Decorator to convert positional arguments into keyword arguments and warn upon use.
    
    .. code-block:: python

        @deprecate_positional_arguments("0.1", (b, d))
        def function(a, b, c, d, **kwargs):
            returns a + c + sum(kwargs.values())

        # The following two calls would be equivalent
        function(1, 2, 3, 4)  # Would raise deprecation warning for arguments b and d.
        function(a=1, b=2, c=3, d=4)  # Would not raise any warning.
```

qiskit_nature/deprecation.py
```diff
@@ -228,6 +228,59 @@ def wrapper(*args, **kwargs):
...
+            keyword_list = inspect.getfullargspec(func)[0]
+            for i, arg in enumerate(args):
+                kwargs[keyword_list[i]] = arg
+                if keyword_list[i] in kw_pos_deprecated:
+                    msg = (
+                        f"Please, refer from using '{keyword_list[i]}' positionally to"
+                        f"construct a VQEFactory. As of version {version} this are no longer"
+                        " properties of the factories but are passed to the VQE via keyword"
+                        " arguments. The option to pass the arguments positionally will be "
+                        "removed no sooner than 3 months after release."
```
mrossinek, line 271: Please remove hard-coded mentioning of VQE and reword this to be more inline with the other messages raised by this module.

mrossinek (changes requested):
Besides these comments please also:
- [ ] remove `.pylint.d` files which you accidentally commited (you can also include this in the `.gitinore` if you want)
- [ ] fix the CI :upside_down_face:

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
MarcDrudis: No, I left this one by mistake

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
MarcDrudis: Do you mean you would make a copy of self.initial_point, modify it and then pass it to the vqe?

qiskit_nature/deprecation.py
MarcDrudis: Good catch

MarcDrudis: By the way, what version should I indicate in the decorator?

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
mrossinek: Simply write `initial_point = self.initial_point.to_numpy_array()` and then use this variable below, instead. This would obviously need to handle the `else` clause to this if-statement, too.

qiskit_nature/deprecation.py
MarcDrudis: The message now looks like this:
```python
f"{func_name}: {keyword_list[i]} is no longer a positional argument "
f"as of version {version} and will be removed no sooner "
"than 3 months after the release. Instead use it as a keyword argument"
```
And we can add some additionla text if wanted.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
MarcDrudis: Got it

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/numpy_minimum_eigensolver_factory.py
MarcDrudis: Removed

mrossinek:
@woodsp-ibm The CI is failing with a rather cryptic error message: https://github.com/Qiskit/qiskit-nature/runs/6402562094?check_suite_focus=true#step:8:802

Do you know how to resolve this? I have not seen this one before.

woodsp-ibm:
> The CI is failing with a rather cryptic error message:
> Do you know how to resolve this? I have not seen this one before.

It seems that a `print()` statement was left in the factory. This generated an empty output cell (just a new line in it) that seems to end up causing the error that was seem around raw. Removing it and it works fine - the notebooks as checked in built fine it was only when the html is built from the re-run version that the issue arose. I pushed a changed to remove the print so it should go through now ok.

In regards of the two notebooks, 03 and 04, they should be updated as they emit deprecation warnings. They still work which is of course the intent of the deprecation, but they should be switched over to the new way i.e. keyword versus positional argument, so they show the right way of doing things now.

MarcDrudis:
> > The CI is failing with a rather cryptic error message:
> > Do you know how to resolve this? I have not seen this one before.
> 
> It seems that a `print()` statement was left in the factory. This generated an empty output cell (just a new line in it) that seems to end up causing the error that was seem around raw. Removing it and it works fine - the notebooks as checked in built fine it was only when the html is built from the re-run version that the issue arose. I pushed a changed to remove the print so it should go through now ok.
> 
> In regards of the two notebooks, 03 and 04, they should be updated as they emit deprecation warnings. They still work which is of course the intent of the deprecation, but they should be switched over to the new way i.e. keyword versus positional argument, so they show the right way of doing things now.

Thank you for fixing the CI! I fixed the tutorials. However, I am wondering, is there a good way of seeing that the tutorials are throwing deprecation warnings? I have to check manually but I am sure there is an easier way.

woodsp-ibm:
> However, I am wondering, is there a good way of seeing that the tutorials are throwing deprecation warnings? I have to check manually but I am sure there is an easier way.

I do not believe we gather these in CI at all - @manoelmarques rather they just end up in the notebook output when run which ends up in the html. Now we do capture deprecations warnings from the logs when running tests and CI gathers these up and  you can see this in build details section. For any tests we run, where we know we are testing deprecated code, its our policy to suppress the warning for that test knowing it will occur. The intent is to only have messages come out from places we cannot avoid, or places that we have to change in the code. At present the factory testing seems to be emitting a large number of these that need to be supressed - seems that a number of warnings from various places are in the output. Some it maybe code that needs to be updated, some may be from tests that we know will do this and we can therefore ignore them in the test. That deprecations file ought to be checked over before the next release and appropriate remedies enacted in the code.

MarcDrudis:
Hello, I have been taking care of deprecation warnings. One of them brought me an interesting problem.
The warning in question happened in GroundStateEigensolver.evaluate_operators(). Here we were trying to get some attributes of the solver:
` # try to get a QuantumInstance from the solver
   quantum_instance = getattr(self._solver.minimum_eigensolver, "quantum_instance", None)
   # and try to get an Expectation from the solver
   expectation = getattr(self._solver.minimum_eigensolver, "expectation", None)`
Note that self._solver is `Union[MinimumEigensolver, MinimumEigensolverFactory]`. Before, this was okay since every subclass of MinimumEigensolverFactory happenned to have quantum_instance and expectation (although we could have happened to create some new factory that didn't need `expectation` and we would have run into an error here). In that sense, we could use MinimumEigensolver instances or their factories interchangebly and we were also avoiding dealing with mypy.
However, now we need to make the difference between these two cases, i.e.:
`if isinstance(self._solver, MinimumEigensolverFactory):
# try to get a QuantumInstance from the solver
quantum_instance = getattr(self._solver.minimum_eigensolver, "quantum_instance", None)
# and try to get an Expectation from the solver
expectation = getattr(self._solver.minimum_eigensolver, "expectation", None)
else:
quantum_instance = getattr(self._solver, "quantum_instance", None)
expectation = getattr(self._solver, "expectation", None)`

This code would work every time since all three instances of MinimumEigensolverFactory {NumpyMinimumEigensolverFactory,VQEUCCFactory,VQEUVCCFactory} happen to have a minimum_eigensolver() method with some subclass of MiniumEigensolver as an output.
However, this creates a new problem since now we are calling MinmumEigensolverFactory.minimum_eigensolver(), which doesn't exist a priori.
Is there a reason against adding a getter for minimum_eigensolver() to the base class MinimumEigensolverFactory? It would look like this:
`
@abstractproperty   
@property
def minimum_eigensolver(self) -> MinimumEigensolver:
    """Returns the solver instance."""
    raise NotImplementedError
`
Thank you for your time.

mrossinek:
The base-class `MinimumEigensolverFactory` should indeed already specify the `minimum_eigensolver` attribute. You can go ahead and add that :+1:

docs/tutorials/03_ground_state_solvers.ipynb
```diff
@@ -339,23 +339,14 @@
...
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
    "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.8.12"
+   "version": ""
```
mrossinek, line 347: I'm not exactly sure if these changes have any significant impact. I'm willing to just let them be merged but we can also arrange for the tutorials to be executed once on a UNIX system to try and minimize the differences here.

qiskit_nature/deprecation.py
```diff
@@ -228,6 +228,61 @@ def wrapper(*args, **kwargs):
...
+    version: str,
+    func_name: str,
+    pre_positional_kw: List[str],
+    additional_msg: Optional[str] = None,
+    stack_level: int = 3,
+) -> Callable:
+    """Decorator to convert positional arguments into keyword arguments and warn upon use.
+    .. code-block:: python
+
+        @deprecate_positional_arguments("0.1", (b, d))
```
mrossinek, line 241: This example appears invalid because `func_name` is not provided

mrossinek: Also, the `(b, d)` will likely not work as intended.. See also next comment.

qiskit_nature/deprecation.py
```diff
@@ -228,6 +228,61 @@ def wrapper(*args, **kwargs):
...
+
+        # The following two calls would be equivalent
+        function(1, 2, 3, 4)  # Would raise deprecation warning for arguments b and d.
+        function(a=1, b=2, c=3, d=4)  # Would not raise any warning.
+
+    Args:
+        version: Version to be used
+        func_name: Name of the function where the deprecation takes place will be used to
+        write the deprecation message.
+        pre_positional_kw: List of arguments of the function before the deprecation.
```
mrossinek, line 253: If `pre_positional_kw` is the list of positional arguments prior to the deprecation, how do you actually specify which arguments are to be deprecated?

test/algorithms/ground_state_solvers/minimum_eigensolver_factories/test_vqe_ucc_factory.py
```diff
@@ -46,40 +47,93 @@ def setUp(self):
...
+            self.assertEqual(getattr(self._vqe_ucc_factory.minimum_eigensolver, prop), cases[1])
+            # Check vqe setter
+            setattr(self._vqe_ucc_factory.minimum_eigensolver, prop, cases[0])
+            self.assertEqual(getattr(self._vqe_ucc_factory, prop), cases[0])
+            self.assertEqual(getattr(self._vqe_ucc_factory.minimum_eigensolver, prop), cases[0])
+
+    def auxiliary_tester_isinstance(self, title: str, prop: str, cases: tuple):
+        """
+        Tests a getter and setter of a given property.Only checks if the type of the property is correct.
+        Args:
```
mrossinek, line 85:
```suggestion
        Tests a getter and setter of a given property. Only checks if the type of the property is correct.
        
        Args:
```

test/algorithms/ground_state_solvers/minimum_eigensolver_factories/test_vqe_ucc_factory.py
```diff
@@ -46,40 +47,93 @@ def setUp(self):
...
+            seed_transpiler=self.seed,
+        )
 
-        self._vqe_ucc_factory = VQEUCCFactory(self.quantum_instance)
+        self._vqe_ucc_factory = VQEUCCFactory(quantum_instance=self.quantum_instance)
+
+    def auxiliary_tester(self, title: str, prop: str, cases: tuple):
+        """
+        Tests the setter and getter of a given property.
+        Args:
```
mrossinek, line 62:
```suggestion
        Tests the setter and getter of a given property.
        
        Args:
```

test/algorithms/ground_state_solvers/minimum_eigensolver_factories/test_vqe_uvcc_factory.py
```diff
@@ -44,39 +45,92 @@ def setUp(self):
...
+            seed_simulator=self.seed,
+            seed_transpiler=self.seed,
+        )
+
+        self._vqe_uvcc_factory = VQEUVCCFactory(quantum_instance=self.quantum_instance)
+
+    def auxiliary_tester(self, title: str, prop: str, cases: tuple):
+        """
+        Tests the setter and getter of a given property.
+        Args:
```
mrossinek, line 60:
```suggestion
        Tests the setter and getter of a given property.
        
        Args:
```

test/algorithms/ground_state_solvers/minimum_eigensolver_factories/test_vqe_uvcc_factory.py
```diff
@@ -44,39 +45,92 @@ def setUp(self):
...
+            self.assertEqual(getattr(self._vqe_uvcc_factory.minimum_eigensolver, prop), cases[1])
+            # Check vqe setter
+            setattr(self._vqe_uvcc_factory.minimum_eigensolver, prop, cases[0])
+            self.assertEqual(getattr(self._vqe_uvcc_factory, prop), cases[0])
+            self.assertEqual(getattr(self._vqe_uvcc_factory.minimum_eigensolver, prop), cases[0])
+
+    def auxiliary_tester_isinstance(self, title: str, prop: str, cases: tuple):
+        """
+        Tests a getter and setter of a given property.Only checks if the type of the property is correct.
+        Args:
```
mrossinek, line 83:
```suggestion
        Tests a getter and setter of a given property. Only checks if the type of the property is correct.
        
        Args:
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -22,34 +23,62 @@
 from qiskit.opflow.gradients import GradientBase
 from qiskit.utils import QuantumInstance
 from qiskit_nature.circuit.library import UVCC, UVCCSD, VSCF
+
+
```
mrossinek, line 27:
```suggestion
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -88,77 +117,105 @@ def __init__(
...
+        self._initial_state = initial_state
         self.initial_point = initial_point if initial_point is not None else VSCFInitialPoint()
-        self.gradient = gradient
-        self.expectation = expectation
-        self.include_custom = include_custom
-        self.ansatz = ansatz
-        self.initial_state = initial_state
-        self.callback = callback
-        self._kwargs = kwargs
+        self._factory_ansatz = ansatz
```
mrossinek, line 123:
```suggestion
        self._ansatz = ansatz
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -88,77 +117,105 @@ def __init__(
...
-
-    @property
+        """DEPRECATED. Use the constructor instead. Setter of the ``include_custom``
+        setting for the ``expectation`` setting."""
+        self.minimum_eigensolver.include_custom = include_custom
+
+    @property  # type: ignore
+    @deprecate_property(
+        "0.4", additional_msg="Use `minimum_eigensolver` and 'solver properties' instead."
+    )
```
mrossinek, line 206: I don't think the `ansatz` should be deprecated. Just like `initial_point` and `initial_state` also remain properties of the factory.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -208,7 +267,7 @@ def get_solver(  # type: ignore[override]
         if initial_state is None:
             initial_state = VSCF(num_modals)
 
-        ansatz = self.ansatz
+        ansatz = self._factory_ansatz
```
mrossinek, line 270:
```suggestion
        ansatz = self._ansatz
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -88,77 +118,108 @@ def __init__(
...
+        self._initial_state = initial_state
         self.initial_point = initial_point if initial_point is not None else HFInitialPoint()
-        self.gradient = gradient
-        self.expectation = expectation
-        self.include_custom = include_custom
-        self.ansatz = ansatz
-        self.initial_state = initial_state
-        self.callback = callback
-        self._kwargs = kwargs
+        self._factory_ansatz = ansatz
```
mrossinek, line 124:
```suggestion
        self._ansatz = ansatz
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/numpy_minimum_eigensolver_factory.py
```diff
@@ -25,48 +28,69 @@
...
-                has the smallest eigenvalue among feasible states. The callable has the signature
-                `filter(eigenstate, eigenvalue, aux_values)` and must return a boolean to indicate
-                whether to consider this value or not. If there is no
-                feasible element, the result can even be empty.
             use_default_filter_criterion: whether to use the transformation's default filter
-                criterion if ``filter_criterion`` is ``None``.
+            criterion if ``filter_criterion`` is ``None``.
+            kwargs: keyword arguments passed to NumpyMinimumEigensolver to construct
+            self._minimum_eigensolver. Note that filter_criterion is now accessed through
+            NumpyMinimumEigensolver.
```
mrossinek, line 48:
```suggestion
            kwargs: keyword arguments passed to NumpyMinimumEigensolver to construct
            the internal `MinimumEigensolver`. Note that filter_criterion is now accessed through
            NumpyMinimumEigensolver.
```

mrossinek (commented):
A few comments but other than this, this PR is shaping up to an end :+1:

qiskit_nature/deprecation.py
MarcDrudis: I access the name of each keyword argument and the ones that are in one list and not the other are the ones to be deprecated.
So, I iterate over pre_positional_kw, and for each one I check if that keyword is in `current_positional_kw = inspect.getfullargspec(func)[0]`. Then I raise a message if it isn't and in both cases add it to kwargs dictionary with I use to call the decorated funtion in the end.

qiskit_nature/deprecation.py
MarcDrudis: I made a change in this function and forgot to change the documentation at some point. Thank you for spotting it.

qiskit_nature/deprecation.py
mrossinek: So would it then not make more sense to call this the `deprecated_positiional_kw`?
And if so, why does one of the cases I saw in your PR include `self` in this list? Surely this should not be deprecated

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
MarcDrudis: Indeed.

qiskit_nature/deprecation.py
mrossinek: Right, thanks for the detailed explanation :+1: 
In that case I would suggets:
- calling this argument: `old_function_arguments` or `original_function_arguments`
- expanding the explanation in this docstring to match your explanation here

MarcDrudis: Done

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -168,18 +219,33 @@ def initial_state(self) -> Optional[QuantumCircuit]:
     @initial_state.setter
     def initial_state(self, initial_state: Optional[QuantumCircuit]) -> None:
         """Setter of the initial state. If ``None`` is passed, this factory will default to using
-        the :class:`~.VSCF`."""
+        the :class:`~.HartreeFock`."""
```
mrossinek, line 222:
```suggestion
        the :class:`~.VSCF`."""
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -88,76 +115,100 @@ def __init__(
...
+        Getter of the ansatz"""
         return self._ansatz
 
     @ansatz.setter
     def ansatz(self, ansatz: Optional[UVCC]) -> None:
-        """Setter of the ansatz. If ``None`` is passed, this factory will default to using the
-        :class:`~.UVCCSD` Ansatz."""
+        """DEPRECATED. Use the constructor instead. Setter of the ``include_custom``
+        Setter of the ansatz. If ``None`` is passed, this factory will default to using the
+        :class:`~.UCCSD` Ansatz."""
```
mrossinek, line 211:
```suggestion
        :class:`~.UVCCSD` Ansatz."""
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -29,27 +30,53 @@
...
         initial_point: Optional[Union[np.ndarray, InitialPoint]] = None,
-        gradient: Optional[Union[GradientBase, Callable]] = None,
-        expectation: Optional[ExpectationBase] = None,
-        include_custom: bool = False,
         ansatz: Optional[UVCC] = None,
         initial_state: Optional[QuantumCircuit] = None,
-        callback: Optional[Callable[[int, np.ndarray, float, float], None]] = None,
         **kwargs,
     ) -> None:
+
```
mrossinek, line 79:
```suggestion
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -29,27 +30,53 @@
...
+    """A factory to construct a VQE minimum eigensolver with UVCCSD ansatz wavefunction.
+
+    Note: get_solver will overwrite any value that we directly set onto the vqe for the ansatz.
+    For example:
+    .. code-block:: python
+
+        factory = VQEUVCCFactory()
+        factory.minimum_eigensolver.ansatz = UVCCD()
+        vqe = factory.get_solver()
+        print(type(vqe.ansatz))  # UVCCSD
```
mrossinek, line 51: Maybe you should point out, how the desired behavior can indeed be achieved:
```python
factory.ansatz = UVCC(excitations="d")
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -29,27 +30,53 @@
...
 
+class VQEUVCCFactory(MinimumEigensolverFactory):
+    """A factory to construct a VQE minimum eigensolver with UVCCSD ansatz wavefunction.
+
+    Note: get_solver will overwrite any value that we directly set onto the vqe for the ansatz.
+    For example:
+    .. code-block:: python
+
+        factory = VQEUVCCFactory()
+        factory.minimum_eigensolver.ansatz = UVCCD()
```
mrossinek, line 49:
```suggestion
        factory.minimum_eigensolver.ansatz = UVCC(excitations="d")
```
The class `UVCCD` does not exist.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -33,23 +38,48 @@
...
 
+class VQEUCCFactory(MinimumEigensolverFactory):
+    """A factory to construct a VQE minimum eigensolver with UCCSD ansatz wavefunction.
+
+    Note: get_solver will overwrite any value that we directly set onto the vqe for both
+    the ansatz and the initial point. For example:
+    .. code-block:: python
+
+        factory = VQEUCCFactory()
+        factory.minimum_eigensolver.ansatz = UCCD()
```
mrossinek, line 52:
```suggestion
        factory.minimum_eigensolver.ansatz = UCC(excitations="d")
```
Same a in the vibrational case.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -33,23 +38,48 @@
...
+        vqe = factory.get_solver()
+        print(type(vqe.ansatz))  # UCCSD
```
mrossinek, line 54: And again, add something like the following:
```python
factory.ansatz = UCC(excitations="d")
```

test/algorithms/ground_state_solvers/minimum_eigensolver_factories/test_vqe_uvcc_factory.py
mrossinek: Also don't overlook the remaining comments from this review.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
MarcDrudis: I actually meant UVCCSD()

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
MarcDrudis: Done

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
mrossinek: Well if you were to write `UCCSD()` there, the example would not show any obvious difference. I simply wanted to correct the `UCCD` example :+1:

woodsp-ibm:
@mrossinek Comment on linking this PR to close that issue https://github.com/Qiskit/qiskit-nature/issues/426#issuecomment-1143802032

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -33,23 +38,60 @@
...
+        print(type(factory.minimum_eigensolver.ansatz)) # UCC
+        vqe = factory.get_solver(problem,qubit_converter)
+        print(type(vqe.ansatz))  # UCCSD (and not UCC nor RealAmplitudes)
+        #Expected way
+        factory = VQEUCCFactory()
+        vqe1 = factory.get_solver(problem,qubit_converter)
+        print(type(vqe1.ansatz))    #UCCSD (default)
+        factory.ansatz = UCC()
+        vqe2 = factory.get_solver(problem,qubit_converter)
+        print(type(vqe2.ansatz))    #UCC
```
mrossinek, line 66: Just to be consistent with our black code formatter even here:
```suggestion
        # Wrong way
        factory = VQEUCCFactory()
        print(type(factory.minimum_eigensolver.ansatz))  # RealAmplitudes (by default)
        factory.minimum_eigensolver.ansatz = UCC()
        print(type(factory.minimum_eigensolver.ansatz))  # UCC
        vqe = factory.get_solver(problem, qubit_converter)
        print(type(vqe.ansatz))   # UCCSD (and not UCC nor RealAmplitudes)
        # Expected way
        factory = VQEUCCFactory()
        vqe1 = factory.get_solver(problem, qubit_converter)
        print(type(vqe1.ansatz))  # UCCSD (default)
        factory.ansatz = UCC()
        vqe2 = factory.get_solver(problem, qubit_converter)
        print(type(vqe2.ansatz))  # UCC
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -29,25 +30,62 @@
...
+        print(type(factory.minimum_eigensolver.ansatz)) # UVCC
+        vqe = factory.get_solver(problem,qubit_converter)
+        print(type(vqe.ansatz))  # UVCCSD (and not UVCC nor RealAmplitudes)
+        #Intended way
+        factoryB = VQEUCCFactory()
+        vqe1 = factoryB.get_solver(problem,qubit_converter)
+        print(type(vqe1.ansatz))    # UVCCSD (default)
+        factoryB.ansatz = UVCC()
+        vqe2 = factory.get_solver(problem,qubit_converter)
+        print(type(vqe2.ansatz))    #UVCC
```
mrossinek, line 63:
```suggestion
        # Wrong way
        factory = VQEUVCCFactory()
        print(type(factory.minimum_eigensolver.ansatz))  # RealAmplitudes (by default)
        factory.minimum_eigensolver.ansatz = UVCC()
        print(type(factory.minimum_eigensolver.ansatz))  # UVCC
        vqe = factory.get_solver(problem, qubit_converter)
        print(type(vqe.ansatz))  # UVCCSD (and not UVCC nor RealAmplitudes)
        #Intended way
        factoryB = VQEUCCFactory()
        vqe1 = factoryB.get_solver(problem, qubit_converter)
        print(type(vqe1.ansatz))  # UVCCSD (default)
        factoryB.ansatz = UVCC()
        vqe2 = factory.get_solver(problem, qubit_converter)
        print(type(vqe2.ansatz))  # UVCC
```

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -88,75 +126,99 @@ def __init__(
...
-        self._include_custom = include_custom
+        """DEPRECATED. Use the constructor instead. Setter of the ``include_custom``
+        setting for the ``expectation`` setting."""
+        self.minimum_eigensolver.include_custom = include_custom
 
     @property
     def ansatz(self) -> Optional[UVCC]:
-        """Getter of the ansatz."""
+        """DEPRECATED. Use ``minimum_eigensolver`` method and solver properties instead.
+        Getter of the ansatz"""
```
mrossinek, line 215: Please update this to match the docstring of `VQEUCCFactory`.
Same for the setter below

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_ucc_factory.py
```diff
@@ -212,6 +291,7 @@ def get_solver(  # type: ignore[override]
         ansatz.num_particles = num_particles
         ansatz.num_spin_orbitals = num_spin_orbitals
         ansatz.initial_state = initial_state
+        self._vqe.ansatz = ansatz
```
mrossinek, line 294: I prefer how you access this in the `VQEUVCCFactory` class via `self.minimum_eigensolver`. Especially since you return it via that property getter.
Please align these two classes as much as possible in this method.

MarcDrudis: Thank you for spotting it.

MarcDrudis:
I wanted to leave a comment saying that since the two classes are almost a copy of one another it can get hard to make changes in both of them. If the classes need to be changed again, specially if that is done before finally removing the deprecated properties, it might be worth it to look into making a superclass that these two can inherit from. In the end the only difference they have is in the get_solver() method.

mrossinek (approved):
Imo this is now good to be merged.

@woodsp-ibm please also take a look :+1:

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -63,101 +99,110 @@ def __init__(
...
+        """
         return self._ansatz
 
     @ansatz.setter
     def ansatz(self, ansatz: Optional[UVCC]) -> None:
-        """Setter of the ansatz. If ``None`` is passed, this factory will default to using the
-        :class:`~.UVCCSD` Ansatz."""
+        """
+        Sets the ansatz of future VQEs produced by the factory.
+        If set to ``None`` it defaults to :class:`UVCCSD`.
```
woodsp-ibm, line 204: Something like `Sets the ansatz to be used, by the factory, to create/set the solver. If None is provided then the factory internally will default to an instance of UVCCSD it creates itself.`

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -29,31 +30,66 @@
...
 
-class VQEUVCCFactory(MinimumEigensolverFactory):
-    """A factory to construct a VQE minimum eigensolver with UVCCSD ansatz wavefunction."""
 
+class VQEUVCCFactory(MinimumEigensolverFactory):
+    """Factory to construct a :class:`VQE` minimum eigensolver with :class:`UVCCSD` ansatz wavefunction.
+
+    Note: Changing the ansatz of the minimum_eigensolver of the factory won't affect
+        the ansatz of new solvers produced by the factory. On top of that the default
+        type for :class:`VQE` is not the same as the factory.
```
woodsp-ibm, line 46: > On top of that the default type for :class:`VQE` is not the same as the factory.

What does this intend to say. I am having a hard time with this bit.

The first part is really saying that any ansatz a user might directly set into VQE via the minumum_eigensolver, will be overwritten by the factory when producing a solver via get_solver, as the factory set's the ansatz via what is defined to the factory here. (Maybe this is clearer than the text that has `won't affect`?)

- - - 

And we also take over initial_point the same way too right as in it should be managed via the factory, similar to the ansatz, as the factory will overwrite whatever might be set in VQE,

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -29,31 +30,66 @@
...
+
+    .. code-block:: python
+
+        # Wrong way
+        factory = VQEUVCCFactory()
+        print(type(factory.minimum_eigensolver.ansatz))  # RealAmplitudes (by default)
+        factory.minimum_eigensolver.ansatz = UVCC()
+        print(type(factory.minimum_eigensolver.ansatz))  # UVCC
+        vqe = factory.get_solver(problem, qubit_converter)
+        print(type(vqe.ansatz))  # UVCCSD (and not UVCC nor RealAmplitudes)
```
woodsp-ibm, line 56: This "Wrong way" is basically just proof, that as the note above says, the factory overwrites any ansatz that the VQE may have. Do we need/want to show that in code here? Maybe if its clear that this is what is being shown - to me it was not. The intended way - is a simple example showing how to use the factory right.

woodsp-ibm (commented):
Looks good. My comments are all around docs with a view to improvements there. I started at the reno and then went up in list of files. So thats the reason I commented in the one I did but I think the comments apply similarly to the other factory too.

manoelmarques (approved):

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -63,101 +99,110 @@ def __init__(
...
+
+    @property  # type: ignore
+    @deprecate_property(
+        "0.4", additional_msg="Use `minimum_eigensolver` and 'solver properties' instead."
+    )
     def quantum_instance(self) -> QuantumInstance:
-        """Getter of the quantum instance."""
-        return self._quantum_instance
+        """DEPRECATED. Use ``minimum_eigensolver`` and solver properties instead.
+        Returns quantum instance."""
```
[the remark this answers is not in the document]
MarcDrudis: Done.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -63,101 +99,110 @@ def __init__(
...
+        return self.minimum_eigensolver.quantum_instance
 
-    @quantum_instance.setter
+    @quantum_instance.setter  # type: ignore
+    @deprecate_property("0.4", additional_msg="Use the constructor instead.")
     def quantum_instance(self, q_instance: QuantumInstance) -> None:
-        """Setter of the quantum instance."""
-        self._quantum_instance = q_instance
+        """DEPRECATED. Use the constructor instead. Sets the quantum instance."""
+        self.minimum_eigensolver.quantum_instance = q_instance
```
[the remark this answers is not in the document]
MarcDrudis: Done.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
```diff
@@ -63,101 +99,110 @@ def __init__(
...
+        """DEPRECATED. Use the constructor instead. Setter of the ``include_custom``
+        setting for the ``expectation`` setting."""
+        self.minimum_eigensolver.include_custom = include_custom
 
     @property
     def ansatz(self) -> Optional[UVCC]:
-        """Getter of the ansatz."""
+        """
+        Gets the ansatz of future VQEs produced by the factory.
+        If value is ``None`` it defaults to :class:`UVCCSD`.
```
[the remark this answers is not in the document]
MarcDrudis: What about:
```
Gets the ansatz from the factory.
This property will only affect ansatz generated from :meth:`get_solver`.
```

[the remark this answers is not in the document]
woodsp-ibm: When I looked at the code it was returning self._ansatz if I recall correctly. That is set just to what the user supplies right - if they supply None, the getter will return None - thats why I was suggesting the 'user provided' bit. Whatever the factory builds out they do not see via this getter - if it returns None they do know the factory will create some default but that cannot be seen via this getter.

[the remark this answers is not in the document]
MarcDrudis: Okay.

qiskit_nature/algorithms/ground_state_solvers/minimum_eigensolver_factories/vqe_uvcc_factory.py
MarcDrudis: Changed.

releasenotes/notes/vqe-factory-improvements-02f411513f11907c.yaml
```diff
@@ -0,0 +1,9 @@
+---
+issues:
```
woodsp-ibm, line 2:
```suggestion
deprecations:
```
This should be under a `deprecations` section - `issues` section is for known bugs i.e. issues that are present in the release that we want to make users aware of that we know are there.

mrossinek (commented):
Besides Steve's comment I believe this looks good to go :+1:

releasenotes/notes/vqe-factory-improvements-02f411513f11907c.yaml
MarcDrudis: Good to know.

releasenotes/notes/vqe-factory-improvements-02f411513f11907c.yaml
```diff
@@ -0,0 +1,9 @@
+---
+deprecations:
+  - |
+    The factories had setter/getter properties that set/get the corresponding property of the solver (VQE).
```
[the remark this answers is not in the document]
MarcDrudis: On it. Also properly referencing some :class: and so on.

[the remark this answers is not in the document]
MarcDrudis: Done!

mrossinek (approved):
I think this LGTM! Thanks, @MarcDrudis for all the work on this!

@woodsp-ibm please also take another final look :slightly_smiling_face:

releasenotes/notes/vqe-factory-improvements-02f411513f11907c.yaml
```diff
@@ -0,0 +1,30 @@
...
+    In order to avoid duplicating all of these attributes, potentially leading to inconsistencies
+    between the attributes of the different classes, these are now deprecated and require you to use
+    the attributes from
+    :attr:`~qiskit_nature.algorithms.VQEUCCFactory.minimum_eigensolver`
+    instead.
+    
+    For this reason, the constructor of
+    :class:`~qiskit_nature.algorithms.VQEUCCFactory` has been changed as well.
+    Now the constructor only has the following positional arguments:
+    * initial_point
```
woodsp-ibm, line 17:
```suggestion

    * initial_point
```
Sphinx does not treat this as a bullet list and needs a blank line to avoid wrapping the text into the paragraph
image

I will also note, that references to VQE etc, in another repo/set of docs, get high highlighted, but since its not all built together the do not end up as links. I think this is ok, not sure if you were aware of that.

Issue #647, opened 2022-04-26
Performance regression of `FermionicOp.reduce`


t-imamichi:
### Environment

- **Qiskit Nature version**: main
- **Python version**:  3.9.12
- **Operating system**: macOS 12.3.1

### What is happening?

`FermionicOp.reduce` of main is significantly slower than that of stable version.
`FermionicOp.normal_ordered` and `FermionicOp.to_normal_order` are slow due to the same reason.

### How can we reproduce the issue?

```python
import random
from timeit import timeit

from qiskit_nature.operators.second_quantization import FermionicOp


def random_op(k, n):
    random.seed(123)
    op = sum(
        (
            FermionicOp("".join(random.choices("+-ENI", k=k)), display_format="dense")
            * random.random()
        )
        for _ in range(n)
    )
    return op


def bench(k, n):
    op = random_op(k, n)
    print(f"{timeit(lambda: op.reduce(), number=1)} sec")

bench(12, 1000)
```

main
```
6.409043849 sec
```

stable (v0.3.2)
```
1.1105954169999999 sec
```

### What should happen?

The method should be as fast as that of the stable version.

### Any suggestions?

I took a performance profile with cProfile. It implies that `FermionicOp.__init__` takes more than 70% of time, especially the following code.
https://github.com/Qiskit/qiskit-nature/blob/286c93e0751b069e3543d88561407e4f297dd9e3/qiskit_nature/operators/second_quantization/fermionic_op.py#L270-L272
This part used be a direct assignment as follows. #607 changed this part into a list comprehension, which is the root cause of the performance regression, I guess.
https://github.com/Qiskit/qiskit-nature/blob/be91c8e9657082d8670fd16fd79d02b565fa9f9c/qiskit_nature/operators/second_quantization/fermionic_op.py#L257

I think we need a fast path to set `_data` without any conversion and internal methods should use the fast path as much as possible.

cprofile graph with main branch
![image](https://user-images.githubusercontent.com/31178928/165312654-bdd03feb-bfc5-4e0c-8df6-46319d8fc9dc.png)

mrossinek:
Closed by #649

PR #649, opened 2022-04-27
Fix issue #647

ikkoham:
### Summary

This PR fixes the performance regression: #647 

### Details and comments

Benchmark result:
main:
```
11.40104109 sec
```
stable/0.3 
```
1.3304943169999999 sec
```
This PR
```
1.3513821619999997 sec
```


qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -229,7 +229,8 @@ def __init__(
             tuple[str, complex],
             list[tuple[str, complex]],
             list[tuple[str, float]],
-            Sequence[tuple[Iterable[tuple[str, int]], complex]],
+            list[tuple[list[tuple[str, int]], complex]],
+            list[tuple[tuple[tuple[str, int], ...], complex]],
```
mrossinek, line 233: To be clear: this change is not breaking because the `Sequence` variant was not released yet. So compared to stable the difference is merely adding:
```
            list[tuple[tuple[tuple[str, int], ...], complex]],
```

mrossinek: I just want to check this. Other than that this looks straight forward to me :+1:

ikkoham: This is not breaking change because this is not released.

ikkoham: This is a bit of a discussion, but this part was changed from list to tuple in #607. In that sense, list would be the old API, so should we deprecate it? (In other words, Iterable was introduced as a general expression that included list and tuple, but in real it is not that broad, so I explicitly reverted in this PR.)

mrossinek: What would be the benefit of explicitly deprecated the list-case?

pbark (approved):
Thanks @ikkoham

qiskit_nature/operators/second_quantization/fermionic_op.py
ikkoham: Well, I guess it's nothing special, just a simplification of the code.

ikkoham: Then let's keep it.

ikkoham:
@pbark @mrossinek Thank you for your review. I'll merge this PR after @mrossinek approves.

PR #659, opened 2022-05-16
Minor fixes to `ActiveSpaceTransformer`

mrossinek:
### Summary

This fixes some minor bugs in the `ActiveSpaceTransformer`.

### DIFF
--- a/qiskit_nature/transformers/second_quantization/electronic/active_space_transformer.py
+++ b/qiskit_nature/transformers/second_quantization/electronic/active_space_transformer.py
@@ -253,6 +253,9 @@ class ActiveSpaceTransformer(BaseTransformer):
         # construct new GroupedElectronicProperty
         grouped_property_transformed = ElectronicStructureDriverResult()
         grouped_property_transformed = self._transform_property(grouped_property)  # type: ignore
+        grouped_property_transformed.molecule = (
+            grouped_property.molecule  # type: ignore[attr-defined]
+        )
 
         return grouped_property_transformed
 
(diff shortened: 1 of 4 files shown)


qiskit_nature/transformers/second_quantization/electronic/active_space_transformer.py
```diff
@@ -363,28 +364,16 @@ def _transform_property(self, prop: Property) -> Property:
         Raises:
             TypeError: if an unexpected Property subtype is encountered.
         """
-        transformed_property: Property
+        transformed_property: Property = None
```
woodsp-ibm, line 367: I would think this instance var ought to be annotated as Optional now

releasenotes/notes/fix-as-trafo-grouped-property-iteration-bca2096817b1e497.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    Some minor issues in the :class:`~qiskit_nature.transformers.second_quantization.electronic.ActiveSpaceTransformer`
+    were fixed. Namely:
+
+    - iterated properties are no longer deep-copied resulting in non-transformed ones actually being dropped rather than carried on
```
woodsp-ibm, line 7: Can a simple unit test be done to assert the behavior now is/remains as its now expected.

mrossinek: On second thought after my initial "thumb up": adding such a test now would be slightly artificial because the `ASTransformer` actually supports mapping all currently existing properties. This fix is a "backport" from the #640 PR because it also removes some other edge cases in the iteration behavior. I will check again tomorrow with Julian if we can create a unittest for one of those edge cases without needing the Seniority Zero transformer :+1:

mrossinek: I added a simple test. It asserts that the `ElectronicBasisTransform` gets updated correctly which it did not prior to this PR (i.e. said test will fail on current `main`). This is does not cover the entirety of this PRs behavioral change but should be sufficient until we have a transformer which can drop certain properties during the transformation.

pbark (approved):
LGTM! Thanks

woodsp-ibm (approved):

PR #663, opened 2022-05-20
Fix CI PSI4 crash on Ubuntu

manoelmarques:

### DIFF
--- a/.github/actions/install-psi4/action.yml
+++ b/.github/actions/install-psi4/action.yml
@@ -11,7 +11,7 @@
 # that they have been altered from the originals.
 
 name: 'Install PSI4'
-description: 'Installs Miniconda with PSI4'
+description: 'Installs conda PSI4'
 inputs:
   os:
     description: 'OS'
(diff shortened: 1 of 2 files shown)


.github/actions/install-psi4/action.yml
```diff
@@ -25,11 +25,21 @@ runs:
     - run : |
         source "$CONDA/etc/profile.d/conda.sh"
         conda activate psi4env
-        if [ "${{ inputs.os }}" == "windows-2019" ]; then
-          conda install -y psi4=1.5 python=${{ inputs.python-version }} -c conda-forge -c psi4/label/dev
+        if [[ "${{ inputs.os }}" == "windows-2019" ]]; then
+          release=1.5
+          echo "installs psi4 release $release"
+          conda install -y psi4=$release python=${{ inputs.python-version }} -c conda-forge -c psi4/label/dev
           psi4 --test
```
mrossinek, line 32: Why is this being run on Windows?

mrossinek (commented):
It actually looks like Psi4 dropped support for Python 3.7 but added support for 3.10: https://github.com/psi4/psi4/pull/2455

.github/actions/install-psi4/action.yml
manoelmarques: This has been like that since last year release 1.5. It is the required install on Windows using conda-forge.
I run a quick test afterwards to make sure all the dlls load correctly

manoelmarques:
OK, I was able to still make Ubuntu python 3.7 work with PSI4 1.5.

I believe we should merge this PR to have everything back to working and then I will explore PSI4 running on python 3.10 in another PR

mrossinek (approved):

PR #670, opened 2022-05-26
Set main_property_name through the constructor of BaseProblem

a-matsuo:
### Summary
Set `main_property_name` thought  the constructor of `BaseProblem` instead of setting in each sub class. 

Based on this discussion https://github.com/Qiskit/qiskit-nature/pull/620#discussion_r881292144 .

- [x] `ElectronicStructureProblem`
- [x] `VibrationalStructureProblem`

### DIFF
--- a/qiskit_nature/problems/second_quantization/base_problem.py
+++ b/qiskit_nature/problems/second_quantization/base_problem.py
@@ -38,13 +38,14 @@ class BaseProblem(ABC):
         self,
         driver: BaseDriver,
         transformers: Optional[List[BaseTransformer]] = None,
+        main_property_name: Optional[str] = None,
     ):
         """
 
         Args:
             driver: A driver encoding the molecule information.
             transformers: A list of transformations to be applied to the driver result.
-
+            main_property_name: A main property name for the problem
         """
 
         self.driver = driver
@@ -56,7 +57,7 @@ class BaseProblem(ABC):
         self._grouped_property: Optional[GroupedSecondQuantizedProperty] = None
         self._grouped_property_transformed: Optional[GroupedSecondQuantizedProperty] = None
 
-        self._main_property_name: str = ""
+        self._main_property_name: str = main_property_name
 
     @property  # type: ignore[misc]
     @deprecate_property(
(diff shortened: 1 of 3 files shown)


woodsp-ibm:
It looks good to me. If a third party happened to subclass BaseProblem and use the old way, the code would still work as the instance var is the same name, so setting it directly still works. (Your LatticeProblem with the code in the PR as it current stands still works right!)

Since we really need the sub-class to set the main property name I would almost argue that it should not have a default value. But that would require more change to do that, so at present I think what you have done here suffices.

This could be merged before the LatticeProblem PR and the problem there updated and then merged - or I guess you marked this as draft and listed the LatticeProblem in here as you prefer to merge the LatticeProblem first and then change its problem class via this PR to keep the changes together. I think I am fine either way.

a-matsuo:
Actually, I was thinking this to be merged after the LatticeProblem PR. I'am fine either way as well. If we merge this PR first, I need to change the LatticeProblem PR based on this change. Do you have have any preference? @mrossinek

woodsp-ibm:
@a-matsuo The only reason I was thinking the LatticeProblem PR might be updated/merged after this was merged that this PR is easy to review, seems what we agreed etc and looks good already - its just sat waiting for the LatticeProblem PR to be merged right, when it could be approved/merged already I think and the LatticeProblem fixed up afterwards. But as I said I am fine either order, but the LatticeProblem still needs a final review.

mrossinek (approved):
I agree, it does not really make a difference in which order we apply these changes. I'm going to approve this as is and let you take the decision :+1: 
I don't think we need a release note for this, do we? @woodsp-ibm

woodsp-ibm:
As I noted above ` If a third party happened to subclass BaseProblem and use the old way, the code would still work as the instance var is the same name, so setting it directly still works.` Since this is the case, and since its unlikely anyone has even done this, I think this is fine
 without a release note.

a-matsuo:
@woodsp-ibm @mrossinek Thank you for reviewing! I will modify the LatticeModelProblem PR based on this change.

PR #675, opened 2022-05-29
Update 05_Sampling_potential_energy_surfaces.ipynb

omarcostahamido:
- the word `dof` only appears once and without the proper explanation next to it (even though one can infer from the rest of the notebook).
- this problem was raised by Y Zheng (dran.z) on crowdin [here](https://crowdin.com/translate/qiskit-docs/9674/en-zhcn?filter=basic&value=0#272798).


woodsp-ibm (approved):

mrossinek (approved):
Thanks, again!

PR #678, opened 2022-05-31
Remove long-deprecated Objects

mrossinek:
### Summary
Prior to releasing 0.4, we should remove these long-deprecated objects.

### Details and comments
I purposefully left some other (also long-deprecated objects; e.g. QMolecule, WatsonHamiltonian and the bosonic bases), since the stack is going to undergo a re-organization anyways, at which point we can drop them more "cleanly". And this way, users have a slightly longer migration time :+1:


woodsp-ibm:
In order to support legacy HDF5 files we wanted to keep around QMolecule though right but move it and make it more a private class to support that rather than being public as it once was.

mrossinek:
> In order to support legacy HDF5 files we wanted to keep around QMolecule though right but move it and make it more a private class to support that rather than being public as it once was.

Yes, but given that we want to refactor the stack, I don't see a reason for migrating the `QMolecule` to a private class now, if we are going to migrate it anyways. Or would you prefer to already do this?

woodsp-ibm (approved):

PR #680, opened 2022-06-01
Add Bogoliubov transform

kevinsung:
### Summary
Adds a circuit to perform the Bogoliubov transform.

### Details and comments
Uses the Givens rotation strategy to implement a general single-particle basis-change, which can either conserve particle number or not. The circuits are similar to the ones used to prepare Slater determinants and fermionic Gaussian states, but those ones are optimized for state preparation from a computational basis state, while the Bogoliubov transform performs the general unitary transformation.

References:
[arXiv:1711.05395](https://arxiv.org/abs/1711.05395)
[arXiv:1603.08788](https://arxiv.org/abs/1603.08788)


kevinsung:
Test failures seem unrelated.

qiskit_nature/circuit/library/__init__.py
```diff
@@ -68,6 +77,8 @@
 
 from .initial_states import FermionicGaussianState, HartreeFock, SlaterDeterminant, VSCF
 
+from .primitives import BogoliubovTransform
```
woodsp-ibm, line 80: I am not sure about `primitives` - to me this could end up confusing with the Qiskit primitives which to me are different and are more about executing circuits than the circuit themselves. Do you have others in mind? It could go into the root of the circuit library otherwise - if others are also 'transforms' maybe that's a more appropriate name?

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
```diff
@@ -141,6 +141,8 @@ def __init__(
         Raises:
             ValueError: transformation_matrix must be a 2-dimensional array.
             ValueError: transformation_matrix must have shape (n_orbitals, 2 * n_orbitals).
+            ValueError: transformation_matrix does not describe a valid transformation
+                of fermionic ladder operators
```
woodsp-ibm, line 145: I see the ValueError that gets raised does explain why its not considered valid. Which is presumably it fails to meet the specials constraints as referenced by the parameter docstring. This is the first part of the message in the ValueError but I wonder here if it said it does not meet the constraints required of the matrix it would tie more into the docstring of the parameter which notes this need and refers to the detail

I will note that when sphinx builds the docs the class and constructor docstring are merged to become one. So this original text `This matrix must satisfy special constraints; see the main body of the docstring of this class.` might be a little confusing when looking at the resultant html. It might be better/easier to simply say something like `This matrix must satisfy the constraints as detailed above.` (You can see the html if you unzip the documentation.zip asset that is created by the build where the assets are listed and can be downloaded via build Summary

qiskit_nature/circuit/library/primitives/bogoliubov_transform.py
```diff
@@ -0,0 +1,265 @@
...
+            validate: Whether to validate the inputs.
+            rtol: Relative numerical tolerance for input validation.
+            atol: Absolute numerical tolerance for input validation.
+            circuit_kwargs: Keyword arguments to pass to the QuantumCircuit initializer.
+
+        Raises:
+            ValueError: transformation_matrix must be a 2-dimensional array.
+            ValueError: transformation_matrix must have orthonormal rows.
+            ValueError: transformation_matrix does not describe a valid transformation
+                of fermionic ladder operators
```
woodsp-ibm, line 157: Similar comment to above - and I think its the same constraint right though the docstring of the parameter here does not say similar

qiskit_nature/utils/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+==============================================
+
+.. currentmodule:: qiskit_nature.utils
+
+
+.. autosummary::
+   :toctree:
+"""
+
+from qiskit_nature.utils.linalg import (
```
woodsp-ibm, line 24: linalg also has this as a public function `fermionic_gaussian_decomposition_jw` is there a reason we do not have it here too?

Maybe this is too premature but assuming this utils grows and we have different modules in here with different types of functions would it be better to have a Linear Algebra section and list the functions as a table, like we have for the classes elsewhere, where it links over to the detail. As it stands the functions come out together on the one page so it will get busy very quickly

docs/apidocs/qiskit_nature.utils.rst
```diff
@@ -0,0 +1,18 @@
+﻿
```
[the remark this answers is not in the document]
kevinsung: Sure. This file was automatically generated by Sphinx so I wasn't sure if I should edit it.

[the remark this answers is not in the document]
kevinsung: Done.

qiskit_nature/circuit/library/__init__.py
kevinsung: I've moved it to the top level.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
kevinsung: I've updated the docstring. I think the reference to the class docstring looks good in the html and is clearer than your suggestion. Let me know if you'd like me to change it.

qiskit_nature/circuit/library/primitives/bogoliubov_transform.py
kevinsung: Done.

qiskit_nature/utils/__init__.py
kevinsung: I don't want to expose `fermionic_gaussian_decomposition_jw` because its output is somewhat opaque. It's better thought of as an internal helper function.

It makes sense to have `linalg` be a separate module and documented as such. But I would advocate for it being at the top level, i.e. `qiskit_nature.linalg` because this mirrors what Numpy and Scipy have, so it would be more easily discovered by users. What do you think? If you disagree then I can just it a new module under `utils`, but in my opinion this nested structure is unnecessarily convoluted.

docs/apidocs/qiskit_nature.utils.rst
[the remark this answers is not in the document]
woodsp-ibm: When adding a folder that needs to be documented, I have done this enough that I just create one of these, that way Sphinx uses it and will not auto-create a default one. Editing it is fine though - its just something sphinx needs and it did it itself since it wasn't provided it! Thx.

qiskit_nature/utils/__init__.py
woodsp-ibm: I think things are ok in terms of structure. I.e utils sub-package and linalg.py module. I can imagine utils gathering different useful modules. I am not sure we have enough linalg that we want to expose things as a sub-package. Its not that we do this like numpy/scipy where its core function where they have a bunch and it makes sense; for us its more supporting logic right that a normal user of our API doing Nature will probably never go to, though they can, I think its mainly for developers of Nature code right.

I was more looking at the html output and the way how we normally do this. Here for instance is [Terra utils sub-package documentation](https://qiskit.org/documentation/apidoc/utils.html). As you can see the functions are in a table (as we normally do on these pages with classes). And they are separated into separate tables as a logical partitioning rather than throwing them all into one big one. That's all I meant, where the page says these are utilities or whatever and then there is a section header saying linear algebra with the two functions listed under it so they end up formatted as a table that has the summary of function there and link to the full doc (just like the output you see elsewhere where we list things)

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
woodsp-ibm: I don;t know this is how it comes out where the class docstring and constructor docstring are merged such that the args end up right below what you are referring to - and saying see main body of the docstring of this class, while correct, I am not sure that readers of the html necessarily see it that way - after all if you look there its right above where they need to look.

image

qiskit_nature/utils/__init__.py
kevinsung: Done.

qiskit_nature/circuit/library/initial_states/fermionic_gaussian_state.py
kevinsung: Ok, I've taken your suggestion.

qiskit_nature/utils/testing.py
```diff
@@ -0,0 +1,28 @@
...
+def random_antisymmetric_matrix(dim: int) -> np.ndarray:
+    """Return a random antisymmetric matrix.
+
+    Args:
+        dim: The width and height of the matrix.
+
+    Returns:
+        The sampled antisymmetric matrix.
+    """
+    mat = np.random.randn(dim, dim) + 1j * np.random.randn(dim, dim)
```
woodsp-ibm, line 27: Usually with random aspects, especially when used in testing, we often want a reproducible entity/result. Can we swithc this over, like the rest of Nature, to use the algorithm_globals random, then seeding that, as a common place. gives us that possibility. The syntax is equivalent since its backed by a numpy generator instance, so
```
from qiskit.utils import algorithm_globals

mat = algorithm_globals.random.randn(.... etc 
```

woodsp-ibm:
As this adds a new feature, as well as introduces `utils`  with functions we are making public, this needs a release note too.

qiskit_nature/utils/testing.py
kevinsung: Right, the function should take the seed as input. I've done this following the same style as Terra, e.g. https://github.com/Qiskit/qiskit-terra/blob/89c4f6eb5c56f49ec5149e5be39c77d36b182b4e/qiskit/quantum_info/operators/random.py

qiskit_nature/circuit/library/bogoliubov_transform.py
```diff
@@ -0,0 +1,269 @@
...
+                Should be either :math:`N \times N` or :math:`N \times 2N`.
+            qubit_converter: A QubitConverter instance.
+            validate: Whether to validate the inputs.
+            rtol: Relative numerical tolerance for input validation.
+            atol: Absolute numerical tolerance for input validation.
+            circuit_kwargs: Keyword arguments to pass to the QuantumCircuit initializer.
+
+        Raises:
+            ValueError: transformation_matrix must be a 2-dimensional array.
+            ValueError: transformation_matrix must have orthonormal rows.
```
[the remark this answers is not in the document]
kevinsung: It's actually checked in the `N x 2N` case too. There the check is more specific but it does imply that the rows are orthonormal, so a separate orthonormality check is unnecessary.

qiskit_nature/utils/linalg.py
```diff
@@ -0,0 +1,176 @@
...
+    result = target.copy()
+    for i, slice_i in enumerate(slices):
+        result[slice_i] *= matrix[i, i]
+        for j, slice_j in enumerate(slices):
+            if j != i:
+                result[slice_i] += target[slice_j] * matrix[i, j]
+    return result
+
+
+def givens_matrix(a: Union[complex, float], b: Union[complex, float]) -> np.ndarray:
```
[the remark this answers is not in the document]
kevinsung: I don't know what you mean by "encompasses". I think floats should be allowed here.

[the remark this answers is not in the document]
kevinsung: Oh interesting, mypy seems fine with this. I didn't know that would work, it seems like it should be a TypeError since floats are not instances of complex.

qiskit_nature/circuit/library/bogoliubov_transform.py
```diff
@@ -0,0 +1,269 @@
...
+        rtol: float = 1e-5,
+        atol: float = 1e-8,
+        **circuit_kwargs,
+    ) -> None:
+        r"""
+        Args:
+            transformation_matrix: The matrix :math:`W` that specifies the coefficients of the
+                new creation operators in terms of the original creation operators.
+                Should be either :math:`N \times N` or :math:`N \times 2N`.
+            qubit_converter: A QubitConverter instance.
```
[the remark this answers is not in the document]
kevinsung: Done.

qiskit_nature/utils/linalg.py
```diff
@@ -0,0 +1,176 @@
...
+        b: A complex number representing the second row entry
+
+    Returns:
+        The Givens rotation matrix.
+    """
+    # Handle case that a is zero
+    if np.isclose(a, 0.0):
+        cosine = 1.0
+        sine = 0.0
+        phase: Union[complex, float] = 1.0
```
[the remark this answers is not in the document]
kevinsung: Done.

qiskit_nature/circuit/library/__init__.py
```diff
@@ -18,8 +18,14 @@
 
 .. currentmodule:: qiskit_nature.circuit.library
 
+.. autosummary::
+   :toctree: ../stubs/
+   :nosignatures:
+
+   BogoliubovTransform
```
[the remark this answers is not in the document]
kevinsung: Originally I did do that with the title "primitives" (as in algorithmic primitives) but @woodsp-ibm said the name was confusing and suggested simply having it at the top level. Do you have any ideas for the name?

qiskit_nature/utils/linalg.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def givens_matrix(a: Union[complex, float], b: Union[complex, float]) -> np.ndarray:
+    """Compute the Givens rotation to zero out a row entry.
+
+    Returns a 2x2 unitary matrix G that satisfies
+    ```
+    G * [a  b]^T= [0  r]^T
+    ```
+    where `r` is a complex number.
```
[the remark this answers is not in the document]
kevinsung: Done.

qiskit_nature/utils/linalg.py
```diff
@@ -0,0 +1,176 @@
...
+                result[slice_i] += target[slice_j] * matrix[i, j]
+    return result
+
+
+def givens_matrix(a: Union[complex, float], b: Union[complex, float]) -> np.ndarray:
+    """Compute the Givens rotation to zero out a row entry.
+
+    Returns a 2x2 unitary matrix G that satisfies
+    ```
+    G * [a  b]^T= [0  r]^T
```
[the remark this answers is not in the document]
kevinsung: Obsolete.

qiskit_nature/circuit/library/bogoliubov_transform.py
```diff
@@ -0,0 +1,269 @@
...
+                    current_matrix = apply_matrix_to_slices(
+                        current_matrix, givens_mat, [target_index, target_index - 1]
+                    )
+
+    # convert left rotations to right rotations
+    for givens_mat, (i, j) in reversed(left_rotations):
+        givens_mat = givens_mat.T.conj()
+        givens_mat[:, 0] *= current_matrix[i, i]
+        givens_mat[:, 1] *= current_matrix[j, j]
+        new_givens_mat = givens_matrix(givens_mat[1, 0], givens_mat[1, 1])
```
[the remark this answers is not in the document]
kevinsung: I don't think there's anything else here worth extracting as a utility.

qiskit_nature/utils/testing.py
```diff
@@ -0,0 +1,51 @@
...
+        dim: The width and height of the matrix.
+        seed: The pseudorandom number generator or seed. Should be an
+            instance of `np.random.Generator` or else a valid input to
+            `np.random.default_rng`
+
+    Returns:
+        The sampled antisymmetric matrix.
+    """
+    rng = parse_random_seed(seed)
+    mat = rng.standard_normal((dim, dim)) + 1j * rng.standard_normal((dim, dim))
```
woodsp-ibm, line 50: Since this is under testing I can't see the rationale for the complexity here - it makes more seeds etc for the user to deal with to get things reproducible. I suggested using the algorithm_globals random generator for consistency with code elsewhere. It would be random by default but can be seeded. Its something users already are accustomed to using across the repos here. Having some exposed public API `parse_random_seed`, whose function is not immediately apparent by name is just adding to users complexity. If you feel that this matrix method truly needs to have an option to be independent of the common generator that we use, I would suggest  rather than pass seed in let the user pass a generator. And by default if None, use algorithm globals. Though honestly my preference would be to keep it simple and not have that option, ie. just use algorithm globals - it can always be added later in a backwards compatible way should it ever be reqd.

kevinsung: > If you feel that this matrix method truly needs to have an option to be independent of the common generator that we use

I truly feel that the correct way to implement this function is to allow the seed to be passed. This allows the function behavior to be reproduced independently of some global state. The closest analogues of this function in Qiskit are the functions in [`qiskit.quantum_info.operators.random`](https://github.com/Qiskit/qiskit-terra/blob/89c4f6eb5c56f49ec5149e5be39c77d36b182b4e/qiskit/quantum_info/operators/random.py), and I have made my implementation consistent with that.

> I would suggest rather than pass seed in let the user pass a generator.

 The seed _is_ allowed to be a generator. Again, this is the same behavior as [`qiskit.quantum_info.operators.random`](https://github.com/Qiskit/qiskit-terra/blob/89c4f6eb5c56f49ec5149e5be39c77d36b182b4e/qiskit/quantum_info/operators/random.py)

> Having some exposed public API parse_random_seed, whose function is not immediately apparent by name is just adding to users complexity.

Ok, I'll remove it from the public API.

woodsp-ibm: > I truly feel that the correct way to implement this function is to allow the seed to be passed. 

Ok, what you linked defines its own global level default rng in that module and all methods use that by default. Otherwise they use a generator as passed or build a local one using the seed. I can accept an implementation done the same way though I guess I question the notion of calling it a testing module where to me potentially that degree of independence would be less important.

kevinsung: Done.

qiskit_nature/circuit/library/__init__.py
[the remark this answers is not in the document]
mrossinek: Hm I see. I just find it a bit confusing that something called a `Tranform` is apparently a `QuantumCircuit`. How can you reason this? Maybe we can find a better name here?

qiskit_nature/utils/random.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Returns:
+        The sampled antisymmetric matrix.
+    """
+    rng = parse_random_seed(seed)
+    mat = rng.standard_normal((dim, dim)) + 1j * rng.standard_normal((dim, dim))
+    return mat - mat.T
+
+
+def random_quadratic_hamiltonian(
```
mrossinek, line 56: I do not see the need for such a specific method to be exposed here.

qiskit_nature/circuit/library/__init__.py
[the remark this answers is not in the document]
mrossinek: Those initial _states_ are indeed circuits because e.g. the `HartreeFock` case will add `X` gates to a circuit which will place the qubits into `|1>` where we want to initialize a particle (after mapping).

But I see your point about the Fourier transform. Would it even make sense to add a module called `transforms` then?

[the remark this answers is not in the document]
kevinsung: I think `transforms` is both too vague and too specific. It's vague because it's not well-defined, and it's too specific because in practice it will probably only encompass circuits whose common name ends with the word "transform". This is why I prefer the more general term `primitives` for algorithmic primitives. I actually think it's fine to leave it at the top level and worry about creating submodules when we have more examples.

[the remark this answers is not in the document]
mrossinek: Okay in that case let's leave it where it is right now.
Primitives will not be an option due to a name clash in the Qiskit ecosystem which will cause too much confusion.

docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
```diff
@@ -432,7 +497,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.6"
+   "version": "3.6.8"
```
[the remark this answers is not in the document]
kevinsung: I actually do run them with a newer kernel, and only switch it back before committing because my kernel is named which breaks the CI. It's the same issue as before...

qiskit_nature/utils/random.py
kevinsung: This is why I originally named this file `testing.py` because this method is used for testing. But @woodsp-ibm asked to rename the file. @woodsp-ibm I've added another function here; are you okay with changing it back to `testing.py`?

releasenotes/notes/bogoliubov-transform-885eb6442172c76e.yaml
```diff
@@ -0,0 +1,9 @@
+---
+features:
+  - |
+    Adds :meth:`~.BogoliubovTransform`.
```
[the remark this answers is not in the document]
kevinsung: Fixed.

qiskit_nature/utils/random.py
kevinsung: @mrossinek You mean that you don't think these functions should be exposed at the `qiskit.utils` level, right? In that case, I propose we create a new module `qiskit.utils.testing` and expose them there. What do you think? @woodsp-ibm thoughts?

kevinsung:
The test failure is due to https://github.com/Qiskit/qiskit-nature/issues/686. It can be worked around by manually editing the notebook file but that quickly becomes tedious.

qiskit_nature/utils/random.py
woodsp-ibm: In Aqua we had random matrix generating functions and you could use them how you wanted whether that was in unit testing, experimenting with how a different algorithm behaved (which I guess is testing to some degree) or whatever. Question is do you name this/these for what they are or a what you think someone would do with them. A testing module could end up being a  hodge-podge of different things all used for "testing". Current I see its just used for unit testing - maybe rather than create this here for now just move it into the unit test or a support module there. If someone needs similar they can copy/use it from there and if we have enough demand it could be made part of public API rather than integrate it into the public API already.

kevinsung: These functions are used in several different files throughtout the repo, so they should definitely live in a high-level location. Personally I use functions like these quite a lot so I would prefer them to be part of the public API, especially since I see no downside.

kevinsung: >From what I can see the though parse_random_seed is still there. You expressed you wanted to be able to seed these independently. That I can accept, though we never did this before. My preference would be if seed is None that these run off the algorithm_globals random generator - similar to the quantum_info random ones you showed where they use a common generator if no seed is given - otherwise use some local one via the given seed

In [`qiskit.quantum_info.operators.random`](https://github.com/Qiskit/qiskit-terra/blob/89c4f6eb5c56f49ec5149e5be39c77d36b182b4e/qiskit/quantum_info/operators/random.py) the "global" genorator is simply `np.random.default_rng()`. My code also defers to `np.random.default_rng()` when the seed is None, so the behavior is equivalent. I simply factored out the function `parse_random_seed` because the function code was duplicated across the different random generation functions. As you requested, I removed it from the public API.

>In terms of (re-)moving them from here - they could still be there in test folder so I was not proposing getting rid of them entirely so you can still use them. As to part of public API is something we have to support/maintain them, document them, people translate docs etc versus just being there in unit test area so it extra effort/responsibility etc around these if we chose to have them there.

If they are in the test folder I can't import them from `qiskit_nature` which makes them inconvenient to use. I don't anticipate these functions causing a noticeable maintenance burden, and I don't mind taking responsibility for maintaining them.

Since you don't seem to be strongly opposed to the way I have things here, I ask that we please just keep it as it is.

kevinsung: Ok, done. @mrossinek @woodsp-ibm are you all okay with how things are now?

kevinsung:
#686 has been closed without being addressed. There's no point in my fixing the notebook kernel until it has been finalized. So I need you to _first approve_ this PR, and then I will fix the notebook, and then it can be merged.

qiskit_nature/utils/random.py
mrossinek: > If they are in the test folder I can't import them from qiskit_nature which makes them inconvenient to use.

Once the day comes that you may need to use them inside of `qiskit_nature`, we can promote them up. For now, since they are only being used in testing, they should go there.

However, my main point was that I do not see the need for the random `QuadraticHamiltonian`. This is a _very_ specific method. Once we have that, why don't we have methods for other operators? Having methods to generate random matrices is fine if you really have such a high demand for it, but having a wrapper purely around `QuadraticHamiltonian` I do not see the point of.

kevinsung: >Once the day comes that you may need to use them inside of qiskit_nature, we can promote them up. For now, since they are only being used in testing, they should go there.

Most functions and classes exposed by Qiskit Nature are exposed because they are useful and not because they are used elsewhere in the library. There is already precedent in Qiskit Terra for exposing functions for generating random objects.

>Once we have that, why don't we have methods for other operators?

We should have such methods, they are actually very useful. Probably many existing tests could be simplified using them.

Anyway, I don't want to argue anymore so I've moved the functions out of `qiskit_nature`.

kevinsung:
@mrossinek @woodsp-ibm I've addressed all your comments. Is there anything else? Again, I can fix the notebook-related CI failure once this PR is approved.

woodsp-ibm:
It seems fine to me. As far as I can see all that I was discussing with you has been resolved now. The only last observation I might have is for the release note where it says

> See the "Quadratic Hamiltonians and Slater determinants" tutorial for a demonstration of how to use this class

where it might be a nice improvement for the reader were that to be a link to the tutorial itself, E.g this is one there now https://qiskit.org/documentation/nature/tutorials/10_lattice_models.html that you can base the link on. Since release_notes is in nature folder you may be able to do a relative url i.e. tutorials/11_quadratic_....minants.html which could be seen to work in the docs built here ahead of them being published.

kevinsung:
@woodsp-ibm I've added think. Could you approve the PR now?

woodsp-ibm:
The link should reference the .html file that is the result of the docs creation, and what is published out there on qiskit.org, not the original notebook itself (FYI: Each html is built out with a reference back to the notebook in github here so users can go get that if they want)

kevinsung:
@woodsp-ibm I tested that the link works in the locally built docs, so sphinx must be handling it nicely.

qiskit_nature/circuit/library/bogoliubov_transform.py
```diff
@@ -0,0 +1,273 @@
...
+    method of the :class:`~.QuadraticHamiltonian` class.
+
+    Currently, only the Jordan-Wigner Transformation is supported.
+
+    References:
+        - `arXiv:1711.05395`_
+        - `arXiv:1603.08788`_
+
+    .. _arXiv:1711.05395: https://arxiv.org/abs/1711.05395
+    .. _arXiv:1603.08788: https://arxiv.org/abs/1603.08788
```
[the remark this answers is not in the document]
kevinsung: Yes, this is the reference I intended. The implementation here uses the unitary decomposition from this reference.

mrossinek (dismissed):
I think we can merge this now.

Awaiting a final approval from @woodsp-ibm before doing so :+1:

woodsp-ibm (approved):
Looks good to me - thx Kevin.

PR #684, opened 2022-06-04
Bopes excited states

### LINKED ISSUE #683: Add support to ExcitedStatesSolver in the BOPESSampler
### What should we add?

BOPES currently evaluates the Born-Oppenheimer Potential Energy Surface (energy as a function of nuclear coordinates) based on a GroundStateSolver and different properties for the sampling of the nuclear coordinates.
It is mostly a wrapper to the solve() method of the GroundStateSolver which sets up initial points for the variational algorithms.
Currently it supports only the ground state BOPES.

ExcitedStatesSolver like NumPyMinimumEigensolver or QEOM have an attribute _gsc (GroundStateSolver). They also implement a solve() method which wraps the _gsc.solve() and returns a EigenstateResult. The only difference is that the outputs are lists (e.g. of excited energies) instead of values (e.g. the ground state energy alone).

I'm not sure how to do it best but I have a local working version with changes looking like
```
        if isinstance(self._gss, GroundStateSolver) and \
                isinstance(self._gss.solver, VariationalAlgorithm):  # type: ignore
            # Save initial point passed to min_eigensolver;
            # this will be used when NOT bootstrapping
            self._initial_point = self._gss.solver.initial_point  # type: ignore
        elif isinstance(self._gss, ExcitedStatesSolver) and \
            hasattr(self._gss, "_gsc") and \
                isinstance(self._gss._gsc.solver, VariationalAlgorithm):  # type: ignore
            # Save initial point passed to min_eigensolver;
            # this will be used when NOT bootstrapping
            self._initial_point = self._gss._gsc.solver.initial_point  # type: ignore
```
And 
```
            if isinstance(self._gss, ExcitedStatesSolver):
                energy = res.total_energies
            elif isinstance(self._gss, GroundStateSolver):
                energy = res.total_energies[0]
```
Which passes the test_bopes_sampler.py, runs the tutorial 05, and also allows me to create plots like the one attached with the following code

```
solver = VQE(quantum_instance=quantum_instance)
me_gsc = GroundStateEigensolver(qubit_converter, solver)
qeom_solver = QEOM(me_gsc, 'sd')
bs = BOPESSampler(gss=qeom_solver, bootstrap=True, num_bootstrap=None, extrapolator=None)
res = bs.sample(es_problem, points)
```

instead of 

```
solver = VQE(quantum_instance=quantum_instance)
me_gsc = GroundStateEigensolver(qubit_converter, solver)
bs = BOPESSampler(gss=me_gsc, bootstrap=True, num_bootstrap=None, extrapolator=None)
res = bs.sample(es_problem, points)
```
H2_molecule_z2sym_BOPES

It does add a lot of if/elif conditions so there might be better ways to do it

Anthony-Gandon:
### Summary

Fix #683 

Makes BOPES compatible with ExcitedStatesSolver by:
- Splitting the solve() method of excitedstatessolver into a get_qubit_operators() followed by the solve() to match the format of groundstateeigensolvers.
- Adding if/else clauses in BOPES because the VQE parameters are accessible with self._gss for a groundstateeigensovler and self._gss._gsc for a excitedstateseigensolver.

A new section was also added in the tutorial 05 to show typical BOPES curves for excited states.


Anthony-Gandon:
There is a mypy error in the bopes_sampler.py because:
- The first argument of BOPES is of type 'Union[GroundStateSolver, ExcitedStatesSolver]'
- In practice, this argument is going to be among GroundStateEigensolver, ExcitedStatesEigensolver, QEOM (which inherits from ExcitedStatesSolver)
- The boxes_sampler tries to access the Variational 'solver' attribute which are not defined at the level of GroundStateSolver and ExcitedStatesSolver.

I think this error existed already when BOPES was only working with GroundStateSolver.

woodsp-ibm:
> I think this error existed already when BOPES was only working with GroundStateSolver.

Indeed it seems that it was not picked up; and never noticed either as so far all the GroundStateSolvers here are GroundStateEigensolvers. In your logic where we want solver, since if its a GroundStateEigensolver that is variational we can manipulate the initial starting point, then it would be simple to change to check for GroundStateEigensolver instance and only access solver in this case - that seems, as you say, to be an problem from before that was not detected (mpypy or otherwise).

As to accessing the private property property of another class (_gsc) since I guess if the ExcitedStateSolver is based on a GroundStateSolver we do not have an abstraction for that  - and QEOM does this right, i.e computes based off a ground state. Arguably some base class for that would be more the way to go so the GroundStateSolver can be exposed via public API. Then here it would check if its in instance of that, then get the GroundStateSolver whereupon its the same logic, ie is that a GroundStateEigensolver and so on.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -113,11 +125,26 @@ def sample(self, problem: BaseProblem, points: List[float]) -> BOPESSamplerResul
         self._driver = problem.driver
         # We have to force the creation of the solver so that we work on the same solver
         # instance before and after _gss.solve
-        self._gss.get_qubit_operators(problem, None)
-        if isinstance(self._gss.solver, VariationalAlgorithm):  # type: ignore
+        if isinstance(self._gss, ExcitedStatesSolver):
+            if hasattr(self._gss, "_gsc"):
+                self._gss._gsc.get_qubit_operators(problem, None)
+        else:
+            self._gss.get_qubit_operators(problem, None)
```
mrossinek, line 132: If doing the above, this could be significantly simpilfied

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -90,10 +92,20 @@ def __init__(
...
             # this will be used when NOT bootstrapping
             self._initial_point = self._gss.solver.initial_point  # type: ignore
+        elif (
+            isinstance(self._gss, ExcitedStatesSolver)
+            and hasattr(self._gss, "_gsc")
+            and isinstance(self._gss._gsc.solver, VariationalAlgorithm)
+        ):  # type: ignore
+            # Save initial point passed to min_eigensolver;
+            # this will be used when NOT bootstrapping
+            self._initial_point = self._gss._gsc.solver.initial_point  # type: ignore
```
mrossinek, line 108: Might we even expose the `initial_point` setter of the GSS on the ESS?

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -219,9 +265,78 @@ def _run_single_point(self, point: float) -> EigenstateResult:
...
-            # at every point evaluation, the optimal params are updated
+        if isinstance(self._gss, GroundStateSolver) and isinstance(
+            self._gss.solver, VariationalAlgorithm
+        ):  # type: ignore
             optimal_params = result.raw_result.optimal_point
             self._points_optparams[point] = optimal_params
 
         return result
+
+    def _run_single_point_excited_solver(self, point: float) -> EigenstateResult:
```
mrossinek, line 276: Why does this method need to be duplicated for the excited state case?

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -219,9 +265,78 @@ def _run_single_point(self, point: float) -> EigenstateResult:
...
+                    self._gss._gsc.solver.initial_point = param_sets.get(point)  # type: ignore
+
+        # the output is an instance of EigenstateResult
+        result = self._gss.solve(self._problem)
+
+        if (
+            isinstance(self._gss, ExcitedStatesSolver)
+            and hasattr(self._gss, "_gsc")
+            and isinstance(self._gss._gsc.solver, VariationalAlgorithm)
+        ):
```
mrossinek, line 338: I think it would make sense to cache all these checks in another variable. That way you only need to evaluate all three conditions once and can then rely on the boolean result which you stored

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -36,7 +38,7 @@ class BOPESSampler:
 
     def __init__(
         self,
-        gss: GroundStateSolver,
+        gss: Union[GroundStateSolver, ExcitedStatesSolver],
```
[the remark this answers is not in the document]
Anthony-Gandon: In the new version, I opted for the name "solver_wrapper" which seemed appropriate

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -90,10 +92,20 @@ def __init__(
...
+        if isinstance(self._gss, GroundStateSolver) and isinstance(
+            self._gss.solver, VariationalAlgorithm
+        ):  # type: ignore
             # Save initial point passed to min_eigensolver;
             # this will be used when NOT bootstrapping
             self._initial_point = self._gss.solver.initial_point  # type: ignore
+        elif (
+            isinstance(self._gss, ExcitedStatesSolver)
+            and hasattr(self._gss, "_gsc")
+            and isinstance(self._gss._gsc.solver, VariationalAlgorithm)
```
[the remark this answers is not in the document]
Anthony-Gandon: A `@property` solver was added to the ExcitedStatesEigenSolver and QEOM (and their abstractions). This simplifies all the if/else clauses into:
`if isinstance(self._solver_wrapper.solver, VariationalAlgorithm):`

qiskit_nature/algorithms/excited_states_solvers/excited_states_eigensolver.py
```diff
@@ -57,34 +57,15 @@ def solver(self, solver: Union[Eigensolver, EigensolverFactory]) -> None:
         """Sets the minimum eigensolver or factory."""
         self._solver = solver
 
-    def solve(
+    def get_qubit_operators(
```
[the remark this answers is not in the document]
Anthony-Gandon: I have not yet found a way to refactor `ExcitedStatesSolver.get_qubit_operators` to call `self._gss.get_qubit_operators` because in principle, the ExcitedStatesSolver is not always aware of a GroundStateSolver. One way would be to split the get_qubit_operators() logic in two pieces (one getting and transforming the operators, and one instantiating the solver from the factory if needed).

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
Anthony-Gandon: This was significantly simplified but for another reason:
The get_qubit_operators() was added in the abstract GroundStateSolver and ExcitedStatesSolver. Their implementation in QEOM, ExcitedStatesEigensolver and GroundStateEigensolver was provided (qeom calls GroundStateEigensolver.get_qubit_operators, ExcitedStatesEigensolver has its own which is almost a copy of the one of GroundStateEigensolver, GroundStateEigensolver has not changed).

It comes down to one line in the end:
`self._solver_wrapper.get_qubit_operators(problem=problem, aux_operators=None)`

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
Anthony-Gandon: Does not seem needed anymore after having exposed the solver

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
Anthony-Gandon: I rolled back on this addition. Not needed anymore

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
Anthony-Gandon: Still an open question: should we have a local attribute `is_variational = isinstance(self._solver_wrapper.solver, VariationalAlgorithm)` ?

Anthony-Gandon:
Modifications were made following the discussion with @mrossinek. 
Apart from this, I wanted to append another new feature for the BOPESSampler. 
For the moment, BOPESSampler calls `self._wrapper_solver.solve()` without arguments so that aux is always None.
I added an argument  aux_operators: `Optional[ListOrDictType[Union[SecondQuantizedOp, PauliSumOp, Callable]]] = None`, to the BOPESSampler, which is passed to a function `evaluate_callable_aux_operators` and then to `self._wrapper_solver.solve()`. 
This allows auxiliary operators to be evaluated along with the energies.
The callable feature also allows nuclear coordinate dependent observables to be defined (forces for examples).

docs/tutorials/05_Sampling_potential_energy_surfaces.ipynb
```diff
@@ -54,7 +54,7 @@
     "\n",
     "warnings.simplefilter(\"ignore\", np.RankWarning)\n",
     "\n",
-    "algorithm_globals.random_seed = 75"
+    "algorithm_globals.random_seed = 76"
```
mrossinek, line 57: Why did you change the random seed?

docs/tutorials/05_Sampling_potential_energy_surfaces.ipynb
```diff
@@ -178,15 +178,15 @@
...
+    "points = np.linspace(0.20, 2, 30)\n",
     "results_full = {}  # full dictionary of results for each condition\n",
     "results = {}  # dictionary of (point,energy) results for each condition\n",
     "conditions = {False: \"no bootstrapping\", True: \"bootstrapping\"}\n",
     "\n",
     "\n",
     "for value, bootstrap in conditions.items():\n",
     "    # define instance to sampler\n",
-    "    bs = BOPESSampler(gss=me_gsc, bootstrap=value, num_bootstrap=None, extrapolator=None)\n",
+    "    bs = BOPESSampler(solver_wrapper=me_gsc, bootstrap=value, num_bootstrap=None, extrapolator=None)\n",
```
mrossinek, line 189:
```suggestion
    "    bs = BOPESSampler(me_gsc, bootstrap=value, num_bootstrap=None, extrapolator=None)\n",
```
This should be supplied positionally. Same should be done elsewhere

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -90,18 +97,26 @@ def __init__(
...
-            self._initial_point = self._gss.solver.initial_point  # type: ignore
+            self._initial_point = self._solver_wrapper.solver.initial_point
 
-    def sample(self, problem: BaseProblem, points: List[float]) -> BOPESSamplerResult:
+    def sample(
+        self,
+        problem: BaseProblem,
+        points: List[float],
+        aux_operators: Optional[
+            ListOrDictType[Union[SecondQuantizedOp, PauliSumOp, Callable]]
```
mrossinek, line 110: We should document the expectations for this `Callable`. You should do that in two ways:
1. add a more detailed type hint
2. add an explanation to the docstring below

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -152,9 +166,9 @@ def _run_points(self, points: List[float]) -> Dict[float, EigenstateResult]:
             The results for all points.
         """
         raw_results: Dict[float, EigenstateResult] = {}
-        if isinstance(self._gss.solver, VariationalAlgorithm):  # type: ignore
+        if isinstance(self._solver_wrapper.solver, VariationalAlgorithm):
```
mrossinek, line 169: Given your comment on the main thread, I agree that storing `self._is_variational_solver` (or something like that) would make sense :+1:

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -204,24 +218,45 @@ def _run_single_point(self, point: float) -> EigenstateResult:
...
+            if isinstance(self._solver_wrapper, ExcitedStatesSolver):
+                optimal_params = result.raw_result.ground_state_raw_result.optimal_point
+            elif isinstance(self._solver_wrapper, GroundStateSolver):
+                optimal_params = result.raw_result.optimal_point
             self._points_optparams[point] = optimal_params
 
         return result
+
+    def evaluate_callable_aux_operators(self):
```
mrossinek, line 246: Please add a return type hint

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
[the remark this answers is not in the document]
mrossinek: We noticed now, that changing the naming here does not work as CI will fail when running the stable tutorials (because the tutorials used to specify this argument via a keyword.

@manoelmarques @woodsp-ibm what do you think on this matter? I would indeed like to update the name... Can we add a patch to the stable tutorials to use this argument positionally, to ensure that they do not break by such a change? We are going to update them once we release 0.4 anyways..

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -36,15 +43,15 @@ class BOPESSampler:
...
+        gss: Union[GroundStateSolver, ExcitedStatesSolver],
         tolerance: float = 1e-3,
         bootstrap: bool = True,
         num_bootstrap: Optional[int] = None,
         extrapolator: Optional[Extrapolator] = None,
     ) -> None:
         """
         Args:
-            gss: GroundStateSolver
+            gss: GroundStateSolver (or ExcitedStatesSolver).
```
mrossinek, line 54:
```suggestion
            gss: GroundStateSolver or ExcitedStatesSolver.
```

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -61,13 +68,15 @@ def __init__(
                 ``WindowExtrapolator``.
         """
 
-        self._gss = gss
+        self._solver_wrapper = gss
+        self._is_variational_solver = isinstance(self._solver_wrapper.solver, VariationalAlgorithm)
```
mrossinek, line 72: I think this might fail if you supply e.g. a `VQEUCCFactory` to a `GroundStateSolver`. Please double check and fix accordingly

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -111,13 +143,15 @@ def sample(self, problem: BaseProblem, points: List[float]) -> BOPESSamplerResul
         """
         self._problem = problem
         self._driver = problem.driver
+        self._aux_operators = aux_operators
```
mrossinek, line 146: Is it really necessary to store these on `self`? It should suffice to pass them into `evaluate_callable_aux_operator`

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
```diff
@@ -0,0 +1,18 @@
...
+    of nuclear coordinates for Born Oppenheimer Potential Energy Surfaces.
+    Adds the `get_qubit_operators()` in
+    :class:`qiskit_nature.algorithms.ExcitedStatesEigensolver`
+    This matches the format of
+    :class:`qiskit_nature.algorithms.GroundStatesEigensolver`
+    BOPESSampler now also accepts auxiliary operators to pass to the solver.
+    Geometry-dependent observable can also be defined as auxiliaries
+
+upgrade:
+  - |
```
mrossinek, line 15: I don't think we ever really use this section. I would just append this information at the end of the `features` section

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
```diff
@@ -0,0 +1,18 @@
...
+    The tutorial 05 now includes calculations of excited state Born Oppenheimer potential energy
```
mrossinek, line 16: The tutorial numbers are not really visible in the online documentation. Instead, you can link to the tutorial directly, as done e.g. here:
https://github.com/Qiskit/qiskit-nature/blob/e5b5021863a4f7d8658d4f72959217347b078053/releasenotes/notes/bogoliubov-transform-885eb6442172c76e.yaml#L8

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
```diff
@@ -0,0 +1,18 @@
+features:
+  - |
+    The solver of BOPESSampler now accepts solvers of type
+    `Union[GroundStateSolver,ExcitedStatesSolver]` instead of only a `GroundStateSolver`. 
+    This generalizes to excited states the sampling 
+    of nuclear coordinates for Born Oppenheimer Potential Energy Surfaces.
+    Adds the `get_qubit_operators()` in
+    :class:`qiskit_nature.algorithms.ExcitedStatesEigensolver`
+    This matches the format of
+    :class:`qiskit_nature.algorithms.GroundStatesEigensolver`
```
mrossinek, line 10:
```suggestion
    :class:`qiskit_nature.algorithms.GroundStateEigensolver`.
```

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
```diff
@@ -0,0 +1,18 @@
+features:
+  - |
+    The solver of BOPESSampler now accepts solvers of type
+    `Union[GroundStateSolver,ExcitedStatesSolver]` instead of only a `GroundStateSolver`. 
+    This generalizes to excited states the sampling 
+    of nuclear coordinates for Born Oppenheimer Potential Energy Surfaces.
+    Adds the `get_qubit_operators()` in
+    :class:`qiskit_nature.algorithms.ExcitedStatesEigensolver`
```
mrossinek, line 8:
```suggestion
    :class:`qiskit_nature.algorithms.ExcitedStatesEigensolver`.
```

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
```diff
@@ -0,0 +1,18 @@
...
+    This matches the format of
+    :class:`qiskit_nature.algorithms.GroundStatesEigensolver`
+    BOPESSampler now also accepts auxiliary operators to pass to the solver.
+    Geometry-dependent observable can also be defined as auxiliaries
```
mrossinek, line 12:
```suggestion
    Geometry-dependent observable can also be defined as auxiliaries.
```

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
Anthony-Gandon: Indeed, this _is_variational_solver needs to be defined after the get_qubit_operators() which has a line
```
if isinstance(self._solver, MinimumEigensolverFactory):
    # this must be called after transformation.transform
    self._solver = self._solver.get_solver(problem, self._qubit_converter)
```

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
Anthony-Gandon: The method sample() calls run_points() which calls run_single_points() which calls solver.solve() on the problem and the aux_operators. Instead of passing the arguments aux_operators through this chain of methods, I suggested that aux_operators would be stored in self.

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
Anthony-Gandon: Done

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
Anthony-Gandon: Done

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
Anthony-Gandon: Done

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
Anthony-Gandon: Done

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
Anthony-Gandon: Done

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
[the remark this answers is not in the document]
manoelmarques: I am fine with changing the stable tutorials for this as it will still work with released code and 0.4 is near anyway.
If you agree, I will create a main PR changing the main tutorials with a backport option.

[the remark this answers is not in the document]
woodsp-ibm: Nothing should break by doing this. If you want to rename it then gss should be kept as deprecated parameter, maybe add at the end, so if people did it by keyword it will not break. Hence the tutorials should be updated to avoid a deprecated parameter but they should not break if we can at all avoid it. Which I think we can in this case right.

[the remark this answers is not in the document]
mrossinek: #688 addresses this and, once backported, you should be able to rename the first argument of this class as you had originally intended, @Anthony-Gandon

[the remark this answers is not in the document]
mrossinek: Okay, we had some more offline discussion. Here is, what the strategy which we should follow:

1. Change `gss -> solver_wrapper=None` and add `gss=None` as a keyword argument at the end.
This ensures that everything works the same right now.
At the same time, you add a deprecation for the `gss` argument.

2. In the future, once we remove `gss`, we can also remove the default `None` value of the `solver_wrapper`, because in Python even a positional argument can be specified as a keyword argument later

@Anthony-Gandon if any of this is unclear, feel free to ping me here or offline :+1:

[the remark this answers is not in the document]
Anthony-Gandon: So I was not sure how to deprecate the gss parameter. The latest commit:

- Adds a `@deprecate_arguments` to the constructor of BOPESSolver.
- Adds the gss argument at the end of the constructor for BOPES and sets the gss and solver_wrapper argument as `Optional`.
- Modifies the doctoring of the constructor
- Modifies the release note to add the deprecation.

Please let me know if that is the right way / if this is enough, thanks !

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -41,17 +42,20 @@
 class BOPESSampler:
     """Class to evaluate the Born-Oppenheimer Potential Energy Surface (BOPES)."""
 
+    @deprecate_arguments({"gss": "solver_wrapper"})
+    # pylint: disable=unused-argument
     def __init__(
         self,
-        gss: Union[GroundStateSolver, ExcitedStatesSolver],
+        solver_wrapper: Optional[Union[GroundStateSolver, ExcitedStatesSolver]] = None,
```
woodsp-ibm, line 49: From a user perspective do you think `solver_wrapper` is a good name? Since its either  a GSS or ESS why not just `solver` - since its one of GroundState or ExcitedStates solvers that a user is passing in. Yes internally the code might want to get at the raw algorithm solver but the "wrapper" aspect seems awkward to me looking at this purely from an end user using the algo

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -61,14 +65,16 @@ def __init__(
                 all previous points will be used for bootstrapping.
             extrapolator: Extrapolator objects that define space/window
                            and method to extrapolate variational parameters.
+            gss: (DEPRECATED) Old name for the solver when it only supported GroundStateEigensolver.
```
woodsp-ibm, line 68: Since the prior version said GroundStateSolver as the type perhaps this should say that here. Maybe in practice the GroundStateEigensolver subclass was all that we had and that was supported in practice but it said the type was a GroundStateSolver

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
Anthony-Gandon: Indeed we can change it to `solver`. Because we exposed the solver of ExcitedStatesEigensolver, `wrapper_solver` now does not feel necessary. Before that, it was to avoid using solver.solver.solve()

qiskit_nature/results/bopes_sampler_result.py
```diff
@@ -44,9 +44,14 @@ def points(self) -> List[float]:
...
     @property
-    def energies(self) -> List[float]:
+    def energies(self) -> Union[List[float], List[List[float]]]:
         """returns list of energies."""
-        return self._energies
+        formatted_energies: Union[List[float], List[List[float]]]
+        if len(self._energies[0]) == 1:
+            formatted_energies = [self._energies[k][0] for k in range(len(self._energies))]
+        else:
+            formatted_energies = self._energies
```
woodsp-ibm, line 53: Is this logic correct? It just seems odd in that if the len of the element is 1 we take out the only element and make a flat list of it, so self._energies is always a list of lists but of the size of the inner list is 1 then we make it a list of floats. This is simply because of how you format up self._energies from the result? Maybe I am confusing things in my mind with the excited state result from the eigensolver since as I recall that was a list of tuples, each tuple being (value, variance) pair

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
mrossinek: I agree. I guess in theory an `AdaptVQE` could also have been used here, but I don't think I ever tested that

qiskit_nature/results/bopes_sampler_result.py
mrossinek: I did some digging. What is being done here is the differentiation between the ground- and excited-states case.
In both cases, `self._energies` will be a list of numpy arrays. In the case of the ground-state, all of these arrays have length 1 in which case this list is being flattened. In the excited-states case, these arrays are longer and, thus, are left untouched.

I think this is fine as is, if a comment gets added, explaining this behavior, similar to as I did here just now

test/algorithms/pes_samplers/test_bopes_sampler.py
```diff
@@ -194,6 +200,142 @@ def test_h2_bopes_sampler_with_factory(self):
...
+            energies,
+            [
+                [-1.13618945, -0.47845306, -0.1204519, 0.5833141],
+                [-1.10115033, -0.74587179, -0.35229063, 0.03904763],
+                [-1.03518627, -0.85523694, -0.42240202, -0.21860355],
+            ],
+            decimal=2,
+        )
+
+    @unittest.skipIf(not _optionals.HAS_PYSCF, "pyscf not available.")
```
mrossinek, line 251:
```suggestion
    @unittest.skipIf(not _optionals.HAS_PYSCF, "pyscf not available.")
    @unittest.skipUnless(optionals.HAS_AER, "qiskit-aer is required to run this test")
```

test/algorithms/pes_samplers/test_bopes_sampler.py
```diff
@@ -194,6 +200,142 @@ def test_h2_bopes_sampler_with_factory(self):
...
```
mrossinek, line 296:
```suggestion
    @unittest.skipIf(not _optionals.HAS_PYSCF, "pyscf not available.")
    @unittest.skipUnless(optionals.HAS_AER, "qiskit-aer is required to run this test")
```

mrossinek (commented):
Besides Steve's comments and these two fixes to ensure proper test skipping when aer is not available, I think this is good to go :+1:

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
Anthony-Gandon: When trying to access the `initial_point` of the solver we would use `self._solver.solver.initial_point`. I find it confusing so I went for` _state_solver`.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
Anthony-Gandon: Indeed, this was a typo, thanks

qiskit_nature/results/bopes_sampler_result.py
Anthony-Gandon: Indeed, when only `GroundStateSolver` were used, BOPES was flattening the result before giving it to the BOPESResult. Now that it also accepts `ExcitedStatesSolver` for which the output is a list of (lists of size N, with N the Nb of excited states), it is more natural to keep the list of (lists of size 1) and to only flatten it later in `BOPESResult`.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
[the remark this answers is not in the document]
Anthony-Gandon: (`solver_wrapper` was changed to `state_solver` following a later comment)

qiskit_nature/results/bopes_sampler_result.py
mrossinek: I'm actually wondering now, if this flattening is needed at all :thinking:
If a user were to use the resulting `BOPESSamplerResult` object, they would then again also need to differentiate the case. If we were to leave this as a list of numpy arrays, it would always behave the same

Anthony-Gandon: Regarding your comment on the flattening of the BOPESResult energies, it is only a matter of backward compatibility. All the tests and tutorials use List(float) instead of List(List(float)). I wanted to keep it transparent for GroundStateSolver and add the option for ExcitedStatesSolver

docs/tutorials/05_Sampling_potential_energy_surfaces.ipynb
```diff
@@ -465,16 +469,167 @@
...
+    "        marker=\"x\",\n",
+    "        color=cmap(k / nb_excited_states),\n",
+    "    )\n",
+    "    (l2,) = plt.plot(\n",
+    "        results_excited[\"points\"],\n",
+    "        results_excited[\"energies_np\"][:, k],\n",
+    "        linestyle=\"dashed\",\n",
+    "        color=cmap(k / nb_excited_states),\n",
+    "    )\n",
+    "plt.legend([l1, l2], [\"qeom\", \"np\"])\n",
```
woodsp-ibm, line 617: In the pre-existing plots the graph for the NumPy based solver is referred to in the legend as `numpy` - perhaps we should do that here too to keep things consistent.

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
```diff
@@ -89,19 +104,41 @@ def __init__(
...
+    def sample(
+        self,
+        problem: BaseProblem,
+        points: List[float],
+        aux_operators: Optional[
+            ListOrDictType[
+                Union[
+                    SecondQuantizedOp,
+                    PauliSumOp,
+                    Callable[[BaseProblem], Union[SecondQuantizedOp, PauliSumOp]],
```
woodsp-ibm, line 121: I see a unit test that passes in a SecondQuantizedOp, maybe since the conversion is done elsewhere the direct passing in of a qubit operator (PauliSumOp) is already tested as such, though it could not harm to do so from this level. Also the callable path through the code is not exercised by unit test and would be good to cover.

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
```diff
@@ -0,0 +1,21 @@
+features:
+  - |
+    The solver of BOPESSampler now accepts solvers of type
```
woodsp-ibm, line 3: I think here BOPESSampler could a class ref like you have in the deprecations part so it forms a link the reader can click on.

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
```diff
@@ -0,0 +1,21 @@
...
+    `Union[GroundStateSolver,ExcitedStatesSolver]` instead of only a `GroundStateSolver`. 
+    This generalizes to excited states the sampling 
+    of nuclear coordinates for Born Oppenheimer Potential Energy Surfaces.
+    Adds the `get_qubit_operators()` in
+    :class:`qiskit_nature.algorithms.ExcitedStatesEigensolver`.
+    This matches the format of
+    :class:`qiskit_nature.algorithms.GroundStatesEigensolver`.
```
woodsp-ibm, line 10:
```suggestion
    :class:`qiskit_nature.algorithms.GroundStateEigensolver`.
```
No `s` in the ground state one (it was failing to resolve this in the outputs docs and it was not clickable as a link).

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
```diff
@@ -0,0 +1,21 @@
...
+    BOPESSampler now also accepts auxiliary operators to pass to the solver.
+    Geometry-dependent observable can also be defined as auxiliaries.
+    See the `Sampling the potential energy surface <tutorials/05_Sampling_potential_energy_surfaces.ipynb>`_ tutorial
+    for a demonstration of how to use this calculation of excited state Born Oppenheimer potential energy surfaces.
+deprecations:
+  - |
+    The argument `gss` of the constructor of:class:`qiskit_nature.algorithms.BOPESSampler` was deprecated
```
woodsp-ibm, line 17:
```suggestion
    The argument `gss` of the constructor of :class:`qiskit_nature.algorithms.BOPESSampler` was deprecated 
```
There was no space between of and the class ref so it all came out as text in the release note rather than being converted.

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
Anthony-Gandon: Done, thanks

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
Anthony-Gandon: done, thanks

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
Anthony-Gandon: Typo fixed

docs/tutorials/05_Sampling_potential_energy_surfaces.ipynb
Anthony-Gandon: Done

qiskit_nature/algorithms/pes_samplers/bopes_sampler.py
Anthony-Gandon: the test now checks the three possible types: Callable, SecondQuantizedOp, PauliSumOp

releasenotes/notes/support_excitedstatessolver_bopes-2bbb5a282b1ff582.yaml
```diff
@@ -0,0 +1,20 @@
...
+    of nuclear coordinates for Born Oppenheimer Potential Energy Surfaces.
+    Adds the `get_qubit_operators()` in :class:`qiskit_nature.algorithms.ExcitedStatesEigensolver`.
+    This matches the format of :class:`qiskit_nature.algorithms.GroundStateEigensolver`.
+    BOPESSampler now also accepts auxiliary operators to pass to the solver.
+    Geometry-dependent observable can also be defined as auxiliaries.
+    See the `Sampling the potential energy surface <tutorials/05_Sampling_potential_energy_surfaces.ipynb>`_ tutorial
+    for a demonstration of how to use this calculation of excited state Born Oppenheimer potential energy surfaces.
+deprecations:
+  - |
+    The argument `gss` of the constructor of :class:`qiskit_nature.algorithms.BOPESSampler` was deprecated
```
woodsp-ibm, line 15:
```suggestion
    The argument `gss` of the constructor of :class:`~qiskit_nature.algorithms.pes_samplers.BOPESSampler` was deprecated 
```

mrossinek (approved):
I did not do another thorough pass of the documentation links but am assuming based on my previous reviews + Steve's reviews that this is good now.
The rest also looks good to me, so I'm gonna approve this PR and let Steve merge it if he approves, too :+1:

woodsp-ibm (approved):
Looks good to me too. Thanks for your first contribution to Qiskit Nature!

Anthony-Gandon:
Thank you all for the feedbacks and reviews!
Anthony

PR #688, opened 2022-06-13
Change BOPESSasmpler creation first arg. to positional

manoelmarques:


mrossinek (approved):
Thanks Manoel! :slightly_smiling_face:

woodsp-ibm (changes requested):
We should not be backporting a breaking change. The BopesSampler ought to be changed so stable tutorials pass - add gss as a deprecated parameter we are changing the name,

woodsp-ibm:
As these changes now exist in #684, without the need of a backport, since the parameter re-naming will be handled via deprecation, is this needed any more?

Anthony-Gandon:
I agree that this backport is not necessary anymore because of how we chose to handle the deprecation of the constructor of BOPESSampler.

PR #697, opened 2022-06-16
Fix is_hermitian and add atol

kevinsung:
### Summary
- Fixes bug where `FermionicOp.is_hermitian` gives false negatives due to not normal-ordering
- Adds `atol` parameter to `is_hermitian`


qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -766,3 +769,10 @@ def one(cls, register_length: int) -> FermionicOp:
             The unity-operator of the given length.
         """
         return FermionicOp(("", 1.0), register_length=register_length, display_format="sparse")
+
+    def is_hermitian(self, atol: Optional[float] = None) -> bool:
+        """Checks whether the operator is hermitian"""
+        if atol is None:
+            atol = self.atol
+        diff = (self - self.adjoint()).normal_ordered().simplify(atol=atol)
```
mrossinek, line 777: Can you please provide an example where the previous (or current implementation without calling `normal_ordered()`) fails?

test/operators/second_quantization/test_fermionic_op.py
```diff
@@ -374,6 +374,14 @@ def test_hermiticity(self):
             )
             self.assertFalse(fer_op.is_hermitian())
 
+        with self.subTest("test require normal order"):
+            fer_op = FermionicOp("+_0 -_0 -_1") - FermionicOp("+_1 -_0 +_0") + FermionicOp("+_1")
+            self.assertTrue(fer_op.is_hermitian())
+
+        with self.subTest("test passing atol"):
+            fer_op = FermionicOp("+_0 -_1") + (1 + 1e-7) * FermionicOp("+_1 -_0")
+            self.assertTrue(fer_op.is_hermitian(atol=1e-6))
```
mrossinek, line 383: It would be good to also do an `assertFalse` where atol is too tight

mrossinek (commented):
Other than these comments, this PR also needs a release note indicating the bugfix

test/operators/second_quantization/test_fermionic_op.py
kevinsung: Done.

qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -60,8 +61,9 @@ def to_list(self) -> list[tuple[str, complex]]:
         """Returns the operators internal contents in list-format."""
         raise NotImplementedError
 
-    def is_hermitian(self) -> bool:
+    def is_hermitian(self, atol: Optional[float] = None) -> bool:
         """Checks whether the operator is hermitian"""
```
mrossinek, line 65:
```suggestion
        """Checks whether the operator is hermitian."""
```
Besides that, now that you added an argument, you need to add an `Args:` section.
At this point I would also add a `Returns:` section.

The same goes for the `FermionicOp` subclass (or you check how sphinx can deal with documentation string inheritance)

mrossinek (commented):
Other than that, this is good to go.

@woodsp-ibm any comments from your side?

qiskit_nature/operators/second_quantization/second_quantized_op.py
kevinsung: Done.

kevinsung:
@mrossinek I found that sphinx handles inherited docstrings correctly, so I deleted the redundant docstrings for both `is_hermitian` and `simplify`.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -771,7 +771,14 @@ def one(cls, register_length: int) -> FermionicOp:
         return FermionicOp(("", 1.0), register_length=register_length, display_format="sparse")
 
     def is_hermitian(self, atol: Optional[float] = None) -> bool:
-        """Checks whether the operator is hermitian"""
+        """Checks whether the operator is hermitian.
```
woodsp-ibm: Since this text is identical to the parent class method it is overridding, you should not have to replicate the text in a derived class. Sphinx should use the parent class text. Indeed its better too if you don't copy it unnecessarily as any docstring changes/corrections etc would then only need to be in the one place.

kevinsung: I've already deleted the redundant docstrings, see my comment above and the latest commit.

qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -48,7 +49,7 @@ def simplify(self, atol: Optional[float] = None):
         Returns a new operator (the original operator is not modified).
 
         Args:
-            atol: Absolute tolerance for checking if coefficients are zero (Default: 1e-8).
+            atol: Absolute numerical tolerance. The default behavior is to use ``self.atol``.
```
mrossinek, line 52: In Terra's docs, the default value of `atol` can be inferred by Sphinx from the `TolerancesMeta` class, but this does not work within the separate docs built for Nature. Thus, I would like to retain the default value in the docstring here. I.e. say that the default is to use `self.atol`, which in turn defaults to `1e-8`

kevinsung: Done.

qiskit_nature/operators/second_quantization/fermionic_op.py
```diff
@@ -248,7 +248,10 @@ def __init__(
 
         self._data: list[tuple[tuple[tuple[str, int], ...], complex]]
 
-        if (
+        if not isinstance(data, str) and not data:
```
ikkoham, line 251: We already have class method `zero`. Is this change necessary?

kevinsung: This change is needed. It actually fixes a bug in `FermionicOp.simplify` that caused it to fail when the simplified operator is equal to zero. This bug was triggered by the new `is_hermitian` tests, but I've added a separate test for this too. I've updated the release note.

mrossinek (changes requested):
After merging #746, the code has been relocated into the new stack.
Please apply these changes on the new operator classes located in `qiskit_nature/second_q/operators` and leave the code in the old location unchanged.

Please also add a test of passing `atol` to `VibrationalOp.is_hermitian` as you have done for the `SpinOp`, too.

Once these changes are done, imo, this can be merged.

kevinsung:
@mrossinek Done.

releasenotes/notes/is-hermitian-33d1000779fba472.yaml
```diff
@@ -0,0 +1,14 @@
+---
+features:
+  - |
+    Adds `atol` parameter to :meth:`.SecondQuantizedOp.is_hermitian`.
+  - |
+    :class:`.FermionicOp` can now be initialized with an empty list. That is,
+    `FermionicOp([])` returns the zero operator instead of raising an error.
```
mrossinek, line 7: This is rather a bugfix imo

mrossinek (commented):
Other than this single comment, this is good to go

releasenotes/notes/is-hermitian-33d1000779fba472.yaml
kevinsung: Done.

woodsp-ibm (approved):

Issue #644, opened 2022-04-21
Improve performance of `QubitMapper.mode_based_mapping`


mrossinek:
### What is the expected enhancement?

When looking towards larger systems and an increased amount of auxiliary operators (multiple hundreds to thousands), the performance of `QubitMapper.mode_based_mapping` becomes a serious bottleneck in the pre-processing step. Waiting times for all auxiliary operators to be mapped can reach on the order of minutes.

I believe, performance can be significantly improved if we:
- cache the processed `pauli_list` object within `QubitMapper` (requires changing the `mode_based_mapping` from a static to an instance method)
- refactor the iteration to deal with sparse labels rather than dense ones (requires sparse labels to be the default, a conversion of dense to sparse, and updating the `VibrationalOp` to be inline with the `FermionicOp` again; if this gets tackled before said update, we can also handle the mapping `FermionicOp`s separately for the time being)
- potentially do something similar to #638 by constructing a single `SparsePauliOp` from a list of terms rather than performing the full computation each time

I put together a script which benchmarks the performance of the mapping.
```python
import time

from qiskit_nature.second_q.mappers import JordanWignerMapper
from qiskit_nature.second_q.operators import FermionicOp

timings = {}

with open("timings.csv", "w") as file:
    try:
        for size in range(10, 21, 5):
            timings[size] = {}
            for num in range(100, 1001, 100):
                print(f"Mapping {num} operators of length {size}", flush=True)
                start = time.time()
                mapper = JordanWignerMapper()
                for i in range(num):
                    q_ops = mapper.map(FermionicOp(" ".join(f"N_{i}" for i in range(size)),
                                                   display_format="sparse"))
                stop = time.time()
                timings[size][num] = stop - start

                file.write(f"{size},{num},{stop - start}\n")
                file.flush()
    except KeyboardInterrupt:
        pass

```
*EDIT: updated the script above to reflect the new code locations.*
Below is the timing report. Right now, this pretty much scales 1-to-1 with the number of operators but I believe we can significantly reduce the overhead.
```csv
10,100,1.338921308517456
10,200,2.6084437370300293
10,300,3.912541151046753
10,400,5.215270042419434
10,500,6.510892629623413
10,600,7.804781436920166
10,700,9.110262393951416
10,800,10.41125202178955
10,900,11.708135843276978
10,1000,13.017327785491943
15,100,10.335224628448486
15,200,20.665453672409058
15,300,30.995742797851562
15,400,41.318920612335205
[...]

PR #699, opened 2022-06-17
Change mode_based_mapping to instance method

jennifershan:
### Summary

- Changes `mode_based_mapping` from static method to instance method

### Details and comments


qiskit_nature/mappers/second_quantization/jordan_wigner_mapper.py
```diff
@@ -44,4 +44,4 @@ def map(self, second_q_op: FermionicOp) -> PauliSumOp:
             pauli_table.append((Pauli((a_z, a_x)), Pauli((b_z, b_x))))
             # TODO add Pauli 3-tuple to lookup table
 
-        return QubitMapper.mode_based_mapping(second_q_op, pauli_table)
+        return QubitMapper.mode_based_mapping(self, second_q_op, pauli_table)
```
mrossinek, line 47:
```suggestion
        return self.mode_based_mapping(second_q_op, pauli_table)
```
This is how this method should be called as an instance method.
The same update needs to be done in all other subclasses of `QubitMapper`

mrossinek (changes requested):
Thanks for the PR. Is this PR supposed to be recognized as part of Unitary Hack? (https://unitaryhack.dev/) I'm asking because I assume you mean to address the issue https://github.com/Qiskit/qiskit-nature/issues/644 which has a bounty assigned to it.

If so, this will require considerably more work. I outlined three points in the description of that issue which should be addressed. I do not see them being started on here.

Are you interested in continuing this PR to work on those performance improvements? I can also provide some more details on the expectations if the issue description is not too clear to you.

jennifershan:
Hi, thank you so much for leaving comments! I did mean to address that issue and I opened this to get started and make some progress on it. I’m definitely interested in continuing this! Could you provide more details? I'm a bit stuck on making sense of the tasks.

mrossinek:
Okay, here are some more details on the three suggested improvements:

* cache the processed pauli_list object within QubitMapper (requires changing the mode_based_mapping from a static to an instance method)

What I meant here is that the lists `times_creation_op`, `times_annihilation_op`, `times_occupation_number_op`, and `times_emptiness_number_op` should become `self.times_...` variables. You need to specify them as part of `__init__` and set them to empty lists. Then, during `mode_based_mapping`, you only need to generate these lists when they are still empty.

We should also add a way to reset these cached objects but we can look into that a bit later once we have made a bit more progress and have this caching in place.

* refactor the iteration to deal with sparse labels rather than dense ones (requires sparse labels to be the default, a conversion of dense to sparse, and updating the VibrationalOp to be inline with the FermionicOp again; if this gets tackled before said update, we can also handle the mapping FermionicOps separately for the time being)

Right now, during `mode_based_mapping`, there is a call `second_q_op.to_list()` which will produce dense labels. What that means is that you obtain labels that look similar to this: `+-II`, `I+-I`, etc.
Instead, this should be changed to iterate over the sparse labels which would look like `+_0 -_1`, `+_1 -_2`, etc.

* potentially do something similar to https://github.com/Qiskit/qiskit-nature/pull/638 by constructing a single SparsePauliOp from a list of terms rather than performing the full computation each time

This change is a bit more involved and could potentially be done in a separate PR. But the idea here would be to remove the need for doing `ret_op.compose(...)` every time you add an operator. Instead, it might be possible to generate a list of terms and then only initialize a single `SparseLabelOp` at the very end, using that list of labels.
I have not looked into the details of this yet, so I cannot provide more details as of right now.

mrossinek:
Hi @jennifershan! It's been two weeks since my last update so I wanted to check in with you. Are you able to progress this PR further?

I am in contact with another person who is keen to work on this, so please let me know :+1:

jennifershan:
Hi @mrossinek sorry, I have been busy lately but will get back on it!

mrossinek:
> Hi sorry, I have been busy lately but will get back on it!

Glad to hear that! I see that you managed to implement my first suggested improvement. How does the performance look now when running the simple benchmark I used in the description of #644?

Would you also like to tackle the second and/or third suggested improvement as part of this PR or keep this more contained to the first one? I would need some time to flesh out the details of the third suggestion myself :+1:

jennifershan:
Yes, I'd like to keep working on the other improvements! Strangely enough, it seems to be taking slightly longer now with the caching, but it may just be computer issues on my end. I will keep testing it.

mrossinek:
> Strangely enough, it seems to be taking slightly longer now with the caching, but it may just be computer issues on my end.

You will need to run the original benchmark without your changes on your own computer, too. You will not be able to compare it with the numbers that I posted in the original issue description :+1:

mrossinek (changes requested):
A short heads-up: we just merged #746 which majorly restructured the code. The mappers are now located under `qiskit_nature/second_q/mappers`.
Please change your PR to update those files and leave the code in the old location unchanged.

Also: are there any updates after my previous comment about the benchmark?

jennifershan:
Hi, thanks for letting me know; I have updated the files accordingly. I ran the benchmark with the original version and with the new version with only caching implemented and the results are somewhat mixed on my laptop. 

This is the original:
`10,100,1.2470319271087646
10,200,2.4612691402435303
10,300,3.748835802078247
10,400,4.510950803756714
10,500,5.506355047225952
10,600,6.107203006744385
10,700,7.033687114715576
10,800,8.606460809707642
10,900,9.99450397491455
10,1000,11.904584169387817
15,100,4.356929779052734
15,200,10.041445970535278
15,300,13.114410161972046
15,400,16.873305082321167
15,500,19.586038827896118
15,600,28.399156093597412
15,700,28.204347133636475
15,800,32.69708585739136
15,900,37.55910110473633
15,1000,41.40331792831421`

This is the new:
`10,100,1.060662031173706
10,200,2.166015863418579
10,300,3.1215579509735107
10,400,3.9897687435150146
10,500,6.040130615234375
10,600,5.982110977172852
10,700,7.335954904556274
10,800,8.006319761276245
10,900,10.332020998001099
10,1000,10.622812032699585
15,100,4.25658106803894
15,200,8.661841869354248
15,300,12.773032188415527
15,400,16.65694308280945
15,500,21.2469539642334
15,600,25.571537017822266
15,700,28.7299861907959
15,800,34.22808504104614
15,900,39.23568367958069
15,1000,42.464502811431885`

This is the trend when I run them in either order, where the new version runs slightly slower towards the end with more and longer operators. (I haven't let it run the 20 length ones, though, so I will do more thorough testing tomorrow.)

mrossinek:
Did you update the benchmark script in the other issue to also reflect the new code location?
I went ahead and edited the original issue description to include that update. If you did not, then you effectively ran the `main` code twice.

That would also explain why you don't see the issue with your current PR branch. While the caching works in general, you did not update the variable names on lines 134 ff. of the `QubitMapper`. There, the code is still trying to access `times_creation_op` instead of `self.times_creation_op`.
When fixing this, you should be able to see an improvement close to the following:
Original on main:
```csv
10,100,0.7812883853912354
10,200,1.5545504093170166
10,300,2.613691568374634
10,400,3.262303113937378
10,500,4.551577091217041
10,600,7.32957911491394
10,700,10.063297748565674
10,800,13.545177459716797
10,900,14.082147598266602
10,1000,17.573906421661377
```
Your (fixed) PR branch:
```csv
10,100,0.20881175994873047
10,200,0.36502599716186523
10,300,0.5334429740905762
10,400,0.6865215301513672
10,500,0.8933868408203125
10,600,1.0970253944396973
10,700,1.215528964996338
10,800,1.4243927001953125
10,900,1.671576976776123
10,1000,1.7892913818359375
```

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -84,31 +87,30 @@ def mode_based_mapping(
...
-            # The annihilation operator is given by 0.5*(X + 1j*Y)
-            annihilation_op = real_part + imag_part
-            times_annihilation_op.append(annihilation_op)
-
-            # The occupation number operator N is given by `+-`.
-            times_occupation_number_op.append(
-                creation_op.compose(annihilation_op, front=True).simplify()
-            )
+        if not (self.times_creation_op and self.times_annihilation_op \
+        and self.times_occupation_number_op and self.times_emptiness_number_op):
```
mrossinek, line 91: I suggest that this entire code block gets extracted into a private utility method. E.g.
```python
def _cache_pauli_table(self, pauli_table):
    if not (...):
        ...
```

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -34,6 +34,10 @@ def __init__(self, allows_two_qubit_reduction: bool = False):
                 number of qubits in the mapped operator can be reduced accordingly.
         """
         self._allows_two_qubit_reduction = allows_two_qubit_reduction
+        self.times_creation_op = []
```
mrossinek, line 37:
```suggestion
        self.times_creation_op: list[SparsePauliOp] = []
```
I believe this is the type hint you're looking for in all of these cases

mrossinek (commented):
For the sake of merging this PR, I suggest that we keep the changes here to a minimum.
That means:
- only apply the caching as already done (but refactor it into a separate method; see comment)
- add a _public_ method `invalidate_cache()` which resets the stored lists
- add some unittests which ensure the caching and invalidation work as expected
- add a release note

This will already be plenty to do and if you want to add further performance optimizations, you can open another PR in the future :+1:

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -34,6 +34,18 @@ def __init__(self, allows_two_qubit_reduction: bool = False):
                 number of qubits in the mapped operator can be reduced accordingly.
         """
         self._allows_two_qubit_reduction = allows_two_qubit_reduction
+        self.times_creation_op: list[SparsePauliOp] = []
```
mrossinek, line 37: In order to resolve the mypy issues which you are seeing here, you need to add the following import as the first import in this file:
```python
```

mrossinek:
Hi @jennifershan! Just checking in: will you be able to finish working on the list of tasks that I had mentioned [here](https://github.com/Qiskit/qiskit-nature/pull/699#pullrequestreview-1059792810) or would you like to hand this PR over for us to finish internally? I am asking because it would be good if we could wrap this PR up before the end of September and I know that life can be busy and get in the way of things :+1:

jennifershan:
Hey, thanks for reaching out! Yeah, I think it definitely makes sense to hand this over to be finished now as I don't have much time to work on it. Thank you for helping me out on everything!

woodsp-ibm (approved):

mrossinek (approved):
Thank you, @jennifershan, for getting us started on this! :clap:

PR #716, opened 2022-06-21
[After #697] Implement approximate equality for SecondQuantizedOp

### LINKED ISSUE #715: Approximate equality for SecondQuantizedOps

Comparing two SecondQuantizedOps for approximate equality is useful for testing. As far as I can tell there is no simple way to do this yet.

kevinsung:
### Summary
Fixes #715 
Comes after #697

### DIFF
--- a/qiskit_nature/second_q/operators/second_quantized_op.py
+++ b/qiskit_nature/second_q/operators/second_quantized_op.py
@@ -85,7 +85,9 @@ class SecondQuantizedOp(StarAlgebraMixin, TolerancesMixin, ABC):
             True if the operators are equal up to numerical tolerance, False otherwise.
         """
         if not isinstance(other, type(self)):
-            return NotImplemented
+            raise NotImplementedError(
+                f"Cannot compare objects of types {type(self)} and {type(other)}."
+            )
         if atol is None:
             atol = self.atol
         diff = (self - other).simplify(atol=atol)
--- a/qiskit_nature/second_q/operators/second_quantized_op.py
+++ b/qiskit_nature/second_q/operators/second_quantized_op.py
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import Any, Optional
 
 import numpy as np
 from qiskit.opflow.mixins import StarAlgebraMixin
@@ -71,7 +71,22 @@ class SecondQuantizedOp(StarAlgebraMixin, TolerancesMixin, ABC):
         Returns:
             True if the operator is hermitian up to numerical tolerance, False otherwise.
         """
+        return self.equiv(self.adjoint(), atol=atol)
+
+    def equiv(self, other: Any, atol: Optional[float] = None) -> bool:
+        """Checks whether this operator is approximately equal to another operator.
+
+        Args:
+            other: The operator to compare to for approximate equality.
+            atol: Absolute numerical tolerance. The default behavior is to use ``self.atol``,
+                which would be 1e-8 unless changed by the user.
+
+        Returns:
+            True if the operators are equal up to numerical tolerance, False otherwise.
+        """
+        if not isinstance(other, type(self)):
+            return NotImplemented
         if atol is None:
             atol = self.atol
-        diff = (self - self.adjoint()).simplify(atol=atol)
+        diff = (self - other).simplify(atol=atol)
         return all(np.isclose(coeff, 0.0, atol=atol) for _, coeff in diff.to_list())
--- a/qiskit_nature/second_q/operators/second_quantized_op.py
+++ b/qiskit_nature/second_q/operators/second_quantized_op.py
@@ -76,6 +76,9 @@ class SecondQuantizedOp(StarAlgebraMixin, TolerancesMixin, ABC):
     def equiv(self, other: Any, atol: Optional[float] = None) -> bool:
         """Checks whether this operator is approximately equal to another operator.
 
+        Note that this method does not normal-order the operators before comparing them,
+        which may affect the result.
+
         Args:
             other: The operator to compare to for approximate equality.
             atol: Absolute numerical tolerance. The default behavior is to use ``self.atol``,
--- a/qiskit_nature/second_q/operators/second_quantized_op.py
+++ b/qiskit_nature/second_q/operators/second_quantized_op.py
@@ -83,11 +83,12 @@ class SecondQuantizedOp(StarAlgebraMixin, TolerancesMixin, ABC):
 
         Returns:
             True if the operators are equal up to numerical tolerance, False otherwise.
+
+        Raises:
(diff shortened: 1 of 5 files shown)


qiskit_nature/second_q/operators/second_quantized_op.py
```diff
@@ -71,7 +71,22 @@ def is_hermitian(self, atol: Optional[float] = None) -> bool:
...
+        Args:
+            other: The operator to compare to for approximate equality.
+            atol: Absolute numerical tolerance. The default behavior is to use ``self.atol``,
+                which would be 1e-8 unless changed by the user.
+
+        Returns:
+            True if the operators are equal up to numerical tolerance, False otherwise.
+        """
+        if not isinstance(other, type(self)):
+            return NotImplemented
```
mrossinek, line 88: This should `raise NotImplementedError` and provide both, the type of `self` and of `other`.

qiskit_nature/second_q/operators/second_quantized_op.py
```diff
@@ -71,7 +71,22 @@ def is_hermitian(self, atol: Optional[float] = None) -> bool:
         Returns:
             True if the operator is hermitian up to numerical tolerance, False otherwise.
         """
+        return self.equiv(self.adjoint(), atol=atol)
+
+    def equiv(self, other: Any, atol: Optional[float] = None) -> bool:
```
mrossinek, line 76: Since this class extends the `TolerancesMixin`, could we simply use this implementation for the `__eq__` method?
After all, that is what I would expect to happen for "tolerant" objects when comparing them...

(@woodsp-ibm: looping you in on this discussion)

qiskit_nature/second_q/operators/second_quantized_op.py
kevinsung: Done.

qiskit_nature/second_q/operators/second_quantized_op.py
kevinsung: I think `__eq__` should be reserved for "true" equality (comparing coefficients using `==`). This behavior would be consistent with the way `__eq__` is typically defined (e.g. for built-in types and numpy arrays). If `equiv` is added then making `__eq__` redirect to it would be redundant and also preclude adding the "true" equality I just described.

qiskit_nature/second_q/operators/second_quantized_op.py
kevinsung: Actually I realized it should be TypeError (see https://docs.python.org/3/library/exceptions.html) so I changed it.

qiskit_nature/second_q/operators/second_quantized_op.py
mrossinek: How do you suggest we implement "true" equality then? Comparing floats to an exact equality without some tolerance can be very flaky and simply checking whether two objects are actually identical instances is usually not what one is after when doing an `==` comparison.

To strengthen my argument I'd also like to point out, that this is exactly how `TolerancesMixin` is being used in Terra:
https://github.com/Qiskit/qiskit-terra/blob/316eaf5da1c3ff1c34d018c5aaca72e88a43483b/qiskit/quantum_info/states/statevector.py#L109-L112

https://github.com/Qiskit/qiskit-terra/blob/316eaf5da1c3ff1c34d018c5aaca72e88a43483b/qiskit/quantum_info/states/densitymatrix.py#L113-L116

kevinsung: > How do you suggest we implement "true" equality then?

I think this should work:
```python
def __eq__(self, other):
    return self.terms == other.terms
```

> Comparing floats to an exact equality without some tolerance can be very flaky

I agree that when dealing with floats approximate equality is generally more useful than exact equality, but there are still occasions to use exact equality. Note that `a == b` means exact equality when `a` and `b` are `floats`.

> simply checking whether two objects are actually identical instances is usually not what one is after when doing an == comparison.

I disagree. For all built-in types as well as Numpy arrays, `==` means exact equality.

> To strengthen my argument I'd also like to point out, that this is exactly how TolerancesMixin is being used in Terra:

As I pointed out above, Terra is deviating here from the behavior of all built-in types as well as Numpy, so I would argue that this is surprising behavior which we should not emulate.

mrossinek: Alright. I had another discussion on this with @woodsp-ibm offline.

We might add `__eq__` in the future, using true equality checking. However, we think that your new method here would be suitably named `isclose` rather than `equiv`. So please update the naming accordingly

qiskit_nature/second_q/operators/second_quantized_op.py
```diff
@@ -71,7 +71,25 @@ def is_hermitian(self, atol: Optional[float] = None) -> bool:
...
+        """Checks whether this operator is approximately equal to another operator.
```
mrossinek, line 77: Please also add a comment here (similar to the `induced_norm` one) that this method does not normal-order or simplify either side of the comparison, since, again, doing that could affect the outcome of this comparison

releasenotes/notes/approx-eq-15521c2b1ece996b.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Adds `.SecondQuantizedOp.equiv` for checking approximate equality
```
mrossinek, line 4: This naming here will need to be updated accordingly

test/second_q/operators/test_fermionic_op.py
```diff
@@ -367,6 +367,18 @@ def test_hermiticity(self):
             self.assertFalse(fer_op.is_hermitian(atol=1e-8))
             self.assertTrue(fer_op.is_hermitian(atol=1e-6))
 
+    def test_equiv(self):
```
mrossinek, line 370: So do all test method names and docstrings

qiskit_nature/second_q/operators/second_quantized_op.py
kevinsung: I'm fine with changing the name, but I just wanted you to be aware that I named it `equiv` to be consistent with `SparsePauliOp.equiv` which is the corresponding method on `SparsePauliOp`. Are you sure you want to break this consistency?

qiskit_nature/second_q/operators/second_quantized_op.py
kevinsung: Done (note the method actually does simplify).

qiskit_nature/second_q/operators/second_quantized_op.py
mrossinek: Thanks for clarifying, I had overlooked that.
Let's stick with `equiv` then :+1:

mrossinek (approved):
LGTM :+1:

woodsp-ibm (approved):

PR #720, opened 2022-06-23
Update protein_plotter.py

Qubit1718:
### Summary
I have been simulating several proteins and trying to generate their 3D structure. But unfortunately, I am facing an error that says 'Axes3DSubplot' object has no attribute 'set_box_aspect' I have also explored matplotlib's Axes and the only parameter set_box_aspect take is aspect and that should be either a float or none.  

### Details and comments
Here is the link https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_box_aspect.html#matplotlib.axes.Axes.set_box_aspect
I might not be correct by commenting the "ax_graph.set_box_aspect" line out, but I have also tried changing the given list [1,1,1] to a float value which didn't work either. I strongly feel the error is with Axes3Dplot. Please look into this. Thanks!!

### DIFF
--- a/qiskit_nature/results/utils/protein_plotter.py
+++ b/qiskit_nature/results/utils/protein_plotter.py
@@ -118,7 +118,7 @@ class ProteinPlotter:
             side_scatter: Scattering object that we will use for the legend.
         """
 
-        self._ax_graph.set_box_aspect([1, 1, 1])
+        #self._ax_graph.set_box_aspect([1,1,1])
 
         self._ax_graph.grid(grid)
 


mrossinek:
The method should in fact exist and should also take a tuple of length 3, as documented here: https://matplotlib.org/stable/api/_as_gen/mpl_toolkits.mplot3d.axes3d.Axes3D.html?highlight=set_box_aspect#mpl_toolkits.mplot3d.axes3d.Axes3D.set_box_aspect

Looks like this feature was added to `matplotlib` in version 3.3: https://github.com/matplotlib/matplotlib/commit/6915f697f444075681406ce6bfa4b2e908821ff7

What version of `matplotlib` do you have installed? As correctly documented in the setup.py (https://github.com/Qiskit/qiskit-nature/blob/49f1c33f11d89f5bd3b071fdaed0ee6e93b9a115/setup.py#L73), 3.3 is in fact the minimal supported version for Qiskit Nature.

PR #722, opened 2022-06-28
Migrate modules to new location

ElePT:
### Summary
This PR addresses epic #701, and is still a WIP, but I would like to start exposing my progress and getting my changes reviewed by the CI.

### Details and comments
So far, the migrated modules are:
-`qiskit_nature.operators`
-`qiskit_nature.mappers`

PR #726, opened 2022-06-29
Adds documentation for the default_filter_criterion

### LINKED ISSUE #643: `ElectronicStructureProblem.get_default_filter_criterion` assumes singlet spin

Currently, the `filter_criterion` produced by `ElectronicStructureProblem` assumes a singlet spin configuration without acknowledging this. The default could be improved in this regard. There are at least two approaches to do this:
- measure the number of alpha and beta particles separately
- compare the angular momentum observable with the expected value

The former would require some deeper changes as this will also affect what auxiliary operators need to be generated by the `ParticleNumber` property. The latter is more straight forward but would rely on the `AngularMomentum` property to always be present. This second case will also (to my knowledge) not be able to differentiate particle numbers `(2, 1)` from `(1, 2)`.

Anthony-Gandon:
Fix #643 

### Summary

Adds a note in the docstring of the default_filter_criterion() method to explicit that it is assuming singlet spin configuration.
An exemple of how a custom filter could be given for other configurations is also provided (it is only an attempt because I don't yet know how to ensure that the AngularMomentum property exists)

### Details and comments
```
def custom_filter_criterion(problem):
      def filter_criterion(problem, eigenstate, eigenvalue, aux_values):
          try:
              num_particles_aux = aux_values["ParticleNumber"][0]
              total_angular_momentum_aux = aux_values["AngularMomentum"][0]
          except TypeError:
              num_particles_aux = aux_values[0][0]
              total_angular_momentum_aux = aux_values[1][0]
          particle_number = cast(
              ParticleNumber,
              problem.grouped_property_transformed.get_property(ParticleNumber),
          )
          total_angular_momentum = cast(
              AngularMomentum,
              problem.grouped_property_transformed.get_property(AngularMomentum),
          )
          return np.isclose(
              total_angular_momentum.spin, total_angular_momentum_aux
          ) and np.isclose(
              particle_number.num_alpha + particle_number.num_beta, num_particles_aux
          )

      return partial(filter_criterion, problem)

```

At the moment and with a simple setup for the problem, I'm getting `None` when trying to access the AngularMomentum property.

```
qubit_op = QubitConverter(mapper=JordanWignerMapper(),
                                               two_qubit_reduction=False,
                                               z2symmetry_reduction=None)
driver = PySCFDriver(atom="H .0 .0 .0; H .0 .0 0.75",
                     unit=UnitsType.ANGSTROM,
                     charge=0,
                     spin=0,
                     basis="sto3g")
esp = ElectronicStructureProblem(driver)
new_numpy_solver = NumPyEigensolverFactory(filter_criterion=custom_filter_criterion(esp))
new_numpy_excited_states_calculation = ExcitedStatesEigensolver(qubit_op, new_numpy_solver)

```

How could I setup a working test for this?


qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
```diff
@@ -175,6 +175,48 @@ def get_default_filter_criterion(
 
         In the fermionic case the default filter ensures that the number of particles is being
         preserved.
+
+        .. note::
+
+            The default filter_criterion assumes a singlet spin configuration. (the number of alpha
+            particles is equal to the number of beta particles).
```
mrossinek, line 182:
```suggestion
            The default filter_criterion assumes a singlet spin configuration. This means, that the
            number of alpha-spin electrons is equal to the number of beta-spin electrons.
```

qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
```diff
@@ -175,6 +175,48 @@ def get_default_filter_criterion(
...
+            If the AngularMomentum property is available, then working with non singlet spin
+            configuration can be done using a custom filter_criterion in the following manner:
```
mrossinek, line 184:
```suggestion
            If the AngularMomentum property is available, one can correctly filter a non-singlet spin
            configuration with a custom `filter_criterion` similar to the following:
```
Please also add a link to the `AngularMomentum` class (see some other docstrings which use the `:class:` syntax on how to do this)

qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
```diff
@@ -175,6 +175,48 @@ def get_default_filter_criterion(
...
+                        num_particles_aux = aux_values[0][0]
+                        total_angular_momentum_aux = aux_values[1][0]
+                    particle_number = cast(
+                        ParticleNumber,
+                        problem.grouped_property_transformed.get_property(ParticleNumber),
+                    )
+                    total_angular_momentum = cast(
+                        AngularMomentum,
+                        problem.grouped_property_transformed.get_property(AngularMomentum),
+                    )
```
mrossinek, line 212: I think we should keep this example simpler and just encode the expected values as hard-coded variables. For example like so:
```python
expected_spin = 2
expected_num_electrons = 6
```

mrossinek: This would considerably simplify this example in the documentation

qiskit_nature/problems/second_quantization/electronic/electronic_structure_problem.py
```diff
@@ -175,6 +175,28 @@ def get_default_filter_criterion(
 
         In the fermionic case the default filter ensures that the number of particles is being
         preserved.
+
+        .. note::
```
mrossinek, line 179: I think this docstring is good now. However, I am not yet sure about its location. I do agree, that this is related to the `ElectronicStructureProblem`, but maybe this should be placed on the class docstring (instead of this method one) to give it a bit more visibility.
The example should also show how this is used in combination with the `NumPyEigensolver`

test/algorithms/excited_state_solvers/test_excited_states_solvers.py
```diff
@@ -151,6 +151,36 @@ def filter_criterion(eigenstate, eigenvalue, aux_values):
         for idx, energy in enumerate(self.reference_energies):
             self.assertAlmostEqual(computed_energies[idx], energy, places=4)
 
+    def test_custom_filter_criterion(self):
+        """Test NumPyEigenSolverFactory with ExcitedStatesEigensolver and a custom filter criterion"""
+
+        expected_spin = 0
```
mrossinek, line 157: Maybe we should instead set-up a non-singlet example for this test. For example `BeH` which is a doublet (`spin = 1`)

woodsp-ibm:
We will have to ensure this gets this into the relocated classes too from the refactor of #722

mrossinek (changes requested):
As previously mentioned by Steve, we have restructured the code base, meaning that this change now needs to be applied to the class located at `qiskit_nature/second_q/problems` and needs to leave the old code unchanged. A matching test location exists, too.

Other than that, this PR looks good to me :+1:

mrossinek (dismissed):
Thanks, @Anthony-Gandon!

mrossinek (dismissed):

qiskit_nature/second_q/problems/electronic_structure_problem.py
```diff
@@ -43,6 +43,35 @@ class ElectronicStructureProblem(BaseProblem):
...
     again when the problem is being solved.
+
+    In the fermionic case the default filter ensures that the number of particles is being
+    preserved.
+
+        .. note::
+
+            The default filter_criterion assumes a singlet spin configuration. This means, that the
+            number of alpha-spin electrons is equal to the number of beta-spin electrons.
+            If the :class:`~qiskit_nature.properties.second_quantization.electronic.AngularMomentum`
```
woodsp-ibm, line 54: Can you fix up this reference to the new place in second_q

Anthony-Gandon: Done

mrossinek (approved):

qiskit_nature/second_q/problems/electronic_structure_problem.py
```diff
@@ -43,6 +43,35 @@ class ElectronicStructureProblem(BaseProblem):
     The attributes `num_particles` and `num_spin_orbitals` are only available _after_ the
     `second_q_ops()` method has been called! Note, that if you do so, the method will be executed
     again when the problem is being solved.
+
+    In the fermionic case the default filter ensures that the number of particles is being
```
woodsp-ibm, line 47: The way this comes out leaves the code rather narrow 
image

I am wondering of the indenting of note and code-block is somehow ending up with the narrower areas we see. Clearly the note and code blocks are not indented in the html. Can you see if we can improve things a little?

Anthony-Gandon: I fixed the overall indenting of the block by suppressing a degree of indentation for the note+code in the doctoring. Thanks

qiskit_nature/second_q/problems/electronic_structure_problem.py
```diff
@@ -43,6 +43,35 @@ class ElectronicStructureProblem(BaseProblem):
...
+
+        expected_spin = 2
+        expected_num_electrons = 6
+
+        def filter_criterion_custom(eigenstate, eigenvalue, aux_values):
+            num_particles_aux = aux_values["ParticleNumber"][0]
+            total_angular_momentum_aux = aux_values["AngularMomentum"][0]
+
+            return np.isclose(expected_spin, total_angular_momentum_aux) and \
+                   np.isclose(expected_num_electrons, num_particles_aux)
```
mrossinek, line 71:
```suggestion
            return (
                np.isclose(expected_spin, total_angular_momentum_aux) and
                np.isclose(expected_num_electrons, num_particles_aux)
            )
```
Maybe we can try this? Sphinx appears to perform the line concatenation of `\` resulting in the following rendering in the docs:
screenshot_1659601278

mrossinek (commented):
Other than this, this LGTM :+1:

mrossinek (approved):
Thanks, Anthony!

woodsp-ibm (approved):

Issue #701, opened 2022-06-20
[refactor] Towards a non-driver focused future


mrossinek:
### What should we add?

This Epic outlines the plan for the upcoming restructuring of the entire Qiskit Nature code base.
Separate issues will be created outlining the refactoring plan for each existing module in more
detail. (Some modules/files will not be migrated. See at the bottom of this description.)

## Vision
We are undertaking a step *away* from the existing control flow in which Qiskit Nature uses
`drivers` to handle the electronic structure integral generation by calling other classical codes.
Instead, the plan is to design a code- and language-agnostic description (implemented with HDF5) of
a problem for Qiskit Nature to solve, which the classical code can generate directly.
In this way, the responsibility over the control flow of a calculation is handed over to the
classical codes rather than remaining within Qiskit Nature.

This has multiple benefits, a major one being the immediate availability of iterative embedding
techniques already implemented in classical codes (e.g. CASSCF). In Qiskit Nature's current design,
only CASCI calculations are possible and an implementation of CASSCF would require a custom `driver`
class for each classical code, which is not scalable for the future.

Another issue will be opened in the near future outlining this vision in more detail and discussing
the planned HDF5 file structure.

## Code Structure

With such a drastic change in design, we are taking this opportunity to restructure the code base.
The following structure roughly outlines the intended new structure:
```
qiskit_nature/
    second_quantization/
        algorithms/
            excited_state_solvers/
            ground_state_solvers/
            initial_points/
        circuit/
            library/
                ansatzes/
                initial_states/
        operators/
        mappers/
        hamiltonians/
        properties/
        problems/
        transformers/
        drivers/
```
A more detailed overview including the main classes is given by the following UML:

![overview](https://user-images.githubusercontent.com/21973473/180176491-0eec6a90-ee8a-4704-9cac-39aff64e2009.png)

## Workflow

Someone already familiar with Qiskit Nature, should not have much trouble understanding the
relations outlined by the following class diagram:

![relations](https://user-images.githubusercontent.com/21973473/180178506-0988d738-9066-477d-8263-60777474bd2b.png)

Together with this pseudo code, this should roughly outline the intended workflow:
```python
from qiskit_nature.second_quantization.algorithms.ground_state_solvers import (
    GroundStateEigensolver,
    VQEUCCFactory,
)
from qiskit_nature.second_quantization.mappers import JordanWignerMapper, QubitConverter
from qiskit_nature.second_quantization.drivers import HDF5Driver

[...]

PR #728, opened 2022-06-30
Allow driver result to be passed to solver

nathanschnitzerNNL:
The changes in this commit allow for a previously calculated driver
result to be passed to the eigensolver for it to get information on the
molecule. In its current state, the driver is run regardless of whether
or not the result is already known.

As the size of the models increase, it is important that expensive operations
like driver.run() are used only when necessary and avoided when
possible. This change introduces the framework for that.


mrossinek:
Thanks for the attempted contribution. This change will actually be superseded by the implementation proposed in #701.
There, what we now call a `Problem` will effectively become the output of a `Driver`.

Issue #91, opened 2021-03-22
Implement an equivalent to legacy `succ_full` option


mrossinek:
### What is the expected enhancement?

The [legacy `UCCSD`](https://github.com/Qiskit/qiskit-aqua/blob/main/qiskit/chemistry/components/variational_forms/uccsd.py) implementation had a `method_doubles="succ_full"` option which is not yet implemented in Qiskit Nature.
This option is a variant of the (now named) `SUCCD` option which also includes symmetrically mirrored double excitations, however, using the same parameter as the symmetric counter part (as also noted [here](https://github.com/Qiskit/qiskit-nature/blob/aae9c01b4d02bf5dbaa342c757be2cfef60a31d5/qiskit_nature/circuit/library/ansatzes/succd.py#L34-L35)).

The original implementation in Qiskit Aqua came from the work of [this paper](https://aip.scitation.org/doi/10.1063/1.5141835) in which sections II.E.2 and appendix E provide more in-depth information.
The following Qiskit Nature snippet shows the current state of the `SUCCD` ansatz for the example discussed the aforementioned paper's appendix:
```python
from qiskit_nature.circuit.library.ansatzes import SUCCD
from qiskit_nature.converters.second_quantization import QubitConverter
from qiskit_nature.mappers.second_quantization import JordanWignerMapper

converter = QubitConverter(JordanWignerMapper())
ansatz = SUCCD(converter, (1, 1), 6)

print(ansatz.decompose().draw())
print(ansatz.excitation_list)
```
which outputs:
```
     ┌─────────────────────────────────────────────────────────────────────────────────────────────┐»
q_0: ┤0                                                                                            ├»
     │                                                                                             │»
q_1: ┤1                                                                                            ├»
     │                                                                                             │»
q_2: ┤2                                                                                            ├»
     │  exp(-it (IYYIXY + IXXIXY + IXYIYY + IYXIYY + IXYIXX + IYXIXX + IYYIYX + IXXIYX))(1.0*t[0]) │»
q_3: ┤3                                                                                            ├»
     │                                                                                             │»
q_4: ┤4                                                                                            ├»
     │                                                                                             │»
q_5: ┤5                                                                                            ├»
     └─────────────────────────────────────────────────────────────────────────────────────────────┘»
«     ┌─────────────────────────────────────────────────────────────────────────────────────────────┐»
«q_0: ┤0                                                                                            ├»
«     │                                                                                             │»
«q_1: ┤1                                                                                            ├»
[...]

PR #729, opened 2022-07-01
Add the `mirror` setting to the `SUCCD` ansatz

YuNariai:
### Summary
Implement succ_full for issue #91 

### Details and comments
Add SUCC_full option into SUCCD.py.  I did the following things:

I added a init arg: mirror. By setting `mirror=True` we can use SUCC_full ansatz, otherwise we will use SUCCD ansatz.
I added a property "mirror" so that `self.mirror` shows whether we are using SUCC_full.
I modified the `excitation_list` in SUCC (and also in UCC)

I changed the function "generate_excitations" so that it will generate SUCC_full excitation list when `mirror=True`.
I added function `_build_fermionic_excitation_ops` (based on the function "_build_fermionic_excitation_ops" in UCC) so that the number of parameters remain the same.

I add new tests.


qiskit_nature/circuit/library/ansatzes/succ_full.py
```diff
@@ -0,0 +1,208 @@
...
+                second_exc[0],
+                second_exc[1],
+            )
+            # add the excitation tuple
+            occ: Tuple[int, ...]
+            unocc: Tuple[int, ...]
+            occ, unocc = zip(alpha_exc, beta_exc)
+            exc_tuple = (occ, unocc)
+            excitations.append(exc_tuple)
+            logger.debug("Added the excitation: %s", exc_tuple)
```
mrossinek, line 162: This entire part is actually equivalent to the following:
```suggestion
        excitations.extend(
            generate_fermionic_excitations(
                2,
                num_spin_orbitals,
                num_particles,
            )
        )
```

qiskit_nature/circuit/library/ansatzes/succ_full.py
```diff
@@ -0,0 +1,208 @@
...
+        specified number of particles distributed in the number of orbitals.
+        Args:
+            excitations: the list of excitations.
+        Returns:
+            The list of excitation operators in the second quantized formalism.
+        """
+        operators = []
+        excitations_dictionary: Dict[int, List[Tuple[Tuple[int, ...], Tuple[int, ...]]]] = {}
+        for exc in excitations:
+            exc_level = sum(exc[1])
```
mrossinek, line 187: I think this is the key aspect of this PR. I think this would be worth an explanatory comment and we may be able to find a better name for this variable. The idea here is, that `sum(exc[1])` sums over the indices of the virtual orbitals in which we excite into. Thus, the level of an excitations is indicated by this sum.
Maybe add a comment explaning this.

qiskit_nature/circuit/library/ansatzes/succ_full.py
```diff
@@ -0,0 +1,208 @@
...
+                for occ in exc[0]:
+                    label[occ] = "+"
+                for unocc in exc[1]:
+                    label[unocc] = "-"
+                op = FermionicOp("".join(label), display_format="dense")
+                op -= op.adjoint()
+                # we need to account for an additional imaginary phase in the exponent (see also
+                # `PauliTrotterEvolution.convert`)
+                op *= 1j  # type: ignore
+                final_op += op
```
mrossinek, line 205: The other key difference compared to the normal `UCC` is now that you sum the excitations of identical levels into one operator, resulting in them being associated to a single parameter in the parameterized ansatz. A very nice solution indeed :+1: 

I was actually hoping, that we could make this `"full"` variant of `SUCC` a keyword option of the `SUCCD` class, since these are quite related in that sense. This would remove the need for largely duplicating the rest of this class. Do you think you could update this PR to attempt such a change instead?

releasenotes/notes/implement-succ-full-ansatz-c10ac025a3af424c.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+      from qiskit_nature.circuit.library.ansatzes import SUCCfull
+      from qiskit_nature.converters.second_quantization import QubitConverter
+      from qiskit_nature.mappers.second_quantization import JordanWignerMapper
+      converter = QubitConverter(JordanWignerMapper())
+      ansatz = SUCCfull(converter, (1, 1), 6)
+
+issues:
+  - |
+    Need tests
```
mrossinek, line 19:
```suggestion
```
This is not something that goes into the release note

qiskit_nature/circuit/library/ansatzes/succ_full.py
```diff
@@ -0,0 +1,208 @@
...
+        operators = []
+        excitations_dictionary: Dict[int, List[Tuple[Tuple[int, ...], Tuple[int, ...]]]] = {}
+        for exc in excitations:
+            exc_level = sum(exc[1])
+            if exc_level in excitations_dictionary:
+                excitations_dictionary[exc_level].append(exc)
+            else:
+                excitations_dictionary[exc_level] = [exc]
+        for exc_level, exc_level_items in excitations_dictionary.items():
+            final_op = FermionicOp.zero(register_length=self.num_spin_orbitals)
```
mrossinek, line 193: The way you initialize this, will actually result in an additional term which I would like to avoid. Maybe you can gather the individual ops in a list instead, which you then sum over?
E.g.
```python
ops = []
for exc in exc_level_items:
    # ...
    ops.append(op)
operators.append(sum(ops))
```

mrossinek (changes requested):
Thank you for this PR! The functionality here seems to be correct, and I like the solution that you went for :+1: However, I would like to try to integrate this into the `SUCCD` class, instead of another subclass. To me at least, this is a similar configuration of the `SUCCD` as `generalized=True` (obviously a different outcome, but a similar vain).

Having said that, here is a working test case for your current implementation:
```python

    @unpack
    @data(
        (
            6,
            (1, 1),
            [
                FermionicOp([("+-I+-I", 1j), ("-+I-+I", -1j)], display_format="dense"),
                FermionicOp([("+-I+I-", 1j), ("-+I-I+", -1j), ("+I-+-I", 1j), ("-I+-+I", -1j)], display_format="dense"),
                FermionicOp([("+I-+I-", 1j), ("-I+-I+", -1j)], display_format="dense"),
            ],
        ),
    )
    def test_succ_full(self, num_spin_orbitals, num_particles, expect):
        """Tests the generalized SUCCD Ansatz."""

        ansatz = SUCCfull(
            qubit_converter=converter,
            num_particles=num_particles,
        )

        assert_ucc_like_ansatz(self, ansatz, num_spin_orbitals, expect)
```
If you integrate this into the `SUCCD` class, you can also add this test case to the corresponding test file :+1:

qiskit_nature/circuit/library/ansatzes/succ_full.py
YuNariai: Thank you very much for the advices! I just want to make sure, I am going to add one more input in SUCCD named "full",  if "full=False", it gives SUCCD, if "full=True", it gives SUCC_full. Am I right?

mrossinek: Yes, in principal this is correct. I will think about whether we can find a better name than `full`, but for now, please go ahead and implement it like this :+1:

qiskit_nature/circuit/library/ansatzes/succ_full.py
YuNariai: Hi morssinek, for this part, I would like to keep the code I used. Because I tried "**generate_fermionic_excitations( 2, ...)** "but it returns wrong excitation when there are two or more alpha/beta particles. For example, **num_particles,num_spin_orbitals=(2,2), 8**, the excitation_list will include **((0,1),(2,3))** the two excited particles are both alpha particles, which I think is wrong.

qiskit_nature/circuit/library/ansatzes/succ_full.py
YuNariai: I have problem with this part: To avoid the additional term, I used the code you showed here. When I run "make mypy" , I got an error: **Incompatible return value type (got "List[Union[FermionicOp, int]]", expected "List[FermionicOp]")**. I have no idea where the int come from.

woodsp-ibm: In line 233 you have this `operators = []` which has no type. Then in the return, after appending things to this, you have different return paths. If not operators path is clearly a List of FermionicOps. The other path is operators which is this Literal. So in total it thinks the return is a Union of Fermionic and Literal list - but the return type hint on the method definition says FermionicOp list. If you change line 233 to `operators: List[FermionicOp] = []` so you give it the type upfront then I think that will work.


You also have another error related to `pool`. mypy picks up on the type of a local variable from its first use. When you use the same variable below it complains because the type being assigned does not match what the type it thinks the var is. Easiest fix if to call the second pool some other name e.g pool_full in the second case and do for exc in pool_full there. Honestly I am not sure why we even need the local var pool as it used only once so `for exc in itertools.product(....` in each case would get rid of it completely and you would not have the problem with how a local var is used.

qiskit_nature/circuit/library/ansatzes/succ_full.py
mrossinek: Hm you're right that these excitations should be excluded from SUCCD. However, you can already achieve that with the following:
```python
generate_fermionic_excitations(
    2,
    num_spin_orbitals,
    num_particles,
    max_spin_excitation=1,
)
```

qiskit_nature/circuit/library/ansatzes/succd.py
```diff
@@ -37,18 +39,13 @@
 
 class SUCCD(UCC):
     """The SUCCD Ansatz.
-
```
mrossinek, line 40: Please revert the removal of these deleted lines

qiskit_nature/circuit/library/ansatzes/succd.py
```diff
@@ -77,12 +74,15 @@ def __init__(
                 the occupation of the spin orbitals. As such, the set of generalized excitations is
                 only determined from the number of spin orbitals and independent from the number of
                 particles.
-
+            full: boolean flag whether or not to use SUCC_full ansatz, which is a variant of SUCCD
+                ansatz that including also the symmetrically mirrored double excitations, but
+                assigns the same circuit parameter to them.
```
mrossinek, line 79:
```suggestion
            full: boolean flag whether or not to include the symmetrically mirrored double
                  excitations, while keeping the original number of circuit parameters.
                  This results in mirrored excitations having identical parametrization values.
```

qiskit_nature/circuit/library/ansatzes/succd.py
```diff
@@ -61,9 +58,9 @@ def __init__(
         initial_state: Optional[QuantumCircuit] = None,
         include_singles: Tuple[bool, bool] = (False, False),
         generalized: bool = False,
+        full: bool = False,
```
mrossinek, line 61: Maybe `mirror` is a better naming option :thinking:

qiskit_nature/circuit/library/ansatzes/succd.py
```diff
@@ -106,27 +106,43 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
...
+        """Sets whether to use SUCC_full."""
+        self._full = full
+
+    @property
+    def excitation_list(self) -> List[Tuple[Tuple[int, ...], Tuple[int, ...]]]:
+        """The excitation list that SUCC is using, in the required format."""
+        return self.generate_excitations(
+            self.num_spin_orbitals,
+            self.num_particles,
+        )
```
mrossinek, line 125: There should be no need for this code. `self.generate_excitations` is already injected as the excitation generator during `__init__`.

qiskit_nature/circuit/library/ansatzes/succd.py
```diff
@@ -106,27 +106,43 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
...
+
     def generate_excitations(
         self, num_spin_orbitals: int, num_particles: Tuple[int, int]
     ) -> List[Tuple[Tuple[int, ...], Tuple[int, ...]]]:
         """Generates the excitations for the SUCCD Ansatz.
-
```
mrossinek, line 113: Again, please revert these docstring changes

qiskit_nature/circuit/library/ansatzes/succd.py
```diff
@@ -179,3 +221,53 @@ def _validate_num_particles(self, num_particles):
...
+            # exc_level is a 4-number string. First two numbers are occupied indices. And
+            # last two numbers indicate the indices changing during excitation. Thus,
+            # the level of an excitations is indicated by this string. The symmetrically
+            # mirrored double excitations have the same exc_level string, and the
+            # excitations with the same level will be assigned the same parameter.
+
+            if exc_level in excitations_dictionary:
+                excitations_dictionary[exc_level].append(exc)
+            else:
+                excitations_dictionary[exc_level] = [exc]
```
mrossinek, line 253:
```suggestion
            excitations_dictionary[exc_level].append(exc)
```

qiskit_nature/circuit/library/ansatzes/succd.py
```diff
@@ -179,3 +221,53 @@ def _validate_num_particles(self, num_particles):
...
+    def _build_fermionic_excitation_ops(self, excitations: Sequence) -> List[FermionicOp]:
+        """Builds all possible excitation operators with the given number of excitations for the
+        specified number of particles distributed in the number of orbitals.
+        Args:
+            excitations: the list of excitations.
+        Returns:
+            The list of excitation operators in the second quantized formalism.
+        """
+        operators = []
+        excitations_dictionary: Dict[str, List[Tuple[Tuple[int, ...], Tuple[int, ...]]]] = {}
```
mrossinek, line 234: If you make this a `defaultdict(list)` (using `from collections import defaultdict`), you can apply the previous suggestion

qiskit_nature/circuit/library/ansatzes/succd.py
```diff
@@ -179,3 +221,53 @@ def _validate_num_particles(self, num_particles):
...
+                str(exc[0][0])
+                + str(exc[0][1])
+                + str(abs((exc[1][0] - exc[0][0]) - (exc[1][-1] - exc[0][-1])))
+                + str(exc[1][0] - exc[0][0] + exc[1][-1] - exc[0][-1])
+            )
+            # exc_level is a 4-number string. First two numbers are occupied indices. And
+            # last two numbers indicate the indices changing during excitation. Thus,
+            # the level of an excitations is indicated by this string. The symmetrically
+            # mirrored double excitations have the same exc_level string, and the
+            # excitations with the same level will be assigned the same parameter.
```
mrossinek, line 248: This is an interesting solution. However, some updates will be required:
- this will break when `include_singles=True`. So you will need to guard against that (ideally add a unittest for this)
- could you maybe rephrase this comment in form of an equation that is a bit easier to read?

mrossinek (changes requested):
Besides my above in this review, yesterday we merged #746. This restructured the code in a major way and means you need to update your PR accordingly.
To do so, please apply your changes on the class located at `qiskit_nature/second_q/circuit/library/ansatzes` and leave the code in the old location unchanged. The matching location also exists in the test folder.

EDIT: I am also adding a link to this comment, so it does not get overlooked: https://github.com/Qiskit/qiskit-nature/pull/729#discussion_r924091022

qiskit_nature/circuit/library/ansatzes/succd.py
YuNariai: I kept this part, because if I remove it, then "**self.excitation_list**" will return SUCCD excitation list even we set "**full=True**"

mrossinek: This is actually not true. Although it looks surprisingly similar (simply because the length matches what you expected).
However, you are running into a different issue here. During the operator conversion from second-quantization to qubit space, the UCC class filters the operators based on their commutation properties: https://github.com/Qiskit/qiskit-nature/blob/b0be0fbe4ece34a643339a1d838a0a86e752a8b1/qiskit_nature/second_q/circuit/library/ansatzes/ucc.py#L287-L300

More specifically, the `zip(...)` operation on this line: https://github.com/Qiskit/qiskit-nature/blob/b0be0fbe4ece34a643339a1d838a0a86e752a8b1/qiskit_nature/second_q/circuit/library/ansatzes/ucc.py#L297
is what is causing issues for you, because your `operators` has length 3, whereas `self._excitation_list` has length 4 (for the specific test case in your unittest). This results in the last excitation being dropped.

I don't have a perfect solution right now, but I believe what needs to be done here, is that this validation logic needs to be extracted into a private method. `SUCCD` can then overwrite this method and use the `excutations_dictionary` instead of `self._excitation_list`, since the former should have the correct length again.

qiskit_nature/circuit/library/ansatzes/succd.py
mrossinek: Pinging this comment again. Please restore this entire file to the status of `main`

test/second_q/circuit/library/ansatzes/test_succd.py
```diff
@@ -150,3 +150,31 @@ def test_puccd_ansatz_generalized(self, num_spin_orbitals, num_particles, expect
...
+                FermionicOp([("+-I+-I", 1j), ("-+I-+I", -1j)], display_format="dense"),
+                FermionicOp(
+                    [("+-I+I-", 1j), ("-+I-I+", -1j), ("+I-+-I", 1j), ("-I+-+I", -1j)],
+                    display_format="dense",
+                ),
+                FermionicOp([("+I-+I-", 1j), ("-I+-I+", -1j)], display_format="dense"),
+            ],
+        ),
+    )
+    def test_succ_full(self, num_spin_orbitals, num_particles, expect):
```
mrossinek, line 169: Please add another test case which checks the compatibility with `include_singles=True`

qiskit_nature/circuit/library/ansatzes/succd.py
YuNariai: If I keep the code showing here, that means I override **self._excitation_list**  in SUCCD, so it doesn't have to have the same length as **operators**. In this way, **self.excitation_list** will return the correct list for both SUCCD and SUCC_full. Do you think this is a feasible solution?

qiskit_nature/circuit/library/ansatzes/succ_full.py
YuNariai: Thanks for the solution! Now `operators` is fine, but I face the same problem on `ops`. No matter I use `ops = []` or , `ops: List[FermionicOp] = []`, it warns **Argument 1 to "append" of "list" has incompatible type "int"; expected "FermionicOp"  [arg-type]**

woodsp-ibm: That's from line 264 i.e. `operators.append(sum(ops))` according to CI. `operators` is a List[FermionicOp] and as you indicate so is ops. But I imagine `sum(ops)` perhaps mypy imagines is summing to an int whereas this adds FermionicOps together right for a single FermionicOp. Maybe use the typing cast construct to inform mypy that the type from sum(ops) is a FermionicOp and not an int. At least this is my best guess!

qiskit_nature/circuit/library/ansatzes/succd.py
mrossinek: I would think the better approach would be, that you slightly refactor the `UCC` class and extract that validation logic into a separate private method.
In `SUCCD` you can then overwrite this validation logic, to use your internal excitation dictionary instead of the default validation logic

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -37,18 +40,13 @@
 
 class SUCCD(UCC):
     """The SUCCD Ansatz.
-
```
mrossinek, line 40: Please revert the removal of the empty lines in this docstring

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -106,27 +106,43 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
         """Sets whether to include single excitations."""
         self._include_singles = include_singles
 
+    @property
+    def mirror(self) -> bool:
+        """Whether use SUCC_full"""
```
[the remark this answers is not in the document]
mrossinek: Pinging this comment, so its back on your radar

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -106,27 +107,42 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
...
+            self.num_spin_orbitals,
+            self.num_particles,
+        )
+        self.operators = operators
+
     def generate_excitations(
         self, num_spin_orbitals: int, num_particles: Tuple[int, int]
     ) -> List[Tuple[Tuple[int, ...], Tuple[int, ...]]]:
         """Generates the excitations for the SUCCD Ansatz.
-
```
mrossinek, line 113: Please revert the changes to this docstring (in general, please do not change unrelated docstrings)

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -146,27 +161,40 @@ def generate_excitations(
...
-        for exc in pool:
-            # find the two excitations (Note: SUCCD only works for double excitations!)
-            alpha_exc, second_exc = exc[0], exc[1]
-            # shift the second excitation into the beta-spin orbital index range
-            beta_exc = (
-                second_exc[0] + beta_index_shift,
-                second_exc[1] + beta_index_shift,
+        if self._mirror:
+
```
mrossinek, line 165: Maybe you could add a short explanatory comment here, which shows the difference between the `if self._mirror` and the `else` case

qiskit_nature/second_q/circuit/library/ansatzes/ucc.py
```diff
@@ -293,17 +292,22 @@ def operators(self):  # pylint: disable=invalid-overridden-method
...
-
-                self._excitation_list = valid_excitations
-                self.operators = valid_operators
+                self._check_length_excitations_list(
+                    operators=operators, excitation_list=self._excitation_list
+                )
 
         return super(UCC, self.__class__).operators.__get__(self)
 
+    def _check_length_excitations_list(self, operators, excitation_list):
```
mrossinek, line 301:
```suggestion
    def _filter_operators(self, operators, excitation_list):
```
This method is really filtering the operators as explained in the comment a few lines above the original code:
> The converted operators are maintained at the same index by the converter
> inserting ``None`` as the result if an operator did not commute. To ensure that
> the ``excitation_list`` is transformed identically to the operators, we retain
> ``None`` for non-commuting operators in order to manually remove them in unison.

This naming reflects better, that this method actually changes objects on `self`, which would have been surprising to me in the old name.

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -106,27 +107,42 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
...
+    def mirror(self, mirror: bool) -> None:
+        """Sets whether to use SUCC_full."""
+        self._mirror = mirror
+
+    def _check_length_excitations_list(self, operators, excitation_list):
+        self._excitation_list = self.generate_excitations(
+            self.num_spin_orbitals,
+            self.num_particles,
+        )
+        self.operators = operators
```
mrossinek, line 125: You're not actually performing any of the filtering here. You should be doing the equivalent operation to the one implemented in `UCC`. However, since this class now behaves differently as part of `_build_fermionic_excitation_ops`, so should this method.

To be more concrete:
- there is no need to call `self.generate_excitations` here. You already get that list as an input
- that said, you actually need to use the `exctiations_dictionary` which you are building out during `_build_fermionic_excitation_ops` (that means you need to store it as `self._excitations_dict` in order to be able to access it here)
- you then need to do a similar `zip(...)` operation as in the `UCC` class, but using the dictionary values instead

If this is not clear, please let me know and I will try to provide a pseudocode for you

releasenotes/notes/add-5b9f14b11d8992b9.yaml
```diff
@@ -0,0 +1,12 @@
+---
+prelude: >
+
```
mrossinek, line 3:
```suggestion
```

releasenotes/notes/add-5b9f14b11d8992b9.yaml
```diff
@@ -0,0 +1,12 @@
+---
+prelude: >
+
+features:
+  - |
+    Adds succ_full ansatz option to ``SUCCD`` objects. It can be used in the ``SUCCD`` function,
```
mrossinek, line 6: Please rephrase as to not mention `succ_full` but actually talk about the new `mirror` keyword argument

releasenotes/notes/add-5b9f14b11d8992b9.yaml
```diff
@@ -0,0 +1,12 @@
+---
+prelude: >
+
+features:
+  - |
+    Adds succ_full ansatz option to ``SUCCD`` objects. It can be used in the ``SUCCD`` function,
+    for example::
```
mrossinek, line 7:
```suggestion
   .. code-block:: python
   
```

test/second_q/circuit/library/ansatzes/test_succd.py
```diff
@@ -150,3 +150,131 @@ def test_puccd_ansatz_generalized(self, num_spin_orbitals, num_particles, expect
...
```
mrossinek, line 169: Please replace `succ_full` with `succ_mirror` in all test names and update the docstrings accordingly

test/second_q/circuit/library/ansatzes/test_succd.py
```diff
@@ -150,3 +150,131 @@ def test_puccd_ansatz_generalized(self, num_spin_orbitals, num_particles, expect
...
+                FermionicOp(
+                    [("+-I+I-", 1j), ("-+I-I+", (-0 - 1j)), ("+I-+-I", 1j), ("-I+-+I", (-0 - 1j))],
+                    display_format="dense",
+                ),
+                FermionicOp([("+-II+-", 1j), ("-+II-+", (-0 - 1j))], display_format="dense"),
+                FermionicOp([("+I-+I-", 1j), ("-I+-I+", (-0 - 1j))], display_format="dense"),
+                FermionicOp([("+I-I+-", 1j), ("-I+I-+", (-0 - 1j))], display_format="dense"),
+                FermionicOp([("I+-+-I", 1j), ("I-+-+I", (-0 - 1j))], display_format="dense"),
+                FermionicOp([("I+-+I-", 1j), ("I-+-I+", (-0 - 1j))], display_format="dense"),
+                FermionicOp([("I+-I+-", 1j), ("I-+I-+", (-0 - 1j))], display_format="dense"),
```
mrossinek, line 264: This is actually not yet the desired behavior. The expected outcome should be this:
```suggestion
                FermionicOp([("+-I+-I", 1j), ("-+I-+I", (-0 - 1j))], display_format="dense"),
                FermionicOp(
                    [("+-I+I-", 1j), ("-+I-I+", (-0 - 1j)), ("+I-+-I", 1j), ("-I+-+I", (-0 - 1j))],
                    display_format="dense",
                ),
                FermionicOp(
                    [("+-II+-", 1j), ("-+II-+", (-0 - 1j)), ("I+-+-I", 1j), ("I-+-+I", (-0 - 1j))],
                    display_format="dense",
                ),
                FermionicOp([("+I-+I-", 1j), ("-I+-I+", (-0 - 1j))], display_format="dense"),
                FermionicOp(
                    [("+I-I+-", 1j), ("-I+I-+", (-0 - 1j)), ("I+-+I-", 1j), ("I-+-I+", (-0 - 1j))],
                    display_format="dense",
                ),
                FermionicOp([("I+-I+-", 1j), ("I-+I-+", (-0 - 1j))], display_format="dense"),
```

mrossinek (changes requested):

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
YuNariai: when I run `make black`, it automatically deletes the empty lines. Anyway I'll add the empty lines manually

YuNariai: I would like to add the empty lines back after fixing all the other problems, otherwise `make black` will delete them

mrossinek: I suggest that you update your version of `black` then. You can do so by doing `pip install -r requirements-dev.txt --upgrade` while your Python environment is activated and you are at the root of your Qiskit Nature clone

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -22,6 +23,8 @@
 from qiskit_nature import QiskitNatureError
 from qiskit_nature.second_q.mappers import QubitConverter
 
+from qiskit_nature.operators.second_quantization import FermionicOp
```
mrossinek, line 26:
```suggestion
from qiskit_nature.second_q.operators import FermionicOp
```

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -77,12 +75,16 @@ def __init__(
                 the occupation of the spin orbitals. As such, the set of generalized excitations is
                 only determined from the number of spin orbitals and independent from the number of
                 particles.
-
+            mirror: boolean flag whether or not to include the symmetrically mirrored double
+                  excitations (refer SUCC_full ansatz), while keeping the original number of circuit
```
mrossinek, line 79: Nowhere in the code, should there be a reference to "SUCC_full". This has no context and users do not know what this means.
If you want to refer to the old name in same way, I suggest that you add a sentence at the end of this parameter documentation stating something along the lines of this:
> Enabling this parameter will result in the SUCCD ansatz referred to as "q-UCCSD0-full" in reference [1].

And then you add the same reference link as in this class docstring so that `[1]` works.

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -106,27 +108,46 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
         """Sets whether to include single excitations."""
         self._include_singles = include_singles
 
+    @property
+    def mirror(self) -> bool:
+        """Whether include the symmetrically mirrored double excitations (SUCC_full)."""
```
mrossinek, line 113:
```suggestion
        """Whether include the symmetrically mirrored double excitations."""
```

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -106,27 +108,46 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
...
+        return self._mirror
+
+    @mirror.setter
+    def mirror(self, mirror: bool) -> None:
+        """Sets whether include the symmetrically mirrored double excitations (SUCC_full)."""
```
mrossinek, line 118:
```suggestion
        """Sets whether include the symmetrically mirrored double excitations."""
```

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -106,27 +108,46 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
...
+        self._mirror = mirror
+
+    def _filter_operators(self, operators):
+        excitation_dict = self._excitations_dict
+        valid_operators, valid_excitations = [], []
+        for op, ex in zip(operators, excitation_dict.values()):
```
mrossinek, line 124: This implementation now looks good. But maybe you can just do this:
```suggestion
        valid_operators, valid_excitations = [], []
        for op, ex in zip(operators, self._excitation_dict.values()):
```

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -146,27 +166,40 @@ def generate_excitations(
...
-        pool = itertools.combinations_with_replacement(alpha_excitations, 2)
-
-        for exc in pool:
-            # find the two excitations (Note: SUCCD only works for double excitations!)
-            alpha_exc, second_exc = exc[0], exc[1]
-            # shift the second excitation into the beta-spin orbital index range
-            beta_exc = (
-                second_exc[0] + beta_index_shift,
-                second_exc[1] + beta_index_shift,
+        if self._mirror:  # if include the symmetrically mirrored double excitations
```
mrossinek, line 169: That is not really what I meant. I rather meant a similarly extensive comment as inside the `else` statement which explains why in this case, we can use the "standard" implementation rather than needing to use `combinations_with_replacement` as explained there

qiskit_nature/second_q/circuit/library/ansatzes/ucc.py
```diff
@@ -293,17 +292,21 @@ def operators(self):  # pylint: disable=invalid-overridden-method
...
-                self._excitation_list = valid_excitations
-                self.operators = valid_operators
+                self._filter_operators(operators=operators)
 
         return super(UCC, self.__class__).operators.__get__(self)
 
+    def _filter_operators(self, operators):
+        excitation_list = self._excitation_list
+        valid_operators, valid_excitations = [], []
+        for op, ex in zip(operators, excitation_list):
```
mrossinek, line 302:
```suggestion
        valid_operators, valid_excitations = [], []
        for op, ex in zip(operators, self._excitation_list):
```

releasenotes/notes/add-5b9f14b11d8992b9.yaml
```diff
@@ -0,0 +1,11 @@
+---
+features:
+  - |
+    Adds new keyword argument "mirror" to ``SUCCD`` class.
+    .. code-block:: python
+    
+      from qiskit_nature.circuit.library.ansatzes import SUCCfull
+      from qiskit_nature.converters.second_quantization import QubitConverter
+      from qiskit_nature.mappers.second_quantization import JordanWignerMapper
```
mrossinek, line 9: Please correct these imports

test/second_q/circuit/library/ansatzes/test_succd.py
```diff
@@ -150,3 +150,135 @@ def test_puccd_ansatz_generalized(self, num_spin_orbitals, num_particles, expect
...
+                FermionicOp(
+                    [("+-I+I-", 1j), ("-+I-I+", -1j), ("+I-+-I", 1j), ("-I+-+I", -1j)],
+                    display_format="dense",
+                ),
+                FermionicOp([("+I-+I-", 1j), ("-I+-I+", -1j)], display_format="dense"),
+            ],
+        ),
+    )
+    def test_succ_mirror(self, num_spin_orbitals, num_particles, expect):
+        """Tests the succ_mirror Ansatz."""
```
mrossinek, line 170:
```suggestion
        """Tests the `mirror` option of the SUCCD Ansatz."""
```
Please update accordingly elsewhere

mrossinek (changes requested):

releasenotes/notes/add-5b9f14b11d8992b9.yaml
mrossinek: I also just noticed that the name of this file is rather non-descriptive. I'm not sure what `reno` is doing with those hashes at the end of the file names.. @manoelmarques is it safe to rename this file?

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -77,12 +75,18 @@ def __init__(
...
+                  parameters. This results in mirrored excitations having identical parameter values.
+                  Enabling this parameter will result in the SUCCD ansatz referred to as
+                  "q-UCCSD0-full" in reference [1].
         Raises:
             QiskitNatureError: if the number of alpha and beta electrons is not equal.
         """
         self._validate_num_particles(num_particles)
         self._include_singles = include_singles
+        self._mirror = mirror
+        self._excitations_dict: Dict[str, List[Tuple[Tuple[int, ...], Tuple[int, ...]]]] = None
```
mrossinek, line 89:
```suggestion
        self._excitation_dict: Dict[str, List[Tuple[Tuple[int, ...], Tuple[int, ...]]]] = None
```
You are not being consistent with this variable name. Make sure this is the same everywhere

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -106,27 +110,45 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
         """Sets whether to include single excitations."""
         self._include_singles = include_singles
 
+    @property
+    def mirror(self) -> bool:
+        """Whether include the symmetrically mirrored double excitations."""
```
mrossinek, line 115:
```suggestion
        """Whether to include the symmetrically mirrored double excitations."""
```

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -106,27 +110,45 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
...
+        return self._mirror
+
+    @mirror.setter
+    def mirror(self, mirror: bool) -> None:
+        """Sets whether include the symmetrically mirrored double excitations."""
```
mrossinek, line 120:
```suggestion
        """Sets whether to include the symmetrically mirrored double excitations."""
```

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -179,3 +213,66 @@ def _validate_num_particles(self, num_particles):
...
                 str(num_particles),
             ) from exc
+
+    def _build_fermionic_excitation_ops(self, excitations: Sequence) -> List[FermionicOp]:
+        """Builds all possible excitation operators with the given number of excitations for the
+        specified number of particles distributed in the number of orbitals.
+        Args:
+            excitations: the list of excitations.
+        Returns:
+            The list of excitation operators in the second quantized formalism.
```
mrossinek, line 223:
```suggestion
        """Builds all possible excitation operators with the given number of excitations for the
        specified number of particles distributed in the number of orbitals.
        
        Args:
            excitations: the list of excitations.
            
        Returns:
            The list of excitation operators in the second quantized formalism.
```

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -179,3 +213,66 @@ def _validate_num_particles(self, num_particles):
...
+        # corresponds to a parameter.
+        for exc in excitations:
+            alpha_occ = exc[0][0]  # alpha occupied indices
+            beta_occ = exc[0][
+                -1
+            ]  # beta occupied indices. If include singles, then beta_occ=alpha_occ
+            alpha_unocc = exc[1][0]  # alpha unoccupied indices
+            beta_unocc = exc[1][
+                -1
+            ]  # beta unoccupied indices. If include singles, then beta_unocc=alpha_unocc
```
mrossinek, line 241:
```suggestion
            alpha_occ = exc[0][0]
            # beta occupied indices. If include singles, then beta_occ=alpha_occ
            beta_occ = exc[0][-1]
            alpha_unocc = exc[1][0]
            # beta unoccupied indices. If include singles, then beta_unocc=alpha_unocc
            beta_unocc = exc[1][-1]
```

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -179,3 +213,66 @@ def _validate_num_particles(self, num_particles):
...
+            alpha_exc = int(
+                str(alpha_occ) + str(alpha_unocc)
+            )  # alpha occupied and unoccupied indice
```
mrossinek, line 244:
```suggestion
            )  # alpha occupied and unoccupied indices
```

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -179,3 +213,66 @@ def _validate_num_particles(self, num_particles):
...
+            beta_exc = int(
+                str(abs(beta_occ - beta_index_shift)) + str(abs(beta_unocc - beta_index_shift))
+            )  # beta occupied and unoccupied indices.
+
+            exc_level = str(abs(alpha_exc - beta_exc)) + str(alpha_exc + beta_exc)
+            # exc_level is a 4-number string, which indicate alpha and beta occupied+unoccpied indices.
+            # Thus, the level of an excitations is indicated by this string. The symmetrically
+            # mirrored double excitations have the same exc_level string, and the
+            # excitaions with the same level will be assigned the same parameter.
```
mrossinek, line 253:
```suggestion
            # excitations with the same level will be assigned the same parameter.
```

releasenotes/notes/add-5b9f14b11d8992b9.yaml
```diff
@@ -0,0 +1,11 @@
+---
+features:
+  - |
+    Adds new keyword argument "mirror" to ``SUCCD`` class.
```
mrossinek, line 4:
```suggestion
    Adds the new keyword argument `mirror` to the ``SUCCD`` class.

```

releasenotes/notes/add-5b9f14b11d8992b9.yaml
```diff
@@ -0,0 +1,11 @@
+---
+features:
+  - |
+    Adds new keyword argument "mirror" to ``SUCCD`` class.
+    .. code-block:: python
+
+      from qiskit_nature.second_q.circuit.library.ansatzes import SUCCD
+      from qiskit_nature.converters.second_quantization import QubitConverter
+      from qiskit_nature.mappers.second_quantization import JordanWignerMapper
```
mrossinek, line 9:
```suggestion
      from qiskit_nature.second_q.mappers import JordanWignerMapper, QubitConverter
```

mrossinek (commented):
Some more comments. Please make sure that the CI passes so that we can wrap this up

Things are looking good :+1:

YuNariai:
> 

Hi, I failed two checks and I don't understand the warning. 
`build finished with problems, 1 warning.
make: *** [Makefile:71: html] Error 1
Error: Process completed with exit code 2.`

mrossinek:
> Hi, I failed two checks and I don't understand the warning. `build finished with problems, 1 warning. make: *** [Makefile:71: html] Error 1 Error: Process completed with exit code 2.`

You have some malformed documentation string. See the warning raised here: https://github.com/Qiskit/qiskit-nature/runs/7740319748?check_suite_focus=true#step:12:317

mrossinek (commented):
Besides this final improvement, I think this can be merged :+1: 

I ran black to fix the formatting changes and I also moved the release note to a more aptly named file while also updating the phrasing slightly :+1:

qiskit_nature/second_q/circuit/library/ansatzes/ucc.py
```diff
@@ -490,6 +490,19 @@ def _build_fermionic_excitation_ops(self, excitations: Sequence) -> list[Fermion
...
+        Args:
+            excitations: the list of excitations.
+
+        Returns:
+            The list of excitation operators in the second quantized formalism.
+        """
+
+        return self._build_a_fermionic_excitation_op(excitations)
+
+    def _build_a_fermionic_excitation_op(self, excitations: Sequence) -> list[FermionicOp]:
```
mrossinek, line 502: I had overlooked that we could do it like so. Two comments:
The name here is misleading. It suggest this method only builds a single operator but returns a list.
I actually suggest that you revert the change in the `UCC` class as explained by my second comment below.

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -260,20 +260,10 @@ def _build_fermionic_excitation_ops(self, excitations: Sequence) -> List[Fermion
...
-                op -= op.adjoint()
-                # we need to account for an additional imaginary phase in the exponent (see also
-                # `PauliTrotterEvolution.convert`)
-                op *= 1j  # type: ignore
-                ops.append(op)
-                sum_ops = cast(FermionicOp, sum(ops))
+            sum_ops = cast(
+                FermionicOp,
+                sum(self._build_a_fermionic_excitation_op(exc_level_items))
+            )
```
mrossinek, line 266: Seeing how things turned out over in the `UCC` class, instead of separating this out into another method, we can simply rely on the `super()` implementation here, like so:
```suggestion
            sum_ops = cast(
                FermionicOp, sum(super()._build_fermionic_excitation_ops(exc_level_items))
            )
```

mrossinek (approved):
I think this is good to go now. Thank you, @YuNariai!

@woodsp-ibm if you could review this too, then we can hopefully merge this soon :slightly_smiling_face:

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -106,6 +110,26 @@ def include_singles(self, include_singles: Tuple[bool, bool]) -> None:
...
 
+    @property
+    def mirror(self) -> bool:
+        """Whether to include the symmetrically mirrored double excitations."""
+        return self._mirror
+
+    @mirror.setter
+    def mirror(self, mirror: bool) -> None:
+        """Sets whether to include the symmetrically mirrored double excitations."""
+        self._mirror = mirror
```
woodsp-ibm, line 121: I think this should also call `self._invalidate()` like the other setters do in the base UCC class since this will also alter the ansatz whereby it needs to signal it needs to be re-built according to the new configuration when its needed.

qiskit_nature/second_q/circuit/library/ansatzes/succd.py
```diff
@@ -179,3 +216,55 @@ def _validate_num_particles(self, num_particles):
...
+
+        self._excitations_dict = excitations_dictionary
+
+        for exc_level, exc_level_items in excitations_dictionary.items():
+            sum_ops = cast(
+                FermionicOp, sum(super()._build_fermionic_excitation_ops(exc_level_items))
+            )
+            operators.append(sum_ops)
+        if not operators:
+            return [FermionicOp.zero(self.num_spin_orbitals)]
```
woodsp-ibm, line 268: Is this something we expect in practice or just to cover things from a coding perspective. The unit tests do not go via this path - this is the case where we have no excitations at all I guess so no operators? Do you think we should continue at this point with such an operator or raise an error?

mrossinek: You're right, I don't think this should actually ever happen. Arguably, an empty list of operators should be handled just fine by the stack further down, so I think we can just remove this `if` case

mrossinek (approved):
Just leaving this approval here to indicate that, once @woodsp-ibm approves this PR, it is good to go from my side, too :+1:

woodsp-ibm (approved):
Thanks @YuNariai for making this contribution to Qiskit Nature!

PR #730, opened 2022-07-05
Add method to compute induced norm of SecondQuantizedOp

kevinsung:
### Summary
Adds `SecondQuantizedOp.induced_norm`.

### DIFF
--- a/qiskit_nature/operators/second_quantization/second_quantized_op.py
+++ b/qiskit_nature/operators/second_quantization/second_quantized_op.py
@@ -65,3 +65,24 @@ class SecondQuantizedOp(StarAlgebraMixin, TolerancesMixin, ABC):
         return frozenset(self.simplify().to_list()) == frozenset(
             self.adjoint().simplify().to_list()
         )
+
+    def induced_norm(self, order: int = 1) -> float:
+        r"""Returns the induced p-norm of the operator.
+
+        If the operator is represented as a sum of terms
+
+        .. math::
+            \sum_i w_i H_i
+
+        then the induced :math:`p`-norm is
+
+        .. math::
+            \left(\sum_i |w_i|^p \right)^{1/p}
+
+        Args:
+            order: Order :math:`p` of the norm. The default value is 1.
+
+        Returns:
+            The induced p-norm of the operator.
+        """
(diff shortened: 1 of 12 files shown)


mrossinek (changes requested):
The same comment as in the other PR applies: after we merged #746, please re-apply these changes on the new code location (`qiskit_nature/second_q/operators`) and leave the old location untouched.

Besides that, please also provide a reference for this definition/implementation of the induced p-norm. I seem to not be able to find a matching definition and would like to verify the naming.

kevinsung:
For the reference, would https://en.wikipedia.org/wiki/Norm_(mathematics)#p-norm be okay? This is the p-norm induced by the coefficients of the operator, hence the name. If the operator basis is orthonormal then this is a standard definition, so just `norm` or `p_norm` would be fine. The somewhat non-standard behavior arises for FermionicOps that are not normal-ordered, since the norm may change after normal-ordering. This is the reason to call it `induced_norm` instead of `norm` to make it clear that it is the norm induced by the coefficients. We could choose to always normal-order first, but it's better to give the user the option not to do that.

qiskit_nature/operators/second_quantization/second_quantized_op.py
```diff
@@ -65,3 +65,24 @@ def is_hermitian(self) -> bool:
         return frozenset(self.simplify().to_list()) == frozenset(
             self.adjoint().simplify().to_list()
         )
+
+    def induced_norm(self, order: int = 1) -> float:
+        r"""Returns the induced p-norm of the operator.
```
mrossinek, line 70: I suggest that you add the explanations which you gave in the comment on this PR above in this docstring.
You could even link to the p-norm definition on Wikipedia

mrossinek (commented):
OIther than this simple comment, this LGTM :+1:

qiskit_nature/operators/second_quantization/second_quantized_op.py
kevinsung: Done.

kevinsung:
@mrossinek I've updated this PR for the library refactor and addressed your comment, it should be good to go.

mrossinek (approved):
LGTM, thanks! :+1:

woodsp-ibm (approved):

PR #731, opened 2022-07-05
Add SparseLabelOp base class

### LINKED ISSUE #669: Add a `SparseLabelOp` class
### What should we add?

In #665 I am proposing the addition of a `PolynomialTensor` class, which is a storage container for coefficients in a matrix representation. Our `SecondQuantizedOp` (sub)classes on the other hand, generally store data of the form:
```
{
  "op_idx1 op_idx2": 0.0,
  "op_idx1 op_idx3": 1.0,
}
```
In other words, they are essentially label-coefficient dictionaries, which can be thought of as the result of `np.ravel(matrix)` matches with it's indices.

More specifically, the `FermionicOp` and `VibrationalOp` used to be almost copies of one another, but the former has been improved significantly in terms of capabilities and performance, whereas the latter has not gotten all these ugprades.

Thus, this issue proposes the introduction of a base class, which should handle all the operations which are case-agnostic (i.e. not specific to fermionic, vibrational or spin settings).
Similar to the `PolynomialTensor`, this likely means that the class will look somewhat like the following:
```python
from qiskit.opflow.mixins import StarAlgebraMixin
from qiskit.quantum_info.operators.mixins import TolerancesMixin

class SparseLabelOp(StarAlgebraMixin, TolerancesMixin):

    def __init__(self, data: Dict[str, complex]):
        self._data = data

    # TODO: implement mathematical operations supported by StarAlgebraMixin

    @abstractmethod
    def commutativity(self) -> bool:
        ...

    @abstractmethod
    def normal_ordered(self) -> SparseLabelOp:
        ...

    @abstractmethod
    def simplify(self) -> SparseLabelOp:
        ...

    @abstractclassmethod
    def zero(cls, ...) -> SparseLabelOp:
        # TBD: arguments to this method
        ...

    @abstractclassmethod
    def one(cls, ...) -> SparseLabelOp:
        # TBD: arguments to this method
        ...
```
Some operations may have to fall back on `self.commutativity` to be correctly implemented.

> :warning: This issue will be fleshed out with more details later.

javabster:
### Summary

closes #669


qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,106 @@
+from __future__ import annotations
+from abc import abstractmethod, abstractclassmethod, ABC
+from typing import Dict
+
+from qiskit.opflow.mixins import StarAlgebraMixin
+from qiskit.quantum_info.operators.mixins import TolerancesMixin
+
+class SparseLabelOp(StarAlgebraMixin, TolerancesMixin, ABC):
```
mrossinek, line 8: The above may be accomplished better (i.e. without accessing the private `_data`) it we implement `__iter__`. There has previously been some discussion around operator iteration (cf. #541 and #544) and I would like to quote [this comment](https://github.com/Qiskit/qiskit-nature/pull/544#issuecomment-1044405280):
> it's probably better to respect the general type-system perspective that iterating or getting an item from a collection-like object should reduce the nested types or the logical dimensionality of the object by one

That means, `__iter__` should iterate `self._data.items()` which is the most natural for an end-user. It is also the most natural implementation in the current setting.

mrossinek (changes requested):
I finally had time for some more feedback on the open discussion points.

Besides that, please also relocate (now that #746 has been merged) this class to `qiskit_nature/second_q/operators/sparse_label_op.py` (and the matching test location).

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,150 @@
+from __future__ import annotations
+from abc import abstractmethod, abstractclassmethod, ABC
+from typing import Dict, Iterator
+import math
+from xmlrpc.client import Boolean
+
+from qiskit.opflow.mixins import StarAlgebraMixin
+from qiskit.quantum_info.operators.mixins import TolerancesMixin
+
+class SparseLabelOp(StarAlgebraMixin, TolerancesMixin, ABC):
```
mrossinek, line 10: I just had a look at the beginnings of the complementary `PolynomialTensor` class (https://github.com/Qiskit/qiskit-nature/pull/756#discussion_r926558375).
I think, it may be better if we implemented the following mixins instead of the `StarAlgebraMixin`:
```python
from qiskit.quantum_info.operators.mixins import (
    LinearMixin,  # contains MultiplyMixin -> _add, _multiply
    AdjointMixin,  # conjugate, transpose
)
```
This makes `adjoint` more granular (because it will be implemented in terms of `.conjugate().transpose()` and it avoids the complexity of `compose` for the time being.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,106 @@
...
+    def mul(self, other: complex) -> SparseLabelOp:
+        """Return scalar multiplication of self and other"""
+        if not isinstance(other, (int, float, complex)):
+            raise TypeError(
+                f"Unsupported operand type(s) for *: 'SparseLabelOp' and '{type(other).__name__}'"
+            )
+        return self.__class__( #Max to think about...
+            [(label, coeff * other) for label, coeff in self._data],
+            register_length=self.register_length,
+        )
```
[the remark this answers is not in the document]
mrossinek: We had a discussion on this and have decided the following things:
- `SparseLabelOp` should NOT be abstract. This means, you can instantiate it normally here.
- by extension, all methods which were currently planned to be abstract, can be removed. This is not an issue because:
    - `normaled_ordered` is not even well-defined for all planned subclasses
    - `commuting` is no longer required since we skip the implementation of `compose` on this class
    - `zero` and `one` might be implemented directly, now that this class is no longer abstract (but I honestly don't think we really need these here)

The reason we went with this solution, is that subclasses will have to override the mathematical methods anyways, in order to do further compatibility checks. They then internally call `new_sparse_label_op = super().whatever(other)` and will need methods to construct a new instance of their specific subclass, using the data from `new_sparse_label_op`

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""The Sparse Label Operator base interface."""
```
mrossinek, line 13:
```suggestion
"""The Sparse Label Operator base class."""
```
This wording is better, because interface would suggest that you cannot actually instantiate this class, which you actually can now :+1:

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+from typing import Dict, Iterator
+import math
+from xmlrpc.client import Boolean
+import numpy as np
+
+from qiskit.quantum_info.operators.mixins import LinearMixin, AdjointMixin, TolerancesMixin
+
+
+class SparseLabelOp(LinearMixin, AdjointMixin, TolerancesMixin):
+    """The Sparse Label Operator base interface."""
```
mrossinek, line 25:
```suggestion
    """The Sparse Label Operator base class."""
```
We will need to extend this documentation with some more explanation. But we can do that a bit later in order to progress the refactoring for now :+1:

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __init__(self, data: Dict[str, complex], register_length: int = None):
+        self._data = data
+        self._register_length = register_length
+
+    def _add(self, other, qargs=None) -> SparseLabelOp:
+        """Return Operator addition of self and other"""
```
mrossinek, line 32: You will need to update the docstrings of all methods to include:
- an `Args:` section which details the input arguments. `self` does not need to be documented and if a method takes no arguments other than `self`, you can omit this section.
- a `Returns:` section which says what is being returned (if `None` is returned, this can be skipped)

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class SparseLabelOp(LinearMixin, AdjointMixin, TolerancesMixin):
+    """The Sparse Label Operator base interface."""
+
+    def __init__(self, data: Dict[str, complex], register_length: int = None):
+        self._data = data
+        self._register_length = register_length
+
+    def _add(self, other, qargs=None) -> SparseLabelOp:
```
mrossinek, line 31:
```suggestion
    def _add(self, other: SparseLabelOp) -> SparseLabelOp:
```
Even though the base class has `qargs`, this makes no sense in our context. I would argue to remove it and disable the pylint and mypy warnings resuling from this.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+
+        for key, value in other._data.items():
+            if key in new_data.keys():
+                new_data[key] += value
+            else:
+                new_data[key] = value
+
+        return SparseLabelOp(new_data)
+
+    def _multiply(self, other: complex) -> SparseLabelOp:
```
mrossinek, line 43:
```suggestion
    def _multiply(self, other: Number) -> SparseLabelOp:
```
Maybe in combination with `from numbers import Number` this is a bit cleaner?
This could also be used for the `isintance` check below

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+    def conjugate(self) -> SparseLabelOp:
+        """Return the conjugate of the ``SparseLabelOp``"""
+        new_data = self._data.copy()
+
+        for key, value in self._data.items():
+            new_data[key] = np.conjugate(value)
+
+        return SparseLabelOp(new_data)
+
+    def transpose(self) -> SparseLabelOp:
```
mrossinek, line 65: This should have a docstring, which also explicitly states that this method has no effect on the `SparseLabelOp`.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+        Args:
+            other: the second ``SparseLabelOp`` to compare the first with.
+
+        Returns:
+            Bool: True if operators are equal to, False if not.
+        """
+        if set(self._data.keys()) != set(other._data.keys()):
+            return False
+        for key, val in self._data.items():
+            if not math.isclose(val, other._data[key], rel_tol=self.rtol, abs_tol=self.atol):
```
mrossinek, line 80: This appears to be unable to handle `complex` numbers. For that, we will need https://docs.python.org/3/library/cmath.html#cmath.isclose
So maybe we can just use that method instead of the `math.isclose` one. I assume `cmath.isclose` can handle real numbers just fine, too

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+        if set(self._data.keys()) != set(other._data.keys()):
+            return False
+        for key, val in self._data.items():
+            if not val == other._data[key]:
+                return False
+        return True
+
+    def __iter__(self) -> Iterator[SparseLabelOp]:
+        """Iterate through ``SparseLabelOp`` items"""
+        return iter(self._data.items())
```
mrossinek, line 102: I think we may need to use the following implementation instead:
```suggestion
    for key, value in self._data.items():
        yield key, value
```

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+            Bool: True if operators are equal, False if not.
+        """
+        if set(self._data.keys()) != set(other._data.keys()):
+            return False
+        for key, val in self._data.items():
+            if not val == other._data[key]:
+                return False
+        return True
+
+    def __iter__(self) -> Iterator[SparseLabelOp]:
```
mrossinek, line 100: Accordingly, this is actually not the correct type hint. It should be this instead:
```suggestion
    def __iter__(self) -> Iterator[Tuple[str, Number]]:
```
(assuming you added the `Number` import suggested above :+1:

test/operators/second_quantization/test_sparse_label_op.py
```diff
@@ -0,0 +1,234 @@
...
+            test_op = SparseLabelOp(op1).equiv(
+                SparseLabelOp(
+                    {
+                        "+_0 -_1": 0.000000001,
+                        "+_0 -_2": 1.0,
+                    }
+                )
+            )
+
+            self.assertTrue(test_op)
```
mrossinek, line 223: It would be good to do this same test (which passes `equiv`) and ensure it *fails* true equaility  (`__eq__`)

mrossinek (commented):
Some feedback for now. I think in general this looks good.

It would be good, if you could also add a release note :+1: 

I think, we can leave the following aspects for an extension later during the refactoring process:
- more detailed documentation
- addition of `Parameter` support

qiskit_nature/operators/second_quantization/__init__.py
```diff
@@ -34,11 +34,13 @@
...
 from .vibrational_op import VibrationalOp
+from .sparse_label_op import SparseLabelOp
 
 __all__ = [
     "FermionicOp",
     "QuadraticHamiltonian",
     "SecondQuantizedOp",
     "SpinOp",
     "VibrationalOp",
+    "SparseLabelOp",
```
woodsp-ibm, line 45: This should be added to the dosctring list above to I think - this way sphinx will also include it so it will check the docstring correctness as well.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""The Sparse Label Operator base interface."""
+
+from __future__ import annotations
+from typing import Dict, Iterator
+import math
+from xmlrpc.client import Boolean
```
woodsp-ibm, line 18: This seems like an "unusual" import that I would not have expected to see! In the places where it appears Boolean is used we can simply use `bool` as the return type and eliminate this import.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+import numpy as np
+
+from qiskit.quantum_info.operators.mixins import LinearMixin, AdjointMixin, TolerancesMixin
+
+
+class SparseLabelOp(LinearMixin, AdjointMixin, TolerancesMixin):
+    """The Sparse Label Operator base interface."""
+
+    def __init__(self, data: Dict[str, complex], register_length: int = None):
```
woodsp-ibm, line 27: This needs a docstring for args. Also if register_length should be `int | None` given its assigned None by None. I do not see `_register_length` instance var used anywhere to check further on that type, though I seem to recall in other code that given None register_length was computed so internally it would not be None.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,102 @@
...
+    def _multiply(self, other: complex) -> SparseLabelOp:
+        """Return scalar multiplication of self and other"""
+        if not isinstance(other, (int, float, complex)):
+            raise TypeError(
+                f"Unsupported operand type(s) for *: 'SparseLabelOp' and '{type(other).__name__}'"
+            )
+        new_data = self._data.copy()
+
+        for key, value in self._data.items():
+            new_data[key] = value * other
```
woodsp-ibm, line 52: Do we need this copy here - would it not be more efficient to start with an empty dictionary as it seem to be overwritten i.e. all keys set, in the for loop anyway.

qiskit_nature/operators/second_quantization/sparse_label_op.py
mrossinek: Thanks for noting this Steve! Indeed, the type hint needs to be updated (which I believe was just done now), but the actual handling of `_register_length` (i.e. computing it at runtime) will need to be handled by the subclasses since it requires knowledge of the structure of the dictionary keys :+1:

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -65,7 +64,7 @@ def _multiply(self, other: Number) -> SparseLabelOp:
         Raises:
             TypeError: if ``other`` is not compatible type (int, float or complex)
         """
-        if not isinstance(other, Number):
+        if not isinstance(other, (int, float, complex)):
```
woodsp-ibm, line 67: I would be curious, if you just leave this Number check here if things are ok (not quite sure what the "fun" with mypy was). If you look in PolynomialTensor there it uses Number for the instance check. For input & checking the more flexibility allows the caller flexibility. I recall when we  test things too 'precisely' we have had cases like numpy ints that fail these checks when its reasonable to have expected things to have worked.

javabster: I'll try adding it back and see what happens!

The 'fun' with mypy shows up here if you want more details: https://github.com/Qiskit/qiskit-nature/runs/7862883533?check_suite_focus=true

woodsp-ibm: I think the aspect with iter is that its getting information about the type from data as passed to the constructor Dict[str, complex] so self_.data.items() becomes an iterable over str, complex because that was what was passed in. I would imagine changing the Dict type to [str, Number] would resolve that. As to multiply, that overrides mixin defined method right. Which is why I think in the PoynomialTensor the type was left as complex though the instance check was done on Number. As in the body then is similar too to quantum info operator implementation https://github.com/Qiskit/qiskit-terra/blob/f2d228ecd288c928d074b1f61702b96eb5b66e87/qiskit/quantum_info/operators/operator.py#L401-L414

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+import cmath
+import numpy as np
+
+from qiskit.quantum_info.operators.mixins import LinearMixin, AdjointMixin, TolerancesMixin
+
+
+class SparseLabelOp(LinearMixin, AdjointMixin, TolerancesMixin):
+    """The Sparse Label Operator base class."""
+
+    def __init__(self, data: Dict[str, complex], register_length: int | None = None):
```
mrossinek, line 27:
```suggestion
    def __init__(self, data: Mapping[str, Number], register_length: int | None = None):
```
As per Steve's other comment.
This requires adding `from typing import Mapping` to the imports at the top

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+            other: the second ``SparseLabelOp`` to add to the first.
+
+        Returns:
+            SparseLabelOp: the new summed ``SparseLabelOp``.
+
+        Raises:
+            ValueError: when ``qargs`` argument is not ``None``
+        """
+        if qargs is not None:
+            raise ValueError("`qargs` argument should be `None`")
```
mrossinek, line 44: In Qiskit Nature we try to provide additional runtime information in our raised errors, like so:
```suggestion
            raise ValueError(f"The `qargs` argument must be `None`, not {qargs}")
```
I wonder now, whether instead of `ValueError`, we should maybe raise a `NotImplementedError`? :thinking: @woodsp-ibm any thoughts?

The same applies to similar errors being raised elsewhere :+1:

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+        self._register_length = register_length
+
+    def _add(self, other: SparseLabelOp, qargs=None) -> SparseLabelOp:
+        """Return Operator addition of self and other.
+
+        Args:
+            other: the second ``SparseLabelOp`` to add to the first.
+
+        Returns:
+            SparseLabelOp: the new summed ``SparseLabelOp``.
```
mrossinek, line 38:
```suggestion
            The new summed ``SparseLabelOp``.
```
The return types are documented via the `-> ReturnType` type hints and are no longer needed in these docstrings.
The same applies elsewhere

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+        Returns:
+            SparseLabelOp: the initial ``SparseLabelOp``.
+        """
+        return self
+
+    def equiv(self, other: SparseLabelOp) -> bool:
+        """Check equivalence of two ``SparseLabelOp`` instances to an accepted tolerance
+
+        Args:
+            other: the second ``SparseLabelOp`` to compare the first with.
```
mrossinek, line 104:
```suggestion
            other: the second ``SparseLabelOp`` to compare with this instance.
```
The same suggestion for `__eq__` below

qiskit_nature/operators/second_quantization/sparse_label_op.py
mrossinek: Note, this might affect the mypy types further down again

releasenotes/notes/sparselabelop-class-57ffd57121dadfe5.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Adds :class:`.SparseLabelOp`.
```
mrossinek, line 4: Maybe this can be expanded a bit? Stating that this will is a class which maps sparse labels to coefficient values which will become the new base class for the other operator types.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+        Args:
+            other: the second ``SparseLabelOp`` to add to the first.
+
+        Returns:
+            the new summed ``SparseLabelOp``.
+
+        Raises:
+            ValueError: when ``qargs`` argument is not ``None``
+        """
+        if qargs is not None:
```
woodsp-ibm, line 43: If other is not a SparseLabelOp instance we should raise a ValueError (Python seeing A + B will effectively try A.add(B) if that raises a TypeError will try B.add(A) since B may support adding an A even if A can't add a B). This then analogous to what is done in multiply where other is type checked against what is supported.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+            raise ValueError(f"The `qargs` argument must be `None`, not {qargs}")
```
woodsp-ibm, line 44: For `qargs` this to be gotten from `other` and passed on down. So if a subclass has that as an attribute this would fail. I am not even sure what the intent is there exactly for qargs as the mixin defines. We could choose instead to ignore qargs if it really has no meaning here

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+
+        new_data = self._data.copy()
+
+        for key, value in other._data.items():
+            if key in new_data.keys():
+                new_data[key] += value
+            else:
+                new_data[key] = value
+
+        return SparseLabelOp(new_data, self._register_length)
```
woodsp-ibm, line 54: Shouldn't we be handling the case there other._register_length may be different - perhaps its still the case the data can be combined by maybe we need to select the larger of the register_lengths if that makes sense generically here in the base class.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+            TypeError: if ``other`` is not compatible type (int, float or complex)
+        """
+        if not isinstance(other, Number):
+            raise TypeError(
+                f"Unsupported operand type(s) for *: 'SparseLabelOp' and '{type(other).__name__}'"
+            )
+        new_data = {}
+
+        for key, value in self._data.items():
+            new_data[key] = value * other
```
woodsp-ibm, line 75: This could possibly be done simply in one line using a dictionary comprehension

`new_data = {k: v*other for k, v in self._data.items()`

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+    def conjugate(self) -> SparseLabelOp:
+        """Returns the conjugate of the ``SparseLabelOp``
+
+        Returns:
+            the complex conjugate of the starting ``SparseLabelOp``.
+        """
+        new_data = {}
+
+        for key, value in self._data.items():
+            new_data[key] = np.conjugate(value)
```
woodsp-ibm, line 88: Likewise - perhaps a one line dictionary comprehension

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+
+        return SparseLabelOp(new_data, self._register_length)
+
+    def transpose(self) -> SparseLabelOp:
+        """This method has no effect on ``SparseLabelOp`` and returns itself.
+
+        Returns:
+            the initial ``SparseLabelOp``.
+        """
```
woodsp-ibm, line 97: Does this "make sense" as a base class implementation. Would  it be better to have an abstract method and require a derived class to implement somethings that actually does a transpose rather than a no-op? If a derived class does not override this then a user calling transpose would perhaps not get back what they were expecting.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+        """Check exact equality of two ``SparseLabelOp`` instances
+
+        Args:
+            other: the second ``SparseLabelOp`` to compare with this instance.
+
+        Returns:
+            True if operators are equal, False if not.
+        """
+        if not isinstance(other, SparseLabelOp):
+            return NotImplemented
```
woodsp-ibm, line 126: This can just return False since we can consider equality to be only numerically equal SparseLabelOps

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+        if set(self._data.keys()) != set(other._data.keys()):
+            return False
+        for key, val in self._data.items():
+            if not val == other._data[key]:
+                return False
```
woodsp-ibm, line 131: Assuming these are dicts, if thats how we store data internally in this class, then `self._data == other._data` will do this. If the data remains as Mapping then I think something more like this logic is whats need. Though you may simply be able to compare self._data.items() to other_data.items(). And self._data.keys() is a set already, albeit a view on the dict. So the casting to a set if not needed.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+    def equiv(self, other: SparseLabelOp) -> bool:
+        """Check equivalence of two ``SparseLabelOp`` instances to an accepted tolerance
+
+        Args:
+            other: the second ``SparseLabelOp`` to compare with this instance.
+
+        Returns:
+            True if operators are equal to, False if not.
+        """
+        if set(self._data.keys()) != set(other._data.keys()):
```
woodsp-ibm, line 109: As mentioned in eq cast to set is not needed - keys is a set view already.

qiskit_nature/operators/second_quantization/sparse_label_op.py
```diff
@@ -0,0 +1,137 @@
...
+        if not isinstance(other, SparseLabelOp):
+            return NotImplemented
+        if set(self._data.keys()) != set(other._data.keys()):
+            return False
+        for key, val in self._data.items():
+            if not val == other._data[key]:
+                return False
+        return True
+
+    def __iter__(self) -> Iterator[tuple[str, complex]]:
```
woodsp-ibm, line 134: Having an iter seems fine. Is this how we expect it to be consumed normally? As in is there any reason to have a more random access - it could expose getitem etc and implement Mapping - it would be pretty much just delegating to self._data

qiskit_nature/operators/second_quantization/sparse_label_op.py
mrossinek: Hm yes that's probably the better call here :+1:
@javabster you can simply remove the handling of `qargs` in its entirety in this case. It may require a pylint exception comment like the following:
```python
# pylint: disable=unused-argument
```
(not 100% certain whether that is the right pylint error name, though)

qiskit_nature/operators/second_quantization/sparse_label_op.py
mrossinek: Yes, selecting the larger of the two `register_length` values should be the way to go here :+1:

qiskit_nature/operators/second_quantization/sparse_label_op.py
mrossinek: I liked the idea of being able to test the `SparseLabelOp` easily. But I agree that this implementation is not ideal.
@javabster could you do the following instead, please?

1. add the `@abstractmethod` decorator to this method and simply remove the internals of the method
2. add `ABC` as the first base class of `SparseLabelOp`
3. add the imports: `from abc import ABC, abstractmethod`
4. in the unittests, create a class `DummySparseLabelOp` which derives from `SparseLabelOp`, implementing the `transpose` method to simply not be abstract but do nothing other than that
5. Use that `DummySparseLabelOp` class for all unittests and remove the `test_transpose` test (since it doesn't do anything)

qiskit_nature/operators/second_quantization/sparse_label_op.py
mrossinek: I don't think we need random access right now, since the user may not know the internal keys a priori (i.e. without iterating over them).
If we see that these will be useful, we can still add them down the road :+1:

qiskit_nature/operators/second_quantization/sparse_label_op.py
javabster: the issue with this one wasn't just linting, if I try to add without qargs being handled it throws an error: `TypeError: _add() got an unexpected keyword argument 'qargs'`

javabster: do you have any suggestions to get around this?

woodsp-ibm: The error, is that when omitting qargs from the function signature? - if so it has to be there since the mixin defines that and indeed calls _add from there in the mixin with qargs. I think it was more just to ignore it, since we have no use for it and arguably do not really care whether its None or not.

mrossinek (changes requested):
@woodsp-ibm has pointed out to me offline, that the location of your files is actually still the old one, and not yet aligned with the ongoing refactoring.

Please move all the files to `qiskit_nature/second_q/operators` (and `test/second_q/operators`, resp.)

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,140 @@
...
+from qiskit.quantum_info.operators.mixins import LinearMixin, AdjointMixin, TolerancesMixin
+
+
+class SparseLabelOp(LinearMixin, AdjointMixin, TolerancesMixin, ABC):
+    """The Sparse Label Operator base class."""
+
+    def __init__(self, data: Mapping[str, complex], register_length: int, **kwargs):
+        """
+        Args:
+            data: Data for operator comprising string key to coeff value mapping
```
mrossinek, line 31:
```suggestion
            data: the operator data, mapping string-based keys to numerical values.
```

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,140 @@
...
+            register_length: Length of register needed for data
```
mrossinek, line 32:
```suggestion
            register_length: the length of the operators register. This coincides with the maximum
                index on which an operation may be performed by this operator.
```

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        self._data: Mapping[str, complex] = {}
+        if kwargs.get("_dont_copy", False):
```
mrossinek, line 35: What is the purpose of hiding this inside `kwargs`? Can we not just add this explicitly?
We can enforce it to be used as a keyword argument by inserting `*` behind the `register_length` argument. I will make a suggestion to show how this is done

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,140 @@
...
+import cmath
+import numpy as np
+
+from qiskit.quantum_info.operators.mixins import LinearMixin, AdjointMixin, TolerancesMixin
+
+
+class SparseLabelOp(LinearMixin, AdjointMixin, TolerancesMixin, ABC):
+    """The Sparse Label Operator base class."""
+
+    def __init__(self, data: Mapping[str, complex], register_length: int, **kwargs):
```
mrossinek, line 28:
```suggestion
    def __init__(self, data: Mapping[str, complex], register_length: int, *, _dont_copy: bool = False):
```
The suggestion matching my other comment.
That said, I would actually change this argument to be `_copy: bool = True` instead of the negated version

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,140 @@
...
+        Args:
+            data: Data for operator comprising string key to coeff value mapping
+            register_length: Length of register needed for data
+        """
+        self._data: Mapping[str, complex] = {}
+        if kwargs.get("_dont_copy", False):
+            self._data = data  # Store by reference
+        else:
+            self._data = dict(data.items())  # Store by value
+        self._register_length = register_length
```
mrossinek, line 39: A `@property` for `register_length` should be provided, which will remove the need to access this private variable

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,140 @@
...
+        if not isinstance(other, Number):
+            raise TypeError(
+                f"Unsupported operand type(s) for *: 'SparseLabelOp' and '{type(other).__name__}'"
+            )
+        new_data = {key: val * other for key, val in self._data.items()}
+
+        return self.__class__(new_data, self._register_length, _dont_copy=True)
+
+    def conjugate(self) -> SparseLabelOp:
+        """Returns the conjugate of the ``SparseLabelOp``
```
mrossinek, line 87:
```suggestion
        """Returns the conjugate of the ``SparseLabelOp``.
```

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,140 @@
...
+        new_data = {key: np.conjugate(val) for key, val in self._data.items()}
+
+        return self.__class__(new_data, self._register_length, _dont_copy=True)
+
+    @abstractmethod
+    def transpose(self) -> SparseLabelOp:
+        """This method has no effect on ``SparseLabelOp`` and returns itself.
+
+        Returns:
+            the initial ``SparseLabelOp``.
```
mrossinek, line 101:
```suggestion
        """Returns the transpose of the ``SparseLabelOp``.

        Returns:
            the transpose of the starting ``SparseLabelOp``.
```

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+
+    def equiv(self, other: SparseLabelOp) -> bool:
+        """Check equivalence of two ``SparseLabelOp`` instances to an accepted tolerance
+
+        Args:
+            other: the second ``SparseLabelOp`` to compare with this instance.
+
+        Returns:
+            True if operators are equal to, False if not.
```
mrossinek, line 111:
```suggestion
            True if operators are equivalent, False if not.
```

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,140 @@
...
+    @abstractmethod
+    def transpose(self) -> SparseLabelOp:
+        """This method has no effect on ``SparseLabelOp`` and returns itself.
+
+        Returns:
+            the initial ``SparseLabelOp``.
+        """
+
+    def equiv(self, other: SparseLabelOp) -> bool:
+        """Check equivalence of two ``SparseLabelOp`` instances to an accepted tolerance
```
mrossinek, line 105:
```suggestion
        """Check equivalence of two ``SparseLabelOp`` instances up to an accepted tolerance.
        
        The absolute and relative tolerances can be changed via the `atol` and `rtol` attributes, respectively.
```

releasenotes/notes/sparselabelop-class-57ffd57121dadfe5.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Adds :class:`.SparseLabelOp` which maps sparse labels to coefficient values. This will form a base class for other operator types in future.
```
mrossinek, line 4:
```suggestion
    Adds :class:`.SparseLabelOp` which maps sparse labels to coefficient values.
```
This change will actually be done before the next release, meaning that once these release notes get compiled together, this would be a bit confusing :+1: More details will be provided by the release notes which refactor the other operator classes

mrossinek (commented):
Some more comments, but overall I think this is close to be done :+1:

qiskit_nature/second_q/operators/sparse_label_op.py
woodsp-ibm: If we want it part of exposed public interface I would get rid of the leading underscore too.

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,151 @@
...
+        Returns:
+            True if operators are equal, False if not.
+        """
+        if not isinstance(other, SparseLabelOp):
+            return False
+        return self.register_length == other._register_length and self._data == other._data
+
+    def __iter__(self) -> Iterator[tuple[str, complex]]:
+        """Iterate through ``SparseLabelOp`` items"""
+        for key, value in self._data.items():
```
woodsp-ibm, line 150: I do not think items() is an iterable; keys() is an iterable though so maybe iterate over keys() and `yield key, self._data[key]` Yes this will include a lookup overhead here but maybe its preferable than using items.

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,151 @@
...
+    """The Sparse Label Operator base class."""
+
+    def __init__(self, data: Mapping[str, complex], register_length: int, *, copy: bool = True):
+        """
+        Args:
+            data: the operator data, mapping string-based keys to numerical values.
+            register_length: the length of the operators register. This coincides with the maximum
+                index on which an operation may be performed by this operator.
+            copy: when set to False data will not be copied and will be stored by value rather
+                than by reference as default.
```
mrossinek, line 35:
```suggestion
            copy: when set to False the `data` will not be copied and the dicionary will be stored by reference rather
                than by value (which is the default). Note, that this might requires you to not change the contents of
                the dictionary after constructing the operator. Use with care!
```
You might need to run `black` after applying this because I did not pay attention to the line length while typing this suggestion...

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        self._data: Mapping[str, complex] = {}
+        if copy is True:
+            self._data = data  # Store by reference
+        else:
+            self._data = dict(data.items())  # Store by value
```
mrossinek, line 41:
```suggestion
        if copy:
            self._data = dict(data.items())
        else:
            self._data = data
```
You had the logic reversed...

test/second_q/operators/test_sparse_label_op.py
```diff
@@ -0,0 +1,266 @@
...
+        test_op = iter(DummySparseLabelOp(op1, 2))
+
+        self.assertEqual(next(test_op), ("+_0 -_1", 0.0))
+        self.assertEqual(next(test_op), ("+_0 -_2", 1.0))
+
+    def test_get_register_length(self):
+        """test register length property"""
+        test_val = DummySparseLabelOp(op1, 2).register_length
+        self.assertEqual(test_val, 2)
+
```
mrossinek, line 263: Could you please add a test for `copy`, which validates that when `copy=True`, changing the provided `data` does NOT affect the internal object?

mrossinek (commented):
Some minor comments but other than this, this LGTM :+1:

mrossinek:
Following my suggestion over in #756, I will make a similar comment here:

I had another discussion with @woodsp-ibm and we said that it might be beneficial to directly implement `SparseLabelOp` as a subclass of `Mapping`. That would give full flexibility when accessing the internal `_data` object.
The way that you could achieve this, is by something similar to the following:
```python
from collections.abc import Mapping
from typing import Iterator

class SparseLabelOp(..., Mapping):
    ...

    def __getitem__(self, __k: str) -> Number:
        return self._data.__getitem__(__k)

    def __len__(self) -> int:
        return self._data.__len__()

        return self._data.__iter__()
```
This also shows an even simpler implementation of `__iter__` which is probably better than what we have right now. It will change it such that the default iteration will only return the keys (and not items) but this is inline with what `Mapping` does :+1:

This will directly enable users to use `len(...)`, `.keys()`, `.values()`, `.items()` and even use `op["key"]` to access specific entries :+1:

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,161 @@
...
+        if not isinstance(other, SparseLabelOp):
+            return False
+        return self.register_length == other._register_length and self._data == other._data
+
+    def __getitem__(self, __k: str) -> complex:
+        """Return a specified ``SparseLabelOp`` item"""
+        return self._data.__getitem__(__k)
+
+    def __len__(self) -> int:
+        """Return length of ``SparseLabelOp``"""
```
mrossinek, line 156:
```suggestion
        """Returns the length of the ``SparseLabelOp``."""
```

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,161 @@
...
+        return self._data.__len__()
+
+    def __iter__(self) -> Iterator[str]:
+        """Iterate through ``SparseLabelOp`` items"""
```
mrossinek, line 160:
```suggestion
        """An iterator over the keys of the ``SparseLabelOp``."""
```

test/second_q/operators/test_sparse_label_op.py
```diff
@@ -0,0 +1,286 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021, 2022.
```
mrossinek, line 3:
```suggestion
# (C) Copyright IBM 2022.
```

test/second_q/operators/test_sparse_label_op.py
```diff
@@ -0,0 +1,286 @@
...
+        test_val = DummySparseLabelOp(op1, 2).register_length
+        self.assertEqual(test_val, 2)
+
+    def test_copy(self):
+        """test copy bool"""
+        data = {
+            "+_0 -_1": 0.0,
+            "+_0 -_3": 1.0,
+        }
+        test_op = DummySparseLabelOp(data, 2)
```
mrossinek, line 280:
```suggestion
        test_op = DummySparseLabelOp(data, 2, copy=True)
```
Just to be explicit, as it makes the unittest just a little bit faster to understand :smile:

mrossinek (commented):
Other than these suggestions, this LGTM :+1: 

Note: if you go to the `Files` tab of the PR, you can batch all of these suggestions into a single commit, which will trigger fewer CI runs :+1:

mrossinek (dismissed):
LGTM :+1: 
Thanks a lot, Abby!

I'll await another official review from @woodsp-ibm before merging :+1:

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,161 @@
...
+from typing import Iterator
+from numbers import Number
+import cmath
+import numpy as np
+
+from qiskit.quantum_info.operators.mixins import LinearMixin, AdjointMixin, TolerancesMixin
+
+
+class SparseLabelOp(LinearMixin, AdjointMixin, TolerancesMixin, ABC, Mapping):
+    """The Sparse Label Operator base class."""
```
woodsp-ibm, line 27: Could this be fleshed out a little to explain what this is a bit more. Maybe an example usage to or something to give someone a better idea of whats this all about.

releasenotes/notes/sparselabelop-class-57ffd57121dadfe5.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Adds :class:`.SparseLabelOp` which maps sparse labels to coefficient values.
```
woodsp-ibm, line 4: Same here. Its a single line item that of course they should be able to click and read the class doc but maybe it can say a little more - it seems kinda sparse as a description!

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Args:
+            other: the second ``SparseLabelOp`` to compare with this instance.
+
+        Returns:
+            True if operators are equal, False if not.
+        """
+        if not isinstance(other, SparseLabelOp):
+            return False
+        return self.register_length == other._register_length and self._data == other._data
```
woodsp-ibm, line 149: register_length here is using the getter on self and the private var on other. I think we should be consistent one way or the other

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Args:
+            other: the second ``SparseLabelOp`` to compare with this instance.
+
+        Returns:
+            True if operators are equivalent, False if not.
+        """
+        if not isinstance(other, SparseLabelOp):
+            return False
+        if self.register_length != other._register_length:
```
woodsp-ibm, line 129: Similar comment on register_length getter/instance var

woodsp-ibm (approved):

mrossinek (approved):
Thanks, Abby! :slightly_smiling_face:

Issue #654, opened 2022-05-02
Low rank representations


kevinsung:
### What should we add?

The low rank representations of the quantum chemistry Hamiltonian (https://www.nature.com/articles/s41534-021-00416-z.pdf) are useful for time evolution as well as variational algorithms. We should add functionality to calculate and use these representations.

kevinsung:
I've added some pseudocode to the opening post.

PR #735, opened 2022-07-11
Add low rank decomposition utilities

kevinsung:
### Summary
Part of #654.
Adds utilities to perform the low rank decomposition of a molecular Hamiltonian. Upcoming pull requests will use this to implement Hamiltonian simulation using Trotter formulas.
Credit goes to @robparrish who wrote the original code on which this implementation is based.


kevinsung:
I agree that ideally there would be one clean commit. Unfortunately that won't be possible until #697, #716, and #730 are merged.

woodsp-ibm:
> I agree that ideally there would be one clean commit.
Great, lets make it so once the dependents are dealt with.

kevinsung:
Too stale now.

PR #736, opened 2022-07-12
Create script to check if reno release note was committed to the wrong location

manoelmarques:
### Summary
Following the same check added to Terra at Qiskit/qiskit-terra#8320

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ all_check: spell style lint copyright mypy clean_sphinx html doctest
 lint:
 	python -m pylint -rn --ignore=gauopen qiskit_nature test tools
 	python tools/verify_headers.py qiskit_nature test tools
+	python tools/find_stray_release_notes.py
 
 mypy:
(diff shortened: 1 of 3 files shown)


mrossinek (approved):
Fine by me. Although I would have preferred a solution based on `pathlib` as suggested here: https://github.com/Qiskit/qiskit-terra/pull/8320#pullrequestreview-1034622386

Let's see if Terra updates things down the road..

manoelmarques:
> Fine by me. Although I would have preferred a solution based on `pathlib` as suggested here: [Qiskit/qiskit-terra#8320 (review)](https://github.com/Qiskit/qiskit-terra/pull/8320#pullrequestreview-1034622386)
> 
> Let's see if Terra updates things down the road..

Yes, Terra is checking after the release file has been committed, not before. I will update here if a new version is created there. For CI purposes it works fine, however for a local test it would flag only after the commit.

Issue #711, opened 2022-06-20
Migrate `qiskit_nature.algorithms`


mrossinek:
### What should we add?

As part of #701, all existing modules of Qiskit Nature will be migrated.
This issue outlines the requirements for the existing `algorithms` module.
This issue is blocked by #702 and #706.

> ⚠️ This migration needs to **copy** the code in order to abide the deprecation process!

The current module structure is roughly as following:
```
qiskit_nature/
    algorithms
        excited_states_solvers/
        ground_state_solvers/
        initial_points/
        pes_samplers/
```

The following modules can be migrated as a whole into their new locations:
```
qiskit_nature/
    second_quantization/
        algorithms/
            excited_states_solvers/
            ground_state_solvers/
            initial_points/
```

Just like with the **#Circuits**, this is the case because they actually implicitly depend on
objects from the `second_quantization` stack.
In the future, we may try to extract some base interfaces into `qiskit_nature/algorithms`, but for
the time being, this will not be done.

The new location for the `pes_samplers` module is yet to be determined.

> This will be updated later!

Issue #710, opened 2022-06-20
Migrate `qiskit_nature.circuit`


mrossinek:
### What should we add?

As part of #701, all existing modules of Qiskit Nature will be migrated.
This issue outlines the requirements for the existing `circuit` module.
This issue is blocked by #702.

> ⚠️ This migration needs to **copy** the code in order to abide the deprecation process!

The entire `qiskit_nature/circuit` module can be migrated to
`qiskit_nature/second_quantization/circuit`.

The reason for this, is that all of its current components actually depend on objects from the
`second_quantization` stack implicitly (e.g. the `operators` module).

Issue #709, opened 2022-06-20
Migrate `qiskit_nature.drivers`


mrossinek:
### What should we add?

As part of #701, all existing modules of Qiskit Nature will be migrated.
This issue outlines the requirements for the existing `drivers` module.
This issue is blocked by #706.

> ⚠️ This migration needs to **copy** the code in order to abide the deprecation process!

Below are the current module contents:
```
qiskit_nature/
    drivers/
        Molecule
        QMolecule  (already deprecated)
        UnitsType
        WatsonHamiltonian  (already deprecated)
        bosonic_bases/  (already deprecated)
        second_quantization/
            BaseDriver
            ElectronicStructureDriver
            ElectronicStructureMoleculeDriver
            VibrationalStructureDriver
            VibrationalStructureMoleculeDriver
            fcidumpd/
                FCIDumpDriver
            gaussiand/
                GaussianDriver
                GaussianForcesDriver
                GaussianLogDriver
                GaussianLogResult
            hdf5d/
                HDF5Driver
            psi4d/
                Psi4Driver
            pyquanted/
                PyQuanteDriver
            pyscfd/
                PySCFDriver
```

Here are the new locations for all of these classes.
```
qiskit_nature/
    second_quantization/
        problems/
            factories/
                Molecule
                UnitsType
                FCIDumpDriver
                ElectronicStructureMoleculeDriver
                VibrationalStructureMoleculeDriver
                Psi4Driver
                PySCFDriver
                GaussianForcesDriver
                (maybe more)
```

Unittests should be migrated accordingly.

Although we are moving away from the entire driver interface (where Qiskit controls the classical
codes), we will keep some of the existing drivers as useful educational and testing tools.

PR #746, opened 2022-07-15
refactor: move code to new structure

### LINKED ISSUE #701: [refactor] Towards a non-driver focused future

mrossinek:
### Summary

This is the follow-up to #722 

### Details and comments

This PR uses the approach describes in https://stackoverflow.com/a/44036771 to "copy" files while preserving the blame history in both locations.

:warning: This PR must be merged withOUT squashing all commits (I may however still be able to clean up some of the commits (e.g. the CI changes and `main` merges)

Closes #701 #702 #703 #704 #705 #706 #707 #708 #709 #710 #711


mrossinek:
FYI: I will be adding commits here until the CI is fixed. Once that is the case, I will rebase to minimize the number of commits being merged into `main`.

mrossinek:
My rebase caused the history preservation to not work as seen here: https://github.com/Qiskit/qiskit-nature/blame/refactor/qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_eigensolver.py

I will revisit this on Monday to fix that.

mrossinek:
The history preserving aspect of this PR is now working as can be seen under the following two links:
- https://github.com/Qiskit/qiskit-nature/blame/refactor/qiskit_nature/second_q/algorithms/excited_states_solvers/__init__.py
- https://github.com/Qiskit/qiskit-nature/blame/refactor/qiskit_nature/algorithms/excited_states_solvers/__init__.py

mrossinek:
I also marked this PR to be closing all of the code relocation issues. I will open separate issues for the planned refactorings within the modules in the near future :+1:

ElePT:
Yay! Git blame works for both the old and the new modules!

ElePT:
One small comment, the tutorials I removed (5,6 and 9) are still removed in this PR. It doesn't make sense to migrate them, but shouldn't we keep the old version of these together with the old version of the repo until we fully deprecate it?

mrossinek:
> One small comment, the tutorials I removed (5,6 and 9) are still removed in this PR. It doesn't make sense to migrate them, but shouldn't we keep the old version of these together with the old version of the repo until we fully deprecate it?

Hm that's a good point actually. For the docs I decided (based on a discussion with @manoelmarques and previous input from @woodsp-ibm) that only the new code is documented in the API (i.e. the old code no longer has API documentation). Nonetheless, I agree that it makes sense to keep the legacy tutorials in place. I will re-add those in an upcoming commit :+1: 

> I also took a quick look at the docs and they seem ok to my untrained eye.

Thanks :+1:

manoelmarques (approved):

PR #754, opened 2022-07-20
Ensure MP2 energy correction matches that from PySCF.

### LINKED ISSUE #739: MP2 Energy Correction is slightly off
### Environment

- **Qiskit Nature version**: 2b18a65ec1af71e6d9c066882afee723e72a8231
- **Python version**: 3.9.13
- **Operating system**: Fedora 36 (Kernel 5.18.10)

### What is happening?

The MP2 energy computed by the `MP2InitialPoint` does not match the value computed by e.g. PySCF exactly.
I found this discrepancy while looking into this question: https://quantumcomputing.stackexchange.com/q/27305/13342
I also used the following patch in order to compare the computed t2 amplitudes with those of PySCF:
```diff
diff --git a/qiskit_nature/algorithms/initial_points/mp2_initial_point.py b/qiskit_nature/algorithms/initial_points/mp2_initial_point.py
index 7ac4e757..c33c5321 100644
--- a/qiskit_nature/algorithms/initial_points/mp2_initial_point.py
+++ b/qiskit_nature/algorithms/initial_points/mp2_initial_point.py
@@ -84,6 +84,8 @@ class MP2InitialPoint(InitialPoint):
         self._reference_energy: float = 0.0
         self._corrections: list[_Correction] | None = None
 
+        self._t2: np.ndarray = None
+
     @property
     def ansatz(self) -> UCC:
         """The UCC ansatz.
@@ -241,6 +243,10 @@ class MP2InitialPoint(InitialPoint):
         Returns:
             Dictionary with MP2 coefficients and energy_corrections for each excitation.
         """
+        nmo = self._integral_matrix.shape[0]
+        nocc = len(self._orbital_energies[self._orbital_energies < 0.0])
+        nvir = nmo - nocc
+        self._t2 = np.zeros((nocc, nocc, nvir, nvir))
         corrections: list[_Correction] = []
         for excitation in self._excitation_list:
             if len(excitation[0]) == 2:
@@ -303,6 +309,15 @@ class MP2InitialPoint(InitialPoint):
         coefficient = -expectation_value / orbital_energy_correction
         coefficient = coefficient if abs(coefficient) > threshold else 0.0
 
+        if coefficient:
+            nocc = self._t2.shape[0]
+
+            if self._t2[i, j, a - nocc, b - nocc]:
+                print(i, j, a - nocc, b - nocc, excitation, self._t2[i, j, a - nocc, b - nocc],
+                      coefficient, expectation_value_iajb, expectation_value_ibja)
+
+            self._t2[i, j, a - nocc, b - nocc] += coefficient
+
         energy_correction = coefficient * expectation_value_iajb
         energy_correction = energy_correction if abs(energy_correction) > threshold else 0.0
 
```
See below for further reproducing steps

```python
from pyscf import gto
import numpy as np

from qiskit_nature.algorithms.initial_points import MP2InitialPoint
from qiskit_nature.circuit.library.ansatzes import UCC
from qiskit_nature.circuit.library.initial_states import HartreeFock

### Qiskit Nature
driver = PySCFDriver(
    atom="Li 0 0 0; H 0 0 1",
    basis="sto3g",
)
driver_results = driver.run()

qubit_converter = QubitConverter(mapper=JordanWignerMapper(), two_qubit_reduction=True)

HF = HartreeFock(num_spin_orbitals=12, num_particles=(2, 2), qubit_converter=qubit_converter)
UCC_operator = UCC(
    num_spin_orbitals=12,
    num_particles=(2, 2),
    excitations="sd",
    qubit_converter=qubit_converter,
    initial_state=HF,
)

mp2_initial_point = MP2InitialPoint()
mp2_initial_point.compute(UCC_operator, driver_results)
initial_point = mp2_initial_point.to_numpy_array()

print(
    f"E(MP2) = {mp2_initial_point.get_energy()} "
    f"E_corr = {mp2_initial_point.get_energy_correction()}"
)

### PySCF
mol = gto.M(atom="Li 0 0 0; H 0 0 1", basis="sto3g")
mp = mol.MP2().run()
t2 = mp.t2
t2[np.abs(t2) < 1e-10] = 0

### Comparison
print(t2)
print(mp2_initial_point._t2)

diff = t2 - mp2_initial_point._t2
diff[np.abs(diff) < 1e-10] = 0
print(list(zip(*np.nonzero(diff))))
print(diff[np.nonzero(diff)])
```
The above script produces the following output in combination with the previously provided patch:
```
0 1 0 3 ((0, 7), (2, 11)) 0.006457699858519577 0.006457699858519577 -0.012961561164766152 -0.004112860235425329
0 1 0 3 ((6, 7), (8, 11)) 0.012915399717039155 0.006457699858519577 -0.012961561164766152 -0.004112860235425329
E(MP2) = -7.778966491316218 E_corr = -0.01160435556765448
converged SCF energy = -7.76736213574856
E(MP2) = -7.77879908757281  E_corr = -0.0114369518242533
[[[[-0.0039749   0.          0.         -0.00169106]
   [ 0.         -0.00195561  0.          0.        ]
   [ 0.          0.         -0.00195561  0.        ]
   [-0.00169106  0.          0.         -0.00118934]]

  [[-0.00021832  0.          0.          0.00383773]
   [ 0.         -0.00276735  0.          0.        ]
   [ 0.          0.         -0.00276735  0.        ]
   [ 0.00121776  0.          0.         -0.00214932]]]


 [[[-0.00021832  0.          0.          0.00121776]
   [ 0.         -0.00276735  0.          0.        ]
   [ 0.          0.         -0.00276735  0.        ]
   [ 0.00383773  0.          0.         -0.00214932]]

  [[-0.01202768  0.          0.          0.02187818]
   [ 0.         -0.02863711  0.          0.        ]
   [ 0.          0.         -0.02863711  0.        ]
   [ 0.02187818  0.          0.         -0.06583919]]]]
[[[[-0.0039749   0.          0.         -0.00169106]
   [ 0.         -0.00195561  0.          0.        ]
   [ 0.          0.         -0.00195561  0.        ]
   [-0.00169106  0.          0.         -0.00118934]]

  [[-0.00021832  0.          0.          0.0193731 ]
   [ 0.         -0.00276735  0.          0.        ]
   [ 0.          0.         -0.00276735  0.        ]
   [-0.00140221  0.          0.         -0.00214932]]]


 [[[-0.00021832  0.          0.         -0.00140221]
   [ 0.         -0.00276735  0.          0.        ]
   [ 0.          0.         -0.00276735  0.        ]
   [ 0.0064577   0.          0.         -0.00214932]]

  [[-0.01202768  0.          0.          0.02187818]
   [ 0.         -0.02863711  0.          0.        ]
   [ 0.          0.         -0.02863711  0.        ]
   [ 0.02187818  0.          0.         -0.06583919]]]]
[(0, 1, 0, 3), (0, 1, 3, 0), (1, 0, 0, 3), (1, 0, 3, 0)]
[-0.01553537  0.00261997  0.00261997 -0.00261997]
```

### What should happen?

The computed t2 amplitudes should match the ones computed by PySCF.

### Any suggestions?

It could be, that the excitation generator used by the UCC ansatz is discarding some excitations which are considered to be symmetric. This could cause some amplitudes to be missing entirely (i.e. a symmetric counterpart of them). But this is nothing more than a guess right now.

declanmillar:
### Summary

Closes #739. @mrossinek found that the MP2 energy correction in Qiskit Nature differed from that in PySCF for particular molecules.

This was due to the Qiskit implementation including equivalent symmetric terms and conflating some parts of the energy computation with the amplitude computation. This fix overhauls the calculation and ensures that we exclude symmetric contributions from the total energy correction. It also changes the unit tests to directly compare with PySCF, rather than with fixed data from Qiskit Aqua.


qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -285,25 +282,19 @@ def _compute_correction(
...
         threshold = self._threshold
 
-        # Infer the number of molecular orbitals from the MO matrix.
-        num_molecular_orbitals: int = integral_matrix.shape[0]
-
-        i = excitation[0][0] % num_molecular_orbitals
-        j = excitation[0][1] % num_molecular_orbitals
-        a = excitation[1][0] % num_molecular_orbitals
-        b = excitation[1][1] % num_molecular_orbitals
+        [[i, j], [a, b]] = np.asarray(excitation) % integral_matrix.shape[0]
```
mrossinek, line 285: Nice :slightly_smiling_face:

test/algorithms/initial_points/test_mp2_initial_point.py
```diff
@@ -190,78 +190,63 @@ def get_matrix(value: int | None = None) -> np.ndarray:
...
 
-        Full excitation sequences generated using:
+            # Compute the PySCF result
+            pyscf_mol = gto.M(atom=atom, basis="sto3g", verbose=0)
+            pyscf_mp = pyscf_mol.MP2().run(verbose=0)
 
-        .. code-block:: python
+            driver = PySCFDriver(atom=atom, basis="sto3g")
+        except (ModuleNotFoundError, MissingOptionalLibraryError):
+            self.skipTest("PySCF driver does not appear to be installed.")
```
mrossinek, line 212: There is a decorator which handles this for you automatically:
```python
```

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -322,7 +313,11 @@ def get_energy_corrections(self) -> np.ndarray:
 
     def get_energy_correction(self) -> float:
         """The overall energy correction."""
-        return self.get_energy_corrections().sum()
+        unique_corrections = {
+            str(np.asarray(exc) % self._integral_matrix.shape[0]): corr
+            for corr, exc in zip(self.get_energy_corrections(), self.excitation_list)
+        }
```
mrossinek, line 319: I'm not sure I fully where the filtering is going on here.
I assume, the dictionary keys might clash at some point, but do they do some in a controllable manner? I.e. in terms of order? How do you ensure (if they clash) that the correct one is being used?

test/algorithms/initial_points/test_mp2_initial_point.py
```diff
@@ -190,78 +190,63 @@ def get_matrix(value: int | None = None) -> np.ndarray:
...
-
-        with self.subTest("Test MP2 energy corrections."):
-            np.testing.assert_array_almost_equal(
-                mp2_initial_point.get_energy_corrections(), energy_corrections, decimal=6
-            )
-
         with self.subTest("Test overall MP2 energy correction."):
             np.testing.assert_array_almost_equal(
-                mp2_initial_point.get_energy_correction(), energy_correction, decimal=6
+                mp2_initial_point.get_energy_correction(), pyscf_mp.e_corr, decimal=10
```
mrossinek, line 243: Would it be a lot of effort to also store the computed `t2` amplitudes and compare those with PySCF? I think that could be useful

mrossinek (changes requested):
Besides these initial comments, I need to ask you to apply all these changes on the new code location `qiskit_nature/second_q/algorithms` and leave the `qiskit_nature/algorithms` code entirely unchanged. See #746 for more details

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
declanmillar: My (probably wrong) assumption was these would always be identical. For example, the excitation with occupied and virtual molecular orbital indicies `((0, 1), (2, 5))` has three corresponding excitations when indexed by spin orbital:
```python
((0, 1), (2, 5)):  9.785199275020633e-08
((0, 1), (2, 5)):  9.785199275020633e-08
((0, 1), (2, 5)):  9.785199275020633e-08
```
My limited computational chemistry knowledge led me to believe this would always be the case, so I just take whichever one is computed last.

test/algorithms/initial_points/test_mp2_initial_point.py
declanmillar: No. Shouldn't be, I can also then use it to produce the energy correction summation above in a cleaner way.

qiskit_nature/algorithms/initial_points/mp2_initial_point.py
mrossinek: These are in fact only guaranteed to be identical in the restricted-spin setting (which (right now) is the only setting currently supported by the MP2 initial point anyways)..
Your explicit example did however clear up my question as I failed to recognize before, how the modulo operation was filtering various states.

I'm assuming, that within the assumption of the current restricted-spin formula for the energy correction computation, this works fine. When moving towards unrestricted spin cases, we will need to update both, that formula and this filtering technique however. Maybe leave a comment here in the code, indicating this :+1:

test/algorithms/initial_points/test_mp2_initial_point.py
declanmillar: So, the T2 amplitudes do seem to differ in the PySCF case. Assuming I've converted indices correctly (I keep getting tied in knots by them), for the following excitations they differ by:
```
((0,1),(2,5)): -0.00034652
((0,1),(5,2)):  0.00034652  
((1,0),(2,5)):  0.00034652
((1,0),(5,2)): -0.00034652
```
These excitations are in the excitation list, so I'm going to have to desect the PySCF code a little more to see where this difference in introduced. As you can see, the differences cancel in the energy correction summation, so that does agree with PySCF.

mrossinek: Seeing that, this may boil down to the symmetric excitation filtering being done here:
https://github.com/Qiskit/qiskit-nature/blob/c6c857f18833393f9c68ea60acdf29f27954b662/qiskit_nature/second_q/circuit/library/ansatzes/utils/fermionic_excitation_generator.py#L229
See the `visited_excitations`. Might be worth, looking into that :+1:

declanmillar:
I've totally overhauled the calculation now. I'm not sure the previous method was correct. The new logic lives in `_compute_t2`. The energy correction result is the same as the Aqua-inherited one, but the amplitudes/coefficients are now different and match PySCF. Due to symmetries, the difference cancelled when computing the total energy, so this may be why it was previously missed.

mrossinek (commented):
This is looking very good!
Just some minor comments :+1:

qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -14,75 +14,87 @@
...
+    Returns:
+        T amplitudes t2[i, j, a, b] (i, j in occupied, a, b in virtual).
+        The MP2 energy correction.
+
+    """
+    num_occ = len(orbital_energies[orbital_energies < 0.0])
+
+    # Use NumPy broadcasting to compute all occupied-virtual energy deltas.
+    energy_deltas = orbital_energies[:num_occ, np.newaxis] - orbital_energies[np.newaxis, num_occ:]
+    double_deltas = energy_deltas[:, :, np.newaxis, np.newaxis] + energy_deltas
```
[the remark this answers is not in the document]
declanmillar: I've added a couple of extra comments to try to clarify what I'm doing here. Let me know if they still need some work.

qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -14,75 +14,87 @@
...
+
+    Returns:
+        T amplitudes t2[i, j, a, b] (i, j in occupied, a, b in virtual).
+        The MP2 energy correction.
+
+    """
+    num_occ = len(orbital_energies[orbital_energies < 0.0])
+
+    # Use NumPy broadcasting to compute all occupied-virtual energy deltas.
+    energy_deltas = orbital_energies[:num_occ, np.newaxis] - orbital_energies[np.newaxis, num_occ:]
```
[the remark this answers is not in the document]
declanmillar: I added a quick comment to give them the heads-up.

qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -14,75 +14,87 @@
...
+    Args:
+        integral_matrix: The two-body molecular orbitals matrix.
+        orbital_energies: The orbital energies.
+
+    Returns:
+        T amplitudes t2[i, j, a, b] (i, j in occupied, a, b in virtual).
+        The MP2 energy correction.
+
+    """
+    num_occ = len(orbital_energies[orbital_energies < 0.0])
```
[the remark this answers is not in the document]
declanmillar: We pass in the full `grouped_property`, so we can access it. It'll just need some extra checks.

qiskit_nature/second_q/algorithms/initial_points/hf_initial_point.py
```diff
@@ -147,10 +147,18 @@ def compute(
         self._compute()
 
     def _compute(self) -> None:
-        self._parameters = np.zeros(len(self._excitation_list))
+        """Compute the coefficients given an excitation list.
 
-    def get_energy(self) -> float:
-        """The reference energy.
+        All excitations will have zero coefficient.
```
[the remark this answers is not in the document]
declanmillar: I've rephrased this as suggested.

qiskit_nature/second_q/algorithms/initial_points/hf_initial_point.py
```diff
@@ -147,10 +147,18 @@ def compute(
...
+
+        Returns:
+            The coefficients for each excitation.
```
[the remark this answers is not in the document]
declanmillar: Yep, updated.

qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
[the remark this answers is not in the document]
declanmillar: Out of curiosity, what sort of molecules would feature this?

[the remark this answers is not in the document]
mrossinek: When you start looking at very complex metalorganics, depending on the quality of your initial method, I've seen this happen.
It's rare, but I'd rather not rely on it, if we have a better way :+1:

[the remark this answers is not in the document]
declanmillar: I opted to use `occupied_alpha` as this has a predictable format. Am I right in thinking that this will always be present in a problem, or should I check for it as we do for the ElectronicEnergy etc?

qiskit_nature/second_q/algorithms/initial_points/hf_initial_point.py
```diff
@@ -147,10 +147,18 @@ def compute(
...
+        """Computes the HF initial point array for a given excitation list.
+
+        In the Hartree-Fock case this is simply an all-zero array.
+
+        Returns:
+            An all-zero array with the same length as the excitation list.
+        """
         self._parameters = np.zeros(len(self._excitation_list))
 
-    def get_energy(self) -> float:
```
declanmillar, line 152: Rather than removing this function, should I formally depreciate it in favour of the property?

mrossinek: Since this is fully in the new stack, there is no need. But you can write in the release note that this name changed :+1:

declanmillar: Perfect. I had already noted the change in the release note.

qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -14,75 +14,101 @@
...
+    :attr:`grouped_property` and :attr:`ansatz` to be passed as arguments or the
     :attr:`grouped_property` and :attr:`excitation_list` attributes to be set already.
 
-    ``MP2InitialPoint`` requires the
-    :class:`~qiskit_nature.second_q.properties.ElectronicEnergy`, which should
-    be passed in via the :attr:`grouped_property` attribute. From this it must obtain the two-body
-    molecular orbital electronic integrals and orbital energies. If the Hartree-Fock reference
-    energy is also obtained, it will be used to compute the absolute MP2 energy using the
-    :meth:`get_energy` method.
+    ``MP2InitialPoint`` requires the :class:`~qiskit_nature.second_q.properties.ElectronicEnergy`,
```
mrossinek, line 87: This should be updated to also reflect that the `ParticleNumber` is also being used :+1: 

Note: with the upcoming refactoring, this will change slightly since the number of particles will actually be stored on the problem itself. So for now you can simply use this as done now, but this will get improved a bit more shortly :+1:

qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -124,13 +149,12 @@ def grouped_property(self) -> GroupedSecondQuantizedProperty | None:
...
-        must contain the two-body molecular orbitals matrix and the orbital energies. Optionally,
-        it will also use the Hartree-Fock reference energy to compute the absolute energy.
+        :class:`~qiskit_nature.second_q.properties.ElectronicEnergy`, which must contain the
+        two-body molecular orbitals matrix and the orbital energies. Optionally, it will also use
+        the Hartree-Fock reference energy to compute the absolute energy.
 
         Raises:
-            QiskitNatureError: If
-                :class:`~qiskit_nature.second_q.properties.ElectronicEnergy`
+            QiskitNatureError: If :class:`~qiskit_nature.second_q.properties.ElectronicEnergy`
```
mrossinek, line 157: We should do the same when `ParticleNumber` is missing

mrossinek (commented):
Just two minor comments, other than this, I think this LGTM :+1:

mrossinek (dismissed):
Thank you, @declanmillar! :slightly_smiling_face:

woodsp-ibm:
I know this is not directly related to the change here. But after the refactor/move of the init file in initial_points folder in the new second_q location it no longer has the docstring that lists the classes for the initial points. You can navigate through the initial_points reference in problems but you end up on a blank page now. And as such the links in the release note do not work. I know we are talking about going over the docs later to fix/improve after this second_q move/refactor but maybe we can at least get the initial points linked back in - it should be a simple copy/paste from the original init file right. Otherwise doc changes are not being run via sphinx etc.

mrossinek (dismissed):

woodsp-ibm (dismissed):

mrossinek (approved):
I hope I didn't mess up resolving those conflicts :sweat_smile:

Issue #666, opened 2022-05-25
Implement `FermionicOp.from_polynomial_tensor`


mrossinek:
### What should we add?

#665 proposes the addition of a `PolynomialTensor` class to handle coefficient storage.

This issue tracks the implementation of the method which can construct a `FermionicOp` from those coefficients.
Below, I summarize the proposed structure of the `PolynomialTensor.data` field:

Currently, coefficients that will later become a `FermionicOp` are stored in most cases in the `ElectronicIntegrals`. Those are tailored to the electronic structure problems allowing some memory reduction in scenarios of identical coefficients for spin-up and -down orbitals, rendering them not generally applicable.
Instead, I propose a structure of `data` as follows:
```
{
  "+-": [
    [1, 2],
    [3, 4]
  ],
  "++--": [
    [[
      [1, 2],
      [3, 4]
    ], [
      [5, 6],
      [7, 8]
    ]], [[
      [9, 10],
      [11, 12]
    ], [
      [13, 14],
      [15, 16]
    ]]
  ]
}
```
Here `+-` stores a one-body term where the first axis of the stored matrix will be mapped to `+` operators, and the second axis to `-`. I.e. this will result in a `FermionicOp`:
```
1.0 * "+_0 -_0" + 2.0 * "+_0 -_1" + 3.0 * "+_1 -_0" + 4.0 * "+_1 -_1"
```
Accordingly, `++--` indicates a two-body term, and, even more importantly, unambiguously stores the two-body terms in the setting of electronic structure problems, in the physicist notation:

$$
g_{pqrs} a^{\dagger}_p a^{\dagger}_q a_r a_s
$$

For brevity, I will not expand the 16 coefficients into the operator but will only show a few examples:
- `data["++--"][0, 0, 0, 0] -> 1.0 * "+_0 +_0 -_0 -_0"`
- `data["++--"][0, 1, 0, 1] -> 6.0 * "+_0 +_1 -_0 -_1"`
- `data["++--"][1, 0, 0, 0] -> 9.0 * "+_1 +_0 -_0 -_0"`

For the specific case of electronic structure problems, we can implement a specialized variant of the `PolynomialTensor` which will allow simplified construction from chemistry-notation two body integrals (using the utilities developed in #520), can implement memory-reduction methods to exploit spin symmetries, and (in the future) could also exploit further symmetry information in chemical systems (like 8-fold symmetry of AO ERIs, for example).

PR #756, opened 2022-07-20
adding PolynomialTensor class

SaashaJoshi:
### Summary
Adding a [draft] file for `PolynomialTensor` class.
[WIP] corresponding unittest.

### Details and comments
The PolynomialTensor class, for now, overrides the `mul` and `add` functions from `StarAlgebraMixin`. More brainstorming needs to be done for `compose` and `adjoint` functions.

### DIFF
--- a/qiskit_nature/second_q/operators/polynomial_tensor.py
+++ b/qiskit_nature/second_q/operators/polynomial_tensor.py
@@ -1,0 +1,74 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2022.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Polynomial Tensor class"""
+
+from __future__ import annotations
+import numpy as np
+from typing import Dict
+from qiskit.opflow.mixins import StarAlgebraMixin
+from qiskit.quantum_info.operators.mixins import TolerancesMixin
+
+
+class PolynomialTensor(StarAlgebraMixin, TolerancesMixin):
+    """Polynomial Tensor class"""
+
+    def __init__(self, data: Dict[str, np.ndarray]):
+        self._data = data
+
+        for key, value in self._data.items():
+            if len(np.shape(value)) == len(key):
+                pass
+            else:
+                raise ValueError(
+                    f"data key length {len(key)} and number of value dimensions {np.shape(value)} do not match"
+                )
+
+    def mul(self, other: complex):
+        """scalar multiplication of PolynomialTensor with complex"""
+
+        # other might be float, int or etc. quantum info - typing is Number (check)
+        prod_dict = {}
+
+        for key, matrix in self._data.items():
+            prod_dict[key] = matrix * other
+        return PolynomialTensor(prod_dict)
+
+    def add(self, other: PolynomialTensor):
+        """addition of PolynomialTensors"""
+
+        # Polynomial tensor immutable!
+        sum_dict = self._data  # or try empty dict {}
+
+        if other is not isinstance(other, PolynomialTensor):
+            raise TypeError("Incorrect argument type: other should be PolynomialTensor")
+
+        for key, value in other.items():
+            if key in self._data.keys():
+                if np.shape(value) == np.shape(self._data[key]):
+                    print(f"key {key} of same dimension present in both")
+                    sum_dict[key] = np.add(value, self._data[key])
+                else:
+                    raise ValueError(
+                        f"Dictionary value dimensions {np.shape(value)} and {np.shape(self._data[key])} do not match"
+                    )
+            else:
+                print(f"adding a new key {key}")
+                sum_dict[key] = value
+
+        return PolynomialTensor(sum_dict)
+
+    def compose(self, other):
+        pass
+
+    def adjoint(self):
+        pass


qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,74 @@
...
+                        f"Dictionary value dimensions {np.shape(value)} and {np.shape(self._data[key])} do not match"
+                    )
+            else:
+                print(f"adding a new key {key}")
+                sum_dict[key] = value
+
+        return PolynomialTensor(sum_dict)
+
+    def compose(self, other):
+        pass
```
mrossinek, line 71: I think the implementation of `compose` will not be possible in this generalized form. Since the class can contain differing dimensionalities at the same time, composition in general is highly non-trivial. I suggest, that instead of trying to implement the `StarAlgebraMixin`, we implement the following mixins:
```python
from qiskit.quantum_info.operators.mixins import (
    LinearMixin,  # includes MultiplyMixin  -> add, mul
    AdjointMixin,  # conjugate, transpose
    TolerancesMixin,
)
```
You can then safely remove the `compose` method from this class

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def adjoint(self):
+        pass
```
mrossinek, line 74: If we do the above, you need to implement `conjugate` and `transpose` instead of `adjoint`.
Those should do the following:
- `conjugate`: should simply call `np.conjugate(...)` on all matrices stored in the internal dict
- `transpose`: should simply transpose all the internal matrices

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,119 @@
...
+                    f"data key {key} of length {len(key)} does not match "
+                    f"data value matrix of dimensions {value.shape}"
+                )
+            if len(set(value.shape)) != 1:
+                raise ValueError(
+                    f"For key {key}: dimensions of value matrix are not identical {value.shape}"
+                )
+
+        if len(set(shapes)) != 1:
+            raise ValueError("Dimensions of value matrices in data dictionary are not identical.")
```
mrossinek, line 49: How about the following instead?
```suggestion
        shapes = set()
        for key, value in self._data.items():
            if len(value.shape) != len(key):
                raise ValueError(
                    f"data key {key} of length {len(key)} does not match "
                    f"data value matrix of dimensions {value.shape}"
                )

            shape = set(value.shape)

            if len(shape) != 1:
                raise ValueError(
                    f"For key {key}: dimensions of value matrix are not identical {value.shape}"
                )

            shapes.update(shape)

        if len(shapes) != 1:
            raise ValueError("Dimensions of value matrices in data dictionary are not identical.")
```
This has the following improvements:
- only adds `value.shape` to `shapes`, if the checks pass
- uses `shapes` as a `set` directly, without going through a `tuple` first

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,119 @@
...
+
+        return PolynomialTensor(sum_dict)
+
+    def _add(self, other, qargs=None):
+        return self.add(other)
+
+    def __eq__(self, other):
+        """Check equality of PolynomialTensors"""
+
+        if self._data.keys() == other._data.keys():
```
mrossinek, line 94:
```suggestion
        if set(self._data.keys()) == set(other._data.keys()):
```
I think otherwise, this is dependent on the order of the keys

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,119 @@
...
+            for key in self._data.keys():
+                if np.allclose(self._data[key], other._data[key], atol=self.atol, rtol=self.rtol):
+                    return isinstance(other, PolynomialTensor)
+                else:
+                    return False
```
mrossinek, line 99: This won't work as intended, because this will always return after the first key.

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            return False
```
mrossinek, line 101: Actually, let me do this suggestion:
```suggestion
        if set(self._data.keys()) != set(other._data.keys()):
            return False

        for key in self._data.keys():
            if not np.allclose(self._data[key], other._data[key], atol=self.atol, rtol=self.rtol):
                return False

        return True
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,119 @@
...
+class PolynomialTensor(LinearMixin, AdjointMixin, TolerancesMixin):
+    """Polynomial Tensor class"""
+
+    def __init__(self, data: Dict[str, np.ndarray]):
+        if not isinstance(data, Dict):
+            raise TypeError(f"data {data} must be a Dictionary")
+
+        self._data = data
+
+        shapes = ()
```
mrossinek, line 35: I think it would be useful, after the following verification, to expose the final `shape` value as an attribute of this class.
This could then be used during e.g. `add` to ensure that the shapes of two instances match, without having to check all of the internal matrices (again)

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,119 @@
...
+    LinearMixin,
+    AdjointMixin,
+    TolerancesMixin,
+)
+
+
+class PolynomialTensor(LinearMixin, AdjointMixin, TolerancesMixin):
+    """Polynomial Tensor class"""
+
+    def __init__(self, data: Dict[str, np.ndarray]):
```
mrossinek, line 29:
```suggestion
    def __init__(self, data: Dict[str, complex | np.ndarray]):
```
I'd like to support constant values in this class, too.
This should look like this:
```python
PolynomialTensor({
    "": 1.0,
    "+-": np.eye(2),
})
```

mrossinek: We should also add `sparse` matrix support, but we can do that in the future and not necessarily directly as part of this PR :+1:

mrossinek (commented):
Some suggestions from my side. Other than that, this looks to be shaping up quite nicely :+1:

qiskit_nature/second_q/operators/polynomial_tensor.py
SaashaJoshi: Do we want the constants to be inside the dictionary or have them as a different argument all together?

mrossinek: For one, I would like them to be in the dictionary.
However, we could also envision them as an additional argument.

The difference would be, how they get handled later:
- constant terms inside the dict would get mapped to an all-identity operation inside the SparseLabelOp that gets build from the PolynomialTensor and, by extension, will also get mapped to an all-identity Pauli term
- additional constants as extra arguments should NOT get added to the operator down the line. They would effectively be a constant eigenvalue offset, which will not be measured and, thus, not incur any noise from the measurement

For now, let's only support constants as part of the dict. I think the extra offsets as arguments will mostly be needed by classes which contain a `PolynomialTensor`, not by the `PolynomialTensor` itself

woodsp-ibm: You want to support constants two ways here potentially. I had been thinking for it outside of the dict but still getting mapped to the identity. I think if you want constants that are not being mapped they can be handled outside of the class and stored in conjunction with some instance. Are you going to allow any key value for the constant or just an empty string? You say "them" as constants do you envision more than one. I had figured it would just be one value.

mrossinek: Yes, sorry, I meant a single constant value, too

I agree, a constant which is not to be mapped, should be handled outside of `PolynomialTensor`. All I meant is, that if in the future we see need for it, I would expect that kind of constant outside the dict.

I don't see a good reason for storing the constant outside the dict, when it is supposed to get mapped to the operator. We are iterating over the dict anyways, so why would we force us to look up another attribute if we can get the constant "for free" while iterating the dict items?

qiskit_nature/second_q/operators/polynomial_tensor.py
woodsp-ibm: We need to figure what we want to do to expose the data. As far as checking need for operations, add, mul, etc since the instances should each be valid, it only needs one matrix to be checked against one in the other to check size of dimension matches - then they either all match or they all don't, since they all should match inside a valid instance

mrossinek: I think having access to the `size` (which later translates to the `register_length` and even later to the `num_qubits`) will be useful also during the `SparseLabelOp.from_polynomial_tensor` implementations.

As to exposing `_data`, I would have defined an `__iter__` method which allows iteration over `_data.items()`.
Since `PolynomialTensor` is supposed to be an immutable object, read-only access by means of iterating the internal data should be sufficient.
Although we may also to expose access to individual objects in `_data`, but that would require the user to know the keys he would need to use for accessing..

qiskit_nature/second_q/operators/polynomial_tensor.py
woodsp-ibm: So always an empty key for this - hence it can only be one value and our signal its a constant. While it can be in the dict its kinda the odd one out somewhat. We can put it as a numpy array even though its a single value so things are more consistent perhaps - but internal checks over shapes/dims will have to special case this. In term of accessing the information, maybe it will be clearer that its more consistent as such in the dictionary once we define how we want to expose things

qiskit_nature/second_q/operators/polynomial_tensor.py
woodsp-ibm: Or we can expose keys() and let the user access any entry they want from a key. This would expose the matrix directly by reference (as would an iter I think) so immutability will be there in as much as its treated as read-only. Since its constructed with a dict being able to later access part of it by key seems consistent. (I had thought the constant would have its own access if it were separate)

qiskit_nature/second_q/operators/polynomial_tensor.py
mrossinek: Yes, the shape checking would need to handle this specifically. But I think that is worth the overhead on the internal site with the benefit of having one unified iteration over `_data` available (i.e. not having to check another `constant` attribute of the class instance after having iterated over the data)

woodsp-ibm: I guess I have been thinking random access to the data, not iterator, though we could do both! Ie you get the keys you can get any matrix you want, using any of the keys, and you can get the constant at any time. If we expose it all via an iterator I guess the consumer still has to deal with matrix versus a constant as the value - I guess they see the empty key and switch on that,

mrossinek: That's right. I had not thought of random access, but that might actually be possible like this, too:
```python
p_t = PolynomialTensor({
    "": 1.0,
    "+": np.arange(-2, 0),
    "+-": np.arange(2, 6).reshape((2, 2)),
})

p_t[()]  # 1.0
p_t[(0,)]  # -2.0
p_t[(1,)]  # -1.0
p_t[(0, 0)]  # 2.0
p_t[(0, 1)]  # 3.0
# etc.
```

woodsp-ibm: Yes I had thought about it at that level since it would not expose the matrices directly - but an element at a time made me wonder... maybe externally thats all one ever needs. I had thought the key might be needed too as part of the index, but if keys cannot ever be the same length, ie same dimension matrix, then there is no ambiguity without it.

mrossinek: Oh hold on, there may actually be keys of identical length but different content :/ yeh, then the above does not work

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,123 @@
...
+                sum_dict[other_key] = other_value
+
+        return PolynomialTensor(sum_dict)
+
+    def _add(self, other, qargs=None):
+        return self.add(other)
+
+    def __eq__(self, other):
+        """Check equality of PolynomialTensors"""
+
```
woodsp-ibm, line 99: This should have a check of other data type and if not a PolyTensor instance return False

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -97,6 +97,9 @@ def _add(self, other, qargs=None):
     def __eq__(self, other):
         """Check equality of PolynomialTensors"""
 
+        if not isinstance(other, PolynomialTensor):
+            raise TypeError("Incorrect argument type: other should be PolynomialTensor")
```
woodsp-ibm, line 101: This should simply `return False` since any other type is not equal.

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+    LinearMixin,
+    AdjointMixin,
+    TolerancesMixin,
+)
+
+
+class PolynomialTensor(LinearMixin, AdjointMixin, TolerancesMixin):
+    """Polynomial Tensor class"""
+
+    def __init__(self, data: Mapping[str, np.ndarray | Number]):
```
mrossinek, line 29:
```suggestion
    def __init__(self, data: Mapping[str, np.ndarray | Number]) -> None:
```
This needs to have a docstring added. Also, the docstrings of the other methods are missing quite a few aspects like `Args`, `Returns`, etc.
I suggest you take a look at #731 which will have docstrings very similar to the ones you need here

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+        copy_dict: Dict[str, np.ndarray] = {}
+
+        shapes = set()
```
mrossinek, line 32: Having just reviewed #731 again, too, I am wondering now if we (after all) could indeed benefit from having an explicit `register_length` here, which you avoid having to keep track of `shapes`.

@woodsp-ibm what do you think?

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+
+            dims = set(value.shape)
+
+            if len(dims) > 1:
+                raise ValueError(
+                    f"For key {key}: dimensions of value matrix are not identical {value.shape}"
+                )
+
+            shapes.update(dims)
+            copy_dict[key] = value
```
mrossinek, line 51: Again, also just having seen #731, should we also think of a `_copy: bool` keyword-argument for this initializer? I would think this would be useful, since it would avoid multiple copies during addition and multiplication.

You can take a look at the that PR to see how this is handled there :+1:

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+
+        if len(shapes) != 1:
+            raise ValueError("Dimensions of value matrices in data dictionary are not identical.")
+
+        self._data = copy_dict
+
+    def mul(self, other: complex):
```
mrossinek, line 58:
```suggestion
    def _multiply(self, other: complex) -> PolynomialTensor:
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+        if not isinstance(other, Number):
+            raise TypeError(f"other {other} must be a number")
+
+        prod_dict: Dict[str, np.ndarray] = {}
+        for key, matrix in self._data.items():
+            prod_dict[key] = np.multiply(matrix, other)
+        return PolynomialTensor(prod_dict)
+
+    def _multiply(self, other):
+        return self.mul(other)
```
mrossinek, line 70:
```suggestion
```
Simply implementing `_multiply` above is fine, as the `MultiplyMixin` deals with exposing this method to the public API

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def add(self, other: PolynomialTensor):
```
mrossinek, line 72:
```suggestion
    def _add(self, other: PolynomialTensor) -> PolynomialTensor:
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+                if other_value.shape == np.shape(sum_dict[other_key]):
+                    sum_dict[other_key] = np.add(other_value, sum_dict[other_key])
+                else:
+                    raise ValueError(
+                        f"For key {other_key}: "
+                        f"corresponding data value of shape {np.shape(sum_dict[other_key])} "
+                        f"does not match other value matrix of shape {other_value.shape}"
+                    )
+            else:
+                sum_dict[other_key] = other_value
```
mrossinek, line 90: I think we might be able to implement this nicer if we:
- have a `register_length` attribute which we can check once. This will then tell us that all shapes are correct
- then we can use dictionary comprehensions based on the assumptions that all shapes match

Again, you can see how we did this in the `SparseLabelOp` PR :+1:

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+
+        return PolynomialTensor(sum_dict)
+
+    def _add(self, other, qargs=None):
+        return self.add(other)
```
mrossinek, line 95:
```suggestion
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def __eq__(self, other):
+        """Check equality of PolynomialTensors"""
+
+        if not isinstance(other, PolynomialTensor):
+            raise TypeError("Incorrect argument type: other should be PolynomialTensor")
```
mrossinek, line 101:
```suggestion
            return False
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+
+        if self._data.keys() != other._data.keys():
+            return False
+        for key, value in self._data.items():
+            if not np.allclose(value, other._data[key], atol=self.atol, rtol=self.rtol):
+                return False
```
mrossinek, line 107: This is actually the implementation of the `equiv` method.

The `__eq__` method should check for exact equality. It would be good to implement both

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+        return True
+
+    def conjugate(self):
```
mrossinek, line 110:
```suggestion
    def conjugate(self) -> PolynomialTensor:
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,126 @@
...
+        """Conjugate of PolynomialTensors"""
+
+        conj_dict: Dict[str, np.ndarray] = {}
+        for key, value in self._data.items():
+            conj_dict[key] = np.conjugate(value)
+
+        return PolynomialTensor(conj_dict)
+
+    def transpose(self):
```
mrossinek, line 119:
```suggestion
    def transpose(self) -> PolynomialTensor:
```

mrossinek (commented):
Besides these comments, a release note needs to be added here, too
https://github.com/Qiskit/qiskit-nature/blob/main/CONTRIBUTING.md#release-notes

You can keep it rather minimal for the time being :+1:

qiskit_nature/second_q/operators/polynomial_tensor.py
SaashaJoshi: I did this to avoid unit testing a private method. But I can look for a way around it.

mrossinek: You can still unittest the `mul` method (or actually just use `number * tensor`)

SaashaJoshi: Yes, was just looking into this. Will make the changes soon.

qiskit_nature/second_q/operators/polynomial_tensor.py
mrossinek: You can actually disregard this suggestion. I realized after talking to Steve offline, that you are not copying the (potentially massive) matrices anywhere which is good as it is now :+1:

mrossinek:
I had another discussion with @woodsp-ibm and we said that it might be beneficial to directly implement `PolynomialTensor` as a subclass of `Mapping`. That would give full flexibility when accessing the internal `_data` object.
The way that you could achieve this, is by something similar to the following:
```python
from collections.abc import Mapping
from typing import Iterator

class PolynomialTensor(..., Mapping):
    ...

    def __getitem__(self, __k: str) -> np.ndarray | Number:
        return self._data.__getitem__(__k)

    def __len__(self) -> int:
        return self._data.__len__()

        return self._data.__iter__()
```

This will directly enable users to use `len(...)`, `.keys()`, `.values()`, `.items()` and even use `p_t["key"]` to access specific entries :+1:

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,194 @@
...
+                )
+
+            shapes.update(dims)
+            copy_dict[key] = value
+
+        if len(shapes) != 1:
+            raise ValueError("Dimensions of value matrices in data dictionary are not identical.")
+
+        self._data = copy_dict
+        self._register_length = register_length
```
mrossinek, line 67:
```suggestion
        for key, value in data.items():
            if isinstance(value, Number):
                value = np.asarray(value)

            if len(value.shape) != len(key):
                raise ValueError(
                    f"Data key {key} of length {len(key)} does not match "
                    f"data value matrix of dimensions {value.shape}"
                )

            dims = set(value.shape)

            if len(dims) > 1:
                raise ValueError(
                    f"For key {key}: dimensions of value matrix are not identical {value.shape}"
                )
            elif len(dims) == 1 and dims[0] != register_length:
                raise ValueError(
                    "Dimensions of value matrices in data dictionary do not match the provided "
                    f"register length, {register_length}"
                )


            copy_dict[key] = value

        self._data = copy_dict
        self._register_length = register_length
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,194 @@
...
+
+    @property
+    def register_length(self) -> int:
+        """ Returns register length of the operator key in `Polynomial Tensor` object """
+
+        return self._register_length
```
mrossinek, line 73:
```suggestion
        """Returns the register length of the operator stored in this `PolynomialTensor`."""
        return self._register_length
```

mrossinek (commented):
Please also double check the formatting of the docstrings. Especially `PolynomialTensor` should always be written as a single word

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,217 @@
...
+        return self._data.__len__()
+
+    def __iter__(self) -> Iterator[str]:
+        """
+        Returns iterator of the `PolynomialTensor`.
+        """
+
+        return self._data.__iter__()
+
+    def _multiply(self, other: complex):
```
mrossinek, line 101:
```suggestion
    def _multiply(self, other: complex) -> PolynomialTensor:
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,217 @@
...
+
+        if not isinstance(other, Number):
+            raise TypeError(f"other {other} must be a number")
+
+        prod_dict: Dict[str, np.ndarray] = {}
+        for key, matrix in self._data.items():
+            prod_dict[key] = np.multiply(matrix, other)
+        return PolynomialTensor(prod_dict, self._register_length)
+
+    def _add(self, other: PolynomialTensor, qargs=None):
```
mrossinek, line 120:
```suggestion
    def _add(self, other: PolynomialTensor, qargs=None) -> PolynomialTensor:
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,217 @@
...
+                if other_value.shape == np.shape(sum_dict[other_key]):
+                    sum_dict[other_key] = np.add(other_value, sum_dict[other_key])
+                else:
+                    raise ValueError(
+                        f"For key {other_key}: "
+                        f"corresponding data value of shape {np.shape(sum_dict[other_key])} "
+                        f"does not match other value matrix of shape {other_value.shape}"
+                    )
+            else:
+                sum_dict[other_key] = other_value
```
mrossinek, line 148:
```suggestion
        if self.register_length != other.register_length:
            raise ValueError(
                "The dimensions of the PolynomialTensors which are to be added together, do not "
                f"match: {self.register_length} != {other.register_length}"
            )

        sum_dict = {key: np.value + other._data.get(key, 0) for key, value in self._data.items()}
        other_unique = {key: other._data[key] for key in other._data.keys() - self._data.keys()}
        sum_dict.update(other_unique)
```
This is what I suggested in a previous review. I took the liberty of copying the suggestion from `SparseLabelOp` and adapting it very slightly for the purposes here

SaashaJoshi: Oh yes, I am still working on it.

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -0,0 +1,208 @@
...
+import numpy as np
+from qiskit.quantum_info.operators.mixins import (
+    LinearMixin,
+    AdjointMixin,
+    TolerancesMixin,
+)
+
+
+class PolynomialTensor(LinearMixin, AdjointMixin, TolerancesMixin, Mapping):
+    """PolynomialTensor class"""
```
mrossinek, line 28: For a future reviewer: I will flesh out this docstring in an upcoming PR when working on #666 :+1:

mrossinek (approved):
I'm happy with this as is! Thanks a lot, Saasha, for your continuous effort on this! :slightly_smiling_face: 

I'm sure that there will be some minor changes that I will be doing to this as I work on #666 but overall this looks like it is in very good shape to be merged now :+1:

PR #758, opened 2022-07-21
[stable] fix: ensure tutorials support settings.dict_aux_operators

mrossinek:
### Summary

This PR ensures that the stable tutorials support `settings.dict_aux_operators = True`

### DIFF
--- a/docs/tutorials/01_electronic_structure.ipynb
+++ b/docs/tutorials/01_electronic_structure.ipynb
@@ -8,6 +8,17 @@
    ]
   },
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qiskit_nature.settings import settings\n",
+    "\n",
+    "settings.dict_aux_operators = True"
+   ]
+  },
+  {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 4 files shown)


woodsp-ibm:
So even though there is a deprecation warning for using indexed based access it does not come out in these tutorials. I'm not sure I am entirely happy with exposing the access done either way in these, especially since we are saying we are switching away from list based to dict. Maybe we should set the settings in the tutorial and just do the dict based logic there?

mrossinek:
I'm fine with any approach that will make things consistent. Enabling dict-based operators via the setting in all tutorials seems like a good approach, too :+1: 
I can update this PR accordingly later today.

docs/tutorials/01_electronic_structure.ipynb
```diff
@@ -190,7 +201,7 @@
    "source": [
     "es_problem = ElectronicStructureProblem(driver)\n",
     "second_q_op = es_problem.second_q_ops()\n",
-    "print(second_q_op[0])"
+    "print(second_q_op[\"ElectronicEnergy\"])"
```
woodsp-ibm, line 204: In theory we have the `main_property_name` on the problem that avoids knowing/hard-coding a string value explicitly here. Should we be using that, or at least showing that here. I know the main goal was to go to the dictionary, but we had that way to allow users to get the main property without having to know. Here for instance we could get that name, print it as well as printing the operator.

ParticleNumber usage below evidently cannot be altered.

mrossinek: I figured that this is really just a temporary fix until the 0.5 release. At that point `main_property_name` won't exist any more. So maybe we can keep it simple like here?

woodsp-ibm (approved):

PR #759, opened 2022-07-21
Enable deprecation msg for List-based aux.ops

manoelmarques:
### Summary
Python default behavior is to not show deprecation msgs unless triggered directly inside a method called from `__main__`. 
Deprecation messages triggered by calls inside Nature code won't show.
This PR creates a special Deprecate category for list-based aux.operators and displays the deprecation msg just for this category overriding the default behavior.

Also, displays just once as it is called several times inside Nature.

All this will be removed after the deprecation period ends.

### DIFF
--- a/test/second_q/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
+++ b/test/second_q/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
@@ -47,7 +47,6 @@ from qiskit_nature.second_q.properties.integrals import (
     TwoBodyElectronicIntegrals,
 )
 from qiskit_nature.second_q.transformers import FreezeCoreTransformer
-from qiskit_nature import settings
 from qiskit_nature.second_q.algorithms.initial_points import MP2InitialPoint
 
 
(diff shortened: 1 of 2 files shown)


test/second_q/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
```diff
@@ -150,41 +149,6 @@ def test_aux_ops_reusability(self):
             frozenset(a.to_list()) == frozenset(b.to_list()) for a, b in zip(aux_ops, aux_ops_copy)
         )
 
-    def test_list_based_aux_ops(self):
```
mrossinek, line 153: Is it necessary that we remove this unittest already? I don't have a strong opinion on this, just want to clarify :+1:

mrossinek (approved):
Either way about the unittest, this is okay to be merged for me :+1:

test/second_q/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
manoelmarques: The same test already exists in the deprecated code. Since now the deprecation msg. only shows twice, the second time the test ran, it failed because the msg wouldn't show again (already shown in the first test). I could have found a way to fix it but didn't think necessary to have the exact same test in two places.

mrossinek: Ah I didn't pay enough attention, so I didn't notice that this was the `second_q` file... All good :slightly_smiling_face:

PR #762, opened 2022-07-22
Remove `GroupedProperty` dependency from `ElectronicDipoleMoment`

### LINKED ISSUE #751: refactor: make `ElectronicDipoleMoment` independent of `GroupedProperty`

As part of #701, the `GroupedProperty` class will be removed.
As a first step into that direction, the `ElectronicDipoleMoment` class should be refactored to remove its inheritance from that class.

ElePT:
### Summary
This PR fixes #751 and removes the `GroupedProperty` dependency from `ElectronicDipoleMoment` with a very basic proposal. I am open to adding getters/setters if necessary, but because there will be further refactorings of the properties, it might be better to keep these changes minimal for now.

### Details and comments
- In the future, the` from_legacy_driver_result()` method will be deprecated, but for the moment it is necessary to pass the current unit tests, so I have also refactored it.


qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -359,7 +359,9 @@ def _validate_num_orbitals(self, nelec_inactive: int, particle_number: ParticleN
 
     # TODO: can we efficiently extract this into the base class? At least the logic dealing with
     # recursion is general and we should avoid having to duplicate it.
-    def _transform_property(self, prop: Property) -> Optional[Property]:
+    def _transform_property(
+        self, prop: SecondQuantizedProperty
+    ) -> Optional[SecondQuantizedProperty]:
```
mrossinek, line 364: Is this change truely necessary? Although this will be refactored further in the near future, this is technically incorrect because e.g. `DriverMetadata` is not a `SecondQuantizedProperty`

test/second_q/drivers/test_driver.py
```diff
@@ -175,7 +175,7 @@ def test_driver_result_electronic_dipole(self):
             dipole = cast(ElectronicDipoleMoment, dipole)
 
             with self.subTest("x axis"):
-                mo_x_dip_ints = dipole.get_property("DipoleMomentX").get_electronic_integral(
+                mo_x_dip_ints = dipole._dipole_axes["DipoleMomentX"].get_electronic_integral(
```
mrossinek, line 178: Although we are now accessing a private variable in all of these tests, I am willing to let this pass since this is going to be changed again very soon.

mrossinek (commented):
Thanks, Elena! I think this is almost ready to go.
I don't think we need a reno here, since this will be refactored further very soon :+1:

qiskit_nature/second_q/transformers/active_space_transformer.py
ElePT: Yes, I agree. I cannot remember what motivated the change, so I am reverting it.

mrossinek (approved):
LGTM, thanks Elena!

mrossinek:
Just a minor comment: I don't think this PR needs a reno since the overall refactoring will come with a more detailed reno which will mostly overrule anything that would be written here for the time being anyways

qiskit_nature/second_q/drivers/psi4d/_template.txt
```diff
@@ -153,9 +153,7 @@ for idx, axis in enumerate(["x", "y", "z"]):
     if _q_has_B:
         _q_dip_mo_b = _q_dip_ao.clone()
         _q_dip_mo_b.transform(_q_hf_wavefn.Cb())
-
-    _q_dipole.add_property(
-        DipoleMoment(
+    dip = DipoleMoment(
```
woodsp-ibm, line 156:
```suggestion
    _q_dip = DipoleMoment(
```

qiskit_nature/second_q/drivers/psi4d/_template.txt
```diff
@@ -174,6 +172,5 @@ for idx, axis in enumerate(["x", "y", "z"]):
                 ),
             ],
         )
-    )
-
+    _q_dipole._dipole_axes[dip.name] = dip
```
woodsp-ibm, line 175:
```suggestion
    _q_dipole._dipole_axes[dip.name] = _q_dip
```
The rationale of the naming here - where we have the `_q_` prefix was to try and avoid any naming collision with the part of the PSI4 file that this gets merged with. I.e. the user may define variables themselves in the PSI4 config they give us and we would not want the code we add to that, which is what this template is about, colliding with that if at all possible. Hence the reason for naming of variables here.

mrossinek (approved):

woodsp-ibm (approved):

PR #757, opened 2022-07-21
refactor: Towards a non-driver future

mrossinek:
### Summary

As outlined in #701, Qiskit Nature is working towards a non-driver
focused future. This means, rather than the `BaseProblem` interface
taking in a `BaseDriver` and an optional list of `BaseTransformer`
instances, the problem itself should be a fully standalone description
of the user's problem of interest.
In other words, `BaseProblem` should be the produced result of a
`BaseDriver` and, by extension, `BaseTransformer` instances should
transform problem instances.

This commit works towards such a future by refactoring the code:
- `BaseProblem` is the returned object of `BaseDriver.run`
- `BaseTransformer.transform` has `BaseProblem` as in- and output
- `BaseProblem` takes over the role of `GroupedProperty`
  - in doing so, `Hamiltonian` objects are more clearly separated from
    `Property` objects (further refactoring to come)

### Details and comments

- [ ] update variable naming (oftentimes, a `problem` is stored in variables called `driver_result` or `grouped_property`)
- [ ] update docs

PR #766, opened 2022-08-03
PropertiesContainer

mrossinek:
### Summary

This is another refactoring preparation PR which I am pushing separately in order to simplify the review process.
This `PropertiesContainer` will be used for `BaseProblem.properties` as part of #757.
Note also, that the notion of what a `Property` is, will be simplified quite a bit as part of that later refactoring PR :+1: 

### Details and comments

This adds the `PropertiesContainer` class which is a storage object for
`Property` instances. The key feature is that it only allows a single
instance of any `Property` kind to be stored at any time.
This is sufficient for all application purposes since all auxiliary
operators generated by the `Property` instances must match in dimension
anyways.

```python
from qiskit_nature.second_q.problems import ElectronicPropertiesContainer
from qiskit_nature.second_q.properties import ParticleNumber, Property

container = ElectronicPropertiesContainer()

container.particle_number = ParticleNumber(10, 10)
print(ParticleNumber in container)  # True

container.particle_number = None
print(ParticleNumber in container)  # False

class MyCustomProperty(Property):
    ...

custom = MyCustomProperty()
container.add(custom)
print(custom in container)  # True

container.discard(MyCustomProperty)
print(custom in container)  # False
```


qiskit_nature/second_q/problems/__init__.py
```diff
@@ -25,24 +25,40 @@
    VibrationalStructureProblem
    LatticeModelProblem
 
+   PropertiesContainer
```
mrossinek, line 28: Ideally I would like to modify the documentation of this class to also include `:inherited-members` (rather than the default as to not do this), because as of right now, the docs will not show all of the methods inherited from the `MutableSet` base class:
screenshot_1659550086

Any suggestions on how to achieve this easily are highly welcome :slightly_smiling_face:

manoelmarques: > Ideally I would like to modify the documentation of this class to also include `:inherited-members` (rather than the default as to not do this), because as of right now, the docs will not show all of the methods inherited from the `MutableSet` base class: screenshot_1659550086
> 
> Any suggestions on how to achieve this easily are highly welcome 🙂

I made the `Bases: MutableSet` clickable going directly to  the official python documentation. This way this class doc. behaves like the others, without a need for an exception and one has complete information about `MutableSet` .

qiskit_nature/second_q/problems/electronic_properties_container.py
```diff
@@ -0,0 +1,105 @@
...
+    @property
+    def angular_momentum(self) -> AngularMomentum | None:
+        """Returns the :class:`qiskit_nature.second_q.properties.AngularMomentum` property."""
+        return cast(AngularMomentum, self._properties.get("AngularMomentum", None))
+
+    @angular_momentum.setter
+    def angular_momentum(self, angular_momentum: AngularMomentum | None) -> None:
+        """Sets the :class:`qiskit_nature.second_q.properties.AngularMomentum` property."""
+        if angular_momentum is None:
+            self._properties.pop("AngularMomentum", None)
```
woodsp-ibm, line 41: Wondering about the strings here. Could we use the class name itself, that would save adding extra strings as such. And seeing how all the methods here follow a similar pattern if we had an internal method to which we passed the class it could serve as the implementation for all of these, no?

docs/conf.py
```diff
@@ -26,6 +26,7 @@
 import os
 import sys
 
+sys.path.insert(0, os.path.abspath(".."))
```
woodsp-ibm, line 29: This could have been a separate PR right as overall it makes other types references as clickable links whether the type of a parameter, a base class a return type etc. I guess its ok here, though as a improvement to the docs its something that can go to stable to improve the current docs.

qiskit_nature/second_q/problems/electronic_properties_container.py
mrossinek: I ran into mypy complaints when trying to do this before. It may just need a proper overloaded variant of the `pop` method though. I will check and get back to you :+1:

docs/conf.py
mrossinek: I've extracted this out into #783 in order to allow a simple backporting of it, too :+1:

qiskit_nature/second_q/problems/electronic_properties_container.py
mrossinek: I extracted an internal `_setter` method to handle this implementation in a central location :+1:

qiskit_nature/second_q/problems/electronic_properties_container.py
```diff
@@ -0,0 +1,72 @@
...
+
+from .properties_container import PropertiesContainer
+
+
+class ElectronicPropertiesContainer(PropertiesContainer):
+    """The container class for electronic structure properties."""
+
+    @property
+    def angular_momentum(self) -> AngularMomentum | None:
+        """Returns the :class:`qiskit_nature.second_q.properties.AngularMomentum` property."""
```
woodsp-ibm, line 34: Minor comment. I am thinking that """Returns the angular momentum property.""" as plain text may suffice and be simpler/easier to read/parse - plus potentially can be translated. The above will put the fully qualified class there as a link and since the return type will be listed as AngularMomentum and will be a link anyway taking one to the exact class maybe too many links - actually at present none of them are working links since properties are not coming out in the docs at present - I noticed same with drivers yesterday where there was a non working link).

Also when the properties are built it uses that in the list of all the fields in the class, as below, which looks a little too much. I think plain text descriptions would be better as the link on the left will bring up the details of exact return type. Maybe I could go with using a ~ so its just the name not the fully qualified one, but in my mind plain textual description there would be preferable I think.

image

Same comment applies for all getter/setters like this.

qiskit_nature/second_q/problems/electronic_properties_container.py
```diff
@@ -0,0 +1,72 @@
...
+        return cast(AngularMomentum, self._properties.get("AngularMomentum", None))
```
woodsp-ibm, line 35: In this `self._properties.get("AngularMomentum", None` since elsewhere the class name is used would this be better as AngularMomentum.__name__ to avoid hard coding the string. Maybe some simple _getter  could be done like the setter to centralize/simplify things a little perhaps?

qiskit_nature/second_q/problems/electronic_properties_container.py
mrossinek: Not sure why I missed to add the `~` here, but yes, plain-text docstrings work for me, too.
Will update this tomorrow :+1:

qiskit_nature/second_q/problems/electronic_properties_container.py
mrossinek: Yeh I should have though of checking the getters for consistency, my bad :see_no_evil: 
Will update this tomorrow, too :+1:

woodsp-ibm (approved):

PR #768, opened 2022-08-05
refactor: move QEOM hopping op builders

mrossinek:
### Summary

This moves the QEOM hopping operator building utilities away from the
`BaseProblem` interface and integrates them into the `QEOM` algorithm
itself. This is a step towards the leaner `BaseProblem` interface which
we are targeting during the ongoing refactoring.

This also decouples the problem classes from a very specific algorithm.
Instead, the algorithm itself is now in charge of adding support for
other problem types.

### Details and comments

This commit also removed some left-over `_create_labels` utilities for `VibrationalOp`s,
which have been migrated to the test methods where they are still being used back in b4a46bf49408153603c3d24bfccc15ceb94aac82


qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -46,28 +54,55 @@ class QEOM(ExcitedStatesSolver):
...
             excitations: The excitations to be included in the eom pseudo-eigenvalue problem.
-                If a string ('s', 'd' or 'sd') then all excitations of the given type will be used.
-                Otherwise a list of custom excitations can directly be provided.
+
+                :`str`: which contains the types of excitations. Allowed characters are
+                    + `s` for singles
+                    + `d` for doubles
+                :`int`: a single, positive integer which denotes the number of excitations
+                    (1 == `s`, etc.)
+                :`list[int]`: a list of positive integers generalizing the above
```
Anthony-Gandon, line 76: Can this be detailed with an example like for the int case? 
My understanding is that the int type can be used to specify only excitations of degree n whereas the list can be used to specify multiple degrees.

```
Proposal:
:`int`: a single, positive integer which denotes one number of excitations (1 == `s`, 2== `d`, etc.)
:`list[int]`: a list of positive integer generalizing the above to multiple numbers of excitations ([1, 2]== `sd`, etc.)
```

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -176,14 +211,14 @@ def solve(
...
+    def _prepare_matrix_operators(self, problem: BaseProblem) -> Tuple[dict, int]:
         """Construct the excitation operators for each matrix element.
 
         Returns:
             a dictionary of all matrix elements operators and the number of excitations
             (or the size of the qEOM pseudo-eigenvalue problem)
         """
-        data = problem.hopping_qeom_ops(self._gsc.qubit_converter, self._excitations)
+        data = self._build_hopping_ops(problem)
         hopping_operators, type_of_commutativities, excitation_indices = data
```
Anthony-Gandon, line 222: I'm actually wondering why the excitation_indices is needed. In the code, I think it is only used to compute the size in the following line which is simple to circumvent or to move around.
(It is  overall a useful piece of information, I'm just questioning its role here)

Anthony-Gandon:
I get that this is just a refactoring and not a full rework but I think it could be a great addition to document somewhere the properties that the constructed excitation operators satisfy.
Especially since Edag_k = - E_k which I think is not straightforward from the naming choice.

For an example, see the `test_build_hopping_operators` test case where we find
```
"E_0": PauliSumOp.from_list([("IIXX", 1), ("IIYX", 1j), ("IIXY", -1j), ("IIYY", 1)]),
"Edag_0": PauliSumOp.from_list( [("IIXX", -1), ("IIYX", 1j), ("IIXY", -1j), ("IIYY", -1)])
```

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
mrossinek: Sure, we can do that :+1: I simply copied the docstring from the excitation generator methods and `UCC`-style ansatze. I will update this according to your suggestion in all locations :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
mrossinek: Afaict it is indeed only used to compute `size` 2 lines below. `size` is then used to determine the EOM matrices size. Maybe this can be done in a more clever way.
In fact, this goes to show that #769 (which I opened shortly after this PR) is something that would be worthwhile looking into.

mrossinek:
> Especially since Edag_k = - E_k which I think is not straightforward from the naming choice.

`dag` is short for `dagger` indicating the complex conjugate. Since all these values are real-valued, that amounts to the relation you wrote above. But yes, I agree that documentation of this can be improved. This again goes back to the issue I linked in my smaller comment above :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: Yes indeed, let's make it part of the refactoring

Anthony-Gandon:
> > Especially since Edag_k = - E_k which I think is not straightforward from the naming choice.
> 
> `dag` is short for `dagger` indicating the complex conjugate. Since all these values are real-valued, that amounts to the relation you wrote above. But yes, I agree that documentation of this can be improved. This again goes back to the issue I linked in my smaller comment above 👍

It was my mistake, I meant to write Edag_k = - E_k.adjoint(). This can also be part of the refactoring

woodsp-ibm (approved):

PR #770, opened 2022-08-06
refactor: use TemporaryDirectory where possible

mrossinek:
### Summary

Instead of using a NamedTemporaryFile and juggle its existence state,
this commit refactors all scenarios to use a TemporaryDirectory within
which the file names can be controlled more easily.
The reason for this refactoring was to guard against possible runtime
conflicts when multiple processes might be attempting to access the same
temporary file name, after the object has been unlinked (which might
free up the temporary name for reuse).

### Details and comments

Thanks @woodsp-ibm for this suggestion :+1:

### DIFF
--- a/test/test_logging.py
+++ b/test/test_logging.py
@@ -13,14 +13,14 @@
 """
 Test Nature logging methods
 """
+from pathlib import Path
+from tempfile import TemporaryDirectory
 from typing import List, Dict, Set
 import sys
 import unittest
 import logging
-import tempfile
 import contextlib
 import io
-import os
 from test import QiskitNatureTestCase
 from qiskit_nature import logging as nature_logging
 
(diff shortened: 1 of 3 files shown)


test/test_logging.py
```diff
@@ -110,22 +110,19 @@ def test_logging_to_default_handler(self):
     def test_logging_to_file(self):
```
mrossinek, line 110: I don't quite understand why the changes to this method make the test which gets executed after this one (namely `test_logging_to_handler`) fail..

@manoelmarques could you please take a look at this? I don't seem to find the reason for this cross-test dependence

manoelmarques: Fixed the unit test method

manoelmarques (approved):

PR #772, opened 2022-08-06
Refactor `BaseProblem.second_q_ops()`

mrossinek:
### Summary

This PR is another self-container step part of #701.
It refactors the `BaseProblem.second_q_ops()` method to produce a tuple, with the first element being the main operator, and the second element a dictionary of auxiliary operators. This matches the input expected by Terra algorithms and avoids the common need of extracting the main operator from one larger dictionary of operators.

### Details and comments

This PR also enables `settings.dict_aux_operators` by default and deprecates `ListOrDict` since Qiskit Nature will always be producing dictionaries of auxiliary operators in the new stack/workflow.


qiskit_nature/list_or_dict.py
```diff
@@ -34,6 +35,15 @@ def __init__(self, values: Optional[ListOrDictType] = None):
         Args:
             values: an optional object of `list` or `dict` type.
         """
+        warn_deprecated(
+            "0.5.0",
+            old_type=DeprecatedType.CLASS,
+            old_name="qiskit_nature.ListOrDict",
+            additional_msg=(
+                ". Qiskit Nature will now always produce dictionary-style auxiliary operators."
+            ),
```
woodsp-ibm, line 44: How does this relate to the now deprecated code. Presumably that still uses the settings flag and will produce a list or dict based on the settings. By default it will now produce a dict, but presumably a user could change the settings back to have lists. Its ignored in the new code of course so the above message seems to be true for that. But then the type is not used either right.

mrossinek: The new code ignores this setting now.
The deprecated code now also produces dict by default, but the user can change it back to list.

qiskit_nature/second_q/algorithms/excited_states_solvers/excited_states_eigensolver.py
```diff
@@ -60,25 +61,12 @@ def solver(self, solver: Union[Eigensolver, EigensolverFactory]) -> None:
     def get_qubit_operators(
         self,
         problem: BaseProblem,
-        aux_operators: Optional[ListOrDictType[Union[SecondQuantizedOp, PauliSumOp]]] = None,
-    ) -> Tuple[PauliSumOp, Optional[ListOrDictType[PauliSumOp]]]:
+        aux_operators: Optional[dict[str, Union[SecondQuantizedOp, PauliSumOp]]] = None,
+    ) -> Tuple[PauliSumOp, Optional[dict[str, PauliSumOp]]]:
```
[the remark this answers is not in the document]
mrossinek: I guess one could define a common TypeVar like `DictStr` or whatever, but realistically this does not aid in readability and in the docs, users will need to navigate to the common type alias (unless its apparent by the name like in this example) to find the real type.
I don't think there is much we can do here, but when revisiting this to produce `SparsePauliOp`s directly and when going over the docs in the future, we can definitely clean up our type hints, yes :+1:

woodsp-ibm (dismissed):

woodsp-ibm (approved):

PR #773, opened 2022-08-08
Implement QCSchema.from_legacy_hdf5

mrossinek:
### Summary

This implements a new class-constructor for the QCSchema which can parse
a legacy (`qiskit_nature.propreties.second_quantization`)-based HDF5
file to re-construct as much of a QCSchema object as possible.

This will serve as a migration path for users with legacy HDF5 files.

### Details and comments

I am leaving this PR relatively small on purpose to ease reviewing.
Merging this, will allow us to do the next refactoring step, which is the intertwined refactoring of the `BaseProblem` interface.
This is a requirement for that step, because we need to be able to use our legacy HDF5 files during the unittests until the QCSchema-HDF5 representation has stabilized.

I am not adding a reno here on purpose, since the migration path (using these changes) can be outlined better in a single bigger release note at the end of the refactoring.


qiskit_nature/second_q/formats/qcschema/qc_schema.py
```diff
@@ -127,3 +132,197 @@ def _from_hdf5_group(cls, h5py_group: h5py.Group) -> QCSchemaInput:
         data["keywords"] = dict(h5py_group["keywords"].attrs.items())
 
         return cls(**data)
+
+    @classmethod
+    def from_legacy_hdf5(cls, legacy_hdf5: str | Path) -> QCSchema:
+        # pylint: disable=line-too-long
+        """Constructs a schema object from a legacy HDF5 object.
+
+        Warning: This method will **not** produce a fully valid QCSchema object!
```
woodsp-ibm, line 141: Maybe use `.. note::` or even `.. warning::`` to have sphinx highlight this block.

You say it will **not** produce a fully valid object. Are you able to put any detail/examples here e.g list the required fields that its missing - is that what this data is, or is it other data such as scf fields where maybe you have to supply them all - although that's an assumption since say for scf it just calls it a list of fields. I see the BasisSet has some compensation - maybe that's it (or at least part of it) - a link to that class from here, where it has more explanation might be helpful

mrossinek: Yep, I can outline this with a bit more details. The basis set is the only that will be missing for sure, but other data may or may not be missing depending on the availability of information stored in the original HDF5 file.
I will try to improve the warning to be more detailed in this regard :+1:

woodsp-ibm (approved):

PR #774, opened 2022-08-09
Restore python default behavior for Nature deprecations

manoelmarques:
### Summary
Do not force output of Nature deprecations  and leave default python behavior of showing only on main: https://peps.python.org/pep-0565/
Disable Nature deprecations in unit tests.
Disable additional unit test deprecations found, like the ones for Aer.

### DIFF
--- a/qiskit_nature/settings.py
+++ b/qiskit_nature/settings.py
@@ -43,6 +43,7 @@ class QiskitNatureSettings:
                 ),
                 stacklevel=3,
             )
+            warnings.filterwarnings("ignore", category=ListAuxOpsDeprecationWarning)
             self._deprecation_shown = True
 
         return self._dict_aux_operators
@@ -61,6 +62,7 @@ class QiskitNatureSettings:
                 ),
                 stacklevel=3,
             )
+            warnings.filterwarnings("ignore", category=ListAuxOpsDeprecationWarning)
             self._deprecation_shown = True
 
         self._dict_aux_operators = dict_aux_operators
--- a/test/algorithms/ground_state_solvers/test_adapt_vqe.py
+++ b/test/algorithms/ground_state_solvers/test_adapt_vqe.py
@@ -15,6 +15,7 @@ import contextlib
 import copy
 import io
 import unittest
+import warnings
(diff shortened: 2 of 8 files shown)


qiskit_nature/settings.py
```diff
@@ -43,6 +43,7 @@ def dict_aux_operators(self) -> bool:
                 ),
                 stacklevel=3,
             )
+            warnings.filterwarnings("ignore", category=ListAuxOpsDeprecationWarning)
             self._deprecation_shown = True
```
mrossinek, line 47: Adding this filter should alleviate the need for `_deprecation_shown` in its entirety, right?

test/algorithms/ground_state_solvers/test_adapt_vqe.py
```diff
@@ -108,7 +109,9 @@ def test_delta(self):
             quantum_instance=QuantumInstance(BasicAer.get_backend("statevector_simulator"))
         )
         delta1 = 0.01
-        calc = AdaptVQE(self.qubit_converter, solver, delta=delta1)
+        with warnings.catch_warnings(record=True):
+            warnings.simplefilter("always")
```
mrossinek, line 113: Why is speciifcally this one needed?

manoelmarques: Because it was showing a deprecation warning from AdaptVQE, unnecessary in this unit test. I prefer doing this than adding another filter.

qiskit_nature/settings.py
manoelmarques: No, because it will eventually go inside this if statement again if the flag is not set and it will set the filter to default and output it again.

test/algorithms/ground_state_solvers/test_adapt_vqe.py
mrossinek: Ah so you just found a warning unrelated to this PR which you are now filtering. Got it, thanks :+1:

mrossinek (approved):
Thanks also for this :+1:

PR #775, opened 2022-08-09
Deprecate protein folding

manoelmarques:
### Summary
Waiting for qiskit-research/qiskit-research#22

### DIFF
--- a/qiskit_nature/problems/sampling/protein_folding/__init__.py
+++ b/qiskit_nature/problems/sampling/protein_folding/__init__.py
@@ -124,6 +124,20 @@ from .peptide.chains.main_chain import MainChain
 from .peptide.chains.side_chain import SideChain
 from .peptide.peptide import Peptide
 from .protein_folding_problem import ProteinFoldingProblem
+from ....deprecation import warn_deprecated, DeprecatedType, NatureDeprecationWarning
+
+warn_deprecated(
+    "0.5.0",
+    old_type=DeprecatedType.PACKAGE,
+    old_name="qiskit_nature.problems.sampling.protein_folding",
+    additional_msg=(
+        ". Please refer to repository "
+        "'https://github.com/qiskit-research/qiskit-research'"
+        " for additional information"
+    ),
+    stack_level=3,
+    category=NatureDeprecationWarning,
+)
 
 __all__ = [
     "ProteinFoldingProblem",
--- a/qiskit_nature/problems/sampling/protein_folding/protein_folding_problem.py
+++ b/qiskit_nature/problems/sampling/protein_folding/protein_folding_problem.py
@@ -51,6 +51,7 @@ class ProteinFoldingProblem(SamplingProblem):
             penalty_parameters: Parameters that define the strength of constraints enforcing in
                                 the problem.
         """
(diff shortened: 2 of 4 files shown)


mrossinek (commented):
This LGTM :+1: 
Shall we wait for the corresponding PR to be merged? https://github.com/qiskit-research/qiskit-research/pull/22

qiskit_nature/problems/sampling/protein_folding/__init__.py
```diff
@@ -124,6 +124,20 @@
 from .peptide.chains.side_chain import SideChain
 from .peptide.peptide import Peptide
 from .protein_folding_problem import ProteinFoldingProblem
+from ....deprecation import warn_deprecated, DeprecatedType, NatureDeprecationWarning
+
+warn_deprecated(
+    "0.5.0",
+    old_type=DeprecatedType.PACKAGE,
+    old_name="qiskit_nature.problems.sampling.protein_folding",
```
dlasecki, line 132: Do we want to deprecate protein folding only or the full sampling package?

mrossinek: I was wondering the same. Technically, only protein folding is being moved to that location in Qiskit Research. Afterwards, `sampling` is empty, so I am not sure whether we still need to deprecate this specific (now empty) module explicitly..

@woodsp-ibm any thoughts? Once we have a migration guide which outlines all the code being moved to `second_q`, this could be mentioned there, too

dlasecki: In principle, someone could have been implementing their own algorithm based on the `SamplingProblem` interface, in their fork (which is outside `protein_folding` but inside `sampling`. They might sync with our `main` from time to time. So it would be reasonable to show deprecation warnings regarding `sampling` too, I think (but the impact is minor). 

I'm assuming that the whole `sampling` package will disappear from Nature, as we are narrowing the scope.

mrossinek: Right, that's a good point! @manoelmarques could you please also add a separate deprecation warning to the `sampling` module (or the `SamplingProblem` if it cannot go on the module itself) that simply states, this is being removed in order to focus the scope of Qiskit Nature?

woodsp-ibm: Seems reasonable that the whole `sampling` module is deprecated and eventually removed from here - with the code going elsewhere and being maintained there. `sampling` is not under `second_q` so does not remain here at all. Had someone used SamplingProblem as a base then they will need to  change code to use the version in the new location. The migration guide from 0.4.0 to 0.5.0 will need to include sampling/Protein folding, where it is now, and to migrate over to it etc.

mrossinek (approved):
LGTM :+1:

woodsp-ibm:
I think the new location for the code should be in whatever release note we end up with right. Do we want to do a release note here noting the deprecation and new location of the protein folding. Its kinda separate from the other code refactoring in that maybe just do this now so as to not have to remember later?

woodsp-ibm (approved):

PR #789, opened 2022-08-16
Fix #788

### LINKED ISSUE #788: UCC: `generalized=True` has no effect when `preserve_spin=False`
I am using the generalized UCCD ansatz and I noticed that I got the same result as without the generalized option.
I am sharing a minimal example here, with explanations below.

- **Qiskit version**: 0.21.1
- **Qiskit_nature version**: 0.4.4
- **Python version**: 3.9

As far as I can see, the operators generated by the GUCCD ansatz and the UCCD ansatz are the same. They both depends on the number of particles, based on the initial Hartree-Fock state, while GUCCD should ignore those informations.

```
import qiskit, qiskit_nature
from qiskit import QuantumCircuit
from qiskit_nature.circuit.library import UCC

print(qiskit.__version__)
print(qiskit_nature.__version__)
# System:
n_spinorbs     = 6
n_alpha        = 2
n_beta         = 2
n_particles    = (n_alpha,n_beta)

jw_mapper = JordanWignerMapper()
jw_converter = QubitConverter(jw_mapper)
circuit = UCC(qubit_converter=jw_converter, num_particles=n_particles, num_spin_orbitals=n_spinorbs, excitations='d', alpha_spin=True, beta_spin=True, max_spin_excitation=None, generalized=False, preserve_spin=False, reps=1, initial_state=None)
n_param = circuit.num_parameters
fermionic_operators = circuit.excitation_ops()

print(n_param)
for op in fermionic_operators:
  print(op.to_list(display_format='sparse'))

circuit = UCC(qubit_converter=jw_converter, num_particles=n_particles, num_spin_orbitals=n_spinorbs, excitations='d', alpha_spin=True, beta_spin=True, max_spin_excitation=None, generalized=True, preserve_spin=False, reps=1, initial_state=None)
n_param = circuit.num_parameters
fermionic_operators = circuit.excitation_ops()

print(n_param)
for op in fermionic_operators:
  print(op.to_list(display_format='sparse'))
```

This results in the following:
```
0.21.1
0.4.4
6
((0, 1), (2, 5))
((0, 3), (2, 5))
((0, 4), (2, 5))
((1, 3), (2, 5))
((1, 4), (2, 5))
((3, 4), (2, 5))
6
((0, 1), (2, 5))
((0, 3), (2, 5))
((0, 4), (2, 5))
((1, 3), (2, 5))
((1, 4), (2, 5))
((3, 4), (2, 5))
```

As one can see, the excitations only goes from spin-orbitals 0,1 and 3,4 to 2 and 5. According to the labelling in qiskit, this means that we have the original state written as follow: | 0_up 1_up 2_up 3_down 4_down 5_down >
where 0_up 1_up 3_down 4_down are occupied, and 2_up and 5_down are empty. Which is why we only see those excitations in UCCD. But in GUCCD it should not be the case, we should have excitations/deexcitations on other spin-orbitals, regardless of their occupation.

Is it possible that this option is ignored ?

mrossinek:
### Summary

This commit properly ensures the compatibility of the generalized and
preserve_spin keyword arguments available in the fermionic excitation
generator methods.
A previous attempt to fix this was done in #747, but some edge cases
were overlooked.

### Details and comments

This commit also includes the same patch as a bugfix in the legacy code
location.

Fixes #788


woodsp-ibm:
If we want this backported, we need to be careful since this is also fixing the newly moved/refactored logic in second_q too. Maybe split the PR into 2 with the original code and release note in one, which get backported, and the fix for the moved code in the other? Otherwise it would need doing on stable more manually to limit it to the right set of files.

mrossinek:
I think in that case I will just manually backport it. That is a total of 2 PRs instead of 3 and it makes this PR fix all places at once :+1:

woodsp-ibm (approved):

PR #790, opened 2022-08-16
Updated Example to newest imports and formats

julenl:
### Summary
The example seems to be written for an older version of qiskit, where most of the imports have a different syntax.
Also, `second_q_ops()` produces a dict, so the *convert* function requires the operator to be extracted from it.

### Details and comments
The example can be easily copy/pasted into a python shell with the newest *qiskit_nature* module installed (0.4.4) to verify that now it does work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,9 @@ the ground-state (minimum) energy of a molecule.
 
 ```python
 from qiskit_nature.settings import settings
-from qiskit_nature.second_q.drivers import UnitsType
-from qiskit_nature.second_q.drivers import PySCFDriver
-from qiskit_nature.second_q.problems import ElectronicStructureProblem
+from qiskit_nature.drivers import UnitsType
+from qiskit_nature.drivers.second_quantization import PySCFDriver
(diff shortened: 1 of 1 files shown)


woodsp-ibm:
The ReadMe here on main branch really is the latest version - in that it was altered to work with the code as exists now on main. The code is being refactored/reorganized and the former way, though it would still work, now emits deprecation warnings . The readme on the stable branch here https://github.com/Qiskit/qiskit-nature/tree/stable/0.4 is compatible with the latest release of Nature, i.e. 0.4.4 and would be what you expected to see I guess, It is also what is published with the project description on PyPi https://pypi.org/project/qiskit-nature/  So this readme on main is compatible with the upcoming changes for the next release of 0.5 which unless you install source of main from the repo here it would not work for you.

mrossinek:
Like Steve has outlined, the README is actually correct on both branches. Unfortunately this is a limitation of Github which prevents us from showing the stable README to non-developer users while also ensuring that PRs go against the development branch (`main`).

Thank you, nonetheless, for taking the time to open this PR! I would be happy to include other contributions from you in the future :slightly_smiling_face:

PR #796, opened 2022-08-18
Towards a non-driver future

mrossinek:
### Summary

This is the major PR which untangles the intertwined interfaces of `BaseProblem`, `BaseDriver` and `BaseTransformer`.
There is no clean way to do this unfortunately, so this PR is rather massive...

### Details and comments

I used this opportunity to also do some reshuffling to place the classes into their intended location. Most notably the population of the `qiskit_nature.second_q.hamiltonians` module.
While this does add a bit of complexity to the PR, this is not a significant addition of the overall complexity and it allowed me to not have to introduce some dummy classes to ensure the API of the new `BaseProblem` looks correct.

I do _not_ recommend a commit-wise review of this PR due to its convoluted nature. The commits are not atomic and do not represent fully working states at each step of the way. Churning through everything as a bulk PR is advisable, albeit cumbersome.

The following is a non-exhaustive list of todos, before this PR can be merged:
- [x] iterate all changed files to verify changes
- [x] write/update documentation
- [x] add an exhaustive release note

From a technical perspective, this PR should already be ready to review, though.

> :warning: Note, that this PR currently removes more things than the should be the case in the end. Most notably, more drivers are being removed. These should be re-added after this PR, following the examples as done here for PySCF and Gaussian.


qiskit_nature/second_q/formats/qcschema_translator.py
```diff
@@ -0,0 +1,169 @@
...
+    ElectronicIntegrals,
+    OneBodyElectronicIntegrals,
+    TwoBodyElectronicIntegrals,
+)
+
+from .molecule_info import MoleculeInfo
+from .qcschema import QCSchema
+
+
+# TODO: make use of basis argument, pending further ElectronicStructureProblem refactoring
```
mrossinek, line 39: As indicated by
> pending further ElectronicStructureProblem refactoring

this will be tackled in another refactoring PR after this one. The further refactoring requires the integration of the `PolynomialTensor` (#756) and `SparseLabelOp` (#731) classes

qiskit_nature/second_q/hamiltonians/electronic_energy.py
```diff
@@ -60,7 +61,10 @@ def __init__(
             nuclear_repulsion_energy: the optional nuclear repulsion energy.
             reference_energy: an optional reference energy (such as the HF energy).
         """
-        super().__init__(self.__class__.__name__, electronic_integrals, shift=energy_shift)
+        self._electronic_integrals: dict[ElectronicBasis, dict[int, ElectronicIntegrals]] = {}
+        for integral in electronic_integrals:
+            self.add_electronic_integral(integral)
+        self._shift = energy_shift or {}
         self._nuclear_repulsion_energy = nuclear_repulsion_energy
         self._reference_energy = reference_energy
```
mrossinek, line 69: This will also be further changed in the upcoming refactoring, at which point only a single-basis `PolynomialTensor` will be stored internally here :+1:

qiskit_nature/second_q/problems/electronic_structure_problem.py
```diff
@@ -77,54 +76,44 @@ def filter_criterion_custom(eigenstate, eigenvalue, aux_values):
...
+            TypeError: TODO.
         """
-        super().__init__(driver, transformers, "ElectronicEnergy")
+        if not isinstance(hamiltonian, ElectronicEnergy):
+            raise TypeError("TODO.")
+        super().__init__(hamiltonian)
+        self.properties: ElectronicPropertiesContainer = ElectronicPropertiesContainer()
+        self.molecule: MoleculeInfo = None
+        self.basis_transform: ElectronicBasisTransform = None
+        # TODO: further refactoring:
```
mrossinek, line 94: This will be part of the upcoming refactoring PR, which depends on the aforementioned integration of the `PolynomialTensor`

qiskit_nature/second_q/properties/driver_metadata.py
```diff
@@ -1,80 +0,0 @@
-# This code is part of Qiskit.
```
mrossinek, line 1: The data stored in this object are now handled by the `QCSchema`. Since they are of no further benefit to the calculation down the road, they are not propagated further into the stack, until a valid and well-motivated reason for doing so comes up.

qiskit_nature/second_q/properties/occupied_modals.py
```diff
@@ -41,7 +41,23 @@ def __init__(
                 through which to map the integrals into second quantization. This attribute **MUST**
                 be set before the second-quantized operator can be constructed.
         """
-        super().__init__(self.__class__.__name__, basis)
+        super().__init__(self.__class__.__name__)
+        self._basis: VibrationalBasis = basis
+
+    @property
+    def basis(self) -> VibrationalBasis:
```
mrossinek, line 48: This will also get refactored further, once the `basis` information can be stored on the `VibrationalStructureProblem` directly (requires `PolyomialTensor`, just like before in the electronic case)

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -359,7 +358,9 @@ def _validate_num_orbitals(self, nelec_inactive: int, particle_number: ParticleN
 
     # TODO: can we efficiently extract this into the base class? At least the logic dealing with
     # recursion is general and we should avoid having to duplicate it.
-    def _transform_property(self, prop: Property) -> Optional[Property]:
+    def _transform_property(
```
mrossinek, line 361: This will get refactored further, once the `ElectronicStructureProblem` has reached its final form

test/second_q/algorithms/initial_points/test_hf_initial_point.py
```diff
@@ -22,14 +22,11 @@
...
-    GroupedSecondQuantizedProperty,
-)
-from qiskit_nature.second_q.properties.electronic_energy import (
-    ElectronicEnergy,
-)
+from qiskit_nature.second_q.problems import ElectronicStructureProblem
+from qiskit_nature.second_q.hamiltonians import ElectronicEnergy
 
 
+@unittest.skip("problem properties")
```
mrossinek, line 29: Note to myself: open task of this PR

test/second_q/algorithms/initial_points/test_mp2_initial_point.py
```diff
@@ -27,17 +27,14 @@
...
-    GroupedSecondQuantizedProperty,
-)
+from qiskit_nature.second_q.hamiltonians import ElectronicEnergy
+from qiskit_nature.second_q.properties import ParticleNumber
 from qiskit_nature.second_q.properties.integrals import ElectronicIntegrals
 from qiskit_nature.second_q.mappers import QubitConverter, JordanWignerMapper
 from qiskit_nature.second_q.algorithms.initial_points import MP2InitialPoint
 
 
+@unittest.skip("problem properties")
```
mrossinek, line 37: Same here

test/second_q/drivers/gaussiand/test_driver_gaussian_log.py
```diff
@@ -150,6 +150,7 @@ def test_quartic_force_constants(self):
         ]
         self.assertListEqual(qfc, expected)
 
+    @unittest.skip("HDF5 migration")
```
mrossinek, line 153: This needs to be figured out once we have a format for the `WatsonHamiltonian` in the vibrational stack in place. See also #804

mrossinek (commented):
Just leaving some general comments here, which hopefully also clarify a few aspects for other reviewers :+1:

mrossinek:
I've added a working release note to this PR. It is by no means finalized and more work will need to be done on it as this refactoring progresses, but I think that a lot of it can only be finalized properly, once the refactoring of the problem classes has progressed further (see also my previous comments)

qiskit_nature/second_q/hamiltonians/electronic_energy.py
```diff
@@ -60,7 +61,10 @@ def __init__(
             nuclear_repulsion_energy: the optional nuclear repulsion energy.
             reference_energy: an optional reference energy (such as the HF energy).
         """
-        super().__init__(self.__class__.__name__, electronic_integrals, shift=energy_shift)
+        self._electronic_integrals: dict[ElectronicBasis, dict[int, ElectronicIntegrals]] = {}
+        for integral in electronic_integrals:
+            self.add_electronic_integral(integral)
+        self._shift = energy_shift or {}
```
woodsp-ibm, line 67: Should we allow getter access to these shift values if users want to read them more directly? We expose these after interpret into the final result but we could provide them similarly via a getter here too. Is there a reason we have setters on these - do we ever want to change these after construction where I see they all default to None.

qiskit_nature/second_q/hamiltonians/electronic_energy.py
```diff
@@ -261,6 +245,32 @@ def integral_operator(self, density: OneBodyElectronicIntegrals) -> OneBodyElect
...
 
+    @property
+    def register_length(self) -> int:
+        ints = None
+        if ElectronicBasis.SO in self._electronic_integrals:
+            ints = self._electronic_integrals[ElectronicBasis.SO]
+        elif ElectronicBasis.MO in self._electronic_integrals:
+            ints = self._electronic_integrals[ElectronicBasis.MO]
+
+        return len(list(ints.values())[0].to_spin())
```
woodsp-ibm, line 256: This seems like it might be an expensive conversion just for a number. Could the integrals themselves report this back? For MO it seems that it would just be twice its dim for example which seems lightweight compared to what I think `to_spin` is doing

qiskit_nature/second_q/hamiltonians/ising_model.py
```diff
@@ -0,0 +1,58 @@
...
+import logging
+from fractions import Fraction
+
+import numpy as np
+
+from qiskit_nature.second_q.operators import SpinOp
+
+from .lattice_model import LatticeModel
+
+logger = logging.getLogger(__name__)
```
woodsp-ibm, line 23: Unless we use the logger, which I do not see done here in this module, we usually avoid setting up one which is unused.

qiskit_nature/second_q/hamiltonians/lattice_model.py
```diff
@@ -0,0 +1,42 @@
...
+        """
+        Args:
+            lattice: Lattice on which the model is defined.
+        """
+        self._lattice = lattice
+
+    @property
+    def lattice(self) -> Lattice:
+        """Return a copy of the input lattice."""
+        return self._lattice.copy()
```
woodsp-ibm, line 34: Is there a reason to return a copy (its actually an object deepcopy looking at the implementation), rather than by reference which would allow the lattice to be updated later in the model? I see this is what was done before though.

qiskit_nature/second_q/problems/electronic_structure_problem.py
```diff
@@ -77,54 +76,47 @@ def filter_criterion_custom(eigenstate, eigenvalue, aux_values):
...
-        transformers: Optional[List[BaseTransformer]] = None,
-    ):
+    def __init__(self, hamiltonian: Hamiltonian) -> None:
         """
-
         Args:
-            driver: A fermionic driver encoding the molecule information.
-            transformers: A list of transformations to be applied to the driver result.
+            hamiltonian: the Hamiltonian of this problem. This needs to be an
+                :class:`.ElectronicEnergy`.
```
woodsp-ibm, line 83: this needs to be of type ElectronicEnergy? The error msg is more general right, and the instance check, in allowing sub-classes of this so it does not have to be an instance of exact class.

Does having the type ElectronicEnergy in the constructor cause an issue? That is a Hamiltonian as per the base class though it is narrowing the type. Same would apply to the other problems here. It would be helpful if it was the narrowed type since hopefully things can be detected earlier rather than some type check thrown at run time

qiskit_nature/second_q/problems/electronic_structure_problem.py
```diff
@@ -77,54 +76,47 @@ def filter_criterion_custom(eigenstate, eigenvalue, aux_values):
...
-        super().__init__(driver, transformers, "ElectronicEnergy")
+        if not isinstance(hamiltonian, ElectronicEnergy):
+            raise TypeError(
+                "Only an ElectronicEnergy hamiltonian is supported by the ElectronicStructureProblem,"
+                f" not one of type, {type(hamiltonian)}"
+            )
+        super().__init__(hamiltonian)
+        self.properties: ElectronicPropertiesContainer = ElectronicPropertiesContainer()
+        self.molecule: MoleculeInfo = None
+        self.basis_transform: ElectronicBasisTransform = None
```
woodsp-ibm, line 96: If we are going over to public instance vars instead of setter/getters thats fine. But they need to come out in docs and it can be done with Attrributes e.g see the new evolvers/PVQD that was added to Terra algorithms.

qiskit_nature/second_q/problems/lattice_model_problem.py
```diff
@@ -14,40 +14,40 @@
...
 from .lattice_model_result import LatticeModelResult
+from .lattice_properties_container import LatticePropertiesContainer
 from .eigenstate_result import EigenstateResult
 
 
 class LatticeModelProblem(BaseProblem):
     """Lattice Model Problem class to create second quantized operators from a lattice model."""
 
-    def __init__(self, lattice_model: LatticeModel) -> None:
+    def __init__(self, hamiltonian: Hamiltonian) -> None:
```
woodsp-ibm, line 31: Similar comment to electronic problem in terms of type

qiskit_nature/second_q/properties/bases/electronic_basis_transform.py
```diff
@@ -65,7 +62,7 @@ def is_alpha_equal_beta(self) -> bool:
         return np.allclose(self.coeff_alpha, self.coeff_beta)
 
     def __str__(self) -> str:
-        string = [super().__str__() + ":"]
+        string = ["ElectronicBasisTransform:"]
```
woodsp-ibm, line 65: Should we prefer to use self.name here (since thats set to the class name) rather than hard code this?

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -178,9 +174,7 @@ def _check_configuration(self):
                 str(self._num_electrons),
             )
 
-    def transform(
-        self, grouped_property: GroupedSecondQuantizedProperty
-    ) -> GroupedElectronicProperty:
+    def transform(self, grouped_property: BaseProblem) -> BaseProblem:
         """Reduces the given `GroupedElectronicProperty` to a given active space.
```
woodsp-ibm, line 178:
```suggestion
        """Reduces the given problem to a given active space.
```
Can we not make this an ElectronicStructureProblem typehint here that we can potentially detect any mistmatch earlier etc. Similar to a prior comment

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -196,27 +190,25 @@ def transform(
                                number of selected active orbital indices does not match
                                `num_molecular_orbitals`.
         """
-        if not isinstance(grouped_property, GroupedElectronicProperty):
+        if not isinstance(grouped_property, ElectronicStructureProblem):
             raise QiskitNatureError(
                 "Only `GroupedElectronicProperty` objects can be transformed by this Transformer, "
                 f"not objects of type, {type(grouped_property)}."
```
woodsp-ibm, line 196: Needs updating the error msg

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -196,27 +190,25 @@ def transform(
...
             )
 
-        particle_number = grouped_property.get_property(ParticleNumber)
+        particle_number = grouped_property.properties.particle_number
         if particle_number is None:
             raise QiskitNatureError(
                 "The provided `GroupedElectronicProperty` does not contain a `ParticleNumber` "
                 "property, which is required by this transformer!"
```
woodsp-ibm, line 203: Same here and below too

qiskit_nature/second_q/transformers/base_transformer.py
```diff
@@ -24,9 +24,7 @@ class BaseTransformer(ABC):
     """
 
     @abstractmethod
-    def transform(
-        self, grouped_property: GroupedSecondQuantizedProperty
-    ) -> GroupedSecondQuantizedProperty:
+    def transform(self, grouped_property: BaseProblem) -> BaseProblem:
         """Transforms one :class:`~qiskit_nature.properties.GroupedProperty` into another one.
```
woodsp-ibm, line 28: Comment needs updating

docs/apidocs/qiskit_nature.second_q.hamiltonians.lattices.rst
```diff
@@ -0,0 +1,19 @@
+﻿
+
+lattices
+=================================
+
+.. automodule:: qiskit_nature.second_q.hamiltonians.lattices
+   :no-members:
+   :no-inherited-members:
+   :no-special-members:
+
```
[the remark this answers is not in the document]
mrossinek: I think I just used the auto-generated file again here... the same is the case in some of the other RST files.
Could we leave all of these as they are for now, and revisit all of the RST files when we figure out how the docs render?

qiskit_nature/second_q/drivers/gaussiand/gaussian_forces_driver.py
```diff
@@ -83,7 +81,7 @@ def __init__(
     @staticmethod
     @_optionals.HAS_GAUSSIAN.require_in_call
     def from_molecule(
```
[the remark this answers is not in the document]
mrossinek: I'm open to whether or not we want to keep these. I don't see a downside in having them.
What do you suggest we do?

qiskit_nature/second_q/drivers/pyscfd/pyscfdriver.py
```diff
@@ -588,110 +521,125 @@ def _populate_driver_result_basis_transform(
...
-        )
-        electronic_energy.overlap = OneBodyElectronicIntegrals(
-            ElectronicBasis.AO,
-            (self._calc.get_ovlp(), None),
-        )
-
-        orbs_energy, orbs_energy_b = self._extract_mo_data("mo_energy")
-        orbital_energies = (
-            (orbs_energy, orbs_energy_b) if orbs_energy_b is not None else orbs_energy
+        return self._to_qcschema(
```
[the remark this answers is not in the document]
mrossinek: At that point that seems more like a duplication of the `QCSchema`, does it not?

qiskit_nature/second_q/formats/molecule_info.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@dataclass
+class MoleculeInfo:
+    """A dataclass storing molecule information."""
+
+    symbols: Sequence[str]
+    """The ordered sequence of atoms which make up this molecule."""
+
+    coords: np.ndarray
```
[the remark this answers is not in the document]
mrossinek: Hm yes I see :+1: 
Would an `np.ndarray` fulfill that `Sequence` type hint? Not really because it does not store `tuple`s, right?

qiskit_nature/second_q/formats/molecule_info.py
```diff
@@ -0,0 +1,59 @@
...
+    """The XYZ coordinates of the atoms represented as a matrix."""
+
+    multiplicity: int = 1
+    """The multiplicity of the molecule (`= 2 * spin + 1`)."""
+
+    charge: int = 0
+    """The total charge of the molecule."""
+
+    units: DistanceUnit = DistanceUnit.BOHR
+    """The distance unit in which the XYZ coordinates are stored."""
```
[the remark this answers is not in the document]
mrossinek: I default this to Bohr since that's what the `QCSchema` dictates. It's also technically the atomic unit which is what we use throughout the code, too.

I'm open to changing it back, though. Or we make the drivers default to Bohr, too...

qiskit_nature/second_q/formats/qcschema_translator.py
```diff
@@ -0,0 +1,187 @@
...
+        nuclear_repulsion_energy=e_nuc,
+        reference_energy=e_ref,
+    )
+    if qcschema.wavefunction.scf_eigenvalues_a is not None:
+        hamiltonian.orbital_energies = np.asarray(qcschema.wavefunction.scf_eigenvalues_a)
+
+    natm = len(qcschema.molecule.symbols)
+    molecule = MoleculeInfo(
+        symbols=qcschema.molecule.symbols,
+        coords=np.asarray(qcschema.molecule.geometry).reshape((natm, 3)),
```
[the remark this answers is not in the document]
mrossinek: We stored the coords as numpy array in the past, so that's what I went with here again, yes

I'm happy to make it more general if you want

qiskit_nature/second_q/formats/qcschema_translator.py
```diff
@@ -0,0 +1,187 @@
...
+        multiplicity=qcschema.molecule.molecular_multiplicity or 1,
```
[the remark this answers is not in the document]
mrossinek: `molecular_multiplicity` is an optional keyword and those defaults to `None`. A multiplicity of 0 makes no sense as that would require negative total spin which does not exist

qiskit_nature/second_q/formats/qcschema_translator.py
```diff
@@ -0,0 +1,187 @@
...
+        charge=qcschema.molecule.molecular_charge or 0,
+        units=DistanceUnit.BOHR,
```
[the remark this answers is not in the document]
mrossinek: I commented on this before. Let's have a discussion offline around this :+1:

qiskit_nature/second_q/hamiltonians/electronic_energy.py
mrossinek: The `_shift` will get refactored further once the `PolynomialTensor` is merged. Hence, I left it untouched for now :+1:

qiskit_nature/second_q/hamiltonians/electronic_energy.py
mrossinek: This will also get refactored further once the `PolynomialTensor` is merged. Let's leave it as is for now :+1:

qiskit_nature/second_q/hamiltonians/lattice_model.py
mrossinek: I have not given that much thought tbh. I guess providing write-access would allow flexibility although the user would need to be careful with the changes.

Do you think we should change it to not copy the lattice?

qiskit_nature/second_q/problems/electronic_structure_problem.py
mrossinek: I think having `ElectronicEnergy` in the type hint will error in `mypy` because it will break the Liskov substitution principle (I need to double check though).

Subclasses of `ElectronicEnergy` are fine, too. Is my docstring too stringent the way its written?

qiskit_nature/second_q/problems/electronic_structure_problem.py
mrossinek: The `basis_transform` is here only until the next refactoring with the `PolynomialTensor` can be done. Once that happens, the problem will have a single `basis` associated with it, and the `basis_transform` will no longer be stored inside of it. Instead, a `BasisTransformer` can be extracted which can transform the basis of an `ElectronicStructureProblem`

qiskit_nature/second_q/properties/bases/electronic_basis_transform.py
mrossinek: This will get refactored further to a degree where this will no longer be a property. Instead, this will be implemented as a `Transformer` at which point this method might go away entirely

qiskit_nature/second_q/transformers/active_space_transformer.py
mrossinek: I will try with `mypy` :+1:

qiskit_nature/second_q/problems/electronic_structure_problem.py
mrossinek: It does indeed work as expected. I updated things accordingly :+1:

qiskit_nature/second_q/transformers/active_space_transformer.py
mrossinek: In this case, it is indeed a violation of the Liskov substitution principle, so the type hint cannot be used in this scenario

woodsp-ibm (approved):

woodsp-ibm (dismissed):

woodsp-ibm (approved):

PR #801, opened 2022-08-24
fix typo effect -> affect

q-inho:
effect -> affect

### DIFF
--- a/docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
+++ b/docs/tutorials/11_quadratic_hamiltonian_and_slater_determinants.ipynb
@@ -384,7 +384,7 @@
    "source": [
     "## Time evolution\n",
     "\n",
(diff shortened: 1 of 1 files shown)


manoelmarques (approved):

woodsp-ibm:
I must admit that I would have thought it ok

> The state preparation circuits shown above effect this basis change

effect meaning to bring about. Somehow affect seems like the wrong word as in affects it to alter it in some way.

Maybe we could choose something different, seeing also as these get translated, if its confusing. Would `The state preparation circuits shown above implement this basis change` work or something like this @kevinsung

kevinsung (changes requested):
This is not a typo. Do not merge.

kevinsung:
The wording is correct as is, and the proposed change makes it incorrect. See https://en.wiktionary.org/wiki/effect#Verb.

kevinsung:
Please close this PR without merging it.

q-inho:
Thank you for clarification.

woodsp-ibm:
Yes, I saw its correct as-is. My question was more can/should we do choose some different wording so as to remove any chance of confusion - i.e using the too simplistic effect noun/affect verb thing. We could easily put 'bring about' there, or 'implement' as I suggested above - both are meanings given by your wikipedia link and these hopefully would avoid any future effect/affect confusion. The docs get translated too, so avoiding any potential confusion, like seems to have already happened, might be a good thing.

woodsp-ibm:
The author closed this while I was typing the above and I did not see until I submitted the comment. I guess we can see if we get any feedback from translators going forwards - we have had in the past when things were not clear.

Issue #802, opened 2022-08-25
UCCSD fails for reps>1 (with MP2 initial point)


cpossel:
### Environment

- **Qiskit Nature version**: 0.4.4
- __qiskit_version__ = {'qiskit-terra': '0.21.1', 'qiskit-aer': '0.10.4', 'qiskit-ignis': '0.7.1', 'qiskit-ibmq-provider': '0.19.2', 'qiskit': '0.37.1', 'qiskit-nature': '0.4.4', 'qiskit-finance': None, 'qiskit-optimization': None, 'qiskit-machine-learning': None}
- **Python version**: 3.8.13
- **Operating system**: Windows 10

### What is happening?

Using the MP2 initial point in combination with UCCSD ansatz works fine if the number of repetitions reps=1.
If reps>1 below error will be raised.

Actually the error is raised also with the default initial point with reps>1.

UCCSD should also support multiple repetitions as discussed with @woodsp-ibm in https://qiskit.slack.com/archives/CB6C24TPB/p1661358583372379?thread_ts=1661341203.419219&cid=CB6C24TPB 

### How can we reproduce the issue?

```
from qiskit.algorithms.optimizers import SPSA
from qiskit.providers.aer import AerSimulator
from qiskit.utils import QuantumInstance
from qiskit_nature.algorithms import GroundStateEigensolver, VQEUCCFactory
from qiskit_nature.algorithms.initial_points import MP2InitialPoint
from qiskit_nature.circuit.library.ansatzes import UCCSD
from qiskit_nature.converters.second_quantization import QubitConverter
from qiskit_nature.drivers import Molecule
from qiskit_nature.drivers.second_quantization import ElectronicStructureMoleculeDriver
from qiskit_nature.mappers.second_quantization import JordanWignerMapper
from qiskit_nature.problems.second_quantization import ElectronicStructureProblem

H_2 = Molecule(
    geometry=[
        ("H", [0.0, 0.0, 0.0]),
        ("H", [0.0, 0.0, 0.74279]),
    ],
    multiplicity=1,
    charge=0,
)
driver = ElectronicStructureMoleculeDriver(H_2)  # In my case this chooses PSI4 (on Windows 10)
problem = ElectronicStructureProblem(driver)
qubit_converter = QubitConverter(mapper=JordanWignerMapper())
second_q_ops = problem.second_q_ops()
num_particles = problem.num_particles
[...]

PR #805, opened 2022-08-25
Fix initial points with reps

declanmillar:
### Summary

Fix the new initial point classes to account for the number of times the evolved operators are repeated in the ansatz. Also refactors the code to use the excitation_list only from the ansatz.

~~Closes: #802~~ Now closed by PR #808.


woodsp-ibm:
I am thinking we should fix the original classes too - maybe in a separate PR so that it can be backported.

mrossinek (dismissed):
LGTM, thank you for the quick fix! Could you also prepare a second PR after this one, which applies the same changes to the old code location like Steve has suggested, such that we can backport this bugfix?

test/second_q/algorithms/ground_state_solvers/test_groundstate_eigensolver.py
```diff
@@ -537,17 +538,67 @@ def test_vqe_ucc_factory_with_user_initial_point(self):
...
+        calc = GroundStateEigensolver(self.qubit_converter, solver)
+        res = calc.solve(self.electronic_structure_problem)
+
+        np.testing.assert_array_almost_equal(
+            solver.initial_point.to_numpy_array(), self.mp2_initial_point
+        )
+        self.assertAlmostEqual(res.total_energies[0], self.reference_energy, places=6)
+
+    def test_vqe_ucc_factory_with_reps(self):
+        """Test when using MP2InitialPoint to generate the initial point with repeated evolved
```
cpossel, line 556: This test's docstring is misleading. As far as I can see, this test uses the default initial point which is not MP2 but hartree fock

declanmillar: Good point. Thanks! Fixed.

declanmillar:
I decided to remove the property `excitation_list` so that the private attribute `_excitation_list` can only be set via the `ansatz`, rather than introducing logic to get if `reps` has been specified or not. Given that these changes are internal to the stack, I assume I'm free to remove this without depreciation? @mrossinek?

qiskit_nature/second_q/algorithms/initial_points/hf_initial_point.py
```diff
@@ -37,46 +34,23 @@ class HFInitialPoint(InitialPoint):
...
-
-    The excitation list generated by the :class:`~qiskit.circuit.library.ansatzes.ucc.UCC` ansatz is
-    obtained to ensure that the shape of the initial point is appropriate.
     """
 
     def __init__(self) -> None:
         super().__init__()
         self._ansatz: UCC | None = None
         self._excitation_list: list[tuple[tuple[int, ...], tuple[int, ...]]] | None = None
+        self._reps: int = 1
```
woodsp-ibm, line 43: Is there any reason to pass these in and store them. Once the anstaz is known they can be gotten from the ansatz right - thats what done internally. Why copy them over when you can access them of ansatz?

releasenotes/notes/fix-initial-point-with-reps-4725be71d06cb747.yaml
```diff
@@ -0,0 +1,8 @@
+---
+features:
```
woodsp-ibm, line 2: I think this is more deprecation section and a fixes section. That is what it would need to be if we are going to fix this on the existing code. And if we do it there its not needed here as such. We might want to give some clues in a migration guide but if we do a bug fix and the logic is already deprecated there then I guess to some extent you are already clued-in.

As to your comments, on the changes vis-a-vis deprecation, since this is "all new code" in this location it does not need it. But as I say we may need to give some hints on transitioning from current initial points to these new ones in migration guide.

qiskit_nature/second_q/algorithms/initial_points/hf_initial_point.py
declanmillar: I thought the same, but I was already pulling the excitation list out, so I chose to take the reps too. But, as you pointed it out, I'll just access these via the ansatz :).

mrossinek: Is the only reason you were pulling out the excitation list, its usage in the MP2 ansatz, in the first place? If so, maybe we can remove its duplication there too, and just access it via the ansatz everywhere :thinking:

qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -117,36 +116,17 @@ def __init__(self, threshold: float = 1e-12) -> None:
     def ansatz(self) -> UCC:
         """The UCC ansatz.
 
-        This is used to ensure that the :attr:`excitation_list` matches with the UCC ansatz that
-        will be used with the VQE algorithm.
+        The ``excitation_list`` and ``reps`` used by the
+        :class:`~qiskit.circuit.library.ansatzes.uvcc.UVCC` ansatz is obtained to ensure that the
```
mrossinek, line 120:
```suggestion
        :class:`~qiskit.circuit.library.ansatzes.ucc.UCC` ansatz is obtained to ensure that the
```

mrossinek (commented):
Besides this copy-paste error, I think this is now a lot cleaner approach, when not duplicating the excitation lists stored inside the ansatz.
I agree with Steve's comment regarding the release note. For the backporting PR, we could also keep changes more minimal (i.e. not do the excitation list handling via the ansatz, in order to not deal with deprecation there)

releasenotes/notes/fix-initial-point-with-reps-4725be71d06cb747.yaml
declanmillar: So, am I right in thinking that this PR does not require a releasenote, as this will be covered by the PR on the existing algorithms?

mrossinek: It actually is still needed. Just that you should have two sections: `fixes` and `deprecates` (or whatever the default names are here).
In one you list the actual bug which is being fixed (which is the part that would also need to be backported to the stable branch later), and in the other section you list what you have `deprecated`. However, this part is arguably not needed in this part of the stack

declanmillar: The release notes should only log changes relative to previous releases though, right? The initial points under `second_q` didn't exist in the previous release so we don't need to log the fix and nothing needs to be depreciated. The fix noted in the release note for PR #808 should cover the fix to the old initial points.

mrossinek: Having the fix documented in the reno of the other PR seems fine. But I would still deprecate that the `excitation_list` cannot be set manually, simply because there is no other way the user would find out about this

releasenotes/notes/fix-initial-point-with-reps-6157ede7a3b9924a.yaml
```diff
@@ -0,0 +1,9 @@
+---
+deprecations:
+  - |
+    The public :attr:`excitation_list` property has now been removed from
+    :class:`qiskit_nature.second_q.algorithms.initial_points.HFInitialPoint`,
+    :class:`qiskit_nature.second_q.algorithms.initial_points.MP2InitialPoint`, and
+    :class:`qiskit_nature.second_q.algorithms.initial_points.VSCFInitialPoint`.
+    Thus the excitation list cannot be set directly, and the private attribute
+    :attr:`_excitation_list` is now set exclusively via the :attr:`ansatz` property.
```
mrossinek, line 9: I'm not sure whether we would want to refer to a private attribute in the release note here..
Maybe having the deprecation above is sufficient while also simply pointing out, that one can modify the excitation list of the provided ansatz?

@woodsp-ibm wdyt?

woodsp-ibm: The release notes are intended to be user facing and this I would agree that talking about some private attribute is out of place. Maybe something like this?

    Thus the excitation list cannot be set directly, rather if required this must be done
    directly on the anstaz via the :attr:`ansatz` property.

qiskit_nature/second_q/algorithms/initial_points/hf_initial_point.py
```diff
@@ -34,18 +34,29 @@ class HFInitialPoint(InitialPoint):
...
+        """
+        return self._ansatz
+
+    @ansatz.setter
+    def ansatz(self, ansatz: UCC) -> None:
+        self._invalidate()
+        self._ansatz = ansatz
+
     @property
     def grouped_property(self) -> BaseProblem | None:
```
mrossinek, line 61: I'm noticing just now that I never updated the naming of this variable in #796..
I will do that in a separate PR in order to get this one done :+1:

qiskit_nature/second_q/hamiltonians/electronic_energy.py
```diff
@@ -213,7 +213,7 @@ def transform_basis(self, transform: ElectronicBasisTransform) -> None:
 
     def integral_operator(self, density: OneBodyElectronicIntegrals) -> OneBodyElectronicIntegrals:
         """Constructs the Fock operator resulting from this
-        :class:`~qiskit_nature.second_q.properties.ElectronicEnergy`.
+        :class:`~qiskit_nature.second_q.hamiltonians.ElectronicEnergy`.
```
mrossinek, line 216: Looks like I was a bit over-ambitious with fixing the failing docs after my failed merge conflict resolution.
This change is technically not part of this PR but I'm willing to let this slide

mrossinek (approved):
Thanks, Declan! :slightly_smiling_face:

woodsp-ibm (approved):

PR #808, opened 2022-08-26
Fix existing initial points with reps

declanmillar:
### Summary

Fixes the existing initial point classes to account for the number of times the evolved operators are repeated in the ansatz.

The new initial point classes are fixed with some additional refactoring in PR #805. This PR minimally fixes the issue for the existing classes.

Closes: #802.

### DIFF
--- a/qiskit_nature/algorithms/initial_points/hf_initial_point.py
+++ b/qiskit_nature/algorithms/initial_points/hf_initial_point.py
@@ -147,7 +147,8 @@ class HFInitialPoint(InitialPoint):
         self._compute()
 
     def _compute(self) -> None:
-        self._parameters = np.zeros(len(self._excitation_list))
+        reps = self._ansatz.reps if self._ansatz is not None else 1
+        self._parameters = np.zeros(reps * len(self._excitation_list))
 
     def get_energy(self) -> float:
         """The reference energy.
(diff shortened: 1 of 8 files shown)


qiskit_nature/algorithms/initial_points/hf_initial_point.py
```diff
@@ -147,7 +147,8 @@ def compute(
         self._compute()
 
     def _compute(self) -> None:
-        self._parameters = np.zeros(len(self._excitation_list))
+        reps = self._ansatz.reps if self._ansatz is not None else 1
```
mrossinek, line 150: When do you expect `self._ansatz` to be `None`? Because in such a scenario this could still not work as expected, right?

mrossinek (commented):
Other than this observation, I think this minimal change is good for backporting the larger fix done in #805 :+1:

qiskit_nature/algorithms/initial_points/hf_initial_point.py
declanmillar: Because one can bypass the ansatz and set the excitation list directly in these old classes. I think this would cover such case. In the new classes this is no longer possible.

mrossinek (approved):
Thanks for the quick fix, Declan!

woodsp-ibm (approved):

PR #513, opened 2022-01-20
Add support for Python 3.10

manoelmarques:
### Summary
Following Qiskit/qiskit-terra#7102

### Details and comments
There isn't pyscf for python 3.10 yet. Since the tutorials use pyscf, I am keeping them running in the CI under python 3.7 and python 3.8 as before.

IMPORTANT:
Terra is backporting this. Once they release it to pypi, I will backport it as well. Do not backport before they release otherwise it will fail. The stable CI gets Terra from Pypi.

PR #817, opened 2022-09-06
Loosen python version requirement from pyscf extra dependency

burgholzer:
### Summary

This PR loosens the Python version limitation from the `pyscf` extra. This makes `pip install qiskit-nature[pyscf]` work as expected also under Python 3.10 (at least under systems not running macOS) .

### Details and comments

Starting from `pyscf==2.1.0` (released Sep 3, 2022) the project also provides wheels for Python 3.10 (see https://pypi.org/project/pyscf/2.1.0/#files). As a consequence, the limitation on the `pyscf` extra introduced in https://github.com/Qiskit/qiskit-nature/pull/513 can be loosened. 

Unfortunately, it can't be dropped completely because the installs for macOS systems using Python 3.10 fails.
This is mainly due to 
- `pyscf==2.1.0` pinning `scipy<=1.1.0` (due to https://github.com/scipy/scipy/issues/15362 and https://github.com/scipy/scipy/issues/16151) and `qiskit-nature` requiring `scipy>=1.4.0`. 
- Hence, the install falls back to `pyscf==2.0.1` which works for all systems with Python<3.10.
- However, the source install under Python 3.10 fails due to https://github.com/pyscf/pyscf/issues/1112, which only got fixed in v2.1.0.

The corresponding issue in `pyscf` is being tracked at https://github.com/pyscf/pyscf/issues/1405.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ setuptools.setup(
     include_package_data=True,
     python_requires=">=3.7",
     extras_require={
-        "pyscf": ["pyscf; python_version < '3.10' and sys_platform != 'win32'"],
+        "pyscf": ["pyscf; sys_platform != 'win32'"],
         "mpl": ["matplotlib>=3.3"],
     },
     zip_safe=False,
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ setuptools.setup(
     include_package_data=True,
     python_requires=">=3.7",
     extras_require={
-        "pyscf": ["pyscf; sys_platform != 'win32'"],
+        "pyscf": ["pyscf; sys_platform != 'win32' and (python_version < '3.10' or sys_platform != 'darwin')"],
         "mpl": ["matplotlib>=3.3"],
     },
     zip_safe=False,


manoelmarques:
The pyscf 2.1.0 distrbution for MacOS changed the  scipy requirements from `scipy (!=1.5.0,!=1.5.1)` to `scipy (<=1.1.0)`.
However there is no `scipy (<=1.1.0)` distribution for python 3.8/3.10 under MacOS. 
For the MacOS, under python 3.8, it reverts to pyscf 2.0.1 and works, however for python 3.10, it can't install the current wheel and fails.

burgholzer:
Yeah, that change was introduced in https://github.com/pyscf/pyscf/pull/1249. The reasons for the downgrade of `scipy` are the https://github.com/scipy/scipy/issues/15362 and https://github.com/scipy/scipy/issues/16151.

What makes matters worse is that the source distribution of `scipy==1.1.0` is broken (a `CMakeLists.txt` file is missing, see https://github.com/pyscf/pyscf/issues/1112).

manoelmarques:
> Yeah, that change was introduced in [pyscf/pyscf#1249](https://github.com/pyscf/pyscf/pull/1249). The reasons for the downgrade of `scipy` are the issues [scipy/scipy#15362](https://github.com/scipy/scipy/issues/15362) and [scipy/scipy#16151](https://github.com/scipy/scipy/issues/16151).
> 
> What makes matters worse is that the source distribution of `scipy==1.1.0` is broken (a `CMakeLists.txt` file is missing, see [pyscf/pyscf#1112](https://github.com/pyscf/pyscf/issues/1112)).

Yes, I wasn't aware of the reasons to the scipy downgrade, it is a version from 2018, very old. In any case, pyscf uploaded MacOS wheel files that don't seem to install. I opened an issue but I am not sure when it will be fixed.
It seems to me  the course of action would be to disable the pyscf install for python 3.10 and mac:
`"pyscf": ["pyscf; python_version < '3.10' or sys_platform != 'darwin'"]`

burgholzer:
> > Yeah, that change was introduced in [pyscf/pyscf#1249](https://github.com/pyscf/pyscf/pull/1249). The reasons for the downgrade of `scipy` are the issues [scipy/scipy#15362](https://github.com/scipy/scipy/issues/15362) and [scipy/scipy#16151](https://github.com/scipy/scipy/issues/16151).
> > What makes matters worse is that the source distribution of `scipy==1.1.0` is broken (a `CMakeLists.txt` file is missing, see [pyscf/pyscf#1112](https://github.com/pyscf/pyscf/issues/1112)).
> 
> Yes, I wasn't aware of the reasons to the scipy downgrade, it is a version from 2018, very old. In any case, pyscf uploaded MacOS wheel files that don't seem to install. I opened an issue but I am not sure when it will be fixed. It seems to me the course of action would be to disable the pyscf install for python 3.10 and mac: `"pyscf": ["pyscf; sys_platform != 'win32' and (python_version < '3.10' or sys_platform != 'darwin')"]`

Unfortunately, that dependency specification won't work either since it would try to install `pyscf` under macOS when using Python<3.10. However, as you noted above and in the related issue, these wheels are broken. As a consequence, `pyscf` cannot be properly installed for macOS at all. The following pattern would work
```
"pyscf": ["pyscf; sys_platform != 'win32' and sys_platform != 'darwin'"]
```
or with an explicit version
```
"pyscf": ["pyscf>=2.1; sys_platform != 'win32' and sys_platform != 'darwin'"]
```

This would 
- allow the extra to work properly for linux systems running Python up to 3.10
- disable the install of `pyscf` on platforms where it is currently broken

Let me know what you think

manoelmarques:
> Unfortunately, that dependency specification won't work either since it would try to install `pyscf` under macOS when using Python<3.10. However, as you noted above and in the related issue, these wheels are broken. As a consequence, `pyscf` cannot be properly installed for macOS at all (without falling back to an older version). The following pattern would work
That was exactly my intention. Pyscf 2.0.1 has been working fine for MacOS < 3.10. I want to keep this available. The only thing I would like to limit until Pyscf is fixed, is installing it under MacOS with python >= 3.10 or Windows.

burgholzer:
> That was exactly my intention. Pyscf 2.0.1 has been working fine for MacOS < 3.10. I want to keep this available. The wheel is broken and it automatically reverts to 2.0.1 as per the Github actions logs, since there is no pyscf version limit in the requires. The only thing I would like to limit until Pyscf is fixed, is installing it under MacOS with python >= 3.10 or Windows.

Sounds reasonable to me as a temporary solution until pyscf is fixed. I'll push the respective changes and update the PR title and description accordingly.

mrossinek (approved):
LGTM :+1: 

Awaiting another check by @manoelmarques before merging

manoelmarques (approved):

manoelmarques:
Thanks a lot @burgholzer for keeping an eye on pyscf.

PR #821, opened 2022-09-07
Keep matplotlib installed in the CI

manoelmarques:
### Summary
VQE has a statement `logger.info(self.print_settings())` that ends up calling `__str__` in `QuantumCircuit` which calls `str(self.draw(output="text"))`. This ends up importing `circuit_drawer` from visualization that imports `Bloch` which finally imports `matplotlib` and fails it it is not installed.
Keeping it installed in the CI.


mrossinek (approved):
Thanks a lot!

PR #823, opened 2022-09-13
Refactor `FermionicOp` as `SparseLabelOp`

### LINKED ISSUE #562: Make SecondQuantizedOp inherit from Iterable
### What should we add?

#544 Makes FermionicOp iterable. As @woodsp-ibm mentioned there, the other subclasses of SecondQuantizedOp should also be made iterable by defining `__iter__`. In fact, he suggested that subclasses of SecondQuantizedOp be required to define `__iter__`. This can be done by making SecondQuantizedOp inherit from the Iterable class from the `typing` module.

I think this is a good suggestion. However, it immediately raises the question of why the `to_list` abstract method on SecondQuantizedOp exists. Apparently, `op.to_list()` is supposed to convert `op` to a list. But there is already a built-in function for that: `list(op)`. So at best, `to_list` is a redundant way to do what the built-in function `list` does. The reality is actually worse, because with #544, `list(op)` and `op.to_list()` actually return different things (see #541 for the explanation why). So after making this change, it seems that `to_list` should be deprecated.

### LINKED ISSUE #610: Deprecate dense labels

Dense labels are confusing to use. I'm sure I'm not the only one who thinks this; I have encountered bugs in other people's code that arose from the use of dense labels. There are two sources of confusion:
1. It is easy to forget to account for the accumulated sign due to permutation of fermionic modes
2. Dense labels in Qiskit Nature use big endian ordering, which is the opposite of Qiskit Terra's convention

Dense labels are not necessary because sparse labels can always be used instead.

mrossinek:
### Summary

This PR refactors the `FermionicOp` to become a `SparseLabeOp`.

### Details and comments

Closes #610
Closes #562


qiskit_nature/second_q/operators/fermionic_op.py
```diff
@@ -14,158 +14,68 @@
...
-    The following labels are equivalent to the example given for sparse labels
-    above:
-
-    .. code-block:: python
-
-        [("+", 0)]
-        [("-", 2)]
-        [("+", 0), ("-", 1), ("+", 4), ("+", 10)]
+    A `FermionicOp` represents a weighted sum of fermionic creation/annihilation operator terms.
+    These terms are encoded as strings consisting of a space-separated list of words. Each word must
```
woodsp-ibm, line 31: > These terms are encoded as strings consisting of a space-separated list of words

`words` here seems a bit odd - I had wondered about `expressions`. And expression would then be an operation and index. Or you could have these as operations where an operation comprises an operator creation/annihilation and an index. Though maybe we are trying to avoid too much overloaded use of  operator

.pylintdict
```diff
@@ -436,6 +438,7 @@ refactoring
 regs
 reinstall
 repr
+reproducability
```
woodsp-ibm, line 441: This looks suspect - `reproducibility` maybe is what was wanted and maybe that spelling exists

qiskit_nature/second_q/operators/fermionic_op.py
mrossinek: Thanks, that's a good suggestion! I think the old docstring also used `word` which is where I got that from :smile:

woodsp-ibm (dismissed):

qiskit_nature/second_q/operators/fermionic_op.py
```diff
@@ -14,158 +14,72 @@
...
+                "+_0 -_0": 1.0,
+                "+_1 -_1": -1.0,
+            },
+            register_length=2,
+        )
 
-    Example:
+    By default, this way of initializing will create a full copy of the dictionary of coefficients.
+    If you have very restricted memory resources available, or would like to avoid the additional
+    copy, you can disable it like so:
```
ElePT, line 61:
```suggestion
    copy, the dictionary will be stored by reference if you disable ``copy`` like so:
```

qiskit_nature/second_q/operators/fermionic_op.py
```diff
@@ -14,158 +14,72 @@
...
-    The following labels are equivalent to the example given for sparse labels
-    above:
-
-    .. code-block:: python
-
-        [("+", 0)]
-        [("-", 2)]
-        [("+", 0), ("-", 1), ("+", 4), ("+", 10)]
+    A `FermionicOp` represents a weighted sum of fermionic creation/annihilation operator terms.
+    These terms are encoded as strings consisting of a space-separated list of expressions. Each
```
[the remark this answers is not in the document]
ElePT: sorry, this suggestion makes the line a bit long, if you apply it you will probably have to reformat 🙈

qiskit_nature/second_q/operators/fermionic_op.py
```diff
@@ -14,158 +14,72 @@
...
 
-      print("truncated str output")
-      print(sum(FermionicOp("I", display_format="sparse") for _ in range(25)))
 
-      FermionicOp.set_truncation(0)
-      print("not truncated str output")
-      print(sum(FermionicOp("I", display_format="sparse") for _ in range(25)))
+    .. note::
 
+        It is the users responsibility, that in the above scenario, :code:`some_big_data` is not
```
ElePT, line 80: Either
```suggestion
        It is the user's responsibility, that in the above scenario, :code:`some_big_data` is not
```

ElePT: or 
```suggestion
        It is the users' responsibility, that in the above scenario, :code:`some_big_data` is not
```

ElePT (commented):
Hey Max, here are my 2 cents on this greatly-looking PR, feel free to apply or dismiss my suggestions, I just added a bit of clarification on the parts I had to read twice to get (POV of a noob user).

woodsp-ibm (approved):

PR #826, opened 2022-09-17
Pin matplotlib due to bug in retworkx visualization

manoelmarques:
### Summary
With the recent matplotlib release retworkx visualization fails. Pin matplotlib until it gets fixed:
Qiskit/rustworkx#680


mrossinek (approved):
Thanks, Manoel!

PR #837, opened 2022-09-22
Fix convertlib unit test, add error msg 

manoelmarques:
### Summary

We could not recreate this error on stable or main. We suspect that the stderr could be polluted with other msgs besides the error msg. We test for containment then instead of `start` and also print the msg from stderr if it fails.


mrossinek (approved):
Thanks, Manoel! :+1:

PR #839, opened 2022-09-22
Fix doc.deploy script due to matplotlib, retworkx limitation

manoelmarques:
### Summary
The documentation deploy script was adjusted to account for the matplotlib/retworkx version problem.
Also the matplotlib was reinserted into requirements-dev.txt since this file is used by the doc.translate scripts.

It is not necessary to backport this PR since those changes have already been made in stable/0.4


mrossinek (approved):
Thanks, Manoel!

PR #840, opened 2022-09-23
Cache .stestr in CI

manoelmarques:
### Summary
Caching the `.stestr` database created by `stestr` so that it can be read in later runs in order to improve the grouping of unit test methods per process.
This is already done in qiskit-terra

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -219,6 +219,17 @@ jobs:
           pip install -U -r requirements-dev.txt
           make lint
         shell: bash
+      - name: Stestr Cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            .stestr
+            .stestr1
+          key: stestr-${{ matrix.os }}-${{ matrix.python-version }}-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}
+          restore-keys: |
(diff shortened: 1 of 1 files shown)


mrossinek (approved):

.github/workflows/main.yml
```diff
@@ -219,6 +219,17 @@ jobs:
           pip install -U -r requirements-dev.txt
           make lint
         shell: bash
+      - name: Stestr Cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            .stestr
+            .stestr1
```
mtreinish, line 227: Just something to keep in mind with this stestr stores a few different pieces of data in the repository (this path). There is the run history which are the results streams from all previous runs, the times database, and the metadata database (which isn't used unless you add metadata about a run). For the scheduler you only need the times database since it is used as a scheduling hint to stestr (so it tries to keep the workers balanced).

Caching the entire directory will lead to very large cache growth over time since each time we update the cache we store an entire output stream (including any captured attachments like stdout and stderr). This will add up to gigabytes eventually and cause things to bog down. That's why in terra we delete the run history with: https://github.com/Qiskit/qiskit-terra/blob/main/.azure/test-linux.yml#L127 to only back up the times database

mtreinish (changes requested):

Issue #820, opened 2022-09-06
Ill-defined behavior of stack when using odd `num_spin_orbitals`


kevinsung:
### What is the expected enhancement?

AngularMomentum is currently initialized by specifying the number of spin orbitals. This is confusing because it doesn't make sense to have an odd number of spin orbitals, and passing an odd number seems to give the same operator as the next lowest even number:
```python
from qiskit_nature.second_q.properties import AngularMomentum

op1 = AngularMomentum(2).second_q_ops()['AngularMomentum']
op2 = AngularMomentum(3).second_q_ops()['AngularMomentum']
print(op1.equiv(op2))
```
```
True
```
I think it should take the number of spatial orbitals as input instead.

mrossinek:
This was a design flaw of the `properties` framework. This will be addressed as part of 0.5. It is blocked by:
- merging of the `SparseLabelOp` (#731)
- rewrite of the `FermionicOp` as a subclass of the `SparseLabelOp` (in progress by myself, PR coming soon)
- merging of the `PolynomialTensor` (#756)
- implementing #666 (will be part of the aforementioned PR which I am working on)

Once all of the above is done, all of the existing classes in `qiskit_nature.second_q.properties` which build out a `FermionicOp` will be revisited and improved :+1:

mrossinek:
This issue is actually more general with respect to the usage of odd values for `num_spin_orbitals` against which no guards are in place. A proper solution would be to consistently switch over to `num_spatial_orbitals`, instead :+1:

Issue #831, opened 2022-09-20
register length should be determined at map time instead of initialization


kevinsung:
### What is the expected enhancement?

Currently, `SparseLabelOp.register_length` stores the "register length" of the operator. Typically this would be the number of qubits in the operator when it is mapped to qubits. However, as far as I can tell, there is no need to store the register length with the original operator. Instead, the register length can be specified at the time the operator is mapped to qubits. For example,
```python
qubit_op = JordanWigner().map(fermionic_op, register_length=5)
```
If the register length is not specified, it can be automatically inferred from the indices present in the operator. This approach is more flexible because the user would not need to specify a register length during initialization.

mrossinek:
I don't think it makes sense to specify this as an input to the mapper. Instead, the user could just set the `fermionic_op.register_length` attribute to the desired value before passing it into the mapper.

Many scenarios (e.g. when comping from a set of integrals/coefficients for a Hamiltonian) will have a fixed `register_length` to begin with. In those cases it would be good to be able to guard against user error resulting from an incorrect value.
However, I agree that this will be a useful feature and ease the use of the operator classes. As the `SparseLabelOp` is currently designed to be immutable, this will need a little design discussion.

kevinsung:
>I don't think it makes sense to specify this as an input to the mapper. Instead, the user could just set the fermionic_op.register_length attribute to the desired value before passing it into the mapper.

I guess whether something "makes sense" is subjective. I think this is mostly a question of which of the following APIs are preferable:
```python
qubit_op = JordanWigner().map(fermionic_op, register_length=5)
```
versus
```python
fermionic_op.register_length = 5
qubit_op = JordanWigner().map(fermionic_op)
```

I much prefer the first one. It is much easier to document and discover because the documentation must live with function itself, and the argument can be discovered by tab auto-completion in an IDE. The second method would require one to know ahead of time about the existence of the `register_length` property and the way it is handled by the mapper. Furthermore, the first method is much more readable because all of the logic is contained in one statement rather than split between two statements. For example, it is not clear from inspection that the line
```python
fermionic_op.register_length = 5
```
should have an effect on the line
```
qubit_op = JordanWigner().map(fermionic_op)
```
One could imagine a situation where these lines of code are written together, then at a later point of time some code is added between these lines. The readability would suffer greatly.

mrossinek:
[...]

PR #841, opened 2022-09-23
Electronic Structure Stack Refactoring - Part 2

### LINKED ISSUE #666: Implement `FermionicOp.from_polynomial_tensor`

### LINKED ISSUE #679: Nuclear repulsion energy is tough to find!

It's pretty difficult to find the nuclear repulsion energy (and probably some other interesting properties) if one doesn't already know where to look:
```python
# es_problem is an `ElectronicStructureProblem`
es_problem.grouped_property_transformed.get_property("ElectronicEnergy").nuclear_repulsion_energy
```

It would be great it the tutorials would mention where they are, e.g. [the ground state solver tutorial](https://github.com/Qiskit/qiskit-nature/blob/main/docs/tutorials/03_ground_state_solvers.ipynb).

(Originated from [this Slack convo](https://qiskit.slack.com/archives/CB6C24TPB/p1653900859430849))

mrossinek:
### Summary

This is the second part of the major refactoring of the electronic structure stack.
It does the following:
- implements `FermionicOp.from_polynomial_tensor`
- adds the `ElectronicIntegrals` container based on the `PolynomialTensor` (a similar design can be used for #654)
- refactors the `ElectronicEnergy` and `ElectronicDipoleMoment` based on the above
- refactors the `ElectronicStructureProblem` to be single-basis only
- updates the electronic transformers to work with the refactored problem class
- removes the now obsolete `IntegralProperty` and `properties.intergals.ElectronicIntegrals` classes (+ subclasses)

### Details and comments

- [x] missing functionality:
  - [x] settle on implementation details of `FermionicOp.from_polynomial_tensor` (see inline comments)
  - [x] ~(optionally may become part of this PR) update `AngularMomentum`, `Magnetization` and `ParticleNumber` (e.g. as suggested in #820)~ see #851
  - [x] data validation in `ElectronicIntegrals`
  - [x] key validation in `SparseLabelOp.register_length.setter`
- [x] documentation
  - [x] module/class/method level
  - [x] ~release note listing the major features (similar to the list above) but see next point~ (makes more sense to have with the migration guide)
  - [x] I plan to write the migration guide in one big go for the entire 0.5 release
  - [x] explicit example in `ElectronicDipoleMoment` for `nuclear_dipole_moment` as constant shift in `PolynomialTensor` objects
- [x] new unittests:
  - [x] for `ElectronicIntegrals`
  - [x] for `BasisTransformer`

Closes #666
Closes #679


qiskit_nature/second_q/formats/fcidump/fcidump.py
```diff
@@ -41,6 +41,7 @@ class FCIDump:
 
     hij: np.ndarray
     """The alpha 1-electron integrals."""
+    # TODO: automatically handle to-chemist index order conversion?
```
mrossinek, line 44: I am not sure what the FCIDump format dictates. Will look into this in more detail next week

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -0,0 +1,313 @@
...
+        alpha: PolynomialTensor | None = None,
+        beta: PolynomialTensor | None = None,
+        mixed: PolynomialTensor | None = None,
+    ) -> None:
+        """TODO."""
+        self._register_length = alpha.register_length if alpha is not None else None
+        self._data: dict[str, PolynomialTensor] = {}
+        self.alpha = alpha
+        self.beta = beta
+        self.mixed = mixed
```
mrossinek, line 49: Actually, this structure is reminiscent of a previous implementation prototype where `ElectronicIntegrals` was trying to implement a Composite-pattern around the `PolynomialTensor`. This is no longer the case (and also not wanted) so this can be changed to simply be three `PolynomialTensor`-type attributes.

qiskit_nature/second_q/operators/fermionic_op.py
```diff
@@ -161,6 +163,63 @@ def _validate_keys(cls, keys: Collection[str], register_length: int) -> None:
...
+
+            # and yet another solution which uses the same numpy iterator as above, but the C-API
+            # instead of Pythonic iterator, which according to their docs could be beneficial when
+            # requiring access to the iterator coordinates/indices
+            # however, here it is even slower than the variant above
+            # Reference: https://numpy.org/doc/stable/reference/generated/numpy.nditer.html
+            # nditer = np.nditer(tensor[key], flags=["multi_index"], op_flags=["readonly"])
+            # while not nditer.finished:
+            #     data[label_template.format(*nditer.multi_index)] = nditer[0]
+            #     nditer.iternext()
```
mrossinek, line 216: I would like to discuss these various solutions before settling on one

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -43,19 +55,23 @@ def __init__(self, data: Mapping[str, np.ndarray | Number], register_length: int
...
+            if validate and len(value.shape) != len(key):
                 raise ValueError(
                     f"Data key {key} of length {len(key)} does not match "
                     f"data value matrix of dimensions {value.shape}"
                 )
 
             dims = set(value.shape)
 
-            if len(dims) > 1:
+            if validate and len(dims) > 1:
```
mrossinek, line 66: We want to be able to disable this check, in order to have the ability to do e.g. active-space transformations which map from `n` AOs to `m` MOs

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -71,6 +87,15 @@ def register_length(self) -> int:
...
         return self._register_length
 
+    @classmethod
+    def empty(cls, register_length: int | None) -> PolynomialTensor:
+        """TODO."""
+        return PolynomialTensor({}, register_length=register_length)
+
+    def is_empty(self) -> bool:
+        """TODO."""
+        return self == PolynomialTensor.empty(self.register_length)
```
mrossinek, line 97: In similar vain to #831 we could also allow `register_length=None` and determine it lazily once the internal data has contents

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -205,4 +230,49 @@ def transpose(self) -> PolynomialTensor:
         for key, value in self._data.items():
             transpose_dict[key] = np.transpose(value)
 
-        return PolynomialTensor(transpose_dict, self._register_length)
+        return PolynomialTensor(transpose_dict, self._register_length, validate=False)
+
+    def compose(self, other, qargs=None, front=False):
+        # TODO:
+        return self
```
mrossinek, line 237: Reminder for myself. Once this is implemented, we can also add `GroupMixin` to this class

qiskit_nature/second_q/properties/bases/electronic_basis.py
```diff
@@ -18,13 +18,8 @@
...
-    basis into which an
-    :class:`~qiskit_nature.second_q.hamiltonians.ElectronicEnergy` must map its
-    integrals before being able to perform the mapping to a
-    :class:`~qiskit_nature.second_q.operators.SecondQuantizedOp`.
+    This ``Enum`` simply names the available electronic bases.
     """
 
     AO = "atomic"
     MO = "molecular"
-    SO = "spin"
```
mrossinek, line 30: We no longer require this special case so I don't see a reason for having this value here.

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -189,68 +182,67 @@ def transform(self, problem: BaseProblem) -> BaseProblem:
...
         if not isinstance(problem, ElectronicStructureProblem):
             raise QiskitNatureError(
                 "Only an ElectronicStructureProblem can be transformed by this Transformer, "
                 f"not problems of type, {type(problem)}."
             )
 
+        # TODO: re-implement non-MO case
+        if problem.basis != ElectronicBasis.MO:
+            raise NotImplementedError()
+            # TODO: suggest the user to fix this by first running a BasisTransformer
```
mrossinek, line 196: I would like suggestions: do you think raising an error, requesting the user to apply a `BasisTransformer` suffices? Anything else would require additional input parameters to this transformer, which tell us the transformation coefficients, since these are no longer nested inside the problem

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -352,6 +335,50 @@ def _validate_num_orbitals(self, nelec_inactive: int, particle_number: ParticleN
                     "does not match the specified number of active electrons."
                 )
 
+    def transform_hamiltonian(self, hamiltonian: Hamiltonian) -> Hamiltonian:
```
mrossinek, line 338: Note to self: make `transform_hamiltonian` part of `BaseTransformer` interface

qiskit_nature/second_q/transformers/basis_transformer.py
```diff
@@ -0,0 +1,97 @@
...
+from qiskit_nature.second_q.problems import BaseProblem, ElectronicStructureProblem
+from qiskit_nature.second_q.properties.bases import ElectronicBasis
+
+from .base_transformer import BaseTransformer
+
+
+class BasisTransformer(BaseTransformer):
+    """TODO."""
+
+    # TODO: figure out how we can make its interface non-electronic specific
```
mrossinek, line 31: This will only become relevant in the future, once the vibrational stack has more than one basis (currently the harmonic one), at which point transformation from it to another one could also be handled via this transformer.

qiskit_nature/second_q/transformers/basis_transformer.py
```diff
@@ -0,0 +1,97 @@
...
+    def __init__(
+        self,
+        initial_basis: ElectronicBasis,
+        final_basis: ElectronicBasis,
+        coefficients: ElectronicIntegrals,
```
mrossinek, line 36: Given the above, I would like to raise suggestions/questions:
- Shall we have a general `Basis` Enum? Or shall we just unionize `ElectronicBasis` and `VibrationalBasis`? Or shall we leave this as is for now?
- `ElectronicIntegrals` is also custom to electronic stack. I think a better type hint here would be some `Protocol` which states that the object needs to support `.einsum`. Maybe `Einsummable`? That would fit for `ElectronicIntegrals` and `PolynomialTensor` as of now.

qiskit_nature/second_q/transformers/basis_transformer.py
```diff
@@ -0,0 +1,97 @@
...
+            raise QiskitNatureError(
+                f"The problems' basis, {problem.basis}, does not match the initial basis of this "
+                f"transformer, {self.initial_basis}."
+            )
+
+        new_problem = ElectronicStructureProblem(
+            cast(ElectronicEnergy, self.transform_hamiltonian(problem.hamiltonian))
+        )
+        new_problem.basis = self.final_basis
+
```
mrossinek, line 66: TODO: need to handle properties such as `ElectronicDipoleMoment`, too

test/second_q/formats/fcidump/test_methods_fcidump.py
```diff
@@ -67,50 +66,20 @@ def test_oh(self):
...
-    @unittest.skip("Skip until FreezeCoreTransformer supports pure MO cases.")
-    def test_lih_with_atoms(self):
-        """LiH with num_atoms test"""
+    def test_lih_with_active_space(self):
+        """LiH with active space test"""
         fcidump = FCIDump.from_file(
             self.get_resource_path("test_fcidump_lih.fcidump", "second_q/formats/fcidump"),
         )
-        result = self._run_fcidump(fcidump, transformers=[FreezeCoreTransformer()])
+        result = self._run_fcidump(fcidump, transformers=[ActiveSpaceTransformer(4, 6)])
```
mrossinek, line 74: Back when the `FCIDump` was a `driver.second_quantization`, it was valid to provide a list of atoms to it, enabling the use of the `FreezeCoreTransformer`. Instead, I am now specifying a full-space `ActiveSpaceTransformer`, which also results in the same energy as the total system

test/second_q/hamiltonians/resources/electronic_energy_op.json
```diff
@@ -1 +1,38 @@
...
\ No newline at end of file
+{
```
mrossinek, line 1: I had to regenerate this file to get the physicist-ordered integrals

test/second_q/hamiltonians/test_electronic_energy.py
```diff
@@ -40,11 +35,10 @@ def setUp(self):
         super().setUp()
         driver = PySCFDriver()
         self.prop = cast(ElectronicEnergy, driver.run().hamiltonian)
-        self.prop.get_electronic_integral(ElectronicBasis.MO, 1).set_truncation(2)
 
     def test_second_q_op(self):
         """Test second_q_op."""
-        op = self.prop.second_q_op()
+        op = self.prop.second_q_op().chop()
```
mrossinek, line 41: Note to self: this may no longer be needed as the `.chop()` should be part of `Fermionic.from_polynomial_tensor()`

test/second_q/problems/resources/H2_631g_ferm_op_active_space.json
```diff
@@ -0,0 +1,38 @@
+{
```
mrossinek, line 1: TODO: remove the non-json files in this folder, which should no longer be needed now

test/second_q/problems/test_vibrational_properties_container.py
```diff
@@ -35,7 +32,7 @@ def test_occupied_modals(self) -> None:
 
         with self.subTest("wrong setting type"):
             with self.assertRaises(TypeError):
-                container.occupied_modals = ElectronicDipoleMoment()  # type: ignore[assignment]
+                container.occupied_modals = Magnetization(2)  # type: ignore[assignment]
```
mrossinek, line 35: I'm just using some other non-vibrational property, since the `ElectronicDipoleMoment` no longer has an empty-init available

test/second_q/properties/test_dipole_moment.py
```diff
@@ -65,64 +55,12 @@ def setUp(self):
     @unpack
     def test_second_q_ops(self, key: str, expected_op_data: dict[str, float]):
         """Test second_q_ops."""
-        op = self.prop.second_q_ops()[key]
+        op = self.prop.second_q_ops()[key].chop()
```
mrossinek, line 58: Same `.chop` note as before

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -205,4 +230,49 @@ def transpose(self) -> PolynomialTensor:
...
+    def expand(self, other: PolynomialTensor) -> PolynomialTensor:
+        return self._tensor(other, self)
+
+    @classmethod
+    def _tensor(cls, a: PolynomialTensor, b: PolynomialTensor) -> PolynomialTensor:
+        new_data = {}
+        for key in a:
+            if key not in b:
+                continue
+            new_data[key] = np.kron(a[key], b[key])
```
mrossinek, line 251: I had a `# type: ignore` comment here earlier, but the non-windows CIs complained about an unused type-ignore statement (which I found weird, because I was getting it locally. I figured my mypy version was out-of-sync). Having removed it now, the Windows-CI is failing now: https://github.com/Qiskit/qiskit-nature/actions/runs/3114556050/jobs/5050531078#step:11:13

@manoelmarques any ideas on how to resolve this?

qiskit_nature/second_q/formats/fcidump/fcidump.py
mrossinek: In fact, the FCIDump format dictates chemistry notation. Thus, I will handle the conversion directly here in order to be consistent :+1:

qiskit_nature/second_q/transformers/basis_transformer.py
mrossinek: `einsum` was not actually the real requirement here. Instead, we rely on `.transpose` to exist, so I have added a `Transposable` protocol which indicates its existence :+1:

qiskit_nature/second_q/properties/dipole_moment.py
```diff
@@ -267,30 +142,21 @@ def interpret(self, result: "EigenstateResult") -> None:
...
-
-            if isinstance(aux_op_eigenvalues, list) and len(aux_op_eigenvalues) < 6:
+            if not isinstance(aux_op_eigenvalues, dict):
                 continue
 
-            axes_order = {"x": 0, "y": 1, "z": 2}
-            dipole_moment = [None] * 3
-            for prop in self._dipole_axes.values():
-                moment = aux_op_eigenvalues.get(prop.name, None)
+            dipole_moment = [float("NaN")] * 3
```
woodsp-ibm, line 148: The dipole tuple type above was `Optional[Number]` - we went from None to NaN (not a Number) though out of curosity I tried `isinstance(float("NaN'), Number)` and despite not being a number it returns True!

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -91,36 +85,26 @@ class ActiveSpaceTransformer(BaseTransformer):
 
     def __init__(
         self,
-        num_electrons: Optional[Union[int, Tuple[int, int]]] = None,
-        num_molecular_orbitals: Optional[int] = None,
-        active_orbitals: Optional[List[int]] = None,
+        num_electrons: int | tuple[int, int],
+        num_spatial_orbitals: int,
```
woodsp-ibm, line 89: I guess I have to ask - in the past we have molecular and spin. Is this the same as the first? Or maybe this is more ambiguous to cope with AO as well as MO?

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -162,11 +147,11 @@ def _check_configuration(self):
...
             )
 
-        if isinstance(self._num_molecular_orbitals, (int, np.integer)):
-            if self._num_molecular_orbitals < 0:
+        if isinstance(self._num_spatial_orbitals, (int, np.integer)):
+            if self._num_spatial_orbitals < 0:
                 raise QiskitNatureError(
                     "The number of active orbitals cannot be negative, not:",
-                    str(self._num_molecular_orbitals),
+                    str(self._num_spatial_orbitals),
```
woodsp-ibm, line 154: The error message below has the wrong field in it
```
                "The number of active orbitals must be an int, not:",
                str(self._num_electrons),
```
Its referring to num_electrons and should be num_spatial_orbitals. I suppose another way could be to cast it to an int, instead of a type check, which would raise an exception if it could not be cast - but maybe leads to other more subtle problems if it were a float.

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -175,25 +160,39 @@ def _check_configuration(self):
...
         """
-        if not isinstance(problem, ElectronicStructureProblem):
-            raise QiskitNatureError(
-                "Only an ElectronicStructureProblem can be transformed by this Transformer, "
-                f"not problems of type, {type(problem)}."
+        if isinstance(problem, ElectronicStructureProblem):
+            return self._transform_electronic_structure_problem(problem)
+        else:
+            raise NotImplementedError(
+                f"The problem of type, {type(problem)}, is not yet supported by this transformer."
```
woodsp-ibm, line 184:
```suggestion
                f"The problem of type, {type(problem)}, is not supported by this transformer."
```

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -352,75 +350,106 @@ def _validate_num_orbitals(self, nelec_inactive: int, particle_number: ParticleN
...
+            if self._transform_active is None:
+                raise NotImplementedError(
+                    "This transformer does not yet support the standalone use of the "
+                    "transform_hamiltonian method. See also "
+                    "https://github.com/Qiskit/qiskit-nature/issues/847"
+                )
+            return self._transform_electronic_energy(hamiltonian)
+        else:
+            raise NotImplementedError(
+                f"The hamiltonian of type, {type(hamiltonian)}, is not yet supported by this "
```
woodsp-ibm, line 378:
```suggestion
                f"The hamiltonian of type, {type(hamiltonian)}, is not supported by this "
```

qiskit_nature/second_q/transformers/base_transformer.py
```diff
@@ -25,13 +26,32 @@ class BaseTransformer(ABC):
...
+
         Returns:
             A new `BaseProblem` instance.
         """
         raise NotImplementedError()
+
+    @abstractmethod
+    def transform_hamiltonian(self, hamiltonian: Hamiltonian) -> Hamiltonian:
+        """Transforms one :class:`~qiskit_nature.second_q.hamiltonians.Hamiltonian` into another.
+        This may or may not affect the size of the Hilbert space.
```
woodsp-ibm, line 46: Very minor comment, like the above transform `This may or may not affect the size of the Hilbert space.` I think `This may affect the size of the Hilbert space.` is entirely equivalent right - the `or may not` is implicit in `may affect`

qiskit_nature/second_q/transformers/basis_transformer.py
```diff
@@ -0,0 +1,232 @@
...
+else:
+    from typing_extensions import Protocol
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Transposable(Protocol):
+    """A Protocol indicating the existence of a ``.transpose()`` function."""
+
+    def transpose(self) -> Transposable:
```
woodsp-ibm, line 47: I was wondering about the return type here as it ideally it should be the type of the class implementing this protocol right - of which Transposable is just an aspect. I saw this which maybe seems relevant https://peps.python.org/pep-0544/#self-types-in-protocols

qiskit_nature/second_q/transformers/basis_transformer.py
```diff
@@ -0,0 +1,232 @@
...
+            self.initial_basis,
+            self.coefficients.transpose(),
+        )
+
+    def transform(self, problem: BaseProblem) -> BaseProblem:
+        if isinstance(problem, ElectronicStructureProblem):
+            return self._transform_electronic_structure_problem(problem)
+        else:
+            raise NotImplementedError(
+                f"The problem of type, {type(problem)}, is not yet supported by this transformer."
```
woodsp-ibm, line 119:
```suggestion
                f"The problem of type, {type(problem)}, is not supported by this transformer."
```

qiskit_nature/second_q/transformers/basis_transformer.py
```diff
@@ -0,0 +1,232 @@
...
+            )
+
```
woodsp-ibm, line 121: Looking at `invert` and `transform` would an `inverse_transform` method here too be convenient?

qiskit_nature/second_q/operators/fermionic_op.py
mrossinek: Yes indeed. I think there is another comment below the section I selected here which mentions that, once `sparse` is supported, the logic above will need to be adapted to also support that setting

qiskit_nature/second_q/operators/polynomial_tensor.py
mrossinek: Yes that's what is required for the same objects to be usable in a context like basis transforms to a reduced dimension.
I don't think it complicates things much. `numpy` will automatically raise errors when a user is trying to combine mismatching objects. But yes, it does require more cognitive awareness of the user who is doing such operations.

qiskit_nature/second_q/algorithms/initial_points/hf_initial_point.py
```diff
@@ -39,7 +40,7 @@ class HFInitialPoint(InitialPoint):
     def __init__(self) -> None:
         super().__init__()
         self._ansatz: UCC | None = None
-        self._reference_energy: float = 0.0
+        self._reference_energy: Number = 0.0  # type: ignore
```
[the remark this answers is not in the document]
mrossinek: I guess I can revert the type change here, but I recall in having to add `type: ignore` statements elsewhere, where I would extract the public API `reference_energy` value which is type `Number` and tried assigning it to this internal field that would then be `float`.
I'll try it again and see what happens, though :+1:

qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -159,28 +161,28 @@ def grouped_property(self, grouped_property: BaseProblem) -> None:
             )
 
         two_body_mo_integral: ElectronicIntegrals | None = (
-            electronic_energy.get_electronic_integral(ElectronicBasis.MO, 2)
+            electronic_energy.electronic_integrals.two_body
         )
-        if two_body_mo_integral is None:
+        if two_body_mo_integral is None or two_body_mo_integral.alpha.is_empty():
```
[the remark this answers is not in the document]
mrossinek: Yes you're right! I think it was returning an optional set in the past and I forgot to update things here...

qiskit_nature/second_q/algorithms/initial_points/hf_initial_point.py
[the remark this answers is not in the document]
mrossinek: Looking at implementing this change now, I notice that `self._reference_energy` is actually exposed in the public API via `total_energy` which also switched to `Number` to be more general and match the `ElectronicStructureProblem.reference_energy` type

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -0,0 +1,482 @@
...
+    def beta(self) -> PolynomialTensor:
+        """The down-spin electronic integrals."""
+        return self._beta
+
+    @beta.setter
+    def beta(self, beta: PolynomialTensor | None) -> None:
+        self._beta = beta if beta is not None else PolynomialTensor.empty()
+
+    @property
+    def mixed(self) -> PolynomialTensor:
```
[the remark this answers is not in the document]
mrossinek: Yes, I see your point :thinking: I wanted to refrain from using `beta_alpha` for the name of just `mixed`, since `mixed` in theory could also support more than two-spin interactions (although the logic for that would obviously need to be added).
I will think about how to resolve this ambiguity. Do you think, this would suffice in being documented more properly?

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -0,0 +1,482 @@
...
+            return False
+
+        if not self.mixed.equiv(other.mixed):
+            return False
+
+        return True
+
+    def _multiply(self, other: complex) -> ElectronicIntegrals:
+        return ElectronicIntegrals(
+            cast(PolynomialTensor, other * self.alpha),
```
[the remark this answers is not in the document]
mrossinek: I added the type check :+1: 

As per the `cast`: mypy is not able to detect that `PolynomialTensor` has a `__mul__`/`__rmul__` implementation.. I am not sure if that is because of how the Mixins work or whether mypy just does not bother checking for such a case and defaults to assuming that `number * something` must return a `number`..

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -0,0 +1,482 @@
...
+            # We can only perform this operation, when all mixed tensors are non-empty.
+            mixed = PolynomialTensor.einsum(einsum_map, *(op.mixed for op in operands))
+        return ElectronicIntegrals(alpha, beta, mixed)
+
+    # pylint: disable=invalid-name
+    @classmethod
+    def from_raw_integrals(
+        cls,
+        h1: np.ndarray,
+        h2: np.ndarray | None = None,
```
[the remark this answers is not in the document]
mrossinek: Ah the lack of these suffixes is likely an artifact from when these were overloaded to also allow SO-basis matrices..
Now that they no longer do that (because then a user would construct a `PolynomialTensor` directly (something I should add to the documentation)) I can add those suffixes to be consistent :+1:

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -127,32 +163,27 @@ def _add(self, other: PolynomialTensor, qargs=None) -> PolynomialTensor:
...
-
         sum_dict = {key: value + other._data.get(key, 0) for key, value in self._data.items()}
         other_unique = {key: other._data[key] for key in other._data.keys() - self._data.keys()}
         sum_dict.update(other_unique)
 
-        return PolynomialTensor(sum_dict, self._register_length)
+        register_length = None
+        if self.register_length is not None and other.register_length is not None:
+            register_length = max(self.register_length, other.register_length)
+
```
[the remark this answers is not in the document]
mrossinek: Right, I assume we would want it to stay `None` when both are, but take the respective value whenever one of them is not `None`.. :thinking:

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -72,21 +75,52 @@ def register_length(self) -> int:
         """Returns the register length"""
         return self._register_length
 
+    @register_length.setter
+    def register_length(self, reg_length: int | None) -> None:
+        self._register_length = reg_length
+
```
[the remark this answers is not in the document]
mrossinek: Yeh some validation could be done I suppose. In general, setting something smaller will not result in errors, but result in a truncated set of coefficients getting extracted in e.g. `FermionicOp.from_polynomial_tensor` (depending on its implementation we need to decide on).
Setting it to something larger will lead to indexing errors though, which we ought to avoid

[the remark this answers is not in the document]
mrossinek: What do you think we should do here? Should we just check that it matches? Resizing the matrices seems to be too expensive of an operation to be done here

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -98,7 +132,7 @@ def zero(cls, register_length: int) -> SparseLabelOp:
         return cls({}, register_length=register_length, copy=False)
 
     @classmethod
-    def one(cls, register_length: int) -> SparseLabelOp:
+    def one(cls, register_length: int | None = None) -> SparseLabelOp:
```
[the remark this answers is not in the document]
mrossinek: Yes, the other PR is now bumping the version requirement anyways. When I wrote that, it had not been merged to `main` of Terra yet. I can change this now :+1:

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -347,6 +379,35 @@ def sort(self, *, weight: bool = False) -> SparseLabelOp:
             {ind: self[ind] for ind in indices}, register_length=self.register_length, copy=False
         )
 
+    def chop(self, tol: float | None = None) -> SparseLabelOp:
```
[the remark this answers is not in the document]
mrossinek: I'd like to keep things consistent and perform everything out-of-place. This was an inconsistency especially noticeable in the old `qiskit_nature.properties` framework, where some methods were in- and some out-of-place operations, with no clear indicators given a methods name.

I am open to add a `round` method if we find a usecase for it. For now, I would refrain from adding more methods though

qiskit_nature/second_q/properties/angular_momentum.py
```diff
@@ -161,10 +157,11 @@ def second_q_ops(self) -> dict[str, FermionicOp]:
...
         h_2 = x_h2 + y_h2 + z_h2
 
-        h1_ints = OneBodyElectronicIntegrals(ElectronicBasis.SO, h_1)
-        h2_ints = TwoBodyElectronicIntegrals(ElectronicBasis.SO, h_2)
+        tensor = PolynomialTensor(
+            {"+-": h_1, "++--": _chem_to_phys(h_2)}, register_length=self._num_spin_orbitals
+        )
 
-        op = (h1_ints.to_second_q_op() + h2_ints.to_second_q_op()).simplify()
+        op = FermionicOp.from_polynomial_tensor(tensor).simplify()
```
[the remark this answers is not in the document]
mrossinek: Especially `simplify` is one method I would like to avoid of making an in-place operation, since it is a lossy conversion.
If you think this is really needed, we could have an in-place and an out-of-place equivalent method.

qiskit_nature/second_q/properties/dipole_moment.py
```diff
@@ -33,210 +29,85 @@
...
+    .. math::
+        \hat{d} = \sum_{p, q} d^x_{pq} a^\dagger_p a_q ,
+
+    where :math:`x` can be any of the three Cartesian axes (:math:`\{x, y, z\}`).
+
+    Just like in the :class:`qiskit_nature.second_q.hamiltonians.ElectronicEnergy`, the nuclear
+    contribution is stored **separately** in the :attr:`nuclear_dipole_moment` attribute and will
+    **not** be included into the generated operator. It is however possible, to manually add this
+    constant shift to the electronic dipole components as a constant term in the internal
+    :class:`qiskit_nature.second_q.operators.ElectronicIntegrals` instances. Refer to the example
```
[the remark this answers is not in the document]
mrossinek: Well you would need to add the `x`-axis offset to the `x`-axis dipole integrals, etc.
I can make this example explicit if you prefer

qiskit_nature/second_q/properties/dipole_moment.py
mrossinek: So are you saying it should actually be type `Number`, dropping the `Optional` since the values will always be numbers now (although possible `NaN`-valued)?

qiskit_nature/second_q/transformers/active_space_transformer.py
mrossinek: Yes, this is ambiguous w.r.t. AO/MO/NAO/whatever one comes up with.
One spatial orbital is then split into two spin orbitals for the up- and down-spin, respectively.

qiskit_nature/second_q/transformers/basis_transformer.py
mrossinek: Do you mean something like this?
```python
def inverse_transform(self, problem):
    inv_trafo = self.invert()
    return inv_trafo.transform(problem)
```

qiskit_nature/second_q/operators/polynomial_tensor.py
mrossinek: Reading your comment again, let me revise my previous answer:
I do not want to store matrices of different size, I just would like to support non-square matrices. For example:
```python
ao2mo = np.zeros((4, 2))
ao2mo[...] = ...
# this 4x2 matrix, maps from 4 AOs to 2 MOs
tensor = PolynomialTensor({"+-": ao2mo}, validate=False)
```

qiskit_nature/second_q/operators/electronic_integrals.py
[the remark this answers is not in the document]
mrossinek: `mypy` is also not clever enough to infer that `self.alpha * complex` is of type `PolynomialTensor`. It still thinks it must be complex...

qiskit_nature/second_q/properties/angular_momentum.py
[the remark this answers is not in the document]
mrossinek: I have opened #860 to track this :+1:

woodsp-ibm (approved):

PR #843, opened 2022-09-24
Pin sphinx < 5.2.0 & pyscf < 2.1.1

manoelmarques:
### Summary
Latest version of Sphinx 5.2.0 is failing documentation
Latest version of Pyscf 2.1.1 is crashing MacOS python 3.8 unit test. I suspect it is a new requirement `libomp.dylib`. I had to install it locally to make new pyscf work and I suspect the CI `libomp.dylib` may be causing the crash. It needs more investigation. Pin for now

Also update intersphinx_mapping with latest urls.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -209,8 +209,8 @@ html_theme_options = {
 
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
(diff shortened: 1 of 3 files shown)


mrossinek (approved):
Thanks, Manoel!

PR #858, opened 2022-10-04
Added `commutators`

### LINKED ISSUE #848: Implement `commutator` routines for the `SparseLabelOp`
### What should we add?

It would be useful to have (`anti-`/`double-`)`commutator` routines for `SparseLabelOp` instances, similar to those available in [`qiskit.opflow`](https://github.com/Qiskit/qiskit-terra/blob/672e95d089ba0e834915796f2b1d59527f3b4ef2/qiskit/opflow/utils.py#L18).

The implementation is likely almost identical (calling `.simplify()` instead of `.reduce()`) and should be placed into `qiskit_nature.second_q.operators.commutators`. Unittests should be added accordingly.

javabster:
### Summary
fixes #848


javabster:
Some things I'm unsure about:
- documentation - i basically followed the docs structure that was used for `tensor_ordering` in the same folder, but not sure if they should fall under the same category
- currently facing testing issue that my `DummySparseLabelOp` doesn't like the `@` operator, not sure if this should be swapped for a different operator perhaps? I don't know the specifics of the `@` operator well or what alternatives are best so any suggestions appreciated!

mrossinek:
> * documentation - i basically followed the docs structure that was used for `tensor_ordering` in the same folder, but not sure if they should fall under the same category

All under the same category is fine, yes :+1: 

<hr/>

One other suggestions which I wanted to re-raise based on [this comment](https://github.com/Qiskit/qiskit-nature/issues/129#issuecomment-830036390) is the addition of a `SparseLabelOp.is_zero()` method.
This will look similar to the new [`PolynomialTensor.is_empty`](https://github.com/Qiskit/qiskit-nature/blob/ce3e8a58dfd0306ccc41ddc9ff7b78050999cc62/qiskit_nature/second_q/operators/polynomial_tensor.py#L104) check which I recently added. It can do that same check first, but if the length is not 0, this check would also need to call `simplify()` and then check whether all coefficients are zero or not.

If you want to include such a method as part of this PR that would be amazing :slightly_smiling_face:

qiskit_nature/second_q/operators/commutators.py
```diff
@@ -0,0 +1,107 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Commutator Utilities (:mod:`qiskit_nature.second_q.operators.commutators`)
+===================================================================================
+
+.. currentmodule:: qiskit_nature.second_q.operators.commutator
```
mrossinek, line 17:
```suggestion
.. currentmodule:: qiskit_nature.second_q.operators.commutators
```

qiskit_nature/second_q/operators/commutators.py
```diff
@@ -0,0 +1,107 @@
...
+
+Utility functions to generate compute commutators for
+:class:`qiskit_nature.second_q.operators.sparse_label_op`.
```
mrossinek, line 20:
```suggestion
Utility functions to compute commutators of
:class:`qiskit_nature.second_q.operators.SparseLabelOp` instances.
```

qiskit_nature/second_q/operators/commutators.py
```diff
@@ -0,0 +1,107 @@
...
+    r"""
+    Compute commutator of `op_a` and `op_b`.
+    .. math::
+        AB - BA.
+    Args:
+        op_a: Operator A
+        op_b: Operator B
+    Returns:
+        SparseLabelOp: the commutator
+    """
```
mrossinek, line 44:
```suggestion
    r"""Compute commutator of `op_a` and `op_b`.

    .. math::

        AB - BA.

    Args:
        op_a: Operator A
        op_b: Operator B

    Returns:
        SparseLabelOp: the commutator
    """
```

mrossinek: Similar reformatting is required in the other docstrings

mrossinek: And you will need to add exceptions to the `.pylintdict` file to ignore the spelling of `AB`, etc.
Please add those exceptions in all lowercase and at the correct alphabetical position in that file

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -444,3 +444,10 @@ def induced_norm(self, order: int = 1) -> float:
             https://en.wikipedia.org/wiki/Norm_(mathematics)#p-norm
         """
         return sum(abs(coeff) ** order for coeff in self.values()) ** (1 / order)
+
+    def is_zero(self) -> bool:
+        """Returns true if operator length is zero or all co-efficients have value zero"""
+        if len(self) == 0:
+            return True
+        else:
+            return all(val == 0 for val in self._data.values())
```
mrossinek, line 453: Maybe we could add a `tol` keyword argument here which allows to specify a tolerance?
If it is `None`, the current implementation is used, otherwise you could use `np.isclose` instead of `==` for the check.

Please also add a short unittest which tests this method

test/second_q/operators/test_commutators.py
```diff
@@ -0,0 +1,53 @@
...
+op2 = FermionicOp({"-_0 +_0": 2}, register_length=1)
+op3 = FermionicOp({"+_0 -_0": 1, "-_0 +_0": 2 + 0.5j}, register_length=1)
+
+
+class TestCommutators(QiskitNatureTestCase):
+    """Commutators tests."""
+
+    def test_commutator(self):
+        """Test commutator method"""
+        self.assertEqual(commutator(op1, op2), FermionicOp({}, register_length=1))
```
mrossinek, line 37: You could also add some non-empty test cases like e.g. the following:
```python
>>> from qiskit_nature.second_q.operators.commutators import commutator, anti_commutator
>>> from qiskit_nature.second_q.operators import FermionicOp
>>> op1 = FermionicOp({"+_0": 1})
>>> op2 = FermionicOp({"-_0": 1})
>>> commutator(op1, op2)
FermionicOp({'+_0 -_0': (1+0j), '-_0 +_0': (-1+0j)}, register_length=1, )
>>> anti_commutator(op1, op2)
FermionicOp({'+_0 -_0': (1+0j), '-_0 +_0': (1+0j)}, register_length=1, )
```

mrossinek (commented):
You will also need to add the file `docs/apidocs/qiskit_nature.second_q.operators.commutators.rst` with the following content:
```rst
commutators
===============================================

.. automodule:: qiskit_nature.second_q.operators.commutators
   :no-members:
   :no-inherited-members:
   :no-special-members:
```

qiskit_nature/second_q/operators/sparse_label_op.py
javabster: sure, do we need `atol`, `rtol` or both?

mrossinek: `atol` is sufficient here. Since we are comparing to 0.0 the relative tolerance has no effect.
`np.isclose` uses the following equation:
```
absolute(a - b) <= (atol + rtol * absolute(b))
```
where `b` is the reference value (i.e. 0.0 in our case)

qiskit_nature/second_q/operators/commutators.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def double_commutator(
+    op_a: SparseLabelOp,
+    op_b: SparseLabelOp,
+    op_c: SparseLabelOp,
+    sign: bool = False,
+) -> SparseLabelOp:
+    r"""Compute symmetric double commutator of `op_a`, `op_b` and `op_c`.
+    See McWeeny chapter 13.6 Equation of motion methods (page 479)
```
mrossinek, line 74: I am looking into making this resource more clear. I myself did not find it..

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -444,3 +444,19 @@ def induced_norm(self, order: int = 1) -> float:
             https://en.wikipedia.org/wiki/Norm_(mathematics)#p-norm
         """
         return sum(abs(coeff) ** order for coeff in self.values()) ** (1 / order)
+
+    def is_zero(self, tol: int = None) -> bool:
```
mrossinek, line 448:
```suggestion
    def is_zero(self, tol: int | None = None) -> bool:
```

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -444,3 +444,19 @@ def induced_norm(self, order: int = 1) -> float:
...
+        r"""Returns true if operator length is zero or all coefficients have value zero
+
+        Args:
+            tol: tolerance for checking coefficient values
```
mrossinek, line 452:
```suggestion
            tol: tolerance for checking coefficient values. If this is `None`, :attr:`atol` will be used instead.
```

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -444,3 +444,19 @@ def induced_norm(self, order: int = 1) -> float:
...
+
+        Returns:
+            if operator length is zero or all coefficients are zero
+        """
+        if len(self) == 0:
+            return True
+        elif tol is not None:
+            return all(np.isclose(val, 0, atol=tol) for val in self._data.values())
+        else:
+            return all(val == 0 for val in self._data.values())
```
mrossinek, line 462:
```suggestion
        tol = tol if tol is not None else self.atol
        return all(np.isclose(val, 0, atol=tol) for val in self._data.values())
```
One can still basically achieve an identity check by setting the tol to something like `1e-18`. The benefit here, is that we do not do equality comparison with an integer value.
This matches what was done in the `.chop()` method :+1:

qiskit_nature/second_q/operators/commutators.py
mrossinek:
```suggestion

    See also Equation (13.6.18) in [1].
```

qiskit_nature/second_q/operators/commutators.py
```diff
@@ -0,0 +1,116 @@
...
+         = (2ABC - 2CBA - BAC + CAB - ACB + BCA)/2.
+
+    Args:
+        op_a: Operator A
+        op_b: Operator B
+        op_c: Operator C
+        sign: False anti-commutes, True commutes
+
+    Returns:
+        SparseLabelOp: the double commutator
```
mrossinek, line 94:
```suggestion
        SparseLabelOp: the double commutator
        
        References:
            [1]: R. McWeeny.
                Methods of Molecular Quantum Mechanics.
                2nd Edition, Academic Press, 1992.
                ISBN 0-12-486552-6.
```

test/second_q/operators/test_commutators.py
```diff
@@ -0,0 +1,80 @@
...
+        expected: dict,
+    ):
+        """Test double commutator method"""
+        self.assertEqual(
+            double_commutator(op_a, op_b, op_c), FermionicOp(expected, register_length=1)
+        )
+
+    def test_commutator_iszero(self):
+        """Test is_zero function with commutator"""
+        self.assertTrue(commutator(op1, op2).is_zero())
```
mrossinek, line 76: I don't think we need this here, since that is covered by the `is_zero` tests on `SparseLabelOp` :+1:

test/second_q/operators/test_commutators.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+    def test_anti_commutator(self, op_a: FermionicOp, op_b: FermionicOp, expected: dict):
+        """Test anti commutator method"""
+        self.assertEqual(anti_commutator(op_a, op_b), FermionicOp(expected, register_length=1))
+
+    @unpack
+    @data(
+        (op1, op2, op3, {}),
+        (op1, op4, op3, {"+_0": (1 + 0.5j)}),
+    )
```
mrossinek, line 61: Please also add the `sign` as a parameter here and check the output when you change it

test/second_q/operators/test_sparse_label_op.py
```diff
@@ -325,6 +325,33 @@ def test_one(self):
...
+    def test_is_zero_tol(self):
+        """test if coefficients are all zero with tolerance"""
+        test_op = DummySparseLabelOp(
+            {
+                "+_0 -_1": 0.05,
+                "+_0 -_3": 0.0,
+            },
+            2,
+        )
+        self.assertTrue(test_op.is_zero(tol=0.1))
```
mrossinek, line 353: You can combine all these into one method and then use `with self.subTest(...)` to match how the other tests are done in this file

mrossinek (commented):
This is looking good already! I just have a few comments :+1:

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -444,3 +444,18 @@ def induced_norm(self, order: int = 1) -> float:
             https://en.wikipedia.org/wiki/Norm_(mathematics)#p-norm
         """
         return sum(abs(coeff) ** order for coeff in self.values()) ** (1 / order)
+
+    def is_zero(self, tol: int | None = None) -> bool:
+        r"""Returns true if operator length is zero or all coefficients have value zero
+
+        Args:
+            tol: tolerance for checking coefficient values. If this is `None`,
+            :attr:`atol` will be used instead.
```
woodsp-ibm, line 453:
```suggestion
                :attr:`atol` will be used instead.
```
Continuation of arg text needs indenting

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -465,3 +465,18 @@ def induced_norm(self, order: int = 1) -> float:
...
+
+    def is_zero(self, tol: int | None = None) -> bool:
+        r"""Returns true if operator length is zero or all coefficients have value zero
+
+        Args:
+            tol: tolerance for checking coefficient values. If this is `None`,
+                :attr:`atol` will be used instead.
+
+        Returns:
+            if operator length is zero or all coefficients are zero
```
mrossinek, line 477:
```suggestion
            If operator length is zero or all coefficients are zero.
```

test/second_q/operators/test_sparse_label_op.py
```diff
@@ -312,6 +312,30 @@ def test_one(self):
...
+            self.assertTrue(test_op.is_zero())
+
+        with self.subTest("coefficients are all zero with tol"):
+            test_op = DummySparseLabelOp(
+                {
+                    "+_0 -_1": 0.05,
+                    "+_0 -_3": 0.0,
+                }
+            )
+            self.assertTrue(test_op.is_zero(tol=0.1))
```
mrossinek, line 337: We should have one more test to make sure that `is_zero()` can also return `False` when required.
Ideally, you use the same operator as in this last test and do
```python
self.assertFalse(test_op.is_zero())
```
This also checks that the `tol` keyword is working as expected.

test/second_q/operators/test_sparse_label_op.py
```diff
@@ -312,6 +312,30 @@ def test_one(self):
...
+            self.assertTrue(test_op.is_zero())
+
+        with self.subTest("coefficients are all zero"):
+            test_op = DummySparseLabelOp(
+                {
+                    "+_0 -_1": 0.0,
+                    "+_0 -_3": 0.0,
+                }
+            )
+            self.assertTrue(test_op.is_zero())
```
mrossinek, line 328: Maybe you could also add one more test which checks that `is_zero()` is still `True`, when a value is e.g. `1e-18` (i.e. below the value of the `atol` attribute).

test/second_q/operators/test_sparse_label_op.py
```diff
@@ -312,6 +312,57 @@ def test_one(self):
...
+            self.assertTrue(test_op.is_zero(tol=0.1))
+
+        with self.subTest("coefficients are all zero with smaller val"):
+            test_op = DummySparseLabelOp(
+                {
+                    "+_0 -_1": 0.05,
+                    "+_0 -_3": 1e-18,
+                }
+            )
+            self.assertTrue(test_op.is_zero(tol=0.1))
```
mrossinek, line 346: I actually meant to test this case instead:
```suggestion
        with self.subTest("coefficients are all zero with smaller val"):
            test_op = DummySparseLabelOp(
                {
                    "+_0 -_1": 0.0,
                    "+_0 -_3": 1e-18,
                }
            )
            self.assertTrue(test_op.is_zero())
```
Otherwise this was just a duplicate of the test above

mrossinek (approved):
This is looking good, thanks, Abby! :slightly_smiling_face:

PR #862, opened 2022-10-07
feat: implement the ElectronicDensity property

### LINKED ISSUE #174: Charge density
Is there anyway to extract the charge density from the electronic ground state from VQE run ?

mrossinek:
### Summary

The `ElectronicDensity` property will be a key requirement for tighter integration with classical codes since it is the key piece of data that we need to return from the quantum processor to the classical code.

### Details and comments

Closes #174

### DIFF
--- a/qiskit_nature/second_q/operators/electronic_integrals.py
+++ b/qiskit_nature/second_q/operators/electronic_integrals.py
@@ -255,7 +255,7 @@ class ElectronicIntegrals(AdjointMixin, LinearMixin):
                 {"+-": self.beta["+-"]},
                 validate=False,
             )
-        return ElectronicIntegrals(alpha, beta)
+        return self.__class__(alpha, beta)
 
     @property
     def two_body(self) -> ElectronicIntegrals:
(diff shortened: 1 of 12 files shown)


qiskit_nature/properties/second_quantization/electronic/electronic_density.py
```diff
@@ -0,0 +1,209 @@
...
+    IntegralProperty,
+    OneBodyElectronicIntegrals,
+    TwoBodyElectronicIntegrals,
+)
+from .particle_number import ParticleNumber
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ElectronicDensity(IntegralProperty):
```
woodsp-ibm, line 39: Was this class added in error? The existing ones in this folder are being deprecated right.

mrossinek: Oh yes, this was done by mistake will migrating over from an earlier version I had offline :see_no_evil:

PR #868, opened 2022-10-09
Change from retworkx to rustworkx, pylint whitelist rustworkx

manoelmarques:
### Summary
Use rustworkx instead of retworkx
rustworkx is a compiled library, so pylint whitelist it.

### DIFF
--- a/qiskit_nature/problems/second_quantization/lattice/lattices/hyper_cubic_lattice.py
+++ b/qiskit_nature/problems/second_quantization/lattice/lattices/hyper_cubic_lattice.py
@@ -17,7 +17,7 @@ from math import pi
 from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
-from retworkx import PyGraph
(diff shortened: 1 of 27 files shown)


woodsp-ibm (dismissed):

qiskit_nature/problems/second_quantization/lattice/lattices/hyper_cubic_lattice.py
```diff
@@ -17,7 +17,7 @@
 from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
-from retworkx import PyGraph
```
mtreinish, line 20: You probably should put rustworkx in the pylintrc as a compiled module, so you don't need an explicit exclude on each import. pylint is unable to parse all the contents of rustworkx because it's compiled.  We should already have retworkx in there already.

woodsp-ibm (approved):

PR #870, opened 2022-10-10
Sparse `PolynomialTensor`

### LINKED ISSUE #866: Support `sparse` matrices in `PolynomialTensor`
### What should we add?

#668 will be **VERY** inefficient, unless the `PolynomialTensor` class supports `sparse` matrices (which we had already discussed but not formally tracked yet).
Since scipy-sparse matrices only support 2D arrays, we will rely on https://sparse.pydata.org/en/stable/#, the same library also used in the [Qiskit Machine Learning](https://github.com/Qiskit/qiskit-machine-learning/blob/51a91785cdf7e68ed73703f963141e7d2516fbbf/setup.py#L73) module.

mrossinek:
### Summary

This adds support for `sparse` arrays to be used inside of the `PolynomialTensor` class.
It also adds/improves the documentation of that class.

### Details and comments

Closes #866


requirements.txt
```diff
@@ -7,3 +7,4 @@ setuptools>=40.1.0
 typing_extensions
 h5py
 rustworkx
+sparse
```
mrossinek, line 10: I know that in Qiskit Machine Learning this is handled as an optional dependency. Do we really need to do the same here? @woodsp-ibm @manoelmarques

mrossinek:
The current CI failure is unrelated and will soon be fixed by #871

requirements.txt
woodsp-ibm: The underlying reason that prompted this being made optional in ML is described in Qiskit/qiskit-machine-learning#67 where at the time users wanted to use a platform where numba, a dependency of sparse, was not supported. It makes the source a bit more complex to cater to this being optional - you can see what was done in ML in the PR that resolved the issue. If the impact is not too bad we may want to consider optionality to have flexibility though still in the near term.

mrossinek: Hm I see. I'm not sure whether Nature will ever be used on a platform like a Raspberry Pi. But I guess we should try to avoid making the `qiskit` metapackage dependent on numba.

woodsp-ibm: The platform in question is now supported by numba. You can see how the sparse optionality is handled in ML and what a similar approach might entail.

mrossinek: If the platform is now supported, is there any other reason for making `sparse` an optional dependency? I would like to avoid making `sparse` optional since it will (presumably) be more and more useful to us when ironing out bottlenecks in the future.
If however there is a very good reason to make it optional, I guess we need to add the same `optionals.HAS_SPARSE` setting in Nature as we have in ML.

woodsp-ibm: It did not start out as optional in ML but was changed since it would fail to install where numba did not support the platform even if you had no intent/need of using sparse. If the impact seems low to the code it might be worth considering having it optional - e.g. Terra has just made tweedledum optional due to platform issues where it was not before since it ended up preventing installation/use.

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -105,58 +246,97 @@ def is_empty(self) -> bool:
         """Returns whether this tensor is empty or not."""
         return len(self) == 0
 
-    def __getitem__(self, __k: str) -> (np.ndarray | Number):
-        """
-        Returns value matrix in the `PolynomialTensor`.
+    @_optionals.HAS_SPARSE.require_in_call
+    def contains_sparse(self) -> bool:
```
woodsp-ibm, line 250: Would `issparse` and `isdense` methods make sense that check for all being an instance of SparseArray, or all not being respectively. These might be used internally below too perhaps in todense and tosparse to return self instead of making a copy if it were already all dense matrices or all sparse accordingly?

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -105,58 +246,97 @@ def is_empty(self) -> bool:
...
     def __iter__(self) -> Iterator[str]:
-        """
-        Returns iterator of the `PolynomialTensor`.
-        """
+        """Returns an iterator of the ``PolynomialTensor``."""
         return self._data.__iter__()
 
+    @_optionals.HAS_SPARSE.require_in_call
+    def todense(self) -> PolynomialTensor:
+        """Converts all internal matrices to dense numpy arrays."""
```
woodsp-ibm, line 275: Could we say for this and `tosparse` below more like "Returns a new instance where all matrices are now dense numpy array" to make it clear its not an inplace conversion.

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -105,58 +246,97 @@ def is_empty(self) -> bool:
...
+            if isinstance(value, SparseArray):
+                dense_dict[key] = value.todense()
+            else:
+                dense_dict[key] = value
+        return PolynomialTensor(dense_dict, validate=False)
+
+    # TODO: change the following type-hint if/when SparseArray dictates the existence of from_numpy
+    @_optionals.HAS_SPARSE.require_in_call
+    def tosparse(
+        self, *, sparse_type: Type[COO] | Type[DOK] | Type[GCXS] = COO
```
woodsp-ibm, line 287: Maybe this can be made clear in the doctring. This only changes any dense matrices to the given sparse type. Any sparse matrices already existing are unaffected - so you cannot use this for sparse conversions among those types.

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -167,11 +347,13 @@ def _add(self, other: PolynomialTensor, qargs=None) -> PolynomialTensor:
 
         return PolynomialTensor(sum_dict, validate=False)
 
+    # pylint: disable=too-many-return-statements
     def __eq__(self, other: object) -> bool:
-        """Check equality of first PolynomialTensor with other
+        """Check equality of ``PolynomialTensor`` instances.
```
woodsp-ibm, line 352: Just a comment - here equality is based on values of the elements of the matrices - the representation (sparse or dense) does not matter,

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -368,6 +601,11 @@ def einsum(
             )
             # results will be contained in transformed["+-"] and transformed["++--"], respectively
 
+        .. note::
+
+           :class:`sparse.SparseArray` does not support ``numpy.einsum``. Thus, all sparse matrices
+           involved in this operation will be converted to dense matrices!
```
woodsp-ibm, line 607: > Thus, all sparse matrices involved in this operation will be converted to dense matrices!

Or said another way the resultant PolynomialTensor will be all dense matrices. So if a user wanted sparse as result they would have to call tosparse afterwards. Converting the matrix back to sparse might be possible here I guess on the fly.

woodsp-ibm (approved):

PR #874, opened 2022-10-11
refactor: handle subclass keyword arguments of SparseLabelOp

mrossinek:
### Summary

This enables the handling of non-default keyword arguments in subclasses of the SparseLabelOp (for example the spin in SpinOp and num_modes and num_modals in VibrationalOp). To do so, it even removes the register_length argument from the base class in favor of the dedicated num_spin_orbitals argument on the FermionicOp subclass.

### Details and comments

This PR also adds some more documentation details to the `SparseLabelOp` base class.


qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -121,28 +151,22 @@ def _validate_polynomial_tensor_key(cls, keys: Collection[str]) -> None:
...
         """Constructs a zero-operator.
 
-        Args:
-            register_length: the length of the operator.
-
         Returns:
             The zero-operator of the given length.
         """
-        return cls({}, register_length=register_length, copy=False)
+        return cls({}, copy=False)
```
woodsp-ibm, line 160: Is this always ok, for subclasses (same for `one` below) - this suggests all other parameters they might have are defaulted for all subclasses - we could state that as a requirement perhaps.

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -64,33 +75,52 @@ def __init__(
...
+    def register_length(self) -> int | None:
         """Returns the register length"""
-        return self._register_length
 
-    @register_length.setter
-    def register_length(self, reg_length: int | None) -> None:
-        self._validate_keys(self._data.keys(), reg_length)
-        self._register_length = reg_length
+    # pylint: disable=unused-argument
+    def _new_instance(
```
woodsp-ibm, line 89: Should we make this abstract so as to have a sub-class implement this. Its the aspect we expect to change right?

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -166,9 +190,7 @@ def _add(self, other: SparseLabelOp, qargs: None = None) -> SparseLabelOp:
         other_unique = {key: other._data[key] for key in other._data.keys() - self._data.keys()}
         new_data.update(other_unique)
 
-        register_length = max(self.register_length, other.register_length)
-
-        return self.__class__(new_data, register_length, copy=False)
+        return self._new_instance(new_data, other=other)
```
woodsp-ibm, line 193: I cannot comment on this above directly 
```
        if not isinstance(other, SparseLabelOp):
            raise ValueError(
                f"Unsupported operand type(s) for +: 'SparseLabelOp' and '{type(other).__name__}'"
            )
```
But this check would/should depend on sub-class type which may  only be able to add like classes. Should we change this instance test as well as the hard coded SparseLabelOp in the message, since I I try to add something not supported for FermionicOp it would be nice to see that name in the message not some parent class name


Also a comment on equiv and eq below since I cannot comment there directly either. They check other and if not a SparseLableOp they will return False. So its not a check involving subclasses eg FermionicOp to FermionicOp which will pass fine though, but so would FermionicOp and SpinOp in this case. I imagine the latter would most likely always fail anyway since `data` would never be the same - even if in the unlikely case the matrices were the same values the keys would be different. If we are adjusting the add instance checking though maybe we want to consider it in these instances too.

qiskit_nature/second_q/operators/sparse_label_op.py
mrossinek: I don't see a better way honestly :thinking:
I think if a subclass does not support this, it will need to overwrite this method.

qiskit_nature/second_q/operators/sparse_label_op.py
mrossinek: I guess we can do that, yes :+1:

qiskit_nature/second_q/operators/sparse_label_op.py
mrossinek: Yep thanks for noticing :+1:

test/second_q/operators/test_sparse_label_op.py
```diff
@@ -45,9 +45,17 @@
...
     """Dummy SparseLabelOp for testing purposes"""
 
-    @classmethod
-    def _validate_keys(cls, keys: Collection[str], register_length: int | None) -> int:
-        return register_length
+    @property
+    def register_length(self) -> int | None:
+        return None
+
+    def _new_instance(
```
woodsp-ibm, line 52:
```suggestion
    # pylint: disable=unused-argument
    def _new_instance(
```

PR #878, opened 2022-10-13
Detailed sparse library check

manoelmarques:
### Summary
Avoid making the whole method fail if `sparse` is not installed, instead check inside the method,

### DIFF
--- a/qiskit_nature/second_q/operators/polynomial_tensor.py
+++ b/qiskit_nature/second_q/operators/polynomial_tensor.py
@@ -274,7 +274,6 @@ class PolynomialTensor(LinearMixin, AdjointMixin, GroupMixin, TolerancesMixin, M
         """Returns an iterator of the ``PolynomialTensor``."""
         return self._data.__iter__()
 
-    @_optionals.HAS_SPARSE.require_in_call
     def todense(self) -> PolynomialTensor:
         """Returns a new instance where all matrices are now dense numpy arrays.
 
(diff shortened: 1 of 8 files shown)


qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -284,6 +283,7 @@ def todense(self) -> PolynomialTensor:
         if self.isdense():
             return self
 
+        _optionals.HAS_SPARSE.require_now("SparseArray")
```
mrossinek, line 286: Could we even go as far as to change the `isinstance` check to `not isinstance(value, np.ndarray)` and in that case map it?
That would avoid relying on sparse in this matter altogether.

I guess the respective change could also be done in `tosparse` :thinking:

mrossinek (commented):
Thank you, Manoel! Just one minor suggestion from my side :+1:

qiskit_nature/second_q/operators/polynomial_tensor.py
manoelmarques: The `tosparse` from the start already needs sparse just for having the `sparse_type` parameter which should be coming from the sparse library. Also, it calls `issparse` which needs `sparse`.
As for the `todense`, if you remove the explicit `SparseArray` test, you still need to call `todense` which is a method from sparse right ? without the `HAS_SPARSE` check it will raise a standard exception which we are trying to avoid by raising our own custom exception.

mrossinek (approved):
Thanks for clarifying Manoel! :+1:

PR #880, opened 2022-10-14
Finalize electronic structure stack

### LINKED ISSUE #820: Ill-defined behavior of stack when using odd `num_spin_orbitals`

### LINKED ISSUE #851: Finalize `ElectronicStructureProblem` refactoring

#841 goes a long way in refactoring the electronic structure stack. However, there are still some aspects left for another PR which I would like to summarize in this issue:

I would still like to apply the following changes:
- remove the `name` attribute of the `Property` base class (the `PropertiesContainer` handles this consistently via `self.__class__.__name__`
- remove the legacy HDF5 integration from those objects
- properly document that a `Property` at that point just means `SparseLabelOpFactory`
- remove remaining data currently stored on property classes in favor of direct storage on the `ElectronicStructureProblem`:
  - `ElectronicStructureProblem.num_particles: int | tuple[int, int]`
  - `ElectronicStructureProblem.num_spatial_orbitals: int | tuple[int, int]` (generalizes `num_spin_orbitals` to enable non-equal orbital numbers per spin species; we can keep the `num_spin_orbitals` property to return the sum of the spatial ones or twice its value when its an integer)
  - `ElectronicStructureProblem.orbital_occupations`
  - `ElectronicStructureProblem.orbital_occupations_b`
- change `AngularMomentum`, `Magnetization` and `ParticleNumber` such that they are only given the `num_spatial_orbitals` information "shortly before" calling `second_q_ops()` (by extension, this will also fix #820)

mrossinek:
### Summary

This PR performs the final refactoring changes to the electronic structure stack before the 0.5 release.

Closes #851

### Details and comments

- removes the `Property` base class in favor of the `SparseLabelOpFactory` protocol
- actually makes use of the `Interpretable` protocol
- relocates data stored inside of the "properties" onto the `ElectronicStructureProblem`
- switches the entire stack from `num_spin_orbitals` to `num_spatial_orbitals` (Closes #820)
- aligns the signatures of `ansatzes` and `initial_states` (consistent argument ordering and now some keyword-only arguments)


qiskit_nature/second_q/circuit/library/ansatzes/ucc.py
```diff
@@ -150,12 +148,15 @@ def __init__(
...
                 :`Callable`: a function which is used to generate the excitations.
-                    The callable must take the __keyword__ arguments `num_spin_orbitals` and
+                    The callable must take the __keyword__ arguments `num_spatial_orbitals` and
                     `num_particles` (with identical types to those explained above) and must return
                     a `list[tuple[tuple[int, ...], tuple[int, ...]]]`. For more information on how
                     to write such a callable refer to the default method
                     :meth:`~qiskit_nature.second_q.circuit.library.ansatzes.utils.\
                     generate_fermionic_excitations`.
+            qubit_converter: the QubitConverter instance which takes care of mapping a
+                :class:`~.SecondQuantizedOp` to a :class:`PauliSumOp` as well as performing all
```
woodsp-ibm, line 158: Do you want to be explicit about conversion type here - would it not be better to say `to a qubit operator` otherwise when we drop PauliSumOp and go to SparsePauliOp directly we'll have to go through and change the text in these too.

qiskit_nature/second_q/circuit/library/initial_states/hartree_fock.py
```diff
@@ -28,13 +28,13 @@ class HartreeFock(QuantumCircuit):
 
     def __init__(
         self,
-        num_spin_orbitals: int,
+        num_spatial_orbitals: int,
```
woodsp-ibm, line 31: We had discussed making this a BlueprintCircuit so it could be updated later like UCC is in regards of orbitals etc. Maybe a separate issue could be raised so as to capture the thought.

qiskit_nature/second_q/problems/electronic_structure_problem.py
```diff
@@ -69,12 +83,23 @@ def filter_criterion_custom(eigenstate, eigenvalue, aux_values):
...
+                np.isclose(num_particles_aux, expected_num_electrons)
             )
 
         solver = NumPyEigensolverFactory(filter_criterion=filter_criterion_spin)
 
+    Attributes:
+        properties: a container for additional observable operator factories.
+        molecule: a container for molecular system data.
+        basis: the electronic basis of all contained orbital coefficients.
+        num_particles: the number of particles (in this case electrons) in the system. When this is
```
woodsp-ibm, line 96: Is there a reason to have int here rather than require it as a tuple. Elsewhere as in HF and UCC the same named item is of type tuple. If its convenient to take it as an int from some known input that defined elsewhere maybe have a property that always as a getter gives back a tuple to it plugs in elsewhere from the source here without having to check/adapt it

qiskit_nature/second_q/properties/protocols.py
```diff
@@ -0,0 +1,58 @@
...
+
+if TYPE_CHECKING:
+    from qiskit_nature.second_q.problems import EigenstateResult
+
+
+@runtime_checkable
+class SparseLabelOpFactory(Protocol):
+    """A protocol indicating :class:`qiskit_nature.second_q.operators.SparseLabelOp` generators."""
+
+    def second_q_ops(self) -> dict[str, SparseLabelOp]:
```
woodsp-ibm, line 37: Do we need specific `dict` type or would Mapping be better

qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -124,51 +122,47 @@ def ansatz(self, ansatz: UCC) -> None:
...
                 identical.
         """
-        return self._grouped_property
+        return self._problem
 
-    @grouped_property.setter
-    def grouped_property(self, grouped_property: BaseProblem) -> None:
-        if not isinstance(grouped_property, ElectronicStructureProblem):
+    @problem.setter
+    def problem(self, problem: BaseProblem) -> None:
```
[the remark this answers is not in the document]
mrossinek: Yes, the point is that those required attributes must also not be `None` on the problem instance provided to the initial point.
I updated the docstring slightly, hopefully improving things :+1: (will push in a minute)

qiskit_nature/second_q/circuit/library/initial_states/hartree_fock.py
mrossinek: Tracked by https://github.com/Qiskit/qiskit-nature/issues/881

qiskit_nature/second_q/problems/electronic_structure_problem.py
mrossinek: I have implemented a variant based on this. This now means that we have `num_particles` handled via a property getter/setter. How do you suggest we document this? Shall I include it in `Attributes:` anyways?
The question can actually be asked for `orbital_occupations`, too

woodsp-ibm: I would leave it documented via the property docstring - if you include it in Attributes its going to show up twice in the docs, no.

woodsp-ibm (dismissed):

woodsp-ibm (approved):

PR #884, opened 2022-10-15
fix num_spin_orbitals for FermionicOp with zero spin orbitals

kevinsung:
### Summary
Fixes a bug where `op = FermionicOp({"": 1.0})` was being assigned `num_spin_orbitals` equal to 1 instead of 0.


mrossinek (approved):
Thanks :+1:

PR #887, opened 2022-10-16
fix: implement correct PolynomialTensor arithmetics

### LINKED ISSUE #885: Incorrect PolynomialTensor arithmetics
### Environment

- **Qiskit Nature version**: `main`
- **Python version**: Any
- **Operating system**: Any

### What is happening?

Some arithmetics implemented by the `PolynomialTensor` class are incorrect.

These include:
- `adjoint`: which is actually not defined for this class. The simplest example is the `"+-+"` entry, where conjugation + transposing is insufficient, because one would actually need to translate the key `"-+-"` which can only be done by a class that has knowledge of key contents
- `compose`: see below
- `tensor`/`expand`: see below

```python
import numpy as np

from qiskit_nature.second_q.operators import FermionicOp, PolynomialTensor

a = np.arange(4).reshape((2, 2))
b = np.arange(16).reshape((2, 2, 2, 2))

pt = PolynomialTensor({"": 1.0, "+": np.array([0, 1]), "+-": a, "++--": b})
f = FermionicOp.from_polynomial_tensor(pt)

print("@", (f @ f).equiv(FermionicOp.from_polynomial_tensor(pt @ pt)))  # @ False
print("^", (f ^ f).equiv(FermionicOp.from_polynomial_tensor(pt ^ pt)))  # ^ False
```

### What should happen?

Both of those print statements in the previous snippet should print `True`.

For the case of `adjoint`, there is no solution other than removing the `AdjointMixin` from the `PolynomialTensor`.

### Any suggestions?

_No response_

mrossinek:
### Summary

Closes #885

### Details and comments

- removes the `AdjointMixin`
- implements the correct `compose` and `tensor` behavior and unittests it against the `FermionicOp`
- adds the `PolynomialTensor.apply` utility method


qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -504,6 +486,9 @@ def compose(
             front: If True composition uses right matrix multiplication, otherwise left
                 multiplication is used (the default).
 
+        Raises:
+            NotImplementedError: when the two tensors doenot have the same :attr:`register_length`.
```
woodsp-ibm, line 490:
```suggestion
            NotImplementedError: when the two tensors does not have the same :attr:`register_length`.
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -537,6 +530,9 @@ def tensor(self, other: PolynomialTensor) -> PolynomialTensor:
         Args:
             other: the other PolynomialTensor.
 
+        Raises:
+            NotImplementedError: when the two tensors doenot have the same :attr:`register_length`.
```
woodsp-ibm, line 534:
```suggestion
            NotImplementedError: when the two tensors do not have the same :attr:`register_length`.
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -556,6 +552,9 @@ def expand(self, other: PolynomialTensor) -> PolynomialTensor:
...
```
woodsp-ibm, line 556:
```suggestion
            NotImplementedError: when the two tensors do not have the same :attr:`register_length`.
```

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -138,11 +139,16 @@ class PolynomialTensor(LinearMixin, AdjointMixin, GroupMixin, TolerancesMixin, M
 
       print(tensor ^ tensor)
 
-    Adjoint
+    You can also implement more advanced arithmetics via the :meth:`apply` and :meth:`einsum`
```
woodsp-ibm, line 142:
```suggestion
    You can also implement more advanced arithmetic via the :meth:`apply` and :meth:`einsum`
```

woodsp-ibm (approved):

Issue #865, opened 2022-10-07
Update `QuadraticHamiltonian` based on `PolynomialTensor`


mrossinek:
### What is the expected enhancement?

#756 introduced the new `PolynomialTensor` class which serves as a generalization of the old `ElectronicIntegrals` which could not be used for the implementation of the `QuadraticHamiltonian` in #482.
The new `PolynomialTensor` should be more flexibel and prior to the actual release of 0.5, I would like to see its usefulness tested by someone other than me :smile: Thus I kindly ask @kevinsung to check whether the `QuadraticHamiltonian` can benefit from this class :+1:

PR #890, opened 2022-10-16
make QuadraticHamiltonian a subclass of PolynomialTensor

kevinsung:
### Summary
Fixes https://github.com/Qiskit/qiskit-nature/issues/865


qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
```diff
@@ -123,29 +123,40 @@ def __init__(
                 if not _is_antisymmetric(antisymmetric_part, rtol=rtol, atol=atol):
```
mrossinek, line 123: I cannot comment higher up, so I am leaving this here: we started to constrain some arguments to be keyword-only in our API by inserting the `*` catch all positional arguments in the argument list.
I think, this would also be useful here, seeing things like `validate`, `atol`, `rtol`...
Personally, I would make it the following:
```python
    def __init__(
        self,
        hermitian_part: np.ndarray | None = None,
        antisymmetric_part: np.ndarray | None = None,
        *,
        constant: float = 0.0,
        num_modes: int | None = None,
        validate: bool = True,
        rtol: float = 1e-5,
        atol: float = 1e-8,
    ) -> None:
```
But I would be open to having `constant` and/or `num_modes` above the `*` catchall.

mrossinek: Some more observation with respect to types:
- constant is constrained to float and does not allow complex. Is this intentional?
- you constrain the matrices to be dense. It might be useful to open this up to `sparse.SparseArray`, too (this would have more implications down below, which I will not comment on any more than stating it here)

qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
```diff
@@ -123,29 +123,40 @@ def __init__(
...
-        if self._hermitian_part is None:
-            self._hermitian_part = np.zeros((self._num_modes, self._num_modes))
-        if self._antisymmetric_part is None:
-            self._antisymmetric_part = np.zeros((self._num_modes, self._num_modes))
+        data: dict[str, float | np.ndarray] = {"": constant}
+        if hermitian_part is not None:
+            data["+-"] = hermitian_part
+        if antisymmetric_part is not None:
+            data["++"] = antisymmetric_part
+        super().__init__(data)
```
mrossinek, line 131: You say, that this class implements the `TolerancesMixin` and allow `rtol` and `atol` as argument to `__init__`. However, you don't set the internal attributes to the provided values, which means that these will only be considered during the course of `__init__` and will then be lost.

mrossinek, line 131: This will always perform additional validation e.g. ensuring that the provided matrices are square (since your keys are length 2).
Since you don't do this check yourself, I think this is good. But as a user if I were to provide `validate=False`, I would also expect the validation of `PolynomialTensor` to be skipped:
```suggestion
        super().__init__(data, validate=validate)
```

qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
```diff
@@ -123,29 +123,40 @@ def __init__(
...
+        return np.zeros((self._num_modes, self._num_modes), dtype=complex)
 
     @property
     def constant(self) -> float:
         """The constant."""
-        return self._constant
+        return cast(float, self[""])
 
     @property
     def num_modes(self) -> float:
```
mrossinek, line 162: Down below in the `second_q_op` method is a `TODO` comment which can be removed.

Also for that method: if you do decide to allow sparse matrices, the `second_q_op` method can be replaced by `FermionicOp.from_polynomial_tensor` which will prevent duplication of the iteration logic.
You would need to register the `"--'` key prior to passing the PT to that method though.

qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
kevinsung: I added the asterisk to force keyword arguments.

>constant is constrained to float and does not allow complex. Is this intentional?

Yes. The constant must be real because the Hamiltonian is hermitian.

>you constrain the matrices to be dense. It might be useful to open this up to sparse.SparseArray, too (this would have more implications down below, which I will not comment on any more than stating it here)

Let's leave that for the future.

qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
kevinsung: Done.

kevinsung: I deleted the `TolerancesMixin` specification. It's actually redundant because `PolynomialTensor` already inherits it. The `rtol` and `atol` arguments to `__init__` are only for the input validation.

qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
mrossinek: It is only necessary if we want to replace the `second_q_op` method with `Fermionic.from_polynomial_tensor`. The benefit of that (as I said) would be immediate support for sparse tensors, too.
This would look somewhat like the following:
```python
def second_q_op(self):
    self._data[""] = constant
    self._data["--"] = self.antisymmetric_part.conjugate()
    return FermionicOp.from_polynomial_tensor(self)
```
It would be better not to modify `self._data` so making a shallow copy of `self._data` and passing a new `PolynomialTensor(shallow_copy_data, copy=False)` into the `FermionicOp.from_polynomial_tensor` method might be better. But this is just to give you the idea

kevinsung: What I'm confused about is, why does `FermionicOp.from_polynomial_tensor(self)` not work directly? Adding an extra operator
```
self._data["--"] = self.antisymmetric_part.conjugate()
```
actually cancels out the `++` key. This seems like a bug with `FermionicOp.from_polynomial_tensor` or PolynomialTensor.

kevinsung: Thank you. The bug was actually in how I stored the antisymmetric matrix in the first place. I've fixed it.

qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
```diff
@@ -32,7 +31,7 @@ def _is_antisymmetric(mat: np.ndarray, rtol: float = 1e-5, atol: float = 1e-8) -
     return np.allclose(mat, -mat.T, rtol=rtol, atol=atol)
 
 
-class QuadraticHamiltonian(Hamiltonian, TolerancesMixin):
+class QuadraticHamiltonian(PolynomialTensor, Hamiltonian):
```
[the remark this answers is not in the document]
kevinsung: I'm not sure I understand. Are you suggesting to set the class attributes `QuadraticHamiltonian.atol` and `QuadraticHamiltonian.rtol` to the values passed to `__init__`? I don't think that should be done since that would affect other instances too.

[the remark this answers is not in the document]
mrossinek: Yes I had misremembered myself, that `rtol` and `atol` are actually class attributes as dictated by the `TolerancesMixin`. In that case, please document that the `rtol` and `atol` keywords of `__init__` are only used during `__init__` and discarded afterwards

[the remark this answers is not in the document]
kevinsung: Done.

qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
```diff
@@ -84,68 +133,53 @@ def __init__(
             ValueError: Hermitian part must be Hermitian.
             ValueError: Antisymmetric part must have shape num_modes x num_modes.
             ValueError: Antisymmetric part must be antisymmetric.
+
+        Note:
```
woodsp-ibm, line 137: Since the atol & rtol values above are defaulted to the same as the Mixin atol and rtol would it be preferable, like is done in other methods using the mixin where local atol and rtol can be passed for the parameters to default to None. Then if None the class level atol/rtol are used?

qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
```diff
@@ -58,6 +106,7 @@ def __init__(
         hermitian_part: Optional[np.ndarray] = None,
         antisymmetric_part: Optional[np.ndarray] = None,
         constant: float = 0.0,
+        *,
```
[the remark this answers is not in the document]
kevinsung: Done.

qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
kevinsung: Done.

woodsp-ibm (approved):

Issue #828, opened 2022-09-19
Support parameterized operators


mrossinek:
### What should we add?

We would like to support parameterized operators. This would:
- use `qiskit.circuit.Parameter` to enable parameterization
- need to be supported in:
  - `SparseLabelOp`
  - ~`PolynomialTensor` (`Parameter` can actually be injected right into the matrices)~ (see #894)
  - ~classes which contain those operators~ (see #894)
- I think, this should not require any special handling in the `QubitMapper` and `QubitConverter`
- some mechanics may need to be added during our `algorithms` to deal with parameters (can be dealt of as a follow-up feature to this one)

This depends on:
- https://github.com/Qiskit/qiskit-terra/pull/8620
- `Hamiltonian` classes being able to have their returned objects of `second_q_op()` be modifiable (i.e. parameterizable). I have some ideas for this but will provide more details once #666 has a PR; during its development I will pay particular attention to this feature. (see #894)

kevinsung:
@mrossinek FYI I currently do not plan to work on adding Parameter support for PolynomialTensor, I suggest renaming this issue so it's specific to SparseLabelOp and creating a separate one for PolynomialTensor if necessary.

PR #891, opened 2022-10-16
add support for Parameters to SparseLabelOp

kevinsung:
### Summary
Part of https://github.com/Qiskit/qiskit-nature/issues/828

### Details and comments
Also fixes https://github.com/Qiskit/qiskit-nature/issues/892


qiskit_nature/second_q/operators/fermionic_op.py
```diff
@@ -148,6 +148,11 @@ class FermionicOp(SparseLabelOp):
             considered a lower bound, which means that mathematical operations acting on two or more
             operators will result in a new operator with the maximum number of spin orbitals of any
             of the involved operators.
+    .. note::
+
+        A FermionicOp can contain Parameters. However, a FermionicOp containing Parameters
+        does not support the following methods:
+        - ``is_hermitian``
```
mrossinek, line 155:
```suggestion
        - ``is_hermitian``
        - ``to_matrix``
```
This method also does not work as long as we do not refactor the `to_matrix` implementation to use the `sparse` library instead of `scipy.sparse`

qiskit_nature/second_q/operators/fermionic_op.py
```diff
@@ -148,6 +148,11 @@ class FermionicOp(SparseLabelOp):
             considered a lower bound, which means that mathematical operations acting on two or more
             operators will result in a new operator with the maximum number of spin orbitals of any
             of the involved operators.
+    .. note::
+
+        A FermionicOp can contain Parameters. However, a FermionicOp containing Parameters
```
mrossinek, line 153:
```suggestion
        A FermionicOp can contain :class:`qiskit.circuit.Parameter` objects as coefficients. However, a FermionicOp containing Parameters
```
Note, you will need to reformat the line to obey the maximum line length

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -16,12 +16,12 @@
 
 from abc import ABC, abstractmethod
 from collections.abc import Collection, Mapping
-from numbers import Number
-from typing import Iterator, Sequence
+from numbers import Complex
```
mrossinek, line 19: Why did you change this to `Complex`? I chose `Number` on purpose to leave this as general as possible.

mrossinek (commented):
Thank you, Kevin!

Besides these minor comments just a question: did you happen to also look into how the `PolynomialTensor` class can handle `Parameter` objects? In theory, this should be possible and I had done some rough testing a while back but not looked into it in more detail.

We can add support for that as part of another PR and a separate tracking issue, but I wanted to ask first, whether you had already looked into this :+1:

qiskit_nature/second_q/operators/sparse_label_op.py
kevinsung: Changing it to `Number` gives these mypy errors:
```python
qiskit_nature/second_q/operators/sparse_label_op.py: note: In member "_multiply" of class "SparseLabelOp":
qiskit_nature/second_q/operators/sparse_label_op.py:227: error: Unsupported operand types for * ("complex" and "Number")  [operator]
qiskit_nature/second_q/operators/sparse_label_op.py:227: note: Both left and right operands are unions
```

qiskit_nature/second_q/operators/fermionic_op.py
kevinsung: Done.

qiskit_nature/second_q/operators/fermionic_op.py
kevinsung: Done.

qiskit_nature/second_q/operators/sparse_label_op.py
mrossinek: Looks like a compatibility limitation of the numbering stack and typing.
I suggest you use `# type: ignore[operator]` but leave the type to be `Number` as originally intended.

kevinsung: I think `Complex` is actually correct. The error
```
Unsupported operand types for * ("complex" and "Number")
```
explains why. Having `Number` there was a bug that slipped through mypy because the type was incorrect in the function signature (it had `complex`, so mypy assumed `other` was `complex`.
https://github.com/Qiskit/qiskit-nature/blob/ba3a191b30606d6c1da5c378de2eea7bceaeab57/qiskit_nature/second_q/operators/sparse_label_op.py#L194

mrossinek:
> I think using Parameters with PolynomialTensor is a very niche use case so it is not a high priority. PolynomialTensor is mainly useful for linear algebra operations, and those would not be possible if it had Parameters.

Quite on the contrary. Enabling parameterized `PolynomialTensor` objects could indeed be very powerful.
Plus, linear algebra would also not be a problem, since the `Parameter` objects do support numerics just fine (otherwise this would also not work for the `SparseLabelOp`).

One example where a parameterized `PolynomialTensor` would be useful is the refactoring of the `ooVQE` algorithm, which applies parameterized orbital rotations.

Just a proof of concept:
```python
import numpy as np

from qiskit.circuit import Parameter
from qiskit_nature.second_q.operators import FermionicOp, PolynomialTensor

a = Parameter("a")
b = Parameter("b")

mata = np.eye(2, dtype=object)
mata[0, 0] = a
matb = np.zeros((2, 2), dtype=object)
matb[0, 1] = b
matb[1, 0] = -b

pt_a = PolynomialTensor({"+-": mata})
pt_b = PolynomialTensor({"+-": matb})
print(pt_a + pt_b)
print(pt_a @ pt_b)
op = FermionicOp.from_polynomial_tensor(pt_a + pt_b)
print(op)
op = FermionicOp.from_polynomial_tensor(pt_a @ pt_b)
print(op)
```
Outputs (omitting zero-terms because chop needs a little work):
```
Polynomial Tensor
 "+-":
[[ParameterExpression(a) ParameterExpression(b)]
 [ParameterExpression(-1.0*b) 1]]
Polynomial Tensor
 "+-+-":
[[[[ParameterExpression(0) ParameterExpression(a*b)]
   [ParameterExpression(-1.0*a*b) ParameterExpression(0)]]

  [[0 ParameterExpression(0)]
   [ParameterExpression(0) 0]]]


 [[[0 ParameterExpression(0)]
   [ParameterExpression(0) 0]]

  [[0 ParameterExpression(b)]
   [ParameterExpression(-1.0*b) 0]]]]
Fermionic Operator
number spin orbitals=2, number terms=4
  a * ( +_0 -_0 )
+ b * ( +_0 -_1 )
+ -1.0*b * ( +_1 -_0 )
+ 1 * ( +_1 -_1 )
Fermionic Operator
number spin orbitals=2, number terms=4
a*b * ( +_0 -_0 +_0 -_1 )
+ -1.0*a*b * ( +_0 -_0 +_1 -_0 )
+ b * ( +_1 -_1 +_0 -_1 )
+ -1.0*b * ( +_1 -_1 +_1 -_0 )
```

kevinsung:
By linear algebra I meant eigenvalue decompositions and the like, not just matrix multiplication. Isn't that needed for computing orbital rotations?

mrossinek:
> By linear algebra I meant eigenvalue decompositions and the like, not just matrix multiplication. Isn't that needed for computing orbital rotations?

Those would not work, yes. The way that the orbital rotations are done there, I do not think its needed. Although this may not hold generally speaking.

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -409,16 +426,16 @@ def chop(self, tol: float | None = None) -> SparseLabelOp:
...
-            zero_imag = cmath.isclose(value.imag, 0.0, abs_tol=tol)
-            if zero_real and zero_imag:
-                continue
-            if zero_imag:
-                new_data[key] = value.real
-            elif zero_real:
-                new_data[key] = value.imag
-            else:
-                new_data[key] = value
+            if not isinstance(value, ParameterExpression):
```
mrossinek, line 429: This `isinstance` check is a bit too agressive. I expect the following to print `True` which it does not currently:
```python
from qiskit.circuit import Parameter

a = Parameter("a")
op1 = FermionicOp({"+_0": a})
print((op1 - op1).chop() == FermionicOp.zero())  # False
```

qiskit_nature/second_q/operators/sparse_label_op.py
kevinsung: @mrossinek I tried replacing everything but that gave other errors. I think it is a bit delicate. In any case it's unrelated to the rest of this PR so I think it's better to do it in a separate PR. I opened https://github.com/Qiskit/qiskit-nature/issues/896 for it.

test/second_q/operators/test_sparse_label_op.py
```diff
@@ -312,6 +363,22 @@ def test_one(self):
...
+
+    def test_chop(self):
+        """Test chop."""
+        op = DummySparseLabelOp({"+_0 -_1": 1 + 1e-12j, "+_0 -_2": a})
+        self.assertEqual(op.chop(), DummySparseLabelOp({"+_0 -_1": 1, "+_0 -_2": a}))
+
+        op = DummySparseLabelOp({"+_0 -_1": 1e-12 + 1j, "+_0 -_2": a})
+        self.assertEqual(op.chop(), DummySparseLabelOp({"+_0 -_1": 1j, "+_0 -_2": a}))
+
+        self.assertEqual((op - op).chop(), DummySparseLabelOp.zero())  # False
```
mrossinek, line 380:
```suggestion
        self.assertEqual((op - op).chop(), DummySparseLabelOp.zero())
```

mrossinek (commented):
Other than this, this LGTM :+1:

test/second_q/operators/test_sparse_label_op.py
kevinsung: Done.

kevinsung:
I think this is broken because of a bug introduced by https://github.com/Qiskit/qiskit-nature/pull/893. See https://github.com/Qiskit/qiskit-nature/pull/893/files#r998269022.

test/second_q/operators/test_fermionic_op.py
```diff
@@ -89,18 +119,41 @@ def test_compose(self):
...
+        with self.subTest("multi compose with parameters"):
+            fer_op = FermionicOp({"+_0 +_1 -_1": self.a, "-_0 +_0 -_1": 1}) @ FermionicOp(
+                {"": 1, "-_0 +_1": self.b}
+            )
+            fer_op = fer_op.simplify()
+            targ = FermionicOp(
+                {
+                    "+_0 +_1 -_1": self.a,
+                    "-_0 +_0 -_1": 1,
+                    "+_0 -_0 +_1": self.a * self.b,
```
mrossinek, line 131:
```suggestion
                    "+_0 +_1 -_0": -self.a * self.b,
```
After #893 the order of terms is preserved. Once #900 is merged, it will also compute the sign correctly.

kevinsung: Done.

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -202,4 +203,4 @@ def mode_based_mapping(
                         )
                 ret_op_list.append(ret_op)
 
-        return PauliSumOp(SparsePauliOp.sum(ret_op_list).simplify().chop())
+        return PauliSumOp(SparsePauliOp.sum(ret_op_list).simplify())
```
mrossinek, line 206: Why is it necessary to remove this `.chop()`? Does it cause problems with parameterized `SparsePauliOp` instances?
If so, a bug report should be filed with Qiskit Terra

mrossinek (commented):
Just one last clarification, other than that this is good to go :+1:

qiskit_nature/second_q/mappers/qubit_mapper.py
kevinsung: Indeed, `chop` is not supported with parameterized `SparsePauliOp`. It is not a bug. See https://github.com/Qiskit/qiskit-terra/pull/8895.

mrossinek (approved):
Thanks, Kevin!

qiskit_nature/second_q/operators/fermionic_op.py
```diff
@@ -148,6 +148,12 @@ class FermionicOp(SparseLabelOp):
             considered a lower bound, which means that mathematical operations acting on two or more
             operators will result in a new operator with the maximum number of spin orbitals of any
             of the involved operators.
+    .. note::
+
+        A FermionicOp can contain :class:`qiskit.circuit.Parameter` objects as coefficients.
+        However, a FermionicOp containing Parameters does not support the following methods:
+        - ``is_hermitian``
+        - ``to_matrix``
```
woodsp-ibm, line 156: What does this mean exactly - we don't seem to check if it contains Parameters and raise an error. It just fails in someway - be nice to raise an error with a suitable message about parameters

kevinsung: We could check if any coefficients are Parameters before performing the operation. However, that would incur some overhead. I decided to follow the design of SparsePauliOp from Terra and not perform this check. The error message gets raised in a different part of the stack, but the error is still about parameters, so I think it's still helpful enough.

kevinsung: Ok this is the error, it doesn't exactly mention parameters, but it's still reasonably helpful.
```
  File "/home/kjs/worktrees/qiskit-nature/qiskit-nature-parameters/test.py", line 7, in <module>
    op.is_hermitian()
  File "/home/kjs/worktrees/qiskit-nature/qiskit-nature-parameters/qiskit_nature/second_q/operators/fermionic_op.py", line 509, in is_hermitian
    return all(np.isclose(coeff, 0.0, atol=atol) for coeff in diff.values())
  File "/home/kjs/worktrees/qiskit-nature/qiskit-nature-parameters/qiskit_nature/second_q/operators/fermionic_op.py", line 509, in <genexpr>
    return all(np.isclose(coeff, 0.0, atol=atol) for coeff in diff.values())
  File "<__array_function__ internals>", line 180, in isclose
  File "/usr/local/venvs/qiskit-nature-parameters-YbHAS_rw/lib/python3.10/site-packages/numpy/core/numeric.py", line 2372, in isclose
    xfin = isfinite(x)
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

kevinsung: In any case what I did is consistent with Terra and also more performant.

woodsp-ibm: I guess I might be inclined to sacrifice a little performance for a check to raise a more meaningful error.

woodsp-ibm:
This is a more general question. What do we expect of such parameterized operators. If I look at a parameterized quantum circuit I have methods to get parameters and also to bind values etc. Might I want this at this level - I see that you've done similar to SparsePauliOp but I guess I could ask the same question of that too.

qiskit_nature/second_q/operators/fermionic_op.py
mrossinek: I agree. In the same way that `PolynomialTensor` has two methods `is_sparse` and `is_dense`, we could add a method `is_parameterized` which could check for this.
All methods which do not support parameterized ops should then call said method and raise a more meaningful error than they do now

mrossinek:
> This is a more general question. What do we expect of such parameterized operators. If I look at a parameterized quantum circuit I have methods to get parameters and also to bind values etc. Might I want this at this level - I see that you've done similar to SparsePauliOp but I guess I could ask the same question of that too.

That is a very good point actually! I raised https://github.com/Qiskit/qiskit-terra/issues/8966 to bring this up. We should also add an `assign_parameters` method here. Inspiration can be drawn from the equivalent methods in `qiskit.opflow` and `qiskit.circuit`

qiskit_nature/second_q/operators/fermionic_op.py
kevinsung: Done.

kevinsung:
I've added `assign_parameters`

kevinsung:
Let me know if there's anythiing else. I'll fix the conflicts after this is approved.

woodsp-ibm:
We were thinking that it could be useful to have a `parameters` property like quantum circuit has https://github.com/Qiskit/qiskit-terra/blob/6aad91418951484ee5cf808a0c147f96bc951223/qiskit/circuit/quantumcircuit.py#L2525 though here it seemed that sorting would most likely not be needed - in Terra it was done so when a parameterized circuit was used as an ansatz the parameters would always be sorted the same e.g. for when used by with an optimizer in VQE .

kevinsung:
@woodsp-ibm I think that should be left for a separate PR, what do you think?

woodsp-ibm:
Since this was all about adding parameter support, and I think it should be a simple method right, could we keep it all done by the same PR.

kevinsung:
In QuantumCircuit it's not a simple method, it maintains an internal "parameter table" that's constantly updated with every operation that could introduce or remove parameters. So it's actually a bit involved which is why I suggested to leave it for another PR.

But if you just want me to add a simple method that iterates through the dictionary and returns the parameters, that's easy enough. Please advise.

woodsp-ibm:
>  a simple method that iterates through the dictionary and returns the parameters

I think that would suffice right - I can't see any reason at present for anything more complex than that.

kevinsung:
Ok, I've added the method. Also, in order to support assigning parameters in place, I changed the type of `data` from `Mapping` to `dict`.

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -463,9 +488,39 @@ def induced_norm(self, order: int = 1) -> float:
...
+        Returns:
+            The operator with the parameters assigned, or None if ``inplace=True``.
+        """
+        data = self._data if inplace else self._data.copy()
+        for key, value in data.items():
+            if value in parameters:
+                data[key] = parameters[value]
+        if not inplace:
+            return self._new_instance(data, other=self)
+        return None
```
mrossinek, line 522:
```suggestion
        return self
```
This is the expected behavior for `inplace=True` operations. The return-type typehint should be adapted accordingly

mrossinek (changes requested):
> Also, in order to support assigning parameters in place, I changed the type of the internal data container from Mapping to dict.

I do not understand why you think this is necessary. The public API can report `Mapping` just fine. The fact that we use a `dict` internally for storage does not affect this.

qiskit_nature/second_q/operators/sparse_label_op.py
kevinsung: Done.

kevinsung:
>I do not understand why you think this is necessary. The public API can report Mapping just fine. The fact that we use a dict internally for storage does not affect this.

When `copy=False`, previously the type of the internal storage object could be `Mapping` rather than `dict`. This precludes in-place operations because `Mapping` does not support item assignment. I updated the internal storage to always be `dict`. But if the input to `__init__` is `Mapping` then we can't support `copy=False`. So I changed the input type to also be `dict`.

mrossinek:
Alright I have also discussed with @woodsp-ibm offline. We would like to do the following:

- use `Mapping` for all public-facing API which is aligned with the immutable operator idea and allows users to specifically provide immutable data if they want to further enforce this
- when `inplace=True` is requested by the user (which we want to support to avoid additional copies if a user so desires), proceed with an `isinstance(..., MutableMapping)` check. If this check fails, raise a `TypeError`. Otherwise proceed normally since the user specifically requested `inplace=True`

kevinsung:
In accordance with our offline discussion, I've reverted the change from `Mapping` to `dict` and removed the `inplace` argument to `assign_parameters`. We will make SparsePauliOp mutable in the future; this is tracked at https://github.com/Qiskit/qiskit-nature/issues/917.

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -32,6 +32,16 @@
 from .polynomial_tensor import PolynomialTensor
 
 
+_TCoeff = Union[complex, ParameterExpression]
+
+
+def _to_number(a: SupportsComplex | ParameterExpression) -> complex | float:
```
mrossinek, line 38:
```suggestion
def _to_number(a: SupportsComplex | ParameterExpression) -> complex:
```
`complex` implies `float` so this type hint union should not be necessary

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -32,6 +32,16 @@
...
+    if isinstance(a, ParameterExpression):
+        sympified = a.sympify()
+        return complex(sympified) if sympified.is_Number else np.nan
+    return complex(a)
```
mrossinek, line 42: I am not happy about specifically casting everything to `complex` here. Is that truely necessary? We should be able to just do:
```suggestion
    return a
```

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -463,9 +489,33 @@ def induced_norm(self, order: int = 1) -> float:
...
+        return any(isinstance(coeff, ParameterExpression) for coeff in self.values())
+
+    def assign_parameters(self, parameters: Mapping[ParameterExpression, _TCoeff]) -> SparseLabelOp:
+        """Assign parameters to new parameters or values.
+
+        Args:
+            parameters: The mapping from parameters to new parameters or values.
+
+        Returns:
+            The operator with the parameters assigned, or None if ``inplace=True``.
```
mrossinek, line 511:
```suggestion
            A new operator with the parameters assigned.
```

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -463,9 +489,33 @@ def induced_norm(self, order: int = 1) -> float:
...
+        """
+        data = dict(self._data.items())
+        for key, value in data.items():
+            if value in parameters:
+                data[key] = parameters[value]
```
mrossinek, line 516: This currently iterates the entire dictionary twice. Maybe the following would be better:
```suggestion
        data = {
            key: parameters[value] if value in parameters else value for key, value in self._data.items()
        }
```
Black will very likely want to reformat this

mrossinek (commented):
Just some minor comments. Other than this, this is looking good :+1:

qiskit_nature/second_q/operators/sparse_label_op.py
kevinsung: Done.

qiskit_nature/second_q/operators/sparse_label_op.py
kevinsung: Done.

qiskit_nature/second_q/operators/sparse_label_op.py
kevinsung: Done.

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -16,12 +16,12 @@
 
 from abc import ABC, abstractmethod
 from collections.abc import Collection, Mapping
-from numbers import Number
-from typing import Iterator, Sequence
+from numbers import Complex
+from typing import Iterator, Sequence, SupportsComplex, Union
```
mrossinek, line 20:
```suggestion
from typing import Iterator, Sequence, Union
```

mrossinek (approved):
This LGTM :+1:

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -296,8 +314,8 @@ def equiv(
     ) -> bool:
         """Check equivalence of two ``SparseLabelOp`` instances up to an accepted tolerance.
 
-        The absolute and relative tolerances can be changed via the `atol` and `rtol` attributes,
-        respectively.
+        The default absolute and relative tolerances can be changed via the `atol` and `rtol`
+        attributes, respectively.
```
woodsp-ibm, line 318: I am wondering what this really means. atol and rtol are described below - the values here, if provided, override the defaults from the mixin. Those can only be changed via the class attributes. I see this text was here before but just changed a little. To me its kinda confusing and I think the args captures it well enough that this could probably just be deleted.

qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -306,7 +324,13 @@ def equiv(
 
         Returns:
             True if operators are equivalent, False if not.
+
+        Raises:
+            ValueError: Operator contains parameters.
```
woodsp-ibm, line 329: `Operator contains parameters.` is a bit terse - should this be `If either operator contains parameters` ? And should the check check below also check the what its being compared to, i.e. other, and  be more say
```
        if self.is_parameterized() or other.is_parameterized():
            raise ValueError("Cannot perform equivalence as operator(s) contain parameters.")
``` 

But.... we can do `__eq__` below with no similar parameter check like this it seems?  It seems is they can be equal they should be equiv too, no?

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -182,7 +183,7 @@ def mode_based_mapping(
         else:
             for terms, coeff in second_q_op.terms():
                 # 1. Initialize an operator list with the identity scaled by the `coeff`
-                ret_op = SparsePauliOp("I" * nmodes, coeffs=[coeff])
+                ret_op = SparsePauliOp("I" * nmodes, coeffs=np.array([coeff]))
```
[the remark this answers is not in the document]
kevinsung: It needs to copy the data anyway because `[coeff]` is a Python list.

qiskit_nature/second_q/operators/sparse_label_op.py
kevinsung: Done.

qiskit_nature/second_q/operators/sparse_label_op.py
kevinsung: Done.

>It seems if they can be equal they should be equiv too, no?

No. Equality for Parameters is well-defined. However, approximate numerical equality (which is what `equiv` is for) is not supported.

qiskit_nature/second_q/mappers/qubit_mapper.py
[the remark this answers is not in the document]
woodsp-ibm: Yes, but at least doing an asarray() signals code wise we do not need a copy even if numpy needs to make a copy internally. It was only a minor comment since the data there is nothing much anyways!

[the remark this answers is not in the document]
kevinsung: Not sure what you mean by the signaling. I could also add `copy=False` but we know it would just be ignored. If you don't mind, let's keep it how I wrote it.

qiskit_nature/second_q/operators/sparse_label_op.py
woodsp-ibm: So you are saying an operator can be equal to itself but not equivalent to itself if it has parameters. Just seems strange. If the other operator is a different instance and equal its the same deal. Yes equiv can be defined as you have done not to valid in the presence of parameters - but equally, like the equals function, it could be defined the same way when it came to parameters - they would have to be equal since there is no notion of equiv as such. In the end to me it depends on how its equiv is defined here not just that parameters do not define it hence it can't be defined here. As I said it just seems logically odd to me. I guess we can see what users make of things - should the equiv in this light.

kevinsung: >So you are saying an operator can be equal to itself but not equivalent to itself if it has parameters.

This sounds strange but in my opinion it's really due to the fact that `equiv` is a poorly named method since "equivalence" is an ambiguous notion. It should really be named `approx_eq` or something similar which indicates that it is for checking approximate numerical equality. We could choose to make anything that compares `==` equal also `equiv` equal, but I think that would be conflating two separate notions in an undesirable way (exact object equality on the one hand, and approximate numerical equality on the other). Also, note that the decision I made is consistent with the behavior of `SparsePauliOp` in Terra, which does not support `equiv` for parameterized operators.

woodsp-ibm: We can be define what equiv means here in terms of approximate equality. As it stands its not defined in the presence of Parameter - unlike equal. I just find it a bit "weird". The fact that they have an identical/equal placeholder (parameter) for final value seems to me to make them equiv just the same way it makes them equal.  As it stands equality is a mix too already, right, of object versus values. I agree its a more limiting definition of equiv what I am thinking since if one operator had Parameter A and the other and expression with Parameter A plus some very small constant value which was within tolerance it would not pass under the rule of equality though you might consider it equiv. But I think thats too much to take care of. I guess its degrees here. We can see how it goes with equiv as it stands - is possible to revisit/relax this later if it seems to be a common use case where people are say more expecting it more to behave as I think I am.

kevinsung: Perhaps a discussion about how to define "equality" and "equivalence" in Qiskit is in order, but that is not the scope of this PR. Here I've simply remained consistent with the rest of Qiskit; if we decide to change these definitions down the line then we should do it consistently.

woodsp-ibm: In terms of SparsePauliOp its more complex in what it potentially supports for parameters (objects - right). It may have a harder time dealing with a similar situation where complexity is an issue there that is not here. Its equiv is first doing some super._eq() check and returning False so I might question consistency as it stands. Anyway the addition of parameters to these operators is new, lets see how it goes as I mentioned above

kevinsung: >In terms of SparsePauliOp its more complex in what it potentially supports for parameters (objects - right).

What do you mean by this? The type of parameters is `ParameterExpression`, both there and here.

qiskit_nature/second_q/mappers/qubit_mapper.py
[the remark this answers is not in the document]
woodsp-ibm: signal - I meant if one read the code it would be clear we did not logically need to do a copy even if numpy internally needs to do it, that was all. Its ok as is.

qiskit_nature/second_q/operators/sparse_label_op.py
woodsp-ibm: https://github.com/Qiskit/qiskit-terra/issues/8966

kevinsung: Ah, interesting, I hadn't realized that. I think it's desirable to specify a type more specific than `Any` for the parameters, but we could probably expand the scope to include sympy Symbols at least.

woodsp-ibm: I think ParameterExpression here is fine. If there's a good use case to expand beyond, to support more, then I think that's something for the future consideration.

kevinsung:
Is there anything else? If not I'll go ahead and fix the merge conflicts.

woodsp-ibm:
> Is there anything else?

Not from me, let's get the conflict sorted then.

mrossinek (approved):
LGTM, thanks Kevin!

PR #893, opened 2022-10-17
fix: make FermionicOp.simplify preserve the term order

### LINKED ISSUE #888: FermionicOp `normal_ordered` and `simplify` should commute

It is surprising that simplifying a normal-ordered operator changes it so that it is no longer normal-ordered. Simplification should conserve normal order. The code below should not raise AssertionError.

```python

op = FermionicOp({"+_1 -_0": -1})
assert op.simplify().normal_ordered() == op.normal_ordered().simplify()
```

mrossinek:
### Summary

Closes #888

### DIFF
--- a/test/second_q/operators/test_fermionic_op.py
+++ b/test/second_q/operators/test_fermionic_op.py
@@ -84,7 +84,7 @@ class TestFermionicOp(QiskitNatureTestCase):
             ) @ FermionicOp({"": 1, "-_0 +_1": 1}, num_spin_orbitals=2)
             fer_op = fer_op.simplify()
             targ = FermionicOp(
-                {"+_0 +_1 -_1": 1, "-_0 +_0 -_1": 1, "+_0 -_0 +_1": 1, "-_0 -_1 +_1": -1},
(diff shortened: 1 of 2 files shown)


woodsp-ibm (approved):

test/second_q/operators/test_fermionic_op.py
```diff
@@ -84,7 +84,7 @@ def test_compose(self):
             ) @ FermionicOp({"": 1, "-_0 +_1": 1}, num_spin_orbitals=2)
             fer_op = fer_op.simplify()
             targ = FermionicOp(
-                {"+_0 +_1 -_1": 1, "-_0 +_0 -_1": 1, "+_0 -_0 +_1": 1, "-_0 -_1 +_1": -1},
+                {"+_0 +_1 -_1": 1, "-_0 +_0 -_1": 1, "+_0 +_1 -_0": 1, "-_0 -_1 +_1": -1},
```
kevinsung, line 87: You swapped the order of the terms here, but you didn't add a minus sign.

mrossinek: Thanks for spotting this! https://github.com/Qiskit/qiskit-nature/pull/900 should address this bug :+1:

PR #897, opened 2022-10-18
Implements `HartreeFock` and `VSCF` initial states a `BlueprintCircuit`

### LINKED ISSUE #881: Make initial states a `BlueprintCircuit`

The ansatze like `UCC` and alike have been implemented as `BlueprintCircuit` which allows "lazy" attribute setting to complete the circuit definition as used for example in the `VQEUCCFactory`.
The same could be done for the initial states like `HartreeFock` and `VSCF`.

Unathi-Skosana:
### Summary

Fixes #881 

### Details and comments

I have tried to preserve the original code as best as I can rather than a complete overhaul, however I'm unsure about one thing in this regard; should the checks for unsupported mappers remain in `__init__` or be moved to `_check_configuration` for both initial states?

### DIFF
--- a/.pylintdict
+++ b/.pylintdict
@@ -268,6 +268,7 @@ ket
 kitaev
 knowles
 kohn
+kron
 kronecker
 kwarg
 kwargs
--- a/qiskit_nature/second_q/operators/electronic_integrals.py
+++ b/qiskit_nature/second_q/operators/electronic_integrals.py
@@ -14,12 +14,13 @@
 
 from __future__ import annotations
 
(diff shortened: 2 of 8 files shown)


mrossinek:
> should the checks for unsupported mappers remain in __init__ or be moved to _check_configuration for both initial states?

These need to be moved to `_check_configuration` or even the `qubit_converter.setter` to disallow it early :+1: 

Please also leave the files in `qiskit_nature.circuit` unchanged and instead apply your changes to the files in `qiskit_nature.second_q.circuit`

PR #898, opened 2022-10-18
fix: some problems with the new FCIDump

mrossinek:
### Summary

This fixes some issues with the new FCIDump class.

### Details and comments

- fixes the beta-shift of the integral indices in the dump method
- reorders the dataclass arguments to match the rest of the stack (order of h1, h2, etc.)
- makes non-required arguments actually optional
- removes `num_orbitals` in favor of always inferring it from the shape of `h1`


test/second_q/formats/fcidump/test_fcidump_dumper.py
```diff
@@ -23,7 +23,7 @@
 from qiskit_nature.second_q.operators.tensor_ordering import _phys_to_chem
 from qiskit_nature.second_q.problems import ElectronicStructureProblem
 from qiskit_nature.units import DistanceUnit
-from qiskit_nature.second_q.drivers import PySCFDriver
+from qiskit_nature.second_q.drivers import PySCFDriver, MethodType
```
mrossinek, line 26:
```suggestion
from qiskit_nature.second_q.drivers import PySCFDriver
```

woodsp-ibm (approved):

PR #905, opened 2022-10-19
Refactor `SpinOp` as `SparseLabelOp`

### LINKED ISSUE #863: Refactor `SpinOp` to be a subclass of `SparseLabelOp`

#667 tracks the implementation of a specific method of what this issue implies, but this issue is actually a requirement before #667 can be tackled.

The [current `FermionicOp`](https://github.com/Qiskit/qiskit-nature/blob/ce3e8a58dfd0306ccc41ddc9ff7b78050999cc62/qiskit_nature/second_q/operators/fermionic_op.py) should be able to serve as a decent guide in combination with the existing `SpinOp` to tackle this issue.
Only sparse labels should be supported.

Care must be taken in this scenario, because the `SpinOp` carries additional data in the form of the spin associated with the operator.
#515 should also be kept in mind when refactoring this operator (it does not need to be solved immediately since that also requires changes to the mappers) but it would be worthwhile to think whether we want direct support inside of `SpinOp` or whether we want to handle this via for example a #795.

ElePT:
### Summary
This PR refactors the `SpinOp` to become a `SparseLabeOp`.

### Details and comments
Closes  #863

TO-DOs:
- [x] Update `.simplify()`, Implement `.index_ordered()` https://github.com/Qiskit/qiskit-nature/pull/902
- [x] Run eigensolver computation check
- [x] Fix references of `SpinOp` throughout stack
- [x] Finish docs


mrossinek (commented):
Things are already looking very good. I did not focus on the unittests as of now, but the core functionality appears to be working as expected :+1:

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,539 @@
...
+                    if index > max_index:
+                        max_index = index
+                elif index >= num_s:
+                    raise QiskitNatureError(
+                        f"The index, {index}, from the label, {key}, exceeds the number of "
+                        f"spins, {num_s}."
                     )
-                self._register_length = register_length
-                self._from_sparse_label(labels)
+            # TODO: do we need to validate the exponent?
```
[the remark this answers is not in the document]
ElePT: It looks like `0` is already a valid exponent and it's simplified to identity as expected, so I will go for option 1.

qiskit_nature/second_q/mappers/logarithmic_mapper.py
```diff
@@ -128,12 +116,16 @@ def _logarithmic_encoding(
...
 
         # Get the spin matrices
-        spin_matrices = [SpinOp(symbol, spin=spin).to_matrix() for symbol in "XYZ"]
+        spin_matrices = [
+            SpinOp.x(spin).to_matrix(),
+            SpinOp.y(spin).to_matrix(),
+            SpinOp.z(spin).to_matrix(),
+        ]
         # Append the identity
         spin_matrices.append(np.eye(dspin))
```
mrossinek, line 130:
```suggestion
        spin_matrices = [
            SpinOp.x(spin).to_matrix(),
            SpinOp.y(spin).to_matrix(),
            SpinOp.z(spin).to_matrix(),
            np.eye(dspin),
        ]
```

qiskit_nature/second_q/mappers/linear_mapper.py
```diff
@@ -86,7 +73,7 @@ def _linear_encoding(self, spin: Union[Fraction, float]) -> List[PauliSumOp]:
             to represent the embedded 'X' operator
         """
 
-        spin_op_encoding: List[PauliSumOp] = []
+        spin_op_encoding: list[PauliSumOp] = []
```
mrossinek, line 76: I wonder whether instead of appending to this list we could just construct it once in the end, the same way to do it in the `LogarithmicMapper`

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,582 @@
...
 
-    :code:`"+_i -_i"` is supported.
-    This pattern is parsed to :code:`+_i -_i = X_i^2 + Y_i^2 + Z_i`.
+        SpinOp({
+                "X_0 X_1": -1,
+                "Y_0 Y_1": -1,
+                "Z_0 Z_1": -1,
+                "Z_0": -0.3,
+                "Z_1": -0.3,
+            ],
```
mrossinek, line 82:
```suggestion
            },
```
I think this is what is causing the docs to fail building

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,582 @@
...
+        SpinOp({"X_0 Z_1": 1j}, num_spins=2).adjoint()
+
+
+    **Iteration**
+
+    Instances of `SpinOp` are iterable. Iterating a SpinOp yields (term, coefficient)
+    pairs describing the terms contained in the operator. Labels containing powers/exponents
+    will be expanded into multiple (term, coefficient) pairs.
 
+    Attributes:
```
mrossinek, line 181: This should also include the `spin` attribute

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,582 @@
...
 
-        ~(1j * z)
+        SpinOp({"X_0 Z_1": 1j}, num_spins=2).adjoint()
+
+
+    **Iteration**
+
+    Instances of `SpinOp` are iterable. Iterating a SpinOp yields (term, coefficient)
+    pairs describing the terms contained in the operator. Labels containing powers/exponents
+    will be expanded into multiple (term, coefficient) pairs.
```
mrossinek, line 180:
```suggestion

    The following attributes can be set via the initializer but can also be read and updated once
    the ``SpinOp`` object has been constructed.

```

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,582 @@
...
+            QiskitNatureError: when spin is not a positive half-integer.
         """
-        self._coeffs: np.ndarray
-        self._spin_array: np.ndarray
-        dtype = np.complex128  # TODO: configurable data type. mixin?
-
+        self.num_spins = num_spins
         spin = Fraction(spin)
-        if spin.denominator not in (1, 2):
+        if spin.denominator not in (1, 2) or spin < 0:
```
mrossinek, line 220:
```suggestion
        if spin.denominator not in (1, 2) or spin <= 0:
```
Spin 0 should not be allowed

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,582 @@
...
+        """Helper method to iterate over label splits.
+
+        Yields:
+            A tuple containing the character, index and exponent in each label split.
+        """
+        for lbl in label.split():
+            # if lbl == "":
+            #     yield "", None, 1
+            #     continue
+
```
mrossinek, line 334:
```suggestion
```
Removing this, also fixes the mis-spelled word `lbl` being found

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,582 @@
...
+                if isinstance(mat, sp.SparseArray):
+                    coo = sp.as_coo(mat)
+                    for value, *index in zip(coo.data, *coo.coords):
+                        data[label_template.format(*index)] = value
+
+        return cls(data, copy=False, num_spins=tensor.register_length).chop()
+
+    @staticmethod
+    def _split_label(label) -> Iterator[tuple[str, int | None, int]]:
```
mrossinek, line 324:
```suggestion
    def _split_label(label) -> Iterator[tuple[str, int, int]]:
```
I don't think this can now be the case anymore :+1:

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,582 @@
...
+            #     continue
+
+            char, index = lbl.split("_")
+            index_split = index.split("^")
+            idx = int(index_split[0])
+            exp = int(index_split[1]) if len(index_split) > 1 else 1
+            yield char, idx, exp
+
+    @classmethod
+    def x(cls, spin: float | Fraction = Fraction(1, 2)) -> SpinOp:
```
mrossinek, line 342: We should also overwrite the `one` and `zero` classmethods from the `SparseLabelOp` to allow the `spin` settings :+1:

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,582 @@
...
-            if n_z >= 1:
-                labels_list.append(f"Z_{pos}" + (f"^{n_z}" if n_z > 1 else ""))
-        if not labels_list:
-            return f"I_{self.register_length - 1}"
-        return " ".join(labels_list)
-
-    @lru_cache(maxsize=128)
     def to_matrix(self) -> np.ndarray:
+        # TODO: use scipy.sparse.csr_matrix() and add parameter `sparse: bool`.
+
```
mrossinek, line 543:
```suggestion
```

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,602 @@
...
-    This is similar to Qiskit's string-based representation of qubit operators.
-    For example,
+    These operator terms are encoded as sparse labels, strings consisting of a
+    space-separated list of expressions. Each expression must look like
+    :code:`[XYZ]_<index>^<power>`, where the :code:`<index>` is a non-negative
+    integer representing the index of the spin mode  where the `X`, `Y` or `Z`
+    component of the spin operator is to be applied.
 
-    .. code-block:: python
+    The value of :code:`index` is bound by the number of spins(`num_spins`) of the operator
```
mrossinek, line 52:
```suggestion
    The value of :code:`index` is bound by the number of spins (``num_spins``) of the operator
```

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,602 @@
...
+    (Note: since Python indices are 0-based, the maximum value an index can take is given by
+    :code:`num_spins-1`). The the :code:`<power>` is a positive integer indicating the number
```
mrossinek, line 54:
```suggestion
    :code:`num_spins-1`). The :code:`<power>` is a positive integer indicating the number
```

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,602 @@
...
+    A `SpinOp` represents a weighted sum of spin operator terms with a certain spin
+    value associated to them. This value can be an integer for bosonic particles or
+    a half-integer (fraction) for fermions.
 
-    Dense labels are strings in which each character maps to a unique spin mode.
-    This is similar to Qiskit's string-based representation of qubit operators.
-    For example,
+    These operator terms are encoded as sparse labels, strings consisting of a
+    space-separated list of expressions. Each expression must look like
+    :code:`[XYZ]_<index>^<power>`, where the :code:`<index>` is a non-negative
```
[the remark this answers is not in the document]
mrossinek: I see its written later but maybe we can already mention it here

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,602 @@
...
 
     :class:`SpinOp` supports the following basic arithmetic operations: addition, subtraction,
-    scalar multiplication, and adjoint.
+    scalar multiplication, adjoint, composition and tensoring.
+
+    As of now, operations that involve two different instances of `SpinOp` (i.e. addition,
+    subtraction, composition, and tensoring) are only supported for identical spins
+    (`op_1.num_spins == op_2.num_spins`). In the future, multiple spin lengths within one
+    `SpinOp` will be supported, which might enable certain operations to also support
+    different `num_spins`.
```
mrossinek, line 131: We should remove the second sentences. We should not promise any features (even if they are _very_ likely).
Let's simply state that only identical spin operations are supported (and then in the future we rephrase this)

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,602 @@
...
 
     .. jupyter-execute::
 
-        ~(1j * z)
+        SpinOp({"X_0 Z_1": 1j}, num_spins=2).adjoint()
+
+
+    **Iteration**
 
+    Instances of `SpinOp` are iterable. Iterating a SpinOp yields (term, coefficient)
```
mrossinek, line 177:
```suggestion
    Instances of ``SpinOp`` are iterable. Iterating a SpinOp yields (term, coefficient)
```

mrossinek (commented):
Things are looking good now! Last thing is to polish the docs a little bit :slightly_smiling_face: 

I did some suggestions and some comments below.
Please also check whether we consistently use RST code blocks (double backticks) instead of Markdown ones (single backticks). I saw a few more occurrences than I commented on.. :see_no_evil:

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,602 @@
...
-            )
+    @classmethod
+    def zero(cls, spin: float | Fraction = Fraction(1, 2)) -> SpinOp:
+        # pylint: disable=arguments-differ
+        """Constructs the "zero" spin operator for a given spin.
 
-        if self.register_length != other.register_length:
-            raise TypeError("Incompatible register lengths for '+'.")
+        Returns:
+            The "one" spin operator for `spin`.
```
woodsp-ibm, line 385:
```suggestion
            The "zero" spin operator for `spin`.
```
Copy/paste just not updated I imagine

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,602 @@
...
-          - :math:`S^-`
-          - Lowering operator
 
-    There are two types of label modes for :class:`SpinOp`.
-    The label mode is automatically detected.
+class SpinOp(SparseLabelOp):
+    """XYZ Spin operator.
 
-    1. Dense Label (default, `register_length = None`)
+    A `SpinOp` represents a weighted sum of spin operator terms with a certain spin
```
woodsp-ibm, line 42: We have a mix of single and double backticks in how these are done in this file. If its quick enough to fix maybe we can make it consistent

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -18,532 +18,602 @@
...
-    def compose(self, other):
-        # TODO: implement
-        raise NotImplementedError
+    def conjugate(self) -> SpinOp:
+        """Returns the conjugate of the ``SpinOp``.
 
-    def mul(self, other: complex) -> "SpinOp":
-        if not isinstance(other, (int, float, complex)):
+        Returns:
+            The complex conjugate of the starting ``SpinOp``.
```
woodsp-ibm, line 430:
```suggestion
            The complex conjugate of this ``SpinOp``.
```
"the starting" seems a bit abstract

qiskit_nature/second_q/operators/spin_op.py
```diff
@@ -12,538 +12,606 @@
...
-                labels_list.append(f"Y_{pos}" + (f"^{n_y}" if n_y > 1 else ""))
-            if n_z >= 1:
-                labels_list.append(f"Z_{pos}" + (f"^{n_z}" if n_z > 1 else ""))
-        if not labels_list:
-            return f"I_{self.register_length - 1}"
-        return " ".join(labels_list)
-
-    @lru_cache(maxsize=128)
     def to_matrix(self) -> np.ndarray:
+        # TODO: use scipy.sparse.csr_matrix() and add parameter `sparse: bool`.
```
woodsp-ibm, line 561: This method used to cache its result and this has been removed. I checked out of band with @mrossinek as I was not sure whether it was intentional or not. It seems the removal was not intended.

mrossinek: The problem is indeed that `dict` is not hashable and since we use that internally, we cannot hash our operator classes. One solutions would be the following:
```python
class SparseLabelOp:
    ...

    def __hash__(self) -> int:
        return hash(repr(self))
```
This would make all operators hashable (again) and enables the caching again

mrossinek (approved):
Despite the caching discussion, I will merge this PR now in order to proceed with remaining tasks for the release. Re-adding the hashing can be done in a separate PR but this way the next steps are no longer blocked.

Thanks a lot, Elena, for your work here! :+1:

PR #908, opened 2022-10-21
docs: add How-To section and basic example

mrossinek:
### Summary

This prepares the addition of a set of very short How-To snippets to the docs.
The idea is to use simple RST files, written in `doctest` format which will allow their validity to be tested during CI


woodsp-ibm (approved):
Looks fine to get the basics in place so we can start/improve from here.

Issue #867, opened 2022-10-07
Remove the `VibrationalIntegrals` in favor of `PolynomialTensor`


mrossinek:
### What is the expected enhancement?

Once #668 and #866 are implemented, the old `VibrationalIntegrals` class should be removed in favor of using the `PolynomialTensor`. This matches what I have done for the electronic-structure stack in #841.

This will also enable the extraction of the `basis` attribute from the `VibrationalEnergy` hamiltonian and `OccupiedModals` property onto the `VibrationalStructureProblem` (again; the same was done for the electronic structure stack). This in turn, will enable the extension of the `BasisTransformer` to transform between different vibrational structure bases in the future.

Issue #864, opened 2022-10-07
Refactor `VibrationalOp` to be a subclass of `SparseLabelOp`


mrossinek:
### What is the expected enhancement?

#668 tracks the implementation of a specific method of what this issue implies, but this issue is actually a requirement before #668 can be tackled.

The [current `FermionicOp`](https://github.com/Qiskit/qiskit-nature/blob/ce3e8a58dfd0306ccc41ddc9ff7b78050999cc62/qiskit_nature/second_q/operators/fermionic_op.py) should be able to serve as a decent guide in combination with the existing `VibrationalOp` to tackle this issue.
Only sparse labels should be supported.

A big question to be discussed, is whether we want to keep the `mode, modal` "dual indexing" or whether we want to expand this immediately into its expanded form indexing purely `register_length`.

Since the original `VibrationalOp` was almost a copy of the `FermionicOp` at the time, the actual implementation should not be too difficult with the good template already in place.

PR #909, opened 2022-10-21
Migration Guide

mrossinek:
### Summary

This PR adds a v0.4 -> v0.5 migration guide to the documentaiton.

### Details and comments

I will also be (partially) updating the tutorials as part of this PR.

Below is an overview of the current status:
- [x] Electronic Stack
- [x] Vibrational Stack (pending #864, #867, etc.)
- [x] Lattice Stack (pending #863, etc.)
- [x] Algorithms
- [x] Protein Folding


docs/migration/00a_Solving_problems_with_v0.5.ipynb
```diff
@@ -0,0 +1,431 @@
...
+   "cell_type": "markdown",
+   "id": "3f1f4687",
+   "metadata": {},
+   "source": [
+    "### `qiskit_nature.algorithms.ground_state_solvers`\n",
+    "\n",
+    "This module was moved to `qiskit_nature.second_q.algorithms.ground_state_solvers`.\n",
+    "Besides updating your imports, you need to pay attention to the following:\n",
+    "\n",
+    "- `AdaptVQE` was migrated to Qiskit Terra: `qiskit.algorithms.minimum_eigensolver.AdaptVQE`\n",
```
woodsp-ibm, line 213: Could we state here that instead of a dedicated adapt vqe ground state solver, in migrating it to Terra, it becomes an algorithm there like VQE that is used with the GroundStateSolver.

(Made me thing about the AdaptVQE factory capability that we don;t have yet - maybe the existing factory could switch between VQE and AdaptVQE - anyway outside the scope of this guide)

docs/migration/00b_Electronic_structure_with_v0.5.ipynb
```diff
@@ -0,0 +1,729 @@
...
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "<ipython-input-1-aea959c80ea0>:7: NatureDeprecationWarning: The qiskit_nature.problems.second_quantization.electronic package is deprecated as of version 0.5.0 and will be removed no sooner than 3 months after the release. Instead use the qiskit_nature.second_q.problems package.\n",
+      "  from qiskit_nature.problems.second_quantization import ElectronicStructureProblem\n",
+      "<ipython-input-1-aea959c80ea0>:7: NatureDeprecationWarning: The qiskit_nature.problems.second_quantization.vibrational package is deprecated as of version 0.5.0 and will be removed no sooner than 3 months after the release. Instead use the qiskit_nature.second_q.problems package.\n",
+      "  from qiskit_nature.problems.second_quantization import ElectronicStructureProblem\n"
```
woodsp-ibm, line 60: It seems strange in the notebook about ElectronicStructure for this to have an message about vibrational. Maybe its the way we have things imported in the init files but I would this it would seem odd to a user.

Also is this all I would expect to see - e.g. its using an older PySCF driver and seemingly no message about that or the Molecule etc

docs/migration/00b_Electronic_structure_with_v0.5.ipynb
```diff
@@ -0,0 +1,729 @@
...
+   "execution_count": 2,
+   "id": "19449c21",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Fermionic Operator\n",
+      "number spin orbitals=4, number terms=36\n",
```
woodsp-ibm, line 131: The new operator has no truncation control over the string output it seems. The previous operator had a truncated output by default.

docs/migration/00b_Electronic_structure_with_v0.5.ipynb
```diff
@@ -0,0 +1,729 @@
...
+    "A few notes are worth adding:\n",
+    "\n",
+    "- The driver for [pyquante2](https://github.com/rpmuller/pyquante2) was removed, because its source code has not been updated for more than 3 years. Originally, it was supported to enable usage of Qiskit Nature on Windows, however, since then [psi4](https://github.com/psi4/psi4) has also added Windows support and the integration of the [Windows Subsystem for Linux](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux) has opened up further development possibilities.\n",
+    "- The `HDF5Driver` was removed in favor of supporting the standardized [QCSchema](https://github.com/MolSSI/QCSchema) (see `qiskit_nature.second_q.formats.qcschema`).\n",
+    "- The `ElectronicStructureMoleculeDriver` was removed because we are steering towards a future with tighter, plugin-like integration with classical codes, making the concept of drivers where Qiskit starts a classical simulation outdated. You can still use the `.from_molecule(...)` methods of the remaining drivers in combination with the `MoleculeInfo` class.\n",
+    "- The `MoleculeInfo` has become a pure data container and no longer supports _degrees of freedom_."
```
woodsp-ibm, line 243: Also it seems the way one defines atoms and distances has changed.

docs/migration/00b_Electronic_structure_with_v0.5.ipynb
```diff
@@ -0,0 +1,729 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "64cfb910",
+   "metadata": {},
+   "source": [
+    "### Using drivers that were migrated to `qiskit_nature.second_q.drivers`"
```
woodsp-ibm, line 251: FYI These were not run to see the results here (of course CI will but its not here in this I am looking at to review thats all)

docs/migration/00b_Electronic_structure_with_v0.5.ipynb
```diff
@@ -0,0 +1,729 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5e041f05",
+   "metadata": {},
+   "source": [
+    "## `qiskit_nature.properties`\n",
+    "\n",
+    "The properties module has been refactored and split into multiple locations. In this section, we will only be focusing on its _electronic_ components."
```
woodsp-ibm, line 620: For people that want to inject integrals they got from elsewhere - or get the integrals from a driver do/should we have anything for this. I know we get users that want to do either and these aspects are changed again right though these are more advanced users so maybe just seeing where/how now in some tutorial or other would suffice.

docs/migration/00c_Vibrational_structure_with_v0.5.ipynb
```diff
@@ -0,0 +1,1180 @@
...
+    "\n",
+    "| Legacy component | New location |\n",
+    "|:-----------------|:-------------|\n",
+    "| BaseDriver | `qiskit_nature.second_q.drivers.BaseDriver` |\n",
+    "| VibrationalStructureDriver | `qiskit_nature.second_q.drivers.VibrationalStructureDriver` |\n",
+    "| VibrationalStructureDriverType | **removed** |\n",
+    "| VibrationalStructureMoleculeDriver | **removed** |\n",
+    "| GaussianForcesDriver | `qiskit_nature.second_q.drivers.GaussianForcesDriver` |\n",
+    "| GaussianLogDriver | `qiskit_nature.second_q.drivers.GaussiaLogDriver` |\n",
+    "| GaussianLogResult | `qiskit_nature.second_q.drivers.GaussianLogResult` |\n",
```
woodsp-ibm, line 496: Strictly speaking these are not vibrational - the ForcesDriver would run things in a way to have the information come out of a log driver instance where it was vibrational. But I think they are ok here, at least they get a mention somewhere.

docs/migration/00d_Lattice_models_with_v0.5.ipynb
```diff
@@ -0,0 +1,466 @@
...
+     ]
+    },
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "<ipython-input-1-b4df7055da5b>:1: NatureDeprecationWarning: The qiskit_nature.problems.second_quantization.electronic package is deprecated as of version 0.5.0 and will be removed no sooner than 3 months after the release. Instead use the qiskit_nature.second_q.problems package.\n",
+      "  from qiskit_nature.problems.second_quantization.lattice.lattices import LineLattice\n",
+      "<ipython-input-1-b4df7055da5b>:1: NatureDeprecationWarning: The qiskit_nature.problems.second_quantization.vibrational package is deprecated as of version 0.5.0 and will be removed no sooner than 3 months after the release. Instead use the qiskit_nature.second_q.problems package.\n",
+      "  from qiskit_nature.problems.second_quantization.lattice.lattices import LineLattice\n"
```
woodsp-ibm, line 90: I commented on these messages before - here it seems even more weird since it mentions electronic and vibrational but nothing specific to lattices

docs/migration/00d_Lattice_models_with_v0.5.ipynb
```diff
@@ -0,0 +1,466 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "from qiskit_nature.second_q.hamiltonians.lattices import LineLattice\n",
+    "from qiskit_nature.second_q.hamiltonians import FermiHubbardModel\n",
+    "\n",
+    "line = LineLattice(2)\n",
+    "fermi = FermiHubbardModel(line.uniform_parameters(2.0, 4.0), 3.0)\n",
+    "print(fermi.second_q_op())"
```
woodsp-ibm, line 144: So once the Model is constructed the method to get the operator no longer is plural ie has the **s** at the end just second_q_op, not second_q_ops. Should we note this -its a bit more subtle if you don't pay attention

docs/tutorials/05_problem_transformers.ipynb
```diff
@@ -0,0 +1,461 @@
...
+   "source": [
+    "## The `BasisTransformer`\n",
+    "\n",
+    "This transformer allows you to transform an `ElectronicStructureProblem` from one basis into another.\n",
+    "This is useful in certain settings such as:\n",
+    "- when you obtained a problem description or the Hamiltonian coefficients in the AO basis from an external source\n",
+    "- when you explicitly generated the problem in the AO basis to modify it manually before transforming into the MO basis\n",
+    "- etc.\n",
+    "\n",
+    "Since we can achieve the second scenario directly using Qiskit Nature, that is what we will be doing here. To learn more about how to do this, we recommend that you read the tutorial on the [`QCSchema`](08_qcschema.ipynb)."
```
woodsp-ibm, line 28: For the three code blocks here it might be helpful to note what they are doing. Eg. We first create a problem in the AO basis to demonstrate the basis transformation, where the objective is to have it in the MO basis. We then create a transformer to go from AO to MO. Finally the problem is transformed and the result can be seen to be in the MO basis.

docs/tutorials/05_problem_transformers.ipynb
```diff
@@ -0,0 +1,461 @@
...
+   "execution_count": 6,
+   "id": "9a37235a",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "(1, 1)\n",
+      "5\n"
```
woodsp-ibm, line 227: Could we note that for LiH the core orbitals is just a the single lowest energy orbital. So as a result the output is one less orbital and the two electrons from that orbital are removed as well. Can we say something about the frozen core energy we extract which gets added back to the overall result at the end.

docs/tutorials/05_problem_transformers.ipynb
```diff
@@ -0,0 +1,461 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "92863022",
+   "metadata": {},
+   "source": [
+    "Furthermore, you can provide a list of orbital indices (0-based) which are to be removed from the system.\n",
+    "\n",
+    "> **Note:** these orbitals *must* be unoccupied, otherwise you will incur a large error in your computation. No guards are in place to check that the provided orbital indices are indeed unoccupied, so this is up to you!"
```
woodsp-ibm, line 248: Even if they are unoccupied it may still be a significant impact dependent on what part, if any, those orbitals play in the dynamics. This we often used with LiH simply because the orbitals in question we would remove were known to play a minimal part - the result was still within chemical accuracy as the energy difference was so small.

docs/tutorials/06_qubit_mappers.ipynb
```diff
@@ -0,0 +1,476 @@
...
+   "id": "48ff72a5",
+   "metadata": {},
+   "source": [
+    "## The `QubitConverter`\n",
+    "\n",
+    "We would like to end this tutorial with a few more notes on the `QubitConverter`.\n",
+    "\n",
+    "There are multiple reasons for its existence:\n",
+    "- it handles the qubit reduction techniques which are applied on the qubit space *after* the mapping process\n",
+    "- it **stores** the symmetry information which you find during the conversion of your *Hamiltonian* such that you can recall and use the same symmetry reductions when mapping additional observables later one\n",
```
woodsp-ibm, line 315: ... when mapping additional observables later one

At the end should be `on` not `one`

I imagine this can be expanded with the other mappers. Though I guess we might want to think about this in regards of document structure going forwards.

docs/tutorials/06_qubit_mappers.ipynb
```diff
@@ -0,0 +1,476 @@
...
+      "-1.0410931415166251 * I\n",
+      "- 0.7958748496863578 * Z\n",
+      "+ 0.1809311997842312 * X\n"
+     ]
+    }
+   ],
+   "source": [
+    "converter = QubitConverter(JordanWignerMapper(), z2symmetry_reduction=\"auto\")\n",
+    "\n",
+    "qubit_op = converter.convert(fermionic_op, sector_locator=problem.symmetry_sector_locator)\n",
```
woodsp-ibm, line 292: We only have this auto symmetry sector locator for Fermionic systems - ElectronicStructureProblem  - but I guess this is all this tutorial covers for now but maybe its worth stating anyway.

docs/tutorials/08_qcschema.ipynb
```diff
@@ -0,0 +1,227 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9f29061c",
+   "metadata": {},
+   "source": [
+    "There are two things to note here:\n",
+    "- the problem is specifically requested in the MO basis\n",
+    "- dipole integrals are handled separately (because the QCSchema does not yet support these coefficients)\n",
```
woodsp-ibm, line 81: > because the QCSchema does not yet support these coefficients

How about `because the current QCSchema standard does not support these coefficients`

docs/tutorials/08_qcschema.ipynb
```diff
@@ -0,0 +1,227 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "from qiskit_nature.second_q.drivers import PySCFDriver\n",
+    "\n",
+    "driver = PySCFDriver()\n",
+    "\n",
+    "problem = driver.run()\n",
+    "print(problem)"
```
woodsp-ibm, line 41: Does this not print anything? - I do not see any output (aside from what might be an extra blank line) - maybe you edited the cell and added this and did not re-run it.

docs/tutorials/09_properties.ipynb
```diff
@@ -0,0 +1,219 @@
...
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "<qiskit_nature.second_q.properties.particle_number.ParticleNumber object at 0x14bfa1aa5b80>\n",
+      "<qiskit_nature.second_q.properties.angular_momentum.AngularMomentum object at 0x14bfa1aa5a60>\n",
+      "<qiskit_nature.second_q.properties.magnetization.Magnetization object at 0x14bfa1aa5ac0>\n",
+      "<qiskit_nature.second_q.properties.dipole_moment.ElectronicDipoleMoment object at 0x14bfa1aa5f70>\n",
+      "None\n"
```
woodsp-ibm, line 110: This shows up as blank lines and None when I look at the output here. Ie the output of the source below.

I guess the None is expected though as it then says

>"You can see, that all properties come pre-populated except for the `ElectronicDensity`

Maybe this is not for this tutorial but the fields on the `properties` like `electronic_density` are they dynamic in that I can add some new property and it will have a field like this. Maybe this is more for some user defined property bit later.

docs/tutorials/09_properties.ipynb
```diff
@@ -0,0 +1,219 @@
...
+   ],
+   "source": [
+    "from qiskit_nature.second_q.drivers import PySCFDriver\n",
+    "\n",
+    "es_problem = PySCFDriver().run()\n",
+    "\n",
+    "print(es_problem.properties.particle_number)\n",
+    "print(es_problem.properties.angular_momentum)\n",
+    "print(es_problem.properties.magnetization)\n",
+    "print(es_problem.properties.electronic_dipole_moment)\n",
```
woodsp-ibm, line 122: In the QCSchema tutorial it mentioned the include dipole setting - should we note that via the way the driver is run here that this is set which is why the dipole moment exists here above

docs/tutorials/09_properties.ipynb
```diff
@@ -0,0 +1,219 @@
...
+    "print(es_problem.properties.electronic_density)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "4dca06da",
+   "metadata": {},
+   "source": [
+    "You can see, that all properties come pre-populated except for the `ElectronicDensity`. But this is easy to add:"
```
woodsp-ibm, line 131: If user adds such a property what should be their expectation of what happens when the problem is run. These all end up producing auxillary operators whose observable is measured and returned in the result?

docs/tutorials/01_electronic_structure.ipynb
```diff
@@ -108,159 +111,478 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "<qiskit_nature.second_q.problems.electronic_structure_problem.ElectronicStructureProblem object at 0x15287045b700>\n"
```
woodsp-ibm, line 126: Having got this far, though print(problem) seems to produce nothing in the output (well a blank line) its just the basic instance print like this - I guess we have no str defined for it

docs/tutorials/01_electronic_structure.ipynb
```diff
@@ -108,159 +111,478 @@
...
+     "execution_count": 10,
+     "metadata": {},
+     "output_type": "execute_result"
     }
    ],
    "source": [
-    "qubit_converter = QubitConverter(mapper=JordanWignerMapper())\n",
-    "qubit_op = qubit_converter.convert(main_op)\n",
-    "print(qubit_op)"
+    "problem.basis"
```
woodsp-ibm, line 385: A lot of these are not outputting anything that can be seen in the notebook (at least when I view it here)

docs/tutorials/01_electronic_structure.ipynb
```diff
@@ -108,159 +111,478 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Now that the Hamiltonian is ready, it can be used in a quantum algorithm to find information about the electronic structure of the corresponding molecule. Check out our tutorials on Ground State Calculation and Excited States Calculation to learn more about how to do that in Qiskit!"
+    "## Solving the `ElectronicStructureProblem`\n",
+    "\n",
+    "In the following, we will compute the ground-state of our problem instance. To learn more about the individual components that go into the `GroundStateSolver`, please refer to the corresponding tutorials."
```
woodsp-ibm, line 517: > To learn more about the individual components that go into the GroundStateSolver, please refer to the corresponding tutorials.

Does this add much - if it linked to some more specifically it might be helpful. For a new user figuring corresponding stuff may be more of a challenge

docs/tutorials/03_ground_state_solvers.ipynb
```diff
@@ -15,7 +15,7 @@
     "\n",
     "<img src=\"aux_files/H2_gs.png\" width=\"200\">\n",
     "\n",
-    "In this tutorial we are going to discuss the ground state calculation interface of Qiskit Nature. The goal is to compute the ground state of a molecular Hamiltonian. This Hamiltonian can be electronic or vibrational. To know more about the preparation of the Hamiltonian, check out the Electronic structure and Vibrational structure tutorials. \n",
+    "In this tutorial we are going to discuss the ground state calculation interface of Qiskit Nature. The goal is to compute the ground state of a molecular Hamiltonian. This Hamiltonian can be electronic or vibrational. To know more about the preparation of the Hamiltonian, check out the [Electronic structure](01_electronic_structure.ipynb) and [Vibrational structure tutorials](02_vibrational_structure.ipynb). \n",
```
woodsp-ibm, line 18: As this is being updated should we take the chance to note, at least briefly in this tutorial, that we compute the electronic part, using the quantum algorithms like VQE, and with that then computing and returning the electronic part of the ground state we add in the constant nuclear repulsion part in the nature stack to form the total ground state energy result.

docs/migration/00_Migration_Guide_to_v0.5.ipynb
```diff
@@ -0,0 +1,314 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "84e77b76",
+   "metadata": {},
+   "source": [
+    "## How to use this guide\n",
+    "\n",
+    "This guide is split across multiple files allowing you to discover the changes to each submodule as you find the need. That means, you do not need to read all of the details at once and jumping around is highly encouraged.\n",
```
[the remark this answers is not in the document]
mrossinek: I am not sure if an overall summary will be all that useful. That will be a looong table. Having the guide split by category/topic seemed the most natural to me

docs/migration/00a_Solving_problems_with_v0.5.ipynb
```diff
@@ -0,0 +1,431 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "a4689533",
+   "metadata": {},
+   "source": [
+    "# Solving Problems with v0.5\n",
+    "\n",
+    "### Overview\n",
+    "The major focus of the Qiskit Nature v0.5 refactoring was the description and representation of second-quantized operators and problems. Nonetheless, the `algorithms` module and supporting modules did also receive significant updates. Most importantly, all algorithms are now based on [Qiskit Primitives](https://qiskit.org/documentation/apidoc/primitives.html) support for which was added in Qiskit Terra 0.22. It is not the intention to provide detailed explanations of the primitives in this migration guide. We suggest that you read the [corresponding resources](https://qiskit.org/documentation/apidoc/primitives.html) of the Qiskit Terra documentation instead.\n",
```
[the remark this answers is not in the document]
mrossinek: Indeed, this is not a hard dependence. However, we do not (and will not) test the `qiskit_nature.second_q` stack against the non-primitive based algorithms (why should we if its going away). This also gives directly more visibility to primitives :+1:

docs/migration/00a_Solving_problems_with_v0.5.ipynb
mrossinek: Yes, I actually have been working on an example on how to use `AdaptVQE`. I might make that a `HowTo` :+1: 

I also think that a dedicated factory may not be needed. But also thinking of `VQD` I was wondering if we can revisit our factory classes and instead make them `AnsatzFactory` that are re-usable for whatever algorithm..

But as you said, this is off-topic for here. Let's discuss this another time :+1:

docs/migration/00_Migration_Guide_to_v0.5.ipynb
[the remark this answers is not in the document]
woodsp-ibm: I was more thinking summary (a paragraph or two) not detail (so not a huge table) such that a reader has an overall idea before jumping into the each section. The relocation sentence I outlined could be part of that. Maybe there are a couple of other pointer sentences that might be helpful in seeing aspects of the big picture of the changes before diving into detail, not sure.

docs/migration/00a_Solving_problems_with_v0.5.ipynb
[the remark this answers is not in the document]
woodsp-ibm: In that quoted bit maybe I am not reading it right - in the first sentence 'algorithms' there refers to the algorithms in Nature I believe. In the second sentence there is algorithms the Nature ones, ie following on from the previous sentence about significant updates, or is it referring to algorithms in Terra. If I would read it as algorithms in Nature, following on the what was said in the prior sentence - this is why I made the comment I did.

docs/migration/00_Migration_Guide_to_v0.5.ipynb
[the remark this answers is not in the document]
mrossinek: Ah that makes sense :+1: I started a sentence like that. Will see if I can expand on that a little and see where we go from there :+1:

docs/migration/00b_Electronic_structure_with_v0.5.ipynb
mrossinek: That might go into how the deprecation warnings are being handled.

@manoelmarques
- do you have any insights why we see the vibrational one appearing here where no vibrational parts are being used?
- and what about the deprecation of other parts like driver, transformer, etc.?

docs/migration/00b_Electronic_structure_with_v0.5.ipynb
mrossinek: Yes this is on purpose. The truncation we had done there was very unnatural imo (by number of characters rather than lines for example) and I found myself disabling it more often than I ended up wanting it.

If we find it useful in the future, we could re-add that :+1:

docs/migration/00b_Electronic_structure_with_v0.5.ipynb
mrossinek: Yes, I was thinking that the tutorial should suffice for this aspect. I can add a specific mention of this in the migration guide though :+1:

docs/migration/00c_Vibrational_structure_with_v0.5.ipynb
```diff
@@ -0,0 +1,1180 @@
...
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "  NIIIIIIIIII * (1268.0676746875001+0j)\n",
+      "+ INIIIIIIIII * (3813.8767834375008+0j)\n",
+      "+ IINIIIIIIII * (705.8633818750001+0j)\n",
+      "+ II+-IIIIIII * (-46.025705898886045+0j)\n",
```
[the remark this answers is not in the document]
mrossinek: Right, I forgot that the Sphinx output does not do the output folding which Jupyter notebooks do.. I will do some manual truncation for now :+1:

docs/migration/00c_Vibrational_structure_with_v0.5.ipynb
```diff
@@ -0,0 +1,1180 @@
...
+      "+ (42.67527310283147+0j) * ( +_3_0 -_3_0 +_1_0 -_1_1 )\n",
+      "+ (42.67527310283147+0j) * ( +_3_0 -_3_0 +_1_1 -_1_0 )\n",
+      "+ (-12.582898125000007+0j) * ( +_3_0 -_3_0 +_1_1 -_1_1 )\n",
+      "+ (7.0983500000000035+0j) * ( +_3_0 -_3_0 +_2_0 -_2_0 )\n",
+      "+ (10.038582840471063+0j) * ( +_3_0 -_3_0 +_2_0 -_2_2 )\n",
+      "+ (21.29505000000001+0j) * ( +_3_0 -_3_0 +_2_1 -_2_1 )\n",
+      "+ (10.038582840471063+0j) * ( +_3_0 -_3_0 +_2_2 -_2_0 )\n",
+      "+ (35.49175000000002+0j) * ( +_3_0 -_3_0 +_2_2 -_2_2 )\n",
+      "+ 0j * ( +_3_0 -_3_1 +_2_0 -_2_1 )\n",
+      "+ 0j * ( +_3_0 -_3_1 +_2_1 -_2_0 )\n",
```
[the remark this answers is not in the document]
mrossinek: Yes, in-line with the new `FermionicOp`, simplification will only occur as late as possible once going through the mappers. Thus, the new `VibrationalOp` might still have non-reduced terms appearing :+1: 
One can easily get rid of those using either `.chop()` or `.simplify()` (but the latter has more side effects)

I will add a note to the guide :+1:

docs/migration/00d_Lattice_models_with_v0.5.ipynb
mrossinek: @manoelmarques another (weirder) occurrence of the same "problem" I pinged you about in another thread above. I'll leave this unresolved until we figure out what is going on here

docs/migration/00d_Lattice_models_with_v0.5.ipynb
mrossinek: Yes, I will do that :+1: 

The reason for that is that all lattice models now implement the `Hamiltonian` interface rather than the `SecondQuantizedProperty` one (which still has `second_q_ops()` because they produce dicts of operators rather than single ones).

docs/tutorials/06_qubit_mappers.ipynb
mrossinek: Yeh that's a good point. Maybe this tutorial could become more of a general guide on how to use our mappers whereas for each specific one there is its own documentation page + maybe a HowTo for discussing some more examples? :thinking:

But this can be done later :+1:

docs/tutorials/08_qcschema.ipynb
mrossinek: Right now these just print
```
<qiskit_nature.second_q.problems.electronic_structure_problem.ElectronicStructureProblem object at 0x155041be1250>
```
which should also show in Sphinx rendered version. While not super insightful, it at least reflects the output type that's why I left these in.

In the future I would like to add some string-representations of problems similar to what we had for the properties and then the tutorials will be able to benefit from this directly :+1:

docs/tutorials/09_properties.ipynb
mrossinek: These fields are not dynamic. The tutorial states:

> The specific subclasses of BaseProblem for the various stacks such as <...> make things even simpler for you, because they provide direct attributes for the built-in properties of Qiskit Nature:

So the fields only exist for the built-in properties of Nature. User-provided ones need to be accessed via sets

docs/tutorials/09_properties.ipynb
mrossinek: Yes indeed. The `ElectronicStructureDriver.run()` method does not expose the `include_dipole` setting so it will always default to `True`

One would need to run the driver more manually to disable that and this depends on the driver, for example:
```python
driver = PySCFDriver()
driver.run_pyscf()
problem = driver.to_problem(include_dipole=False)
```

mrossinek: I've added a comment to the tutorial :+1:

docs/tutorials/09_properties.ipynb
mrossinek: Yes exactly :+1: 

But the user will need to extract the evaluated auxiliary operators manually since the result classes do not expose arbitrary data fields (and unless the user also implement `.interpret()` they would not be populated anyways

docs/tutorials/01_electronic_structure.ipynb
mrossinek: See similar previous threads. I'll resolve this one

docs/tutorials/01_electronic_structure.ipynb
mrossinek: Again, the instance strings... I'll leave the discussion to a single thread.

But for the basis in particular this is actually meaningful:
> <ElectronicBasis.MO: 'molecular'>

docs/tutorials/01_electronic_structure.ipynb
mrossinek: I'll make a specific link :+1:

mrossinek (commented):
Open tasks:

- [x] figure out how to fix the stable tutorials in CI (tries linking to non-existent files)
- [x] add overall release note

I was referring to some HowTos in my comments on the review above, but I would like to leave those for a separate PR (if time permits)

mrossinek:
The reason that the CI is failing the stable tutorials check is because of this line: https://github.com/Qiskit/qiskit-nature/blob/8e1da0a3d59dfd17c1741e6f1cdf90903b0c57ae/.github/workflows/main.yml#L341

The CI recursively removes the new tutorials and places the downloaded stable tutorials in their place. This breaks CI, because the `docs/migration` notebooks are trying to link to tutorial pages which no longer exist (because of the recursive deletion).

We have two options for fixing this:
1. we can also `rm -rf docs/migration/*` in this CI step
2. instead of `rm -rf docs/tutorials/*` we simply overwrite the new with the stable ones, leaving new files which did not exist before unchanged (i.e. only do the `cp -R ...` step

@manoelmarques what do you think?

releasenotes/notes/0.5-prelude-a1d20e9b76893ece.yaml
```diff
@@ -0,0 +1,111 @@
+---
+prelude: >
+    Qiskit Nature 0.5 comes with a major redesign of the :class:`.BaseProblem`
+    layer of its stack. Rather than being tightly integrating both, drivers and
+    transformers, problems are now a lot more standalone containers, being
+    generated by the various drivers (or built out by a user to their custom
+    needs directly). As a short example, compared to the old initialization:
```
woodsp-ibm, line 7:
```suggestion
    layer of its stack. Rather than tightly integrating drivers and transformers,
    problems are now a lot more standalone and are generated by the
    various drivers (or built out by a user to their custom needs directly).
    As a short example comparing the previous to the new problem creation:
```

releasenotes/notes/0.5-prelude-a1d20e9b76893ece.yaml
```diff
@@ -0,0 +1,111 @@
...
+      vqe = VQEUCCFactory(Estimator(), UCCSD(), SLSQP())
+
+      algorithm = GroundStateEigensolver(converter, vqe)
+
+      cas.fcisolver = QiskitSolver(algorithm)
+
+      cas.run()
+
+    Qiskit Nature still provides ``drivers`` to enable simple testing and provide
+    a more accessible entry to Qiskit Nature for users which do not come from a
```
woodsp-ibm, line 76:
```suggestion
    a more accessible entry to Qiskit Nature for users who do not come from a
```

woodsp-ibm (approved):

PR #911, opened 2022-10-21
Suppress warning messages on deprecated unit test methods

manoelmarques:
### Summary
The unit tests that test deprecated functions should not emit deprecate messages.


woodsp-ibm (approved):
Thanks @manoelmarques that certainly cleaned up the CI deprecated messages. I see there are a couple of second_q tests touched but it seems just relative to absolute imports for other parts from test.

PR #915, opened 2022-10-25
Refactor of VibrationalOp as SparseLabelOp subclass

### LINKED ISSUE #864: Refactor `VibrationalOp` to be a subclass of `SparseLabelOp`

declanmillar:
### Summary

Refactor VibrationalOp to be a SparseLabelOp subclass.

Closes #864.

### Details and comments

Currently keeping the mode, modal "dual indexing" with aim to expand this into its expanded form indexing purely register_length inside the `simplify` method.


qiskit_nature/second_q/operators/_bits_container.py
```diff
@@ -0,0 +1,154 @@
...
+
+    This is a utility object used during the simplification process of a operators.
+    It manages access to an internal data container, which maps from integers to bytes.
+    Each integer key corresponds to a vibrational mode of an operator term.
+    Each value consists of four bits which encoding for the corresponding index:
+
+        1. if a `+` has been applied
+        2. if a `-` has been applied
+        3. whether a `+` or `-` was applied first
+        4. whether the last applied operator was a `+` or `-`
```
dlasecki, line 33:
```suggestion
    Each value consists of four bits which encoding for the corresponding index:

        1. if a `+` has been applied
        2. if a `-` has been applied
        3. whether a `+` or `-` was applied first
        4. whether the last applied operator was a `+` or `-`.
```

mrossinek: Thanks! I will also correct the wording a bit :+1:

qiskit_nature/second_q/operators/__init__.py
```diff
@@ -53,6 +53,7 @@
     "SecondQuantizedOp",
     "SpinOp",
     "VibrationalOp",
+    "build_dual_index",
```
mrossinek, line 56: I will redo this slightly and make `build_dual_index` a `staticmethod` of `VibrationalOp` instead. That shows a lot better that this method is a specific utility for that class.

test/second_q/problems/test_vibrational_structure_problem.py
```diff
@@ -55,7 +41,7 @@ def test_second_q_ops_without_transformers(self):
         """Tests that the list of second quantized operators is created if no transformers
         provided."""
         expected_num_of_sec_quant_ops = 4
-        expected_len_of_vibrational_op = 47
+        expected_len_of_vibrational_op = 46
```
mrossinek, line 44: In case somebody comes across this change: that is because the all-identity with coefficient `0` was dropped :+1:

mrossinek (approved):
Thanks, Declan, for all the work here!

I'm merging this greedily now in order to unblock the next refactoring steps of the vibrational stack. Anything we may want to change we can also do as part of that PR :+1:

Issue #919, opened 2022-10-27
`ElectronicDensity` enhancements


mrossinek:
### What is the expected enhancement?

The new `ElectronicDensity` property is very naive in its current form. It was intended to be a starting point for further improvements, some ideas I will be listing below.

- [ ] renormalization techniques based on expected quantities, for example particle number, spin, etc.
    - this may become particularly important once implementing algorithms iterating over the `ElectronicDensity`. The noise we incur when measuring these expectation values on a quantum device can hinder convergence of such algorithms significantly
- [ ] symmetry-recovery (similar problem as in the above)
- more initialization methods, for example:
    - [ ] based purely on number of spatial orbitals and particle number
    - [ ] `ElectronicDensity.zero`
    - [ ] `ElectronicDensity.identity`

PR #920, opened 2022-10-27
perf: some performance improvements for ElectronicDensity

mrossinek:
### Summary

This PR adds some minor enhancements to the new `ElectronicDensity` property:
- only adds operators for the required indices (skips e.g. alpha-beta 2-RDM indices in favor of only beta-alpha 2-RDM; it also fully skips off-diagonal entries which will be zero in any case (besides noise)
- adds the `include_rdm2` boolean kwarg which allows disabling the 2-RDM entirely (since this would mean an overhead of N**4 operators which may not always be needed)

### Details and comments

The current class is still quite minimal, but further enhancements to it are tracked in #919


qiskit_nature/second_q/properties/electronic_density.py
```diff
@@ -103,35 +114,83 @@ def interpret(
...
+                        rdm2_ba[_shifted_index(index, n_spatial)] = aux_value.real
 
         result.electronic_density = ElectronicDensity.from_raw_integrals(
-            rdm1[n_spatial:, n_spatial:],
-            rdm2[n_spatial:, n_spatial:, n_spatial:, n_spatial:],
-            rdm1[:n_spatial, :n_spatial],
-            rdm2[:n_spatial, :n_spatial, :n_spatial, :n_spatial],
-            rdm2[:n_spatial, n_spatial:, n_spatial:, :n_spatial],
+            rdm1_a,
+            rdm2_aa if contains_rdm2 else None,
```
woodsp-ibm, line 166: Is this check here needed? Above, lines 126 thru 132 rdm2_aa etc are initialized to None and then set based contains_rdm2 - so these vars should already be None or some valid value, based on contains_rdm2 logic above, that this ternary seems like it should not be needed again here.

woodsp-ibm (approved):

PR #924, opened 2022-11-01
Fixed styles for transformers docs

javabster:
### Summary
I reviewed all pages under `second_q/transformers`, mostly looking good just fixed a few things

Changes in this PR:
- added superscript to `h1` in `active_space_transformer` (not sure if this is correct)
- replaced top bullet points in `active_space_transformer` and `freeze_core_transformer` with raw html bc it wasn't rendering correctly wirth the normal sphinx notation
- general styles: added custom CSS to make headings in methods render the same as in classes
- general styles: adjusted class tables so it doesn't have a weird scroll

### Screenshots

methods now have styled headings:
Screenshot 2022-11-01 at 1 38 12 PM

now weird scroll inside table:
Screenshot 2022-11-01 at 1 42 24 PM

Freeze core new styles:
Screenshot 2022-11-01 at 2 05 22 PM

active space new styles:
Screenshot 2022-11-01 at 2 08 25 PM


docs/_static/custom.css
```diff
@@ -25,3 +25,23 @@
 .toggle{
   background: #FBFBFB;
 }
+
```
woodsp-ibm, line 28: The css styles file here is common with Qiskit to keep the same style throughout the docs. I think we should keep it that way - if there some style improvement it should get done Qiskit in general I think.

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -46,36 +46,46 @@ class ActiveSpaceTransformer(BaseTransformer):
...
-        - finally, it is possible to select a custom set of active orbitals via their indices using
-          `active_orbitals`. This allows selecting an active space which is not placed around the
-          Fermi level as described in the first case, above. When using this keyword argument, the
-          following criteria must be met *in addition* to the ones listed above:
-
-            #. the length of `active_orbitals` must be equal to `num_spatial_orbitals`. Note, that
-               we do **not** infer the number of active orbitals from this list of indices!
-
-            #. the sum of electrons present in `active_orbitals` must be equal to `num_electrons`.
+    .. raw:: html
```
woodsp-ibm, line 57: Can we go back to standard formatting syntax for Sphinx and not embed html. Maybe it looks better but it should be following the standard way so its formatted by Sphinx - if needed we should improve things throughout  as mentioned above.

qiskit_nature/second_q/transformers/freeze_core_transformer.py
```diff
@@ -32,13 +32,17 @@ def __init__(
...
         configured active space.
 
-        The orbitals to be removed are specified in two ways:
-            1. When `freeze_core` is enabled (the default), the "core" orbitals will be determined
-               automatically according to `count_core_orbitals`. These will then be made inactive
-               and removed in the same fashion as in the :class:`ActiveSpaceTransformer`.
-            2. Additionally, unoccupied spatial orbitals can be removed via a list of indices
-               passed to `remove_orbitals`. It is the user's responsibility to ensure that these are
-               indeed unoccupied orbitals, as no checks are performed.
+        .. raw:: html
```
woodsp-ibm, line 35: Similar comment in regards raw html

qiskit_nature/second_q/transformers/active_space_transformer.py
javabster: The reason I went with the raw html solution was because this is what it looks like with the sphinx generated styles:

Screenshot 2022-11-01 at 1 38 02 PM


I wasn't able to find a way to get around this without using raw html. If you would still prefer the sphinx generated style lmk and I'll remove the raw html

woodsp-ibm: I think that's because the number list is indented and its treating that whole thing as a definition list. Try outdenting the numbered list, but you will need to add a blank line preceding it (after that text) to ensure its seen correctly.

docs/_static/custom.css
javabster: I opened [this issue](https://github.com/Qiskit/qiskit_sphinx_theme/issues/117) in the `qiskit_sphinx_theme` to fix the class and method header styles, but i don't know when it will get prioritised and it will still be a while before the next version of the theme is released and ready for all repos to use. The fix in this PR would work for the time being if you want the styles fixed in time for the nature release.

javabster: So I guess it depends how soon you want the style fix? In this case I would advocate for changing it here (and also I can open PRs for a similar fix in some of the other repos maybe?) so at least the users get the benefit of the better styles straight away? But if you would rather wait lmk

woodsp-ibm: I would prefer to keep things consistent and wait for if/when the theme etc gets updated for all Qiskit so that its picked up everywhere, whatever is done, rather than have specifics here so its different.

mrossinek: I am indifferent to this. Keeping things consistent might be easier down the road but this change seems so minor and yields much better results that I would also not see this as a hassle. Especially since this is aligned with the rest of the theme and really just fixing a bug by adding attributes that were missing before

docs/_static/custom.css
```diff
@@ -25,3 +25,23 @@
...
+    word-wrap: break-word;
+}
+
+article .method dd {
+    padding-left: 3em;
+}
+
+.wy-table-responsive {
+    max-height: unset;
+}
```
mrossinek, line 47: I would like to stress that even though I am fine with keeping the style consistent, **this** particular block I find very important. If there is no other way of achieving the same, I would highly argue to keep this in here.

This block removes the ability to scroll inside of tables which list e.g. the classes in a module doc page. I find this particularly important, because without it **there is NO visual indicator** that the table is actually scrollable in the first place. E.g. in the following screenshot (from the current `main` docs), a user will only learn about the `transformers` module, if they accidentally scroll within the table:
screenshot_1667373268

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -46,36 +46,46 @@ class ActiveSpaceTransformer(BaseTransformer):
     :math:`F^I_{pq} = h_{pq} + \sum_i 2 g_{iipq} - g_{iqpi}` and the inactive energy which is
     given by :math:`E^I = \sum_j h_{jj} + F ^I_{jj}`, where :math:`i` and :math:`j` iterate over
     the inactive orbitals.
-    By using the inactive Fock operator in place of the one-electron integrals, `h1`, the
+    By using the inactive Fock operator in place of the one-electron integrals, :math:`h_{1}`, the
```
mrossinek, line 49: Actually, `h1` was not defined anywhere so I would just do the following:
```suggestion
    By using the inactive Fock operator in place of the one-electron integrals the
```

qiskit_nature/second_q/transformers/active_space_transformer.py
mrossinek: I agree that we should not embed raw html. I also verified that dedenting and adding a blank line between the paragraph and first list item fixes the rendering :+1:

docs/_static/custom.css
woodsp-ibm: I am not sure about the scrollable - when I look at things they all have horiz and vert scrollbars and with different tables on the same page they are all different widths which does look ugly. We have the same css as Terra yet it seems the outcome is different. Here is something from Terra where you can see it has scroll bars (even if not needed) and the widths are different with the highlighted bars
image




Things were setup so we had the same look/style for docs, we use same css same these etc - I am not sure why things are different - maybe we need to tweak something elsewhere perhaps in conf.py @manoelmarques any chance of taking a look.

mrossinek: In that case it may be my browser because the same snippet does not have scroll bars for me:
screenshot_1667395446

woodsp-ibm: This is the [current published Nature ](https://qiskit.org/documentation/nature/apidocs/qiskit_nature.html) that looks the same - all sections have scroll bars including the long sub modules one

image

javabster: this difference is most likely because of the browser. I use chrome which doesn't show those scroll bars but if you're looking in safari/firefox maybe it does?

woodsp-ibm: It could be browser specific - I am using Firefox - not good that something so basic renders differently. So you already have issues with whats published in Qiskit in general then in this regard - not good at all if that;s the case.

qiskit_nature/second_q/transformers/active_space_transformer.py
javabster: Ah THANK YOU i was struggling for ages with indenting and rebuilding but couldnt get the right combo to make it work. I'll give this a try 😄

docs/_static/custom.css
javabster: > not good that something so basic renders differently

It's because we rely on the default HTML components for this so the browser takes over the styling. And the styles differ across browsers because ✨ branding ✨

javabster:
@mrossinek @woodsp-ibm I've removed the raw hmtl and the custom css (which i'll add in a separate PR) 👍 lmk if theres anything else needed

docs/_static/custom.css
woodsp-ibm: Perhaps also environment - I am on Windows, doesn't Mac hide scrollbars by default (its been a while since I touched one)

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -55,27 +55,28 @@ class ActiveSpaceTransformer(BaseTransformer):
     https://arxiv.org/abs/2009.01872.
 
     The active space can be configured in one of the following ways through the initializer:
-        - when only `num_electrons` and `num_spatial_orbitals` are specified, these integers
-          indicate the number of active electrons and orbitals, respectively. The active space will
-          then be chosen around the Fermi level resulting in a unique choice for any pair of
-          numbers.  Nonetheless, the following criteria must be met:
 
-            #. the remaining number of inactive electrons must be a positive, even number
+    - when only ``num_electrons`` and `num_spatial_orbitals` are specified, these integers
```
mrossinek, line 59:
```suggestion
    - when only ``num_electrons`` and ``num_spatial_orbitals`` are specified, these integers
```

qiskit_nature/second_q/transformers/active_space_transformer.py
```diff
@@ -55,27 +55,28 @@ class ActiveSpaceTransformer(BaseTransformer):
...
-          following criteria must be met *in addition* to the ones listed above:
+        #. the number of active orbitals must not exceed the total number of orbitals minus the
+           number of orbitals occupied by the inactive electrons
 
-            #. the length of `active_orbitals` must be equal to `num_spatial_orbitals`. Note, that
-               we do **not** infer the number of active orbitals from this list of indices!
+    - when, ``num_electrons`` is a tuple, this must indicate the number of alpha- and beta-spin
+      electrons, respectively. The same requirements as listed before must be met.
+    - finally, it is possible to select a custom set of active orbitals via their indices using
+      `active_orbitals`. This allows selecting an active space which is not placed around the
```
mrossinek, line 72:
```suggestion
      ``active_orbitals``. This allows selecting an active space which is not placed around the
```

mrossinek (approved):

PR #931, opened 2022-11-02
Docs: Fix method and scroll styles

javabster:
### Summary

This PR fixes the styles on method pages so that the headers are formatted with styles that match the class headers. It also removes the scroll functionality for large tables so that users in browsers like Chrome can see the full table without the "hidden" scroll

Before:
parameters etc. don't have nice headings
Screenshot 2022-11-02 at 1 58 52 PM

With this PR:
Screenshot 2022-11-02 at 1 54 26 PM

Before:
in chrome you can't see that you need to scroll to see the bottom of the table.
Screenshot 2022-11-02 at 1 58 25 PM

With this PR:
no need to scroll because table is shown in full
Screenshot 2022-11-02 at 1 54 37 PM

TODO: open same PR in qiskit and link here


mrossinek (approved):
Thank you, Abby, for this PR! I'm in favor of these changes so I am going to already approve these.

Given our offline discussion yesterday, I assume that Steve would first like to see these same changes be applied to Terra before merging things here though :+1:

javabster:
So I was taking a deeper look into the styles for terra yesterday, and I noticed that there are style inconsistencies even within a single repo. In short I found the following:

- Classes have styled headings with grey boxes, **_centred text_**,  and a purple line **_on top_**
- Functions have styled headings with grey boxes, **_left aligned text_**, and a purple line **_on the left_**
- Methods have no styled headings
- Properties have not styled headings

In this PR I attempted to add styles to the method headings in line with the styles for the class headings, but now I am wondering if instead I should have matched the styles for the function headings? You can see the difference in the screenshot below:

Screenshot 2022-11-03 at 10 06 37 AM


I'm going to open an issue in qiskit to discuss more and hopefully come to a decision, so maybe hold off on merging until then

woodsp-ibm:
>  I assume that Steve would first like to see these same changes be applied to Terra before merging things here though

More that I would just like it agreed. The docs were done with view to everyone using them, having the same style etc hence my reluctance. If we agree on the changes and they will get applied overall to Qiskit the fact that it may be staged, where maybe we do this Nature and ML first - seems fine.

manoelmarques:
@Mergifyio refresh

mrossinek:
I have converted this to draft until we figure out the timeline for this change. Please correct me if I'm wrong, but I am assuming that we wanted to await a matching PR in Terra?
@javabster any updates on that front?

javabster:
> Is this still needed after #979?

Yes, the latest sphinx release doesn't include the style fixes required here unfortunately. [This issue](https://github.com/Qiskit/qiskit/issues/1626) is still open to fix the styles but haven't gotten round to it yet. It might be worth closing this issue and fixing it in the theme then just bumping the theme version in nature, what do you think?

PR #938, opened 2022-11-04
Remove pin from matplotlib

manoelmarques:
### Summary
Remove pin as it has been fixed in Qiskit/rustworkx#681


mrossinek (approved):
Thanks, Manoel! :+1:

PR #939, opened 2022-11-04
Add support for Python 3.11

manoelmarques:
### Summary
Following Qiskit Terra, add support to python 3.11

### Details and comments

The following libraries don't support python 3.11 yet:
- ~PySCF~ 2.2.1 has 3.11 support for Linux and Mac
- Psi4
- PyQuante2
- Sparse because of numba numba/numba#8304
- Gaussian, it fails the build of file qcmatrixio.F: `f2py -c -m qcmatrixio qcmatrixio.F`
```
File "/home/manoel/anaconda3/envs/Qiskitenv311/lib/python3.11/site-packages/numpy/f2py/crackfortran.py", line 3369, in traverse
    new_result = visit(obj, parents, result, *args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/manoel/anaconda3/envs/Qiskitenv311/lib/python3.11/site-packages/numpy/f2py/crackfortran.py", line 3428, in character_backward_compatibility_hook
    assert parents[-3][0] == 'vars'
 ```

The pylint minimum  version now is 2.15.0

### CODE REVIEW

**Comment by manoelmarques** [comment]
@Mergifyio refresh

**Comment by manoelmarques** [comment]
@Mergifyio refresh

**Comment by mrossinek** [comment]
Some minor updates:
- I was able to generate the gaussian files under Python 3.11 using `numpy<1.24`. I will need to investigate more (and check the numpy repo for filed issues) why numpy 1.24 fails to do this properly.
- I was also able build a wheel for PySCF under 3.11 locally which passes at least all the driver tests

Progress on numba support for Python 3.11 also seems to be converging. I will revisit this PR in a few weeks with more time by when numba should hopefully have merged their 3.11 support and we then also will be able to use `sparse` under 3.11 :+1:

**Comment by mrossinek** [comment]
I have marked this as ready for review. However, tomorrow I still need to add:
- [ ] the compiled qcmatrixio library for Mac OS
- [ ] the compiled qcmatrixio library for Windows

File: releasenotes/notes/add-python311-support-3a939c4697ecd896.yaml [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+upgrade:
+  - |
+    Added support for running with Python 3.11.
+    At the the time of the release, Psi4, PyQuante2 and Sparse didn't have a python 3.11 version.
```
woodsp-ibm, line 5: Since PyQuante2 driver is part of deprecated code, and we do not have any newer driver based on it goings forwards, should we even mention it here. Maybe this should be a new feature section too rather than upgrade - I think upgrade is more about things the user should do to upgrade their code - e.g. the removal of deprecated code means they now have to upgrade it if they did not before.

woodsp-ibm (approved):

PR #964, opened 2022-11-15
Leverage opt_einsum for sparse-einsum support

### LINKED ISSUE #955: Leverage `opt_einsum` (if available) for sparse-einsum support
### What should we add?

The package `opt_einsum` supports einsum summation directly on `sparse` objects as described in its documentation: https://optimized-einsum.readthedocs.io/en/stable/index.html

This would allow to avoid the `.to_dense()` call for example here: https://github.com/Qiskit/qiskit-nature/blob/de01b704254592b7e311ac88b6a1430767e9cb43/qiskit_nature/second_q/operators/polynomial_tensor.py#L743

Two requirements should be met before attempting to use `opt_einsum.contract` in favor of `np.einsum`:

1. `opt_einsum` must be installed (optional)
2. `qiskit_nature.settings.optimize_einsum` must be enabled (default; but a toggle because the performance gain comes at the expense of memory consumption)

When implementing this not only the code in `PolynomialTensor` needs to be updated, but also the docstrings which mention the lacking `.einsum` support for sparse tensors.

manoelmarques:
### Summary
Closes #955

### DIFF
--- a/qiskit_nature/optionals.py
+++ b/qiskit_nature/optionals.py
@@ -88,4 +88,6 @@ HAS_SPARSE = LazyImportTester(
     install="pip install 'qiskit-nature[sparse]'",
 )
 
-HAS_OPT_EINSUM = LazyImportTester("opt_einsum", install="pip install opt_einsum")
+HAS_OPT_EINSUM = LazyImportTester(
+    "opt_einsum", name="Optimized einsum", install="pip install opt_einsum"
+)
--- a/qiskit_nature/optionals.py
+++ b/qiskit_nature/optionals.py
@@ -87,3 +87,5 @@ HAS_SPARSE = LazyImportTester(
     name="sparse",
     install="pip install 'qiskit-nature[sparse]'",
 )
+
+HAS_OPT_EINSUM = LazyImportTester("opt_einsum", install="pip install opt_einsum")
--- a/qiskit_nature/second_q/operators/polynomial_tensor.py
+++ b/qiskit_nature/second_q/operators/polynomial_tensor.py
@@ -30,6 +30,7 @@ from qiskit.quantum_info.operators.mixins import (
 
 from qiskit_nature.settings import settings
 import qiskit_nature.optionals as _optionals
+from qiskit_nature.utils import get_einsum
 
 if _optionals.HAS_SPARSE:
     # pylint: disable=import-error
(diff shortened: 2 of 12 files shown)


qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -740,12 +741,20 @@ def einsum(
         Returns:
             A new ``PolynomialTensor``.
         """
-        dense_operands = [op.to_dense() for op in operands]
+        if _optionals.HAS_OPT_EINSUM:
+            # pylint: disable=import-error
+            from opt_einsum import contract
```
woodsp-ibm, line 746: Do we always want to use when installed or just if the PT is not dense (i.e. has sparse array(s) in it). Elsewhere we use numpy einsum - so this is really only for this local use, because of sparse, unless things were changed to say have a more central einsum that worked either way that all the other usages could switch over to. `dense_operands` seems a bit misleading as name now - maybe more simply say `einsum_operands` as in the opt_einsum case these can be sparse.

Should we have/update a unit test covering this new path to assert when installed that the result if things were sparse going into the einsum it's sparse coming out of it too (as it stood before they would all be dense right - the PT equality test does not care about that aspect hence we might want to add a check to ensure things are as we expect when opt_einsum is sued.

qiskit_nature/optionals.py
```diff
@@ -97,3 +97,5 @@ def _is_available(self):
     name="sparse",
     install="pip install 'qiskit-nature[sparse]'",
 )
+
+HAS_OPT_EINSUM = LazyImportTester("opt_einsum", install="pip install opt_einsum")
```
mrossinek, line 101: This should have a `name` specified, too

qiskit_nature/second_q/operators/polynomial_tensor.py
mrossinek: I think if `opt_einsum` is available, we should always use it. Their docs state that `np.einsum(..., optimize=True)` is the upstreamed version of a lot of the features originally implemented by `opt_einsum`. But that their package still provides more up-to-date algorithms ([source](https://optimized-einsum.readthedocs.io/en/stable/#features)).

I agree that `dense_operands` is misleading now. Maybe we can just rename it to `_operands` or `operand_list` or similar.

I agree that the unittests should also be updated to test this.

woodsp-ibm: > I think if opt_einsum is available, we should always use it.

At present its only used here in PT right, where it was wanted to handle sparse better - we have other places numpy einsum is used, but there we do not use sparse at all. If we are using it on dense here too, if installed, did we try checking performance on dense - if its faster/more performant maybe it would be useful elsewhere too?

mrossinek (approved):
Thanks, Manoel! :+1:

PR #966, opened 2022-11-16
Remove negative coeff. from `SpinOp.index_order`

### LINKED ISSUE #962: Coefficient change in `SpinOp.index_order`
### Environment

- **Qiskit Nature version**: 0.5
- **Python version**: any
- **Operating system**: any

### What is happening?

The `SpinOp.index_order` method introduces a faulty sign change:
```python
>>> from qiskit_nature.second_q.operators import SpinOp
>>> op = SpinOp({"Z_1 Z_0": 1.0})
>>> op.index_order()
SpinOp({'Z_0 Z_1': (-1+0j)}, spin=1/2, num_spins=2, )
```

(see above)

### What should happen?

The sign should not be affected by the term order of the labels.

### Any suggestions?

The following line needs to be removed: https://github.com/Qiskit/qiskit-nature/blob/3f556988a51f9d8fd30b8628b13dbf6178d9da40/qiskit_nature/second_q/operators/spin_op.py#L555

There might potentially be other aspects to this so please double-check that and add a unittest for the `SpinOp.index_order` method.

ElePT:
### Summary
Fixed #962, added corresponding unit tests.


mrossinek (commented):
Generally this is looking good, but please also add a release note such that we can backport this :+1:

releasenotes/notes/fix-spin-op-index-order-024e76d21b08bb9a.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Removes sign change in :meth:`qiskit_nature.second_q.operators.spin_op.index_order` carried over from
+    :class:`~qiskit_nature.second_q.operators.FermionicOp`'s implementation. Adds corresponding unit
+    tests to verify behavior.
```
mrossinek, line 6:
```suggestion
    Removes wrong sign change from :meth:`qiskit_nature.second_q.operators.SpinOp.index_order`.
```
- fixes the method link
- removes the reference to `FermionicOp` as I think it is rather misleading here
- I do not think we need to mention the unittests as the end-user likely does not care about this development detail

mrossinek (approved):
Thanks, Elena!

Issue #967, opened 2022-11-16
`QubitConverter` Refactoring


mrossinek:
### What should we add?

We should refactor the `QubitConverter` and `QubitMapper` classes to achieve the following goals.

### Removal of `qiskit.opflow`

Qiskit Terra plans to deprecate `qiskit.opflow` in favor of `qiskit.quantum_info` in the next release (0.23). The primitives and new algorithms based on them already largely switched over to `quantum_info` objects and only kept support of the `PauliSumOp` because of its coupling to the `Z2Symmetries`. In this Epic, I will outline our plans around `Z2Symmetries` to allow us removing this final opflow dependence.

One step needs to be done in Terra and the corresponding issue is https://github.com/Qiskit/qiskit-terra/issues/9140.

Then we also need get rid of the [`TwoQubitReduction`](https://github.com/Qiskit/qiskit-terra/blob/205c7fe/qiskit/opflow/converters/two_qubit_reduction.py). Since this particular transformation is unique to the `ParityMapper` of Qiskit Nature, I suggest that we fully extract it from Qiskit Terra and migrate it to here.

### Enabling direct usage of `QubitMapper`s

In a far future, the usage of `Z2Symmetries` will not yield as much of a significant advantage as nowadays (think removal of a handful of qubits from a multi-hundred/multi-thousand qubit-sized problem). Thus, we should already prepare for an API where `QubitMapper` objects can be used directly.

Especially if we integrate the `TwoQubitReduction` directly into the `ParityMapper`, each mapper will be more useful as a standalone object, too.

### Making a standalone `TaperedQubitMapper`

With the changes proposed above, one could fully rethink how we can implement the tapering utilities evolving around the `Z2Symmetries` which will remain in Terra. In fact, one could envision a standalone `TaperedQubitMapper`, which takes any other mapper and all the information necessary to determine the Z2Symmetries during instantiation.
This class could then implement a 2-step mapping procedure where the operators are first mapped and then tapered. This would be a stateful mapper since the symmetries are determined based on the hamiltonian and subsequent additional operators are reduced accordingly.

If in the future we decide to remove `Z2Symmetries`, this class can simply be deprecated without further affecting the stack.

### Removing `QubitConverter`

Doing all of the above, will allow us to simply deprecate the `QubitConverter` as it will be replaced (in-place) by any `QubitMapper` instance. This also greatly simplifies the usage of simple mappers for any end-user :+1: 

<hr/>

### `MixedMapper`

[...]

PR #971, opened 2022-11-17
Qeom refactoring and upgrades

### LINKED ISSUE #52: QEOM energies differ for Statevector and Qasm simulator

### LINKED ISSUE #403: Evaluate `aux_operators` on excited states with `QEOM`

The `QEOM` algorithm currently does **not** evaluate the `aux_operators` on the excited states. This would be a good addition.

### LINKED ISSUE #415: QEOM evaluation of operators

### LINKED ISSUE #769: [refactor] improve QEOM code

The implementation of the `QEOM` algorithm looks like it could undergo some refactoring in order to make the code more maintainable. I do not have any specific targets in mind right now, but I am opening this issue to keep this in mind for some point in the future.

Some rough ideas include:
- improve the hopping operator generation and commutativity handling:
    - right now, this is split across 5+ methods which can surely be improved to some degree
    - the code needs to handle tapering manually. This results in a very hard to follow flow of logic. This can surely be streamlined quite a bit better
    - `parallel_map` is being used in multiple places. We should check if we can streamline this better in order to avoid multiple iterations over the same data
- generally add more detailed/explanatory comments in the code

Anthony-Gandon:
Closes #769
Closes #415
Closes #403
Closes #52 

### Summary
Proposition for a modification to QEOM internals to :
- Make the code easier to follow
- Improve the potential for parallelism
- Improve the performances of the classical calculations by using a custom treatment of the tapering.

A new functionality is also added to compute excited state properties and transition amplitudes of arbitrary observables.

The typical interface to use it would be the following:

```python
from qiskit.algorithms.optimizers import COBYLA

from qiskit_nature.units import DistanceUnit
from qiskit_nature.second_q.algorithms import VQEUCCFactory, GroundStateEigensolver
from qiskit_nature.second_q.algorithms.excited_states_solvers import QEOM
from qiskit_nature.second_q.circuit.library import UCCSD
from qiskit_nature.second_q.drivers import PySCFDriver
from qiskit_nature.second_q.mappers import QubitConverter
from qiskit_nature.second_q.mappers import JordanWignerMapper

import numpy as np

optimizer = COBYLA(maxiter=500,disp=False)
qubit_converter = QubitConverter(JordanWignerMapper(), z2symmetry_reduction=None, two_qubit_reduction=False)

new_driver = PySCFDriver(
    atom="H 0 0 0; H 0 0 1.735",
    basis="sto3g",
    charge=0,
    spin=0,
    unit=DistanceUnit.ANGSTROM,
)
new_es_problem = new_driver.run()
hamiltonian_op, _ = new_es_problem.second_q_ops()
aux_ops = {'hamiltonian':hamiltonian_op}

# Qeom results
vqe_solver = VQEUCCFactory(Estimator(), UCCSD(), optimizer)
me_gsc = GroundStateEigensolver(qubit_converter, vqe_solver)
qeom_solver = QEOM(me_gsc, estimator=Estimator(), excitations='sd', aux_eval_rules="all")
results_qeom = qeom_solver.solve(new_es_problem, aux_operators=aux_ops)

print("Aux QEOM")
for aux_op_eval in results_qeom.aux_operators_evaluated:
    print(aux_op_eval)

print("Transi QEOM")
for indice, value in results_qeom.raw_result.transition_amplitudes.items():
    print(indice, value)

```

### Details and comments

The following guides details some of the changes that are proposed
[QEOM_refactoring-3.pdf](https://github.com/Qiskit/qiskit-nature/files/10014630/QEOM_refactoring-3.pdf)


qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -145,97 +240,77 @@ def solve(
...
+
+        # 2. Run ground state calculation with fully tapered custom auxiliary operators
+        # Note that the solve() method includes the `second_q' auxiliary operators
+        tap_custom_aux_ops = self.qubit_converter._symmetry_reduce_clifford(
+            ListOrDict(pre_tap_custom_aux_ops), True
         )
-        matrix_operators_dict, size = self._prepare_matrix_operators(problem)
 
-        # 3. Evaluate eom operators
+        groundstate_result = self._gsc.solve(problem, tap_custom_aux_ops)
```
mrossinek, line 261: Why do you only include the `custom_aux_ops` here?

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -145,97 +240,77 @@ def solve(
...
+        (
+            pre_tap_qubit_op_main,  # Hamiltonian
+            pre_tap_default_aux_ops,  # Default auxiliary observables
+            pre_tap_custom_aux_ops,  # User-defined auxiliary observables
+        ) = self.get_qeom_qubit_operators(problem, aux_operators)
+
+        # 2. Run ground state calculation with fully tapered custom auxiliary operators
+        # Note that the solve() method includes the `second_q' auxiliary operators
+        tap_custom_aux_ops = self.qubit_converter._symmetry_reduce_clifford(
+            ListOrDict(pre_tap_custom_aux_ops), True
```
mrossinek, line 258: Ideally you can isolate the usage of `ListOrDict` to the `QubitConverter` and avoid the need to wrap it here.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -145,97 +240,77 @@ def solve(
...
-        return result
-
-    def _prepare_matrix_operators(self, problem: BaseProblem) -> Tuple[dict, int]:
-        """Construct the excitation operators for each matrix element.
+        # 5. Solve the pseudo-eigenvalue problem
+        energy_gaps, expansion_coefs, commutator_metric = self._compute_excitation_energies(
+            h_mat, s_mat
+        )
+        gammas_square: np.ndarray = np.abs(np.diagonal(commutator_metric))
+        logger.debug("Gamma square... %s", gammas_square)
```
mrossinek, line 277:
```suggestion
        logger.info("Gamma square = %s", gammas_square)
```

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -274,58 +349,36 @@ def _build_hopping_ops(
...
+    def _build_all_eom_operators(
+        self,
+        pre_tap_operator: PauliSumOp,
+        expansion_basis_data: Tuple[Dict[str, PauliSumOp], Dict[str, List[bool]], int],
     ) -> dict:
         """Building all commutators for Q, W, M, V matrices.
 
         Args:
-            hopping_operators: All hopping operators based on excitations_list,
+            pre_tap_operator: Pre tapered Hamiltonian operator
```
mrossinek, line 360:
```suggestion
            pre_tap_operator: The untapered Hamiltonian operator.
```

mrossinek: Maybe you can update the variable names to say `untap` instead of `pre_tap` :thinking:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -387,11 +451,11 @@ def _build_commutator_routine(
                 # theory, one would choose this according to the nature of the problem (i.e.
                 # whether it is fermionic or bosonic), but in practice, always choosing the
                 # anti-commutator has proven to be more robust.
-                q_mat_op = double_commutator(left_op, operator, right_op_1, sign=False)
+                q_mat_op = -double_commutator(left_op1, operator, right_op_1, sign=False)
                 # In the case of the single commutator, we are always interested in the energy
                 # difference of two states. Thus, regardless of the problem's nature, we will
                 # always use the commutator.
-                w_mat_op = commutator(left_op, right_op_1)
+                w_mat_op = -commutator(left_op1, right_op_1)
```
mrossinek, line 458: Where do the `-` signs come from?

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -509,28 +546,105 @@ def _build_eom_matrices(
         logger.debug("\nM:=========================\n%s", m_mat)
         logger.debug("\nV:=========================\n%s", v_mat)
 
-        return m_mat, v_mat, q_mat, w_mat, m_mat_std, v_mat_std, q_mat_std, w_mat_std
+        # Matrix building rules
+        # h_mat = [[M, Q], [P, N]] and s_mat = [[V, W], [T, U]]
+        # N = M.conj() = M.T
```
mrossinek, line 551: Is this the case by construction?

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -509,28 +546,105 @@ def _build_eom_matrices(
...
+        # P = Q.adjoint() = Q.conj()
```
mrossinek, line 552:
```suggestion
        # P = Q.adjoint() = Q.T.conj()
```

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -509,28 +546,105 @@ def _build_eom_matrices(
...
+        s_mat_std: np.ndarray = np.array([[v_mat_std, w_mat_std], [w_mat_std, v_mat_std]])
+
+        return h_mat, s_mat, h_mat_std, s_mat_std
+
+    def _prepare_expansion_basis(
+        self, problem: BaseProblem
+    ) -> Tuple[Dict[str, PauliSumOp], Dict[str, List[bool]], int]:
+        """Prepares the basis expansion operators by calling the builder for second quantized operator
+        and applying transformations (Mapping, Reduction, First step of the tapering).
+
```
mrossinek, line 568:
```suggestion
```

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -542,30 +656,265 @@ def _compute_excitation_energies(
         # Since we may now have
         # small values (positive or negative) take the absolute and then threshold zero.
         logger.debug("... %s", res[0])
-        w = np.sort(np.real(res[0]))
+        order = np.argsort(np.real(res[0]))[::-1][len(res[0]) // 2 : :]
```
mrossinek, line 659: Please add a comment explaining what this does for a future reader of this code

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -542,30 +656,265 @@ def _compute_excitation_energies(
...
+        w = np.real(res[0])[order]
         logger.debug("Sorted real parts %s", w)
-        w = np.abs(w[len(w) // 2 :])
+        w = np.abs(w)
         w[w < 1e-06] = 0
```
mrossinek, line 663: Can we make this threshold configurable?

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -542,30 +656,265 @@ def _compute_excitation_energies(
...
         self._ground_state_raw_result = None
         self._excitation_energies: Optional[np.ndarray] = None
         self._expansion_coefficients: Optional[np.ndarray] = None
-        self._m_matrix: Optional[np.ndarray] = None
-        self._v_matrix: Optional[np.ndarray] = None
-        self._q_matrix: Optional[np.ndarray] = None
-        self._w_matrix: Optional[np.ndarray] = None
-        self._v_matrix_std: float = 0.0
-        self._q_matrix_std: float = 0.0
-        self._w_matrix_std: float = 0.0
```
mrossinek, line 568: Unfortunately I don't think we can simply remove these result objects without formally deprecating them..
Can we extract these back from the `H` and `S` matrix to avoid making the actual runtime code ugly and instead extract the logic of finding these into this result object to formally deprecate these?

mrossinek (commented):
I think this PR is overall already in great shape. Thanks a lot Anthony for all the work here!

These are quite a few comments and I will need to do another pass of the actual QEOM code in a little bit more detail with some more time on my hands. But I figured I should post my review as it is now to give you time to work on these comments.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom_electronic_ops_builder.py
```diff
@@ -107,9 +107,7 @@ def _build_single_hopping_operator(
         label.append(f"+_{occ}")
     for unocc in excitation[1]:
         label.append(f"-_{unocc}")
-    fer_op = FermionicOp(
-        {" ".join(label): 4.0 ** len(excitation[0])}, num_spin_orbitals=2 * num_spatial_orbitals
-    )
+    fer_op = FermionicOp({" ".join(label): 1.0}, num_spin_orbitals=2 * num_spatial_orbitals)
```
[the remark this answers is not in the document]
Anthony-Gandon: Both should work the same, I got used to using the latter because it normalises the S matrix in HX = SXE but I don't mind sticking with the original choice

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_solver.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, qubit_converter: QubitConverter) -> None:
                              according to a mapper it is initialized with.
         """
         self._qubit_converter = qubit_converter
+        self._solver: Any = None
```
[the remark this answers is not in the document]
Anthony-Gandon: It is a quick fix to a mypy issue I had: In qEOM the factories ned an additional step in the form of :
if isinstance(self.solver, MinimumEigensolverFactory):
             self._gsc._solver = self.solver.get_solver(problem, self.qubit_converter)
This used to fail because the groundstatesolver does not have a _solver attribute

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -121,21 +139,98 @@ def excitations(
...
+        aux_default_ops = self.qubit_converter.convert_match(aux_second_q_ops)
+        aux_custom_ops = {}
+        if aux_operators is not None:
+            for name_aux, aux_op in aux_operators.items():
+                if isinstance(aux_op, (SparseLabelOp)):
+                    converted_aux_op = self.qubit_converter.convert_match(aux_op)
+                else:
+                    converted_aux_op = aux_op
+
+                aux_custom_ops[name_aux] = converted_aux_op
```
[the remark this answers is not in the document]
Anthony-Gandon: One issue is that the GroundstateEigensolver returns an error when it is passed a dictionary of aux operators containing the default second_q operators. Else I would also prefer treating them simultaneously

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -80,26 +86,38 @@ def __init__(
...
                     :meth:`generate_vibrational_excitations`, when solving an
                     :class:`.ElectronicStructureProblem` or a :class:`.VibrationalStructureProblem`,
                     respectively.
+
+            aux_eval_rules: The rules determining how observables should be evaluated on excited states.
+
+                :`str`: specific predefined rules. Allowed strings are:
+
+                    + `all` to compute all expectation values and all transition amplitudes
+                    + `diag` to only compute expectation values
```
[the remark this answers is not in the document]
Anthony-Gandon: Sure

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: Previously answered: the `_gsc.solve` fail if the aux_op contains on of the default keys "AngularMomentum, ...". This is the only reason why I'm splitting the custom and default aux_ops

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: Indeed this naming is not ideal: the three step currently are uncap, pretap, tap. THis is due to the splitting of the tapering into two parts

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: If found inconsistencies in the management of signs so my changes aim at following closely the theoretical formulas of the QEOM paper.
In this case the signs come from Eq(4) of the paper https://arxiv.org/abs/1910.12890. Previously, the expansion operators were built as E and -E^dag instead of E and E^dag, which explains why the sign was not needed

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: Yes this comes from analytics. Note that in many cases the previous definitions worked as W is often the 0 matrix, and V is often real, but my comments describe the only expected properties of these matrices for arbitrary expansion operators (ie the most general case).

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: As stated above, Q = Q.T. Maybee I should still remind it

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: I'm fine with extracting M,V,Q,W from H and S in the result class. This seems like an easy fix

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom_electronic_ops_builder.py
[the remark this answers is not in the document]
mrossinek: You can leave it as it is now. I just wanted to know :+1:

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_solver.py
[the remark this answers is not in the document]
mrossinek: Okay I see. Can you constrain the type to more than just `Any`? If yes, we can keep it in :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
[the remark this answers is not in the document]
mrossinek: Oh because of clashing names? Where exactly do you pass these on?

I'm assuming that you provide these via the `aux_operators` to the ground-state solver and then it complains because it also gets the default ones out of the problem. If so I would rather add a precedence logic which means that custom aux operators will overwrite the built-in ones if the names collide.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
mrossinek: Oh okay yes I see now. See the previous thread for a suggestion on how to deal with this

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
mrossinek: Okay thanks for the explanation :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
mrossinek: Okay I see. The reason I made the change suggestion here is because this is what you implemented in the code below :+1:

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_solver.py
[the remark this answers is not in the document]
Anthony-Gandon: Yes I can have "MinimumEigensolver | MinimumEigensolverFactory | None", I thought Any was avoiding unnecessary imports

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -122,20 +151,81 @@ def excitations(
...
+        if isinstance(self.solver, MinimumEigensolverFactory):
+            self._gsc._solver = self.solver.get_solver(problem, self.qubit_converter)
+
+        return untap_main_op, untap_aux_ops
 
     def solve(
         self,
         problem: BaseProblem,
-        aux_operators: Optional[dict[str, SparseLabelOp]] = None,
+        aux_operators: Optional[dict[str, Union[SparseLabelOp, PauliSumOp]]] = None,
```
mrossinek, line 228: Actually we need to update all explicit mentions of `PauliSumOp` to instead use `qiskit_nature.second_q.algorithms.ground_state_solvers.ground_state_solver.QubitOperator`. The reason being that is the union of `PauliSumOp` (which will be pending deprecated soon) and `SparsePauliOp` which we need to focus on here :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -374,8 +425,8 @@ def _build_commutator_routine(
             The indices of the matrix element and the corresponding qubit
             operator for each of the EOM matrices.
         """
-        m_u, n_u, left_op, right_op_1, right_op_2 = params
-        if left_op is None or right_op_1 is None and right_op_2 is None:
+        m_u, n_u, left_op1, right_op_1, right_op_2 = params
```
mrossinek, line 428:
```suggestion
        m_u, n_u, left_op_1, right_op_1, right_op_2 = params
```
Very nitpicky I know.. but you did it so nicely above, too.. :smile:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -509,63 +533,404 @@ def _build_eom_matrices(
...
+        qeom_result.excitation_energies = energy_gaps
+
+        qeom_result.m_matrix = h_mat[: len(h_mat) // 2, : len(h_mat) // 2]
+        qeom_result.v_matrix = s_mat[: len(h_mat) // 2, : len(h_mat) // 2]
+        qeom_result.q_matrix = h_mat[len(h_mat) // 2 :, : len(h_mat) // 2]
+        qeom_result.w_matrix = s_mat[len(h_mat) // 2 :, : len(h_mat) // 2]
+        qeom_result.m_matrix_std = h_mat_std[0, 0]
+        qeom_result.v_matrix_std = s_mat_std[0, 0]
+        qeom_result.q_matrix_std = h_mat_std[0, 1]
+        qeom_result.w_matrix_std = s_mat_std[0, 1]
```
mrossinek, line 881: This is good, thank you :+1: 

@manoelmarques what would the deprecation strategy look like here, if I want to make these read-only properties of the result object. Doing this would allow us to put this computation logic into the property getters and ensure that the matrices stored in the result are always consistent (plus it simplifies the code here a bit).

test/second_q/algorithms/excited_state_solvers/test_excited_states_solvers.py
```diff
@@ -139,6 +99,20 @@ def _solve_with_vqe_mes(self, converter: QubitConverter):
...
+        QubitConverter(ParityMapper()),
+        QubitConverter(ParityMapper(), two_qubit_reduction=True),
+        QubitConverter(ParityMapper(), z2symmetry_reduction="auto"),
+        QubitConverter(ParityMapper(), two_qubit_reduction=True, z2symmetry_reduction="auto"),
+        QubitConverter(BravyiKitaevMapper()),
+        QubitConverter(BravyiKitaevMapper(), z2symmetry_reduction="auto"),
+    )
+    def test_vqe_mes(self, qubit_converter):
+        """Test VQEUCCSDFactory with QEOM + Various mappings"""
+        self._solve_with_vqe_mes(qubit_converter)
```
mrossinek, line 114: I think you can just put the content of that method in here directly. Or is it still being used elsewhere and I am overlooking it?

mrossinek (commented):
A few more comments from my side :+1: 

This is shaping up very nicely! I am skipping the changes done to the `QubitConverter` on purpose, since it will be refactored again as part of #967 at which point we can revert these unreleased changes and update the QEOM code accordingly. I would still like to get this long-standing PR out before the `QubitConverter` refactor hits because it allows us to do these two things largely in parallel :+1:

(the above is just a comment for a potential other reviewer coming across this)

qiskit_nature/second_q/algorithms/excited_states_solvers/excited_states_solver_eval_rules.py
```diff
@@ -0,0 +1,25 @@
...
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""The definition of the available evaluation rules for excited states solver."""
+
+from enum import Enum
+
+
+class EvaluationRules(Enum):
```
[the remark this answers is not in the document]
Anthony-Gandon: Currently I agree that these rules are only used for QEOM, but in the long run they seem general enough to be extended to other solvers as well. For the moment, I'm changing to QEOMEvaluationRule.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -14,38 +14,48 @@
 
 from __future__ import annotations
 
-from typing import Any, Callable, List, Union, Optional, Tuple, Dict
+from typing import Any, Callable, Sequence, Union, Optional, cast
```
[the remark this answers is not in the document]
Anthony-Gandon: Done

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -88,12 +100,29 @@ def __init__(
...
+
+            eigenvalue_treshold: Threshold for the qEOM eigenvalues. This plays a role when one
+                excited state approaches the ground state, in which case it is best to avoid manipulating
+                very small absolute values.
         """
         self._gsc = ground_state_solver
         self._estimator = estimator
         self.excitations = excitations
+        self._aux_eval_rules = aux_eval_rules
+        self.eigenvalue_treshold = eigenvalue_treshold
```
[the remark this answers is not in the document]
Anthony-Gandon: I tried to add some more text to the docstring. I'll keep this comment open until we think it is enough.

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -428,6 +428,37 @@ def _two_qubit_reduce(
 
         return reduced_op
 
+    def two_qubit_reduce(
```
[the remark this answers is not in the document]
Anthony-Gandon: SO actually two_qubit_reduce is closer to `convert_match()` and to make them similar I need 1) that the mapper can accept a PauliSumOp and act as identity 2) that I unset the symmetries before calling convert_match and reset them right after. THis is what I've done in QEOM

releasenotes/notes/refactors-and-upgrades-qeom-bd14d85c94e912d0.yaml
```diff
@@ -0,0 +1,65 @@
...
+      for aux_op_eval in results_qeom.aux_operators_evaluated:
+          print(aux_op_eval)
+  - |
+    Added public methods :meth:`qiskit_nature.second_q.mappers.qubit_converter.symmetry_reduce_clifford` 
+    and :meth:`qiskit_nature.second_q.mappers.qubit_converter.convert_clifford` and 
+    :meth:`qiskit_nature.second_q.mappers.qubit_converter.find_taper_op` to allow a step by step 
+    tapering of operators.
+  - |
+    Changed the behavior of the 
+    :class:`qiskit_nature.second_q.algorithms.ground_state_solvers.ground_state_eigensolver` to not raise
```
mrossinek, line 56:
```suggestion
    :class:`~qiskit_nature.second_q.algorithms.GroundStateEigensolver` to not raise 
```

releasenotes/notes/refactors-and-upgrades-qeom-bd14d85c94e912d0.yaml
```diff
@@ -0,0 +1,65 @@
+---
+features:
+  - |
+    Improves the QEOM code and implements the calculation of excited state properties and transition 
+    amplitudes with QEOM.
```
mrossinek, line 5: Can you please make `QEOM` a link like the ones I corrected further down in the release note?

releasenotes/notes/refactors-and-upgrades-qeom-bd14d85c94e912d0.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - |
+    Changed the behavior of the 
+    :class:`qiskit_nature.second_q.algorithms.ground_state_solvers.ground_state_eigensolver` to not raise 
+    an error when the user specifies a auxiliary operator which name clashes an internally constructed
+    operator's name. The new behavior is to apply precedence to the user-defined operators over the 
+    builtin ones in case of conflicts. A warning will be logged when this case happens.
+  - |
+    Refactored the treatment of symmetries in the QEOM to reduce the classical overhead.
+    The parallelism was also reworked to improve the performances of the current hybrid 
+    sequential+parallel approach.
```
mrossinek, line 63: I don't think this needs to be mentioned explicitly here.

releasenotes/notes/refactors-and-upgrades-qeom-bd14d85c94e912d0.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - |
+    Added a tolerance parameter to control the eigenvalue threshold in the QEOM calculation.
```
mrossinek, line 65: Can you link to `QEOM` or even to the `tol` attribute of that class?
Should be something along the lines of:
```
:attr:`~qiskit_nature.second_q.algorithms.excited_states_solvers.QEOM.tol`
```

mrossinek (commented):
This is looking very good, thanks a lot, Anthony!

I have a last round of comments, mostly on some docstring details :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
manoelmarques: If you put the calculating logic in the init method for each property, then you just need the deprecation msg in the setters. This way the user still can use the deprecated setter which would override the calculation done in the init.

mrossinek: Thanks, that's a good idea :+1: @Anthony-Gandon could you do that, please?

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -14,44 +14,87 @@
...
+
+    ALL = "all"
+    DIAG = "diag"
+
+
 class QEOM(ExcitedStatesSolver):
-    """The calculation of excited states via the qEOM algorithm."""
+    """
+    The calculation of excited states via the qEOM algorithm.
+    This algorithms extends a ``GroundStateSolver`` object to identify good approximations to the excited
```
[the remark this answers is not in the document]
Anthony-Gandon: Done

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -14,44 +14,87 @@
...
+    In addition to the excited state energies, one can calculate other properties of the excited states
+    at the cost of additional measurements to perform on the ground state.
+    This requires providing the desired auxiliary observables to the ``solve`` method and having
+    specified the auxiliary evaluation rules of the ``QEOM`` object.
+
+    The following attributes can be read and updated once the ``QEOM`` object has been
+    constructed.
+
+    Attributes:
+        aux_eval_rules (EvaluationRule, dict[str, list[tuple[int, int]]], None): The rules determining
```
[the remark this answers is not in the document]
Anthony-Gandon: Fixed

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -122,20 +185,91 @@ def excitations(
...
-    ) -> Tuple[PauliSumOp, Optional[dict[str, PauliSumOp]]]:
-        return self._gsc.get_qubit_operators(problem, aux_operators)
+        aux_operators: dict[str, SparseLabelOp | QubitOperator] | None = None,
+    ) -> tuple[QubitOperator, dict[str, QubitOperator] | None]:
+        """
+        Gets the operator and auxiliary operators, and transforms the provided auxiliary operators.
+        If the user-provided ``aux_operators`` contain a name which clashes with an internally
+        constructed auxiliary operator, then the corresponding internal operator will be overridden by
+        the user-provided operator. Note: the names used for the internal auxiliary operators correspond
+        to the `Property.name` attributes which generated the respective operators.
```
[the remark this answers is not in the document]
Anthony-Gandon: Agreed

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -88,12 +134,29 @@ def __init__(
...
+            aux_eval_rules: The rules determining how observables should be evaluated on excited states.
+
+                :`Enum`: specific predefined rules. Allowed rules are:
+
+                    + ALL to compute all expectation values and all transition amplitudes
+                    + DIAG to only compute expectation values
+
+                :`dict[str, list[tuple]]`: Dictionary mapping valid auxiliary operator's name to lists
+                    of tuple (i, j) specifying the indices of the excited states to be evaluated on. By
+                    default, none of the auxiliary operators are evaluated on none of the excited states.
```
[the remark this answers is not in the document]
Anthony-Gandon: I added it in the doctsring of the aux_eval_rule parameter:

```
aux_eval_rules: The rules determining how observables should be evaluated on excited states.
  By default, none of the auxiliary operators are evaluated on none of the excited states.

  :`Enum`: specific predefined rules. Allowed rules are:
  
    + ALL to compute all expectation values and all transition amplitudes.
    + DIAG to only compute expectation values.
    
  :`dict[str, list[tuple[int, int]]]`: Dictionary mapping valid auxiliary operator's name
    to lists of tuple (i, j) specifying the indices of the excited states to be evaluated
    on.

```

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_eigensolver.py
```diff
@@ -87,14 +86,16 @@ def get_qubit_operators(
         problem: BaseProblem,
         aux_operators: dict[str, SparseLabelOp | QubitOperator] | None = None,
     ) -> tuple[QubitOperator, dict[str, QubitOperator] | None]:
-        """Gets the operator and auxiliary operators, and transforms the provided auxiliary operators."""
+        """Gets the operator and auxiliary operators, and transforms the provided auxiliary operators.
+        If the user-provided ``aux_operators`` contain a name which clashes with an internally
+        constructed auxiliary operator, then the corresponding internal operator will be overridden by
+        the user-provided operator. Note: the names used for the internal auxiliary operators correspond
+        to the `Property.name` attributes which generated the respective operators."""
```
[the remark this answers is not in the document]
Anthony-Gandon: Fixed

releasenotes/notes/refactors-and-upgrades-qeom-bd14d85c94e912d0.yaml
```diff
@@ -0,0 +1,65 @@
...
+      )
+      es_problem = driver.run()
+      hamiltonian_op, _ = es_problem.second_q_ops()
+      aux_ops = {"hamiltonian": hamiltonian_op}
+
+      # Qeom results
+      vqe_solver = VQEUCCFactory(Estimator(), UCCSD(), optimizer)
+      me_gsc = GroundStateEigensolver(qubit_converter, vqe_solver)
+      qeom_solver = QEOM(
+          me_gsc, estimator=Estimator(), excitations="sd", aux_eval_rules="all"
```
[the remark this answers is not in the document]
Anthony-Gandon: Should we also add the EvaluationRule in the `excited_states_solver.__init__.py` file ? Else we need a two line import :
```
from qiskit_nature.second_q.algorithms.excited_states_solvers import QEOM
from qiskit_nature.second_q.algorithms.excited_states_solvers.qeom import EvaluationRule

```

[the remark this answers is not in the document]
mrossinek: If the name of the `Enum` class does not reflect it is currently coupled to `QEOM` it should stay only in the `.qeom` module.
If you want to put it in the module `__init__` I would argue it should be called `QEOMEvaluationRule`

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: As discussed with @mrossinek , I could not put the calculating logic in the init method. I changed the getters instead.

mrossinek (commented):
Okay some final (minor) observations but other than that this looks good to me :+1: 
We will just need to keep in mind that, once we progress on #972 and #999, we should revert the public API changes (as much as possible) done to the `QubitConverter` in this PR.
We can see about the details of that as part of working on the epic mentioned above.

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_eigensolver.py
```diff
@@ -87,14 +86,25 @@ def get_qubit_operators(
         problem: BaseProblem,
         aux_operators: dict[str, SparseLabelOp | QubitOperator] | None = None,
     ) -> tuple[QubitOperator, dict[str, QubitOperator] | None]:
-        """Gets the operator and auxiliary operators, and transforms the provided auxiliary operators."""
+        """Gets the operator and auxiliary operators, and transforms the provided auxiliary operators.
```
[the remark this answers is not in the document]
Anthony-Gandon: Done

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -521,6 +540,94 @@ def _symmetry_reduce(
 
         return tapered_qubit_ops
 
+    def symmetry_reduce_clifford(
+        self,
+        converted_ops: ListOrDictType[PauliSumOp],
+        check_commutes: bool,
```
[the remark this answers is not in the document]
Anthony-Gandon: Yes makes sense

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -14,44 +14,87 @@
...
+    projected Hamiltonian will be solved. This method typically works well for calculating the
+    lowest-lying excited states of a problem.
+    The excited-state energies are calculated by default in this algorithm for all excited states.
+    Auxiliary observables can be specified to the ``solve`` method along with auxiliary evaluation
+    rules of the ``QEOM`` object.
+
+    The following attributes can be read and updated once the ``QEOM`` object has been
+    constructed.
+
+    Attributes:
```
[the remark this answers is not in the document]
Anthony-Gandon: I don't really know how to deal with it: excitation is not a public attribute in the sense of aux_eval_rules for example. A setter and getter for excitation were defined and because the setter does an additional check, it is used in the init (self.excitations = excitations)
But if I add a docstring for excitations as an attribute, then it creates a conflict with the getter/setter doctoring

releasenotes/notes/refactors-and-upgrades-qeom-bd14d85c94e912d0.yaml
```diff
@@ -0,0 +1,63 @@
...
+      vqe_solver = VQEUCCFactory(Estimator(), UCCSD(), optimizer)
+      me_gsc = GroundStateEigensolver(qubit_converter, vqe_solver)
+      qeom_solver = QEOM(
+          me_gsc, estimator=Estimator(), excitations="sd", aux_eval_rules=EvaluationRule.ALL
+      )
+      results_qeom = qeom_solver.solve(es_problem, aux_operators=aux_ops)
+
+      print("Aux QEOM")
+      for aux_op_eval in results_qeom.aux_operators_evaluated:
+          print(aux_op_eval)
```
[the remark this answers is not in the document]
Anthony-Gandon: I know iterate over the states and over the operators to show a full example of how to access results.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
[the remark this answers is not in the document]
mrossinek: Hm I actually overlooked that setter, sorry..

Looking at it now though I am a bit confused. The setter retricts the value a lot more than what the `__init__` suggests is possible. Thus, I would argue that this check is in fact obsolete and ought to be outright removed.
With that check removed, you can indeed make this simply a public attribute.

Could you test this quickly by using for example numbers as an input? As things are now I suspect them to break on that check although they should work..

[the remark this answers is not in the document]
Anthony-Gandon: As discussed offline, this test is only required to defend against using triple excitations (ok for UCC but not tested for QEOM)

Anthony-Gandon:
I'm not sure why we did not have a mypy error previously when 
```
if isinstance(self.solver, MinimumEigensolverFactory):
    self._gsc._solver = self.solver.get_solver(problem, self.qubit_converter)
```
(because self._gscc is GroundStateSolver and does not have a _solver attribute, only GroundStateEigensolver has). Can we add a #type: ignore because anyway this part will be subject to changes in the near future.

mrossinek (approved):
I am tentatively approving this PR because I think it looks good as it is.

~I am not quite sure what you meant in your last comment though, so I will hold off with merging until that is cleared up either here or offline.~

EDIT: the questionable type ignore statement is already included in this PR and fine by me :+1: 

Thanks a lot for all your work, Anthony!

PR #981, opened 2022-11-22
Simplify pyscf installation check

manoelmarques:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -115,7 +115,6 @@ html_css_files = ["style.css", "custom.css", "gallery.css"]
 nbsphinx_timeout = 360
 nbsphinx_execute = os.getenv("QISKIT_DOCS_BUILD_TUTORIALS", "never")
 nbsphinx_widgets_path = ""
-exclude_patterns = ["_build", "**.ipynb_checkpoints"]
 nbsphinx_thumbnails = {}
 
 spelling_word_list_filename = "../.pylintdict"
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -115,7 +115,6 @@
 nbsphinx_timeout = 360
 nbsphinx_execute = os.getenv("QISKIT_DOCS_BUILD_TUTORIALS", "never")
 nbsphinx_widgets_path = ""
-exclude_patterns = ["_build", "**.ipynb_checkpoints"]
```
mrossinek, line 118: Why are we removing this? I don't object, I would just like to understand :smile:

qiskit_nature/second_q/drivers/pyscfd/pyscfdriver.py
```diff
@@ -295,7 +295,6 @@ def chkfile(self, chkfile: str) -> None:
         self._chkfile = chkfile
 
     @staticmethod
-    @_optionals.HAS_PYSCF.require_in_call
```
mrossinek, line 298: Just to check: you are removing this because this is already checked via the `require_in_instance` decorator of the class?

mrossinek (commented):
I would just like to clarify two aspects. Overall I'm good with these changes :+1:

docs/conf.py
manoelmarques: This is a duplicate entry :-) There is already the same statement below.

qiskit_nature/second_q/drivers/pyscfd/pyscfdriver.py
manoelmarques: Yes, no need to check twice

mrossinek (approved):
Thanks, Manoel!

PR #983, opened 2022-11-24
create testing module and move random.py in there

kevinsung:
### Summary
This change was discussed offline with @mrossinek . The reasons for this change are:
- Future PRs may use randomness utilities within the library
- Random sampling utilities are useful so having these exposed as part of the library is beneficial to users.


qiskit_nature/testing/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Testing utilities (:mod:`qiskit_nature.testing`)
```
mrossinek, line 14: In order for this section to show up in the docs, you will need to add a reference in `qiskit_nature/__init__.py`.
There is a `Submodules` section which currently contains `second_q` and `utils`. I suggest you add it between these two sections to preserve alphabetical order.

Besides that, we will need to see whether a stubs file is needed in `docs/apidocs`. You can test that by building the docs locally using `make html`

mrossinek (commented):
This looks good to me besides this single suggestion.

Please also add a release note which lists the addition of this module in the `features` section.

qiskit_nature/testing/__init__.py
kevinsung: Done.

mrossinek (commented):
Thanks for fixing the docs :+1: 

Now all that is left is a release note as I mentioned before.

While you are at it, could you please check the three method docstrings for consistency? (ending lines with dots, capitalizing argument descriptions, etc.) There is not much to change but while scrolling through I noticed some minor details.

kevinsung:
I've added the release note and fixed the docstrings.

qiskit_nature/testing/random.py
```diff
@@ -30,10 +30,10 @@ def parse_random_seed(seed: Any) -> np.random.Generator:
     Args:
         seed: The pseudorandom number generator or seed. Should be an
             instance of `np.random.Generator` or else a valid input to
-            `np.random.default_rng`
+            `np.random.default_rng`.
```
woodsp-ibm, line 33: If we are going to expose this more for end user use, can we find a different/better name? Somehow parse_random_seed does not do not for me - maybe its just me.... Some thoughts get_rng, get_random, random. And maybe we can state what happens with seed is None in the docstring since that returns the Terra algorithms rng rather than say seed the np default_rng. Can we address the TODO if making this more public?

kevinsung: Thanks for mentioning the `None` behavior. I forgot that I had intended to change it to match the numpy default RNG; I have made this change now. The reason is that after discussion with the Terra developers, this was not an intended usage of the Terra `algorithms` RNG; see https://github.com/Qiskit/qiskit-terra/pull/8156.

kevinsung: As for the name, I chose `parse_random_seed` because it precisely describes what the method does: it takes a representation of an RNG seed and resolves it into an actual RNG. In my opinion, the names you suggested are too vague, though `get_rng` might be okay.

kevinsung: Actually, this change made me realize that the function is no longer needed since `np.random.default_rng` can be used directly. I'll delete it.

mrossinek (approved):
Thank you, Kevin! :+1:

PR #991, opened 2022-12-06
Fix CI aer build

manoelmarques:
### Summary
Github actions updated Ubuntu-latest OS so it needed a change in the Aer build


mrossinek (approved):
Thanks, Manoel!

Issue #974, opened 2022-11-17
Implement a `TaperedQubitMapper`


mrossinek:
### What should we add?

As the third step of #967 we should implement a `TaperedQubitMapper` which gathers all the `Z2Symmetries`-related tapering techniques currently controlled via the `QubitConverter` in a single `QubitMapper` wrapper:

```python
class TaperedQubitMapper:
    def __init__(self, mapper: QubitMapper, z2symmetries: list[int]):
        ...
        
    @classmethod
    def from_problem(cls, mapper: QubitMapper, problem: ElectronicStructureProblem):
        # implies the previous z2symmetries == "auto" case
        # extract problem.hamiltonian and problem.symmetry_sector_locator
        
    def map(self):
        # uses self.mapper.map + applies tapering
        
    # provide 2-step tapering i.e. one method for _only_ mapping and one to apply tapering on-top
    # this is necessary for cases like the HF bitstring and qEOM
```

Just like with my proposal in #973, this mapper will be stateful.

<hr/>

This is blocked by https://github.com/Qiskit/qiskit-terra/issues/9140

Anthony-Gandon:
- [ ] Once we work on that mapper we need to think about how to handle the suppress_none=False case
which we had in the QubitConverter.

PR #999, opened 2022-12-08
Update API to allow QubitMapper where QubitConverter is used.

### LINKED ISSUE #972: Update API to allow `QubitMapper` where `QubitConverter` is used
### What should we add?

This will be the first step of #967 because it can be largely parallelized with https://github.com/Qiskit/qiskit-terra/issues/9140.

The idea here is to update all occurrences of `QubitConverter` in the public API to also allow direct usage of `QubitMapper`s. This will require potential `isinstance` checks to differentiate the two cases from another, but for the transition period while `QubitConverter` is being deprecated there is no way around this.

Until later parts of #967 are completed, some mappers will not provide the full functionality that is available through the `QubitConverter` but for the sake of not having to do all the changes at once in one massive PR this is okay.

Unittests should be added to assert the working cases.

Anthony-Gandon:
### Summary

Closes #972 

This PR updates the API to allow `QubitMapper` object where `QubitConverter` were previously required. THis is possible because the `QubitConverter` is a wrapper around the `QubitMapper` with additional methods for qubit reduction (two qubit reduction and tapering).  The utility method `map()` was modified in the QubitMapper to act on lists, as the `convert_match()` does for the QubitConverter.

The following gist goes through a the main classes where this functionality is accessible: [gist](https://gist.github.com/Anthony-Gandon/96285641fc2520ddf411d9f466ea7b7e)

This PR is going to be affected by #971, but I wanted to initiate the reflexion and feedback process.


qiskit_nature/second_q/algorithms/excited_states_solvers/excited_states_eigensolver.py
```diff
@@ -34,7 +34,7 @@ class ExcitedStatesEigensolver(ExcitedStatesSolver):
 
     def __init__(
         self,
-        qubit_converter: QubitConverter,
+        qubit_converter: Union[QubitConverter, QubitMapper],
```
mrossinek, line 37:
```suggestion
        qubit_converter: QubitConverter | QubitMapper,
```
We should try to do it in this way consistently. I'll only mention it here once as a stand-in

mrossinek, line 37: In the future I would like to rename this argument to `qubit_mapper`. This will not be an easy process so it will be done in a separate PR once all the API is updated to support `QubitMapper` and works. I'm mentioning it now just as an FYI.
This will not be an easy undertaking but I will start discussing some plans on how to do this properly offline.

qiskit_nature/second_q/algorithms/excited_states_solvers/excited_states_eigensolver.py
```diff
@@ -70,19 +70,27 @@ def get_qubit_operators(
...
+            main_operator = self._qubit_converter.convert(
+                main_second_q_op,
+                num_particles=num_particles,
+                sector_locator=problem.symmetry_sector_locator,
+            )
+            aux_ops = self._qubit_converter.convert_match(aux_second_q_ops)
+
+        else:
+            main_operator = self._qubit_converter.map(main_second_q_op)
+            aux_ops = self._qubit_converter.map_all(aux_second_q_ops)
```
mrossinek, line 83: I think this is pretty much how I would expect it to always look :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/excited_states_solver.py
```diff
@@ -58,7 +58,7 @@ def get_qubit_operators(
     ) -> Tuple[PauliSumOp, Optional[dict[str, PauliSumOp]]]:
         """Construct qubit operators by getting the second quantized operators from the problem
         (potentially running a driver in doing so [can be computationally expensive])
-        and using a QubitConverter to map and reduce the operators to qubit operators.
+        and using a QubitConverter or QubitMapper to map and reduce the operators to qubit operators.
```
mrossinek, line 61: We will need to do a thorough pass of the docs. Just mentioning this so its on our radar (I know that this is still a draft PR)

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom_vibrational_ops_builder.py
```diff
@@ -100,6 +101,13 @@ def _build_single_hopping_operator(
...
     vibrational_op = VibrationalOp({" ".join(label): 1}, num_modals)
-    qubit_op: PauliSumOp = qubit_converter.convert_match(vibrational_op)
+
+    qubit_op: PauliSumOp
+    if isinstance(qubit_converter, QubitConverter):
+        qubit_op = qubit_converter.convert_match(vibrational_op)
+    elif issubclass(type(qubit_converter), QubitMapper):
+        qubit_op = qubit_converter.map_all(vibrational_op)
+    else:
+        raise QiskitNatureError("QubitConverter object is not valid")
```
mrossinek, line 111: This should not be necessary given the type hint. As per my earlier comment a simple:
```python
if isinstance(..., QubitConverter):
    ...
else:
    ...
```
should suffice in all situations :+1:

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -372,58 +322,29 @@ def convert_match(
...
-                qubit_ops = dict(iter(qubit_ops))
-
-        return qubit_ops
-
-    def _map(self, second_q_op: SparseLabelOp) -> PauliSumOp:
-        if self._sort_operators and isinstance(second_q_op, SparseLabelOp):
-            second_q_op = second_q_op.sort()
-        return self._mapper.map(second_q_op)
+    def _map(self, second_q_ops: ListOrDictType[SparseLabelOp]) -> _ListOrDict[PauliSumOp]:
+        return self._mapper.map(second_q_ops)
```
mrossinek, line 326:
```suggestion
        return self._mapper.map_all(second_q_ops)
```

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -61,6 +104,49 @@ def map(self, second_q_op: SparseLabelOp) -> PauliSumOp:
         """
         raise NotImplementedError()
 
+    def map_all(
```
mrossinek, line 107: I am still pondering the necessity of `map` *and* `map_all` to exist. Since `map_all` can also handle a single operator, it begs the question whether we need to differentiate these or whether we can simply extend the functionality of `map` to also support `ListOrDictType`. :thinking:

qiskit_nature/second_q/mappers/spin_mapper.py
```diff
@@ -34,3 +37,46 @@ def map(self, second_q_op: SpinOp) -> PauliSumOp:
             The `PauliSumOp` corresponding to the problem-Hamiltonian in the qubit space.
         """
         raise NotImplementedError()
+
+    def map_all(
```
mrossinek, line 41: Given that `map_all` is implemented on the base class, we should not need to duplicate it just for the type hint. There must be a cleaner solution to this. I will need to think about this longer than I have time today but I am happy to discuss :+1:

releasenotes/notes/update-api-for-qubitmapper-qubitconverter-d237a7b596d50207.yaml
```diff
@@ -0,0 +1,14 @@
+---
+features:
+  - |
+    Updated the API to allow :class:`~qiskit_nature/second_q/mappers/QubitMapper` objects  in places
+    where :class:`qiskit_nature/second_q/mappers/QubitConverter` were previously required. This addition 
+    advances toward a future deprecation of :class:`~qiskit_nature/second_q/mappers/QubitConverter`. 
+    All inputs of type :class:`~qiskit_nature/second_q/mappers/QubitConverter` now support
+    :class:`~qiskit_nature/second_q/mappers/QubitMapper` objects implementing a transformation from 
+    second quantized operators to Pauli operators. Note that the mappers currently do not support 
+    qubit reduction techniques.
```
mrossinek, line 10:
```suggestion
    Updated the API to allow :class:`~qiskit_nature.second_q.mappers.QubitMapper` objects  in places
    where :class:`~qiskit_nature.second_q.mappers.QubitConverter` were previously required. This addition 
    advances toward a future deprecation of :class:`~qiskit_nature.second_q.mappers.QubitConverter`. 
    All inputs of type :class:`~qiskit_nature.second_q.mappers.QubitConverter` now support
    :class:`~qiskit_nature.second_q.mappers.QubitMapper` objects implementing a transformation from 
    second quantized operators to Pauli operators. Note that the mappers currently do not support 
    qubit reduction techniques.
```

test/second_q/circuit/library/test_bogoliubov_transform.py
```diff
@@ -49,7 +49,7 @@ def test_bogoliubov_transform(self, n_orbitals, num_conserving):
             orbital_energies,
             transformed_constant,
         ) = hamiltonian.diagonalizing_bogoliubov_transform()
-        matrix = converter.map(hamiltonian.second_q_op()).to_matrix()
+        matrix = converter._map(hamiltonian.second_q_op()).to_matrix()
```
mrossinek, line 52: We should not call a private method here

mrossinek (commented):
Just some initial observations :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/excited_states_eigensolver.py
Anthony-Gandon: I think it was the only appearance of the old syntax, my mistake

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom_vibrational_ops_builder.py
Anthony-Gandon: Also a file I forgot to update with the if/else syntax

test/second_q/circuit/library/test_bogoliubov_transform.py
Anthony-Gandon: convert_only can be used as replacement here.

qiskit_nature/second_q/mappers/spin_mapper.py
Anthony-Gandon: We can use the fact that the `SpinOp` and `VibrationalOp` class are based on the `SparseLabelOp` class and only keep the `map()` method in the `QubitMapper`. I think this should avoid the duplicates

qiskit_nature/second_q/mappers/qubit_mapper.py
Anthony-Gandon: I have replaced all local definition of `map()` to a private `_map_single()` and then the `qubitmapper()` class implements the `map()` method accepting a `SparseLabelOp | ListOrDictType[SparseLabelOp]` and returning a `PauliSumOp | ListOrDictType[PauliSumOp]`

releasenotes/notes/update-api-for-qubitmapper-qubitconverter-d237a7b596d50207.yaml
```diff
@@ -0,0 +1,13 @@
...
+    Updated the API to allow :class:`~qiskit_nature.second_q.mappers.QubitMapper` objects  in places
+    where :class:`qiskit_nature.second_q.mappers.QubitConverter` were previously required. This addition 
+    advances toward a future deprecation of :class:`~qiskit_nature.second_q.mappers.QubitConverter`. 
+    All inputs of type :class:`~qiskit_nature.second_q.mappers.QubitConverter` now support
+    :class:`~qiskit_nature.second_q.mappers.QubitMapper` objects implementing a transformation from 
+    second quantized operators to Pauli operators. Note that the mappers currently do not support 
+    qubit reduction techniques.
+  - | 
+    The method :meth:`~qiskit_nature.second_q.mappers.QubitMapper.map()` now supports individual operators
+    as well as list and dictionaries of operators.
```
mrossinek, line 13:
```suggestion
    as well as lists and dictionaries of operators.
```

mrossinek (commented):
I think this is now looking a lot better in terms of API :+1: 

Could you please add more unittests to cover the usage of the new API?

Anthony-Gandon:
New unit tests were added following the API changes, however I still need to review properly the docstrings.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -223,12 +237,16 @@ def get_qubit_operators(
...
+        if isinstance(self.qubit_converter, QubitConverter):
+            _, z2symmetries = self.qubit_converter.find_taper_op(
+                main_op, problem.symmetry_sector_locator
+            )
+            self.qubit_converter.force_match(z2symmetries=z2symmetries)
+            untap_main_op = self.qubit_converter.convert_clifford(main_op)
+            untap_aux_ops = self.qubit_converter.convert_clifford(aux_ops)
+        else:
+            untap_main_op = main_op
+            untap_aux_ops = aux_ops
```
mrossinek, line 249: Can we add a TODO comment here referring to https://github.com/Qiskit/qiskit-nature/issues/974?
Once we implement that mapper the logic here will need to be adapted to re-add the same two-step
logic that is currently only possible in the `QubitConverter` case.
The same might apply in other places below so it may be useful to add similar comments elsewhere.

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_eigensolver.py
```diff
@@ -34,7 +34,7 @@ class GroundStateEigensolver(GroundStateSolver):
 
     def __init__(
         self,
-        qubit_converter: QubitConverter,
+        qubit_converter: QubitConverter | QubitMapper,
```
mrossinek, line 37: The docstring below also needs a minor update.
The same goes for some of the other modified classes.

qiskit_nature/second_q/circuit/library/ansatzes/ucc.py
```diff
@@ -310,7 +311,12 @@ def operators(self):  # pylint: disable=invalid-overridden-method
                 # inserting ``None`` as the result if an operator did not commute. To ensure that
                 # the ``excitation_list`` is transformed identically to the operators, we retain
                 # ``None`` for non-commuting operators in order to manually remove them in unison.
-                operators = self.qubit_converter.convert_match(excitation_ops, suppress_none=False)
+                if isinstance(self.qubit_converter, QubitConverter):
+                    operators = self.qubit_converter.convert_match(
+                        excitation_ops, suppress_none=False
+                    )
+                else:
+                    operators = self.qubit_converter.map(excitation_ops)
```
mrossinek, line 319: We should add a comment here, again referring to https://github.com/Qiskit/qiskit-nature/issues/974.
Once we work on that mapper we need to think about how to handle the `suppress_none=False` case
which we had in the `QubitConverter`.

qiskit_nature/second_q/circuit/library/initial_states/vscf.py
```diff
@@ -63,20 +63,21 @@ def __init__(
         )
 
     @property
-    def qubit_converter(self) -> QubitConverter:
+    def qubit_converter(self) -> QubitConverter | QubitMapper | None:
```
mrossinek, line 66: This currently still holds `QubitConverter(DirectMapper())` as the internally-generated default
value. Can we already change this or do we need to keep this as is?
If the latter, shall we log a warning that this default is going to change?
Also pinging @woodsp-ibm

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -312,7 +265,7 @@ def force_match(
 
     def convert_match(
         self,
-        second_q_ops: SparseLabelOp | ListOrDictType[SparseLabelOp],
+        second_q_ops: SparseLabelOp | PauliSumOp | ListOrDictType[SparseLabelOp | PauliSumOp],
```
mrossinek, line 268: Why can this now take a `PauliSumOp` as input?

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -16,14 +16,57 @@
...
+from qiskit.algorithms.list_or_dict import ListOrDict as ListOrDictType
 
 from qiskit_nature import QiskitNatureError
 from qiskit_nature.second_q.operators import SparseLabelOp
 
+# pylint: disable=invalid-name
+T = TypeVar("T")
+
+
+class _ListOrDict(Dict, Iterable, Generic[T]):
```
mrossinek, line 33: I am wondering whether we should extract this class into its own file (e.g. `_list_or_dict.py`). I
did at some point suggest that this iteration utility be extract to Qiskit Terra where e.g.
https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/algorithms/observables_evaluator.py
could also make use of this. We can revisit this later though.

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -61,6 +104,53 @@ def map(self, second_q_op: SparseLabelOp) -> PauliSumOp:
         """
         raise NotImplementedError()
 
+    def map(
+        self,
+        second_q_ops: SparseLabelOp | PauliSumOp | ListOrDictType[SparseLabelOp | PauliSumOp],
```
mrossinek, line 109: The same question applies here: why the `PauliSumOp`?

If it is really needed, we should attempt to use `SparsePauliOp` instead since the goal is to get
rid of opflow objects.

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -61,6 +104,53 @@ def map(self, second_q_op: SparseLabelOp) -> PauliSumOp:
...
+        qubit_ops: _ListOrDict = _ListOrDict()
+        for name, second_q_op in iter(wrapped_second_q_ops):
+            if isinstance(second_q_op, PauliSumOp):
+                qubit_ops[name] = second_q_op
+            else:
+                qubit_ops[name] = self._map_single(second_q_op)
+
+        returned_ops: Union[PauliSumOp, ListOrDictType[PauliSumOp]]
+
+        if issubclass(wrapped_type, (SparseLabelOp, PauliSumOp)):
```
mrossinek, line 142: Can we make this type unwrapping a utility function of `_ListOrDict`? Something like
`_ListOrDict.unwrap(self, _type)`?

test/second_q/algorithms/excited_state_solvers/test_excited_states_solvers.py
```diff
@@ -185,6 +186,42 @@ def custom_filter_criterion(eigenstate, eigenvalue, aux_values):
...
 
+    @unittest.skipIf(not _optionals.HAS_PYSCF, "pyscf not available.")
+    def test_solver_compatibility_with_mappers(self):
+        """Test that solvers can use both QubitConverter and QubitMapper"""
+        with self.subTest("Excited states solver"):
+            # pylint: disable=unused-argument
+            def filter_criterion(eigenstate, eigenvalue, aux_values):
+                return np.isclose(aux_values["ParticleNumber"][0], 2.0)
+
+            # with qubit converter
```
mrossinek, line 197: These cases should be different subtests such that it is easier to identify what test fails.
The same applies elsewhere.

test/second_q/algorithms/excited_state_solvers/test_qeom_electronic_ops.py
```diff
@@ -45,12 +45,13 @@ def setUp(self):
             basis="sto3g",
         )
 
-        self.qubit_converter = QubitConverter(JordanWignerMapper())
+        self.mapper = JordanWignerMapper()
```
mrossinek, line 48: Please add a test case which actually uses this.

test/second_q/algorithms/excited_state_solvers/test_qeom_vibrational_ops.py
```diff
@@ -41,7 +41,9 @@ class TestHoppingOpsBuilder(QiskitNatureTestCase):
     def setUp(self):
         super().setUp()
         algorithm_globals.random_seed = 8
-        self.qubit_converter = QubitConverter(DirectMapper())
+
+        self.mapper = DirectMapper()
```
mrossinek, line 45: Please add a test case which actually uses this.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: Yes, good point

qiskit_nature/second_q/circuit/library/initial_states/vscf.py
Anthony-Gandon: Having the default value set to `DirectMapper()` won't change any of the logic. However, because it is a public attribute, it might be possible that one defines a VSCF object with the default value and then affects symmetries to the qubit converter. In that case, the QubitMapper will fail.

qiskit_nature/second_q/mappers/qubit_converter.py
Anthony-Gandon: We can discuss that again: The idea is that the basic mode based mapping mapper.map() would act as the identity on Pauli Operator and as a mapper on Label operator. This would simplify the logic for aux_operators (which can be a dict of FermionicOp and QubitOperator) and also allow us to call map multiple times without it failing.

test/second_q/algorithms/excited_state_solvers/test_excited_states_solvers.py
Anthony-Gandon: ok

qiskit_nature/second_q/mappers/qubit_converter.py
mrossinek: Ah right I remember now. This is okay by me except for the aspect which I only thought of once commenting on `QubitMapper.map` below:
- we need to include `SparsePauliOp` in favor of `PauliSumOp` (maybe both for the time being? Not 100% sure right now)
- we need to update the docstrings to make this behavior more clear

qiskit_nature/second_q/mappers/qubit_mapper.py
Anthony-Gandon: As discussed, I rolled back these changes

qiskit_nature/second_q/mappers/qubit_converter.py
Anthony-Gandon: As discussed, I rolled back these changes

qiskit_nature/second_q/mappers/qubit_mapper.py
Anthony-Gandon: Not sure I understand this comment

Anthony-Gandon: the `wrapped_type` is defined before the _ListOrDict object can be built.

mrossinek: I was thinking of something along the lines of the following:
```python
class _ListOrDict:

    def unwrap(self, wrapped_type):
        if wrapped_type == list:
            return [op for _, op in iter(self)]
        if wrapped_type == dict:
            return dict(iter(self))
        # only other case left is that it was a single operator to begin with:
        return list(iter(self))[0][1]
```
Optionally this could also include the `suppress_none` argument (if needed for unwrapping in other places)

woodsp-ibm (approved):

PR #1008, opened 2022-12-19
Bind parameters of SparseLabelOp recursively

### LINKED ISSUE #1006: SparseLabelOp fails to assign parameters recursively
### Environment

- **Qiskit Nature version**: de01b704254592b7e311ac88b6a1430767e9cb43
- **Python version**: 3.10.8
- **Operating system**: Arch Linux

### What is happening?

See title.

```python
from qiskit.circuit import Parameter

a = Parameter("a")

op = FermionicOp({"+_0 -_1": a + 1}) + FermionicOp({"+_1 -_0": a})

assigned = op.assign_parameters({a: 1})
print(repr(assigned))
```
```
FermionicOp({'+_0 -_1': ParameterExpression(a + 1), '+_1 -_0': 1}, num_spin_orbitals=2, )
```

### What should happen?

The coefficient `ParameterExpression(a + 1)` should get resolved to 2. Instead, it is left untouched.

### Any suggestions?

_No response_

kevinsung:
### Summary
Fixes #1006.

### DIFF
--- a/qiskit_nature/second_q/operators/sparse_label_op.py
+++ b/qiskit_nature/second_q/operators/sparse_label_op.py
@@ -511,7 +511,11 @@ class SparseLabelOp(LinearMixin, AdjointMixin, GroupMixin, TolerancesMixin, ABC,
             A new operator with the parameters assigned.
         """
         data = {
-            key: value.bind(parameters, allow_unknown_parameters=True)
+            # TODO use this line once
+            # https://github.com/Qiskit/qiskit-terra/pull/9304
+            # is merged:
+            # key: value.bind(parameters, allow_unknown_parameters=True)
+            key: value.bind({k: v for k, v in parameters.items() if k in value.parameters})
             if isinstance(value, ParameterExpression)
             else value
             for key, value in self._data.items()
--- a/qiskit_nature/second_q/operators/sparse_label_op.py
+++ b/qiskit_nature/second_q/operators/sparse_label_op.py
@@ -511,7 +511,9 @@ class SparseLabelOp(LinearMixin, AdjointMixin, GroupMixin, TolerancesMixin, ABC,
             A new operator with the parameters assigned.
         """
         data = {
-            key: parameters[value] if value in parameters else value
+            key: value.bind(parameters, allow_unknown_parameters=True)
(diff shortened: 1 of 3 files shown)


qiskit_nature/second_q/operators/sparse_label_op.py
```diff
@@ -511,7 +511,9 @@ def assign_parameters(self, parameters: Mapping[ParameterExpression, _TCoeff]) -
             A new operator with the parameters assigned.
         """
         data = {
-            key: parameters[value] if value in parameters else value
+            key: value.bind(parameters, allow_unknown_parameters=True)
```
mrossinek, line 514:
```suggestion
            key: value.bind({k: v for k, v in parameters.items() if k in value.parameters})
```
Although more expensive, doing it this way decouples this change from the fix in Terra

test/second_q/operators/test_sparse_label_op.py
```diff
@@ -143,6 +143,12 @@ def test_add(self):
 
             self.assertEqual(test_op, target_op)
 
+        with self.subTest("parameter + parameter"):
+            test_op = DummySparseLabelOp({"+_0 -_1": a}) + DummySparseLabelOp({"+_1 -_0": a})
+            target_op = DummySparseLabelOp({"+_0 -_1": a, "+_1 -_0": a})
+
+            self.assertEqual(test_op, target_op)
+
```
mrossinek, line 151: Please also add a unittest which asserts that the proposed change works correctly when two different parameters are present (which would break without the `allow_unknown_parameters` or my proposed workaround).

mrossinek (commented):
And please also add a release note

qiskit_nature/second_q/operators/sparse_label_op.py
kevinsung: Done.

test/second_q/operators/test_sparse_label_op.py
kevinsung: I added such a test to the test of `assign_parameters`. I realized that this test of addition was unrelated and not broken so I removed it.

mrossinek (approved):
Thanks, Kevin 👍

PR #1009, opened 2022-12-19
refactor: avoid index order conversion for MP2

mrossinek:
### Summary

I want to be able to reuse the MP2 computing functionality elsewhere and would like to avoid the physicist-to-chemist conversion. So this PR refactores the internal `_compute_mp2` utility to rely on physicist-ordered two-body terms instead.

### DIFF
--- a/qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
+++ b/qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
@@ -19,7 +19,6 @@ import numpy as np
 from qiskit_nature.exceptions import QiskitNatureError
 from qiskit_nature.second_q.circuit.library import UCC
 from qiskit_nature.second_q.operators import ElectronicIntegrals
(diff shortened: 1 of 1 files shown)


qiskit_nature/second_q/algorithms/initial_points/mp2_initial_point.py
```diff
@@ -33,7 +32,7 @@ def _compute_mp2(
 
     Args:
         num_occ: The number of occupied molecular orbitals.
-        integral_matrix: The two-body molecular orbitals matrix.
+        integral_matrix: The two-body molecular orbitals matrix in physicist' index order.
```
declanmillar, line 35:
```suggestion
        integral_matrix: The two-body molecular orbitals matrix in physicists' index order.
```

declanmillar (approved):
Other than a tiny typo, this LGTM. It makes sense to be consistent with ordering.

manoelmarques (approved):

PR #1014, opened 2022-12-27
Fix dtype bug in BogoliubovTransform

kevinsung:
### Summary
The test I added throws an error without this fix.

### DIFF
--- a/qiskit_nature/second_q/circuit/library/bogoliubov_transform.py
+++ b/qiskit_nature/second_q/circuit/library/bogoliubov_transform.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2022.
+# (C) Copyright IBM 2022, 2023.
 #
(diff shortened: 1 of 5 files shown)


mrossinek (commented):
Generally this seems fine to me.

Please update the docstring of the `givens_matrix` method to reflect that a complex matrix will be returned.

<hr/>

I am not sure whether we can consider this a bugfix because this "changes"/constrains the return type of a method. @woodsp-ibm what do you think?

kevinsung:
Done. It's at least a bugfix because it fixes the bug in `BogoliubovTransform` which I added a test for. I guess it's technically a breaking change, though I expect the user impact to be negligible.

kevinsung:
Now that I think of it, I could also fix the bug without modifying the `givens_matrix` function. I think that's preferable, so I'll go ahead and do that.

mrossinek (approved):
Thanks, this is indeed better :+1:

PR #1015, opened 2023-01-03
Pending deprecate M,Q,W,V matrix setters in QEOMResult

### LINKED ISSUE #1013: (Pending-)Deprecate some matrix setters of the `QEOMResult`
### What should we add?

#971 heavily refactored and approved the QEOM implementation.
As such, the result object no longer requires/uses the setters for the M, Q, W and V matrices. Thus, these should be (pending-) deprecated.

These matrices can (and will) remain computable from the H and S matrices :+1:

manoelmarques:
### Summary

Closes #1013


mrossinek (approved):
Thanks, Manoel! :+1:

PR #1018, opened 2023-01-04
Implements Two Qubit Reduction in ParityMapper

### LINKED ISSUE #973: Integrating the `TwoQubitReduction` into the `ParityMapper`
### What should we add?

As outlined in #967 we want to extract the `TwoQubitReduction` from `qiskit.opflow` into the `ParityMapper` which is the only mapper that actually supports this hard-coded symmetry.

To achieve this, I envision an API similar to the following:
```python
class ParityMapper:
    def __init__(self, two_qubit_reduction: bool = False, num_particles: tuple[int, int] | None = None):
        ...
```
We need to default the `two_qubit_reduction` to False because we do not know the `num_particles` by default.
Providing the `num_particles` at initialization time ensures that the mapper can be used right away. The downside is that this makes it stateful for the current problem.

I would also be open to discussing the following API (which I personally even prefer):
```python
class ParityMapper:
    def __init__(self, num_particles: tuple[int, int] | None = None):
        ...
```
The idea here is that if `num_particles` is provided we enable the reduction and if it is `None` we do not. Simple and less redundant :+1: 

<hr/>

Implementation of this issue is blocked by https://github.com/Qiskit/qiskit-terra/issues/9140 because we would want to immediately use the new `SparsePauliOp`-based `Z2Symmetries` :+1:

Anthony-Gandon:
### Summary

Fix #973 

This PR "hardcodes" the Two Qubit Reduction symmetry into the ParityMapper.
To do so, it uses the `Z2symmetry` class from qiskit.quantum_info.analysis with custom symmetry specifications (instead of using the method `find_z2_symmetries`)

The two possible ways to use this new mapper are the following:

By default, two_qubit_reduction is set to False so that this mapper only applies the mapping
```
driver = PySCFDriver()
driver_result = driver.run()
fermionic_op, _ = driver_result.second_q_ops()
mapper = ParityMapper() 
qubit_op = mapper.map(fermionic_op)
print(qubit_op)
```
```
-0.8105479805373271 * IIII
+ 0.1721839326191555 * IIIZ
- 0.2257534922240236 * IIZZ
+ 0.1721839326191552 * IZZI
...
```
Even when setting the numbers of particles, two_qubit_reduction is False so only the mapping is applied (we might want to assign the _tapering_values anyway)
```
mapper.num_particles = [1, 1]
print(mapper._tapering_values)
qubit_op = mapper.map(fermionic_op)
print(qubit_op)
```

```
[-1, 1]
-0.8105479805373271 * IIII
+ 0.1721839326191555 * IIIZ
- 0.2257534922240236 * IIZZ
+ 0.1721839326191552 * IZZI
...
```
When setting two_qubit_reduction=True (the variable can also be modified with mapper.two_qubit_reduction = True) and the numbers of particles to None, the tapering produces a list of 4 reduced operators. (2^2 because 2 hardcoded symmetries)

```
mapper_TQR = ParityMapper(two_qubit_reduction=True)
qubit_op_TQR = mapper_TQR.map(fermionic_op)
print(qubit_op_TQR)
```

```
[PauliSumOp(SparsePauliOp(['II', 'IZ', 'ZI', 'ZZ'],
              coeffs=[-0.56872272+0.j, -0.05356956+0.j, -0.05356956+0.j,  0.67586183+0.j]), coeff=1.0), PauliSumOp(SparsePauliOp(['II', 'IZ', 'ZI', 'ZZ'],
              coeffs=[-0.81054798+0.j, -0.05356956+0.j,  0.39793742+0.j, -0.00571589+0.j]), coeff=1.0), PauliSumOp(SparsePauliOp(['II', 'IZ', 'ZI', 'ZZ', 'XX'],
              coeffs=[-1.05237325+0.j,  0.39793742+0.j, -0.39793742+0.j, -0.0112801 +0.j,
  0.1809312 +0.j]), coeff=1.0), PauliSumOp(SparsePauliOp(['II', 'IZ', 'ZI', 'ZZ'],
              coeffs=[-0.81054798+0.j,  0.39793742+0.j,  0.05356956+0.j,  0.00571589+0.j]), coeff=1.0)]
```
However the setter for particle numbers also sets the tapering values (when two_qubit_reduction is true). Using the setter before calling `map` produces a single result instead of a list (because it chooses one symmetry sector).
```
mapper_TQR.num_particles = [1, 1]
print(mapper_TQR._tapering_values)
qubit_op_TQR = mapper_TQR.map(fermionic_op)
print(qubit_op_TQR)
```
```
[-1, 1]
-1.0523732457728594 * II
+ 0.39793742484317896 * IZ
- 0.39793742484317873 * ZI
- 0.01128010425623538 * ZZ
+ 0.18093119978423122 * XX

```
Now for the interplay of the `QubitConverter` and `ParityMapper` arguments (see qubit converter tests) but the main idea is that The arguments of the qubitconverter have priority: 

- calling qubit_converter.convert(op, num_particles) will override the num_particles of the ParityMapper.
- creating a `QubitConverter(ParityMapper(two_qubit_reduction = False), two_qubit_reduction = True)` will override the two_qubit_reduction of the parity mapper to true
- calling qubit_converter.convert(op) without a num_particles will set it to None in the ParityMapper.


qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -583,11 +583,10 @@ def _prepare_expansion_basis(
...
         # Small workaround to apply two_qubit_reduction to a list with convert_match()
-        z2_symmetries = self.qubit_converter.z2symmetries
-        self.qubit_converter.force_match(z2symmetries=Z2Symmetries([], [], []))
-        reduced_hopping_ops = self.qubit_converter.convert_match(hopping_operators)
-        self.qubit_converter.force_match(z2symmetries=z2_symmetries)
-        untap_hopping_ops = self.qubit_converter.convert_clifford(reduced_hopping_ops)
+        if isinstance(self.qubit_converter, QubitConverter):
+            untap_hopping_ops = self.qubit_converter.convert_clifford(hopping_operators)
+        else:
+            untap_hopping_ops = hopping_operators
```
Anthony-Gandon, line 589: Add a TODO for issue #974

qiskit_nature/second_q/mappers/parity_mapper.py
```diff
@@ -14,27 +14,51 @@
...
 
+logger = logging.getLogger(__name__)
+
 
 class ParityMapper(FermionicMapper):  # pylint: disable=missing-class-docstring
-    def __init__(self):
+    def __init__(self, num_particles: tuple[int, int] | None = None):
         """The Parity fermion-to-qubit mapping.
 
         When using this mapper ``two_qubit_reduction`` can optionally be used for the qubit
```
mrossinek, line 38: This is now outdated. Please update

qiskit_nature/second_q/mappers/parity_mapper.py
```diff
@@ -14,27 +14,51 @@
...
         operator that is created, see converter class
         :class:`~qiskit_nature.second_q.mappers.QubitConverter`.
         """
         super().__init__(allows_two_qubit_reduction=True)
```
mrossinek, line 42: Note for a future reviewer: we decided that the super init argument `allows_two_qubit_reduction` will be deprecated in a separate PR

qiskit_nature/second_q/mappers/parity_mapper.py
```diff
@@ -54,5 +78,53 @@ def pauli_table(cls, nmodes: int) -> list[tuple[Pauli, Pauli]]:
...
+
+        if self.num_particles is not None and mapped_op.num_qubits > 2:
+            reduced_op = self.convert(mapped_op)
+        else:
+            if mapped_op.num_qubits <= 2:
+                logger.warning(
+                    "The original qubit operator only contains %s qubits! Skipping the requested "
+                    "two-qubit reduction!",
+                    mapped_op.num_qubits,
+                )
```
mrossinek, line 127: I would expect this to only be logged if `self.num_particles is not None`. Currently this would always be logged regardless of whether or not we actually attempted to apply the reduction

qiskit_nature/second_q/mappers/parity_mapper.py
```diff
@@ -54,5 +78,53 @@ def pauli_table(cls, nmodes: int) -> list[tuple[Pauli, Pauli]]:
 
         return pauli_table
 
+    def convert(self, operator: SparsePauliOp) -> SparsePauliOp:
```
mrossinek, line 81: Does this need to be a public method? If so, what is the reasoning for the naming? Would a more descriptive name like `two_qubit_reduce()` be better?

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -167,6 +170,10 @@ def z2symmetries(self) -> Z2Symmetries:
         """
         return copy.deepcopy(self._z2symmetries)
 
+    def _check_reset_mapper(self) -> None:
```
mrossinek, line 173: Maybe you could add some sort of explanation for a reader of this code?
I.e. that this method is used to preserve the original behavior of the `QubitConverter` class with respect to how it deals with `two_qubit_reduction` and `num_particles` being set, whereas the new `ParityMapper` only has a single argument (namely `num_particles`) to handle this.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -190,26 +190,25 @@ def get_qubit_operators(
...
         if isinstance(self.qubit_converter, QubitConverter):
-            main_op = self.qubit_converter.convert_only(
-                main_operator,
-                num_particles=num_particles,
-            )
+            self.qubit_converter.force_match(num_particles=num_particles)
+            main_op = self.qubit_converter.convert_only(main_operator, num_particles=num_particles)
         else:
+            if isinstance(self.qubit_converter, ParityMapper):
+                self.qubit_converter.num_particles = num_particles
```
mrossinek, line 202: Why is overriding the `num_particles` necessary here? Is this just a precaution against potentially incomplete/inconsistent setup of the mapper by the user?

mrossinek (commented):
A few comments but generally this is looking good :+1:

qiskit_nature/second_q/mappers/parity_mapper.py
Anthony-Gandon: Fixed

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: This is not really needed as future Mappers will be instantiated from the problem and thus won't need this. However at the moment qeom forces the num_particles of its Problem to the mapper and I tried to follow this logic. This is relevant if the Mapper is set with a wrong particle number for example. 
I'm not sure we want to keep it

mrossinek: Okay I see. I suppose we can keep it for now and then see in the future how this plays out :+1:

qiskit_nature/second_q/mappers/parity_mapper.py
```diff
@@ -14,27 +14,55 @@
...
-        When using this mapper ``two_qubit_reduction`` can optionally be used for the qubit
-        operator that is created, see converter class
-        :class:`~qiskit_nature.second_q.mappers.QubitConverter`.
+        When using this mapper, :attr:`num_particles` can optionally be used to apply an additional step
+        of reduction after the mapping to pauli operators. The two-qubit reduction eliminates the central
+        and last qubit in a list of Pauli that has diagonal operators (Z,I) at those positions.
+
+        Chemistry specific method:
+        It can be used to taper two qubits when the spin orbitals are ordered in two spin sectors,
+        (block spin order) according to the number of particles in the system.
```
mrossinek, line 44: Can we rephrase this a little bit? I am thinking of the following aspects:
- since this is now in Nature, we do not need to highlight that this is a chemistry specific method. It should also be clear from context since this is a `FermionicMapper`
- the block ordering assumption is a bit duplicated here. Its pointed out at the end, plus it is stated explicitly that the middle and last qubit are removed. Maybe this can be combined into one statement

qiskit_nature/second_q/mappers/parity_mapper.py
```diff
@@ -54,5 +82,55 @@ def pauli_table(cls, nmodes: int) -> list[tuple[Pauli, Pauli]]:
 
         return pauli_table
 
+    def _two_qubit_reduce(self, operator: SparsePauliOp) -> SparsePauliOp:
+        """
+        Applies the two qubit reduction to the operator. This method hard codes the `Z2Symmetry`
```
mrossinek, line 87:
```suggestion
        Applies the two qubit reduction to the operator. This method hard codes the ``Z2Symmetry``
```
Single backticks in RST are italic. Double backtick is needed for verbatim (unless prefixed with smth like `:attr:`)

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -167,6 +170,15 @@ def z2symmetries(self) -> Z2Symmetries:
...
+        """Resets the `ParityMapper` if the attribute :attr:`two_qubit_reduction` is set to False. This
```
mrossinek, line 174:
```suggestion
        """Resets the ``ParityMapper`` if the attribute :attr:`two_qubit_reduction` is set to False. This
```

mrossinek (commented):
Just some documentation nitpicks but other than that, this LGTM :+1:

qiskit_nature/second_q/mappers/parity_mapper.py
```diff
@@ -14,27 +14,52 @@
...
+    def __init__(self, num_particles: tuple[int, int] | None = None):
         """The Parity fermion-to-qubit mapping.
 
-        When using this mapper ``two_qubit_reduction`` can optionally be used for the qubit
-        operator that is created, see converter class
-        :class:`~qiskit_nature.second_q.mappers.QubitConverter`.
+        When using this mapper, :attr:`num_particles` can optionally be used to apply an additional step
+        of reduction after the mapping to pauli operators. The two-qubit reduction tapers two qubits
+        (middle and last qubit) when the spin orbitals are ordered in two spin sectors (block spin order)
+        and selects the symmetry sector corresponding to the provided number of particles.
```
mrossinek, line 41: It's better now, but how about this:
```suggestion
        When using this mapper, :attr:`num_particles` can optionally be used to apply an additional step
        of reduction after the mapping to pauli operators. The two-qubit reduction tapers two qubits
        (middle and last qubit) because the spin orbitals are ordered in two spin sectors (block spin order).
        Based on the provided number of particles this allows the automatic selection of
        the correct symmetry sector.
```

The reason I suggest this is because the ordering cannot be configured in Qiskit Nature.

mrossinek (approved):
Thanks a lot, Anthony!

I will greedily try to approve and automerge this in order to unblock #1031.
If we want to change some aspects of the `ParityMapper` we can still do this in a follow-up PR :+1:

PR #1022, opened 2023-01-09
Add low rank linalg utils

kevinsung:
### Summary
Part of #654 .

This PR adds low-level linear algebra utilities. A future PR will introduce functions for handling the higher level `ElectronicEnergy` class.

Credit to @mariomotta who provided the first version of the code for the modified Cholesky decomposition.


kevinsung:
CI failure seems unrelated.

woodsp-ibm:
I re-ran and whatever connectivity/issue was there is resolved. Its just errors in spell checker now where it will need Cholesky added to make it happy.

kevinsung:
CI failure seems unrelated

test/utils/test_linalg.py
```diff
@@ -30,3 +37,92 @@ def test_givens_matrix(self, a: complex, b: complex):
...
+        )
+        np.testing.assert_allclose(reconstructed, two_body_tensor, atol=1e-8)
+
+    @unittest.skipIf(not _optionals.HAS_PYSCF, "pyscf not available.")
+    def test_low_rank_two_body_decomposition_error_threshold_max_rank(self):
+        """Test low rank decomposition error threshold and max rank."""
+        driver = PySCFDriver(atom="Li 0 0 0; H 0 0 1.6")
+        driver_result = driver.run()
+        electronic_energy = driver_result.hamiltonian
+        two_body_tensor = to_chemist_ordering(electronic_energy.electronic_integrals.alpha["++--"])
```
mrossinek, line 84: If chemistry ordering is required by the `low_rank_two_body_decomposition`, this needs to be added to its documentation. This should be done in a `..note::` or `..warning::` section to highlight it and should link to the docstring of `to_chemist_ordering`.

test/utils/test_linalg.py
```diff
@@ -30,3 +37,92 @@ def test_givens_matrix(self, a: complex, b: complex):
...
+        reconstructed = np.einsum(
+            "tpk,tqk,tkl,trl,tsl->pqrs",
+            leaf_tensors,
+            leaf_tensors,
+            core_tensors,
+            leaf_tensors,
+            leaf_tensors,
+        )
+        self.assertEqual(len(leaf_tensors), max_rank)
+        np.testing.assert_allclose(reconstructed, two_body_tensor, atol=1e-5)
```
mrossinek, line 99: Please wrap every section which has its own assertion into a `subTest`

mrossinek (commented):
Alright this is a first pass at a review.

In general, it would be good if you could communicate how you expect and enduser to make use of this functionality. Either in form of examples in the docstrings, in the form of a tutorial, or in form of an issue in case you plan to add more features building up on top of this. Without an overview of your plans for this it is hard to see what you are going for.

In short:
- what does a workflow involving these methods look like?
- do you plan to add more public API classes/methods building up on this?

Depending on your answers to the above, this will also be of interest to #1033 which is laying the foundation for the addition of 4- and 8-fold symmetry aware two-body tensors. (see the description of that PR, too)

kevinsung:
> In general, it would be good if you could communicate how you expect and enduser to make use of this functionality. Either in form of examples in the docstrings, in the form of a tutorial, or in form of an issue in case you plan to add more features building up on top of this.

I've expanded the opening post of #654 to include an outline of the features that will be built on top of this. Let me know if that is not sufficients.

qiskit_nature/testing/random.py
```diff
@@ -66,6 +68,29 @@ def random_quadratic_hamiltonian(
...
+    """Sample a random two-body tensor with real-valued orbitals.
+
+    Args:
+        dim: The dimension of the tensor. The shape of the returned tensor will be
+            (dim, dim, dim, dim).
+        rank: Rank of the sampled tensor. The default behavior is to use
+            the maximum rank, which is `n_orbitals * (n_orbitals + 1) // 2`.
+        seed: The pseudorandom number generator or seed. Should be an
+            instance of `np.random.Generator` or else a valid input to
+            `np.random.default_rng`.
```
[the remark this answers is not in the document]
kevinsung: This is the numpy type signature for `default_rng`:
```python
def default_rng(
    seed: None | _ArrayLikeInt_co | SeedSequence | BitGenerator | Generator = ...
) -> Generator: ...
```
Here `_ArrayLikeInt_co` is imported from `numpy._typing`, so it is not part of the public API. I don't think we should rely on it.

In any case, here I followed the pattern of the rest of the functions in the file. If we want to change the type signature, we can do it all at once in a separate PR. What do you think?

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,129 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
     column_i = matrix[:, i].copy()
     column_j = matrix[:, j].copy()
     matrix[:, i], matrix[:, j] = column_j, column_i
+
+
+def modified_cholesky(
```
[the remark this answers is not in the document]
kevinsung: Reference added.

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,129 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
...
+    mat: np.ndarray, *, error_threshold: float = 1e-8, max_vecs: int | None = None
+) -> np.ndarray:
+    r"""Modified Cholesky decomposition.
+
+    The modified Cholesky decomposition of a square matrix :math:`M` has the form
+
+    .. math::
+        M = \sum_{i=1}^N v_i v_i^\dagger
+
+    where each :math:`v_i` is a vector. `M` must be positive definite.
```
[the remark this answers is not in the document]
kevinsung: Done.

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,129 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
...
+    No checking is performed to verify whether `M` is positive definite.
```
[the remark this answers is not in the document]
kevinsung: Done.

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,129 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
...
+
+    References:
+        - `arXiv:1711.02242`_
+
+    Args:
+        mat: The matrix to decompose.
+        error_threshold: Threshold for allowed error in the decomposition.
+            The error is defined as the maximum absolute difference between
+            an element of the original tensor and the corresponding element of
+            the reconstructed tensor.
```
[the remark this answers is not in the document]
kevinsung: I think the max norm is for matrices, not tensors. In any case, I still think it's better to write out the definition explicitly. What do you think?

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,129 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
...
+    r"""Low rank decomposition of a two-body tensor.
+
+    The low rank decomposition is a representation of a two-body tensor
+    :math:`h_{pqrs}` as
+
+    .. math::
+        h_{pqrs} = \sum_{t=1}^N \sum_{k\ell} U^{t}_{pk} U^{t}_{qk}
+            Z^{t}_{k\ell} U^{t}_{r\ell} U^{t}_{s\ell}
+
+    Here each :math:`U^{t}` is a unitary matrix, referred to as a "leaf tensor,"
```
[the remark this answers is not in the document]
kevinsung: Why? The general convention seems to be to put commas inside the quotes. Sources:
- https://owl.purdue.edu/owl/general_writing/punctuation/quotation_marks/more_quotation_mark_rules.html
- https://style.mla.org/punctuation-and-quotation-marks/

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,129 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
...
+    and each :math:`Z^{(t)}` is a Hermitian matrix, referred to as a "core tensor."
+
+    The number of terms :math:`N` in the decomposition depends on the allowed
+    error threshold. A larger error threshold may yield a smaller number of terms.
+    Furthermore, the `max_vecs` parameter specifies an optional upper bound
```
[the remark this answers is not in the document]
kevinsung: I think it makes sense to call it `max_rank` here. One could imagine extending this function to handle non-positive tensors where the modified Cholesky decomposition doesn't work. Then the terms would not necessarily correspond one-to-one with Cholesky vectors. I've fixed the docstring.

test/utils/test_linalg.py
kevinsung: Done.

test/utils/test_linalg.py
kevinsung: Done.

qiskit_nature/utils/linalg.py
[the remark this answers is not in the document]
mrossinek: I suppose its fine

qiskit_nature/testing/random.py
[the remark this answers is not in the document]
mrossinek: Fine I suppose then

qiskit_nature/utils/linalg.py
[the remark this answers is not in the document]
mrossinek: Okay sounds good :+1:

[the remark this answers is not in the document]
kevinsung: I decided to rename it to `max_vecs` after all.

qiskit_nature/testing/random.py
```diff
@@ -66,6 +66,28 @@ def random_quadratic_hamiltonian(
     )
 
 
+def random_two_body_tensor_real(dim: int, rank: int | None = None, seed: Any = None) -> np.ndarray:
+    """Sample a random two-body tensor with real-valued orbitals.
+
+    Args:
+        dim: The dimension of the tensor. The shape of the returned tensor will be
+            ``(dim, dim, dim, dim)``.
```
mrossinek, line 74: I find using the term "dimension" a bit confusing here as I think of a 2-body tensor as a 4-dimensional tensor with a given "size". So maybe `size` would be better here.

qiskit_nature/testing/random.py
```diff
@@ -28,9 +30,8 @@ def random_antisymmetric_matrix(dim: int, seed: Any = None) -> np.ndarray:
 
     Args:
         dim: The width and height of the matrix.
-        seed: The pseudorandom number generator or seed. Should be an
-            instance of `np.random.Generator` or else a valid input to
-            `np.random.default_rng`.
+        seed: The pseudorandom number generator or seed. Should be a valid input to
+            ``np.random.default_rng``.
```
mrossinek, line 34: Can you please change these links to the following?
```suggestion
            :func:`numpy.random.default_rng`.
```
Using our intersphinx mappings this allows them to be clickable and directly lead to the numpy docs

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,136 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
...
+) -> np.ndarray:
+    r"""Modified Cholesky decomposition.
+
+    The modified Cholesky decomposition of a square matrix :math:`M` has the form
+
+    .. math::
+        M = \sum_{i=1}^N v_i v_i^\dagger
+
+    where each :math:`v_i` is a vector. :math:`M` must be positive definite.
+    No checking is performed to verify whether :math:`M` is positive definite.
```
mrossinek, line 195: How will this currently appear to the user when not satisfying this? Will it outright break? If so, what will the error message be? If not, does that mean it silently "works"?

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,136 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
...
+    The number of terms :math:`N` in the decomposition depends on the allowed
+    error threshold. A larger error threshold may yield a smaller number of terms.
+    Furthermore, the ``max_vecs`` parameter specifies an optional upper bound
+    on :math:`N`. The ``max_vecs`` parameter is always respected, so if it is
+    too small, then the error of the decomposition may exceed the specified
+    error threshold.
+
+    .. note::
+        No checking is performed to verify whether the input matrix is positive definite.
+        If the input matrix is not positive definite, then the decomposition returned will be invalid.
```
mrossinek, line 205: Just saw this now. Silently "working" seems like a bad idea, especially given that checking this matrix is fairly simple. I suggest we add a check which can be disabled via a keyword argument if you really do not want this to be checked.

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,136 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
     column_i = matrix[:, i].copy()
     column_j = matrix[:, j].copy()
     matrix[:, i], matrix[:, j] = column_j, column_i
+
+
+def modified_cholesky(
+    mat: np.ndarray, *, error_threshold: float = 1e-8, max_vecs: int | None = None
```
mrossinek, line 185: How will this currently handle say a `sparse.SparseArray` object? It would be nice to test that and (if it works out of the box) add a unittest for this case to ensure that it stays that way.

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,136 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
...
+                cholesky_vecs[:, 0:index] @ cholesky_vecs[max_error_index, 0:index].conj()
+            )
+        cholesky_vecs[:, index] /= np.sqrt(max_error)
+        errors -= np.abs(cholesky_vecs[:, index]) ** 2
+
+    return cholesky_vecs[:, :index]
+
+
+def double_factorized(
+    two_body_tensor: np.ndarray, *, error_threshold: float = 1e-8, max_vecs: int | None = None
```
mrossinek, line 250: Same here, thinking about `sparse.SparseArray`. The output of this method can get _very_ large, so supporting some sparse tensor format is desirable.

mrossinek (commented):
In general this looks good to me :+1: 
I left some comments and besides those below just a few additions:

- this PR needs a release note
- you say in the description that you will add routines to more tightly couple to `ElectronicEnergy` in the future. Since this PR has been open for a while, we have added a `Tensor` class in the meantime. Given my comments on `sparse.SparseArray` below, it might be worthwhile considering supporting `Tensor` as the input type in addition to numpy arrays unless there is a good reason _not_ to support sparse arrays.
- This will likely simplify the integration with the rest of the stack down the road.

qiskit_nature/utils/linalg.py
mrossinek: Sorry, I had misread the condition there as something that could be easily checked :sweat_smile: I suggest changing the admonition from `note` to `warning` to emphasize this a bit more.

qiskit_nature/utils/linalg.py
kevinsung:
```
    @data(4, 5)
    def test_modified_cholesky_sparse(self, dim: int):
        """Test modified Cholesky decomposition on a random tensor using sparse."""
        rng = np.random.default_rng(4640)
        # construct a random positive definite matrix
        unitary = np.array(random_unitary(dim, seed=rng))
        eigs = rng.uniform(size=dim)
        mat = sparse.COO(unitary @ np.diag(eigs) @ unitary.T.conj())
>       cholesky_vecs = modified_cholesky(mat)

test/utils/test_linalg.py:67:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
qiskit_nature/utils/linalg.py:234: in modified_cholesky
    max_error_index = np.argmax(errors)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = (<COO: shape=(4,), dtype=float64, nnz=4, fill_value=0.0>,), kwargs = {}, relevant_args = (<COO: shape=(4,), dtype=float64, nnz=4, fill_value=0.0>, None)

>   ???
E   TypeError: no implementation found for 'numpy.argmax' on types that implement __array_function__: [<class 'sparse._coo.core.COO'>]

<__array_function__ internals>:180: TypeError
```

qiskit_nature/utils/linalg.py
kevinsung: It would fail with the same message as above.

qiskit_nature/testing/random.py
kevinsung: Done.

qiskit_nature/testing/random.py
kevinsung: Done.

qiskit_nature/utils/linalg.py
kevinsung: It silently "works" but returns nonsense. This matches the behavior of e.g. `np.linalg.cholesky` so should not be surprising to users.

qiskit_nature/utils/linalg.py
kevinsung: Done.

qiskit_nature/utils/linalg.py
mrossinek: I suggest adding a `.. warning::` admonition just like it is done in the other docstring to emphasize this further.

qiskit_nature/utils/linalg.py
mrossinek: Hm, that is unexpected. I wonder whether upgrading to numpy >= 1.24 will fix this. But that requires `numba` to release first which I expect to happen in the near future. I will keep an eye on this, but I don't think this blocks this PR at this time. But let's leave this thread open.

qiskit_nature/utils/linalg.py
kevinsung: Done.

kevinsung:
Test failures are unrelated.

mrossinek (commented):
Besides the missing reno this looks good to me.

@woodsp-ibm please also have a look when you have time

kevinsung:
I've added the release note.

releasenotes/notes/low-rank-linalg-86da86096d6122d8.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Added linear algebra utilities for performing the double-factorization of a two-body tensor:
+    - :func:`qiskit_nature.utils.double_factorized`
+    - :func:`qiskit_nature.utils.modified_cholesky`
```
woodsp-ibm, line 6: This does not come out as a sub-list. Maybe needs a space

image

qiskit_nature/utils/linalg.py
```diff
@@ -179,3 +179,142 @@ def _swap_columns(matrix: np.ndarray, i: int, j: int) -> None:
...
+    too small, then the error of the decomposition may exceed the specified
+    error threshold.
+
+    The input tensor is assumed to be positive definite when reshaped into a matrix.
+
+    .. warning::
+        No checking is performed to verify whether the input tensor is positive definite.
+        If the input tensor is not positive definite, then the decomposition returned will be invalid.
+
+    References:
```
woodsp-ibm, line 278: It might be nice to have the titles of the papers here so one has an idea without clicking over to it to find out what the ref is about. Same in the other file with a similar ref.

kevinsung: done.

releasenotes/notes/low-rank-linalg-86da86096d6122d8.yaml
kevinsung: I added a two-space indent, hopefully that fixes it.

releasenotes/notes/low-rank-linalg-86da86096d6122d8.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Added linear algebra utilities for performing the double-factorization of a two-body tensor:
+      - :func:`qiskit_nature.utils.double_factorized`
```
woodsp-ibm, line 5:
```suggestion

      - :func:`qiskit_nature.utils.double_factorized`
```

Lets try adding a space (blank line) before the inner list. It looks better but I think its interpreting the whole thing as a definition list now where the first sentence is the term, hence the bold in the resultant html

image

woodsp-ibm (approved):
Seems it finally worked out with the release note.

image

mrossinek (approved):
Thanks, Kevin!

PR #856, opened 2022-09-30
refactor: switch to using primitives in qiskit_nature/second_q stack

mrossinek:
### Summary

This refactors the entire `qiskit_nature/second_q` stack to use the new Qiskit primitives.

Some further changes are required:
- [x] unify the VQE-Factory signatures with the new `VQE` (positional estimator, ansatz and optimizer)
  - [x] do we want a default optimizier?
  - [x] do we even want a default estimator?
- [x] refactor the `EigenstateResult.eigenstates` once https://github.com/Qiskit/qiskit-terra/pull/8816 is merged
- [x] check the documentation strings for the validity after these changes

### Details and comments

Closes #849

PR #1023, opened 2023-01-11
Use allow_unknown_parameters in SparseLabelOp.assign_parameters

### LINKED ISSUE #1010: Use allow_unknown_parameters option in SparseLabelOp.assign_parameters
### What should we add?

After https://github.com/Qiskit/qiskit-terra/pull/9304 is merged.

See https://github.com/Qiskit/qiskit-nature/pull/1008.

kevinsung:
### Summary
Fixes #1010


mrossinek (approved):
In general I think this is good to go. I am just not sure about some "meta-details".
What I mean by that is that this refactors the change done by the bugfix PR #1008. That PR was backported to stable (but not released yet).

- Does that mean we can simply merge this as is without a release note?
- Should we backport this, too?

Pinging @woodsp-ibm and @manoelmarques for input on this

mrossinek:
Actually, this now uses an unreleased feature of Qiskit Terra so we can*NOT* backport it as such. I am a bit surprised that the CI did not catch this... @manoelmarques are we still testing against Qiskit Terra installed from source even though our requirements file currently only uses `0.22.*`?

woodsp-ibm:
If the change uses/depends on new features from an as yet unreleased version then the requirements.txt needs to get updated too by this PR, if it does not already reflect the version dependence it requires.

kevinsung:
> If the change uses/depends on new features from an as yet unreleased version then the requirements.txt needs to get updated too by this PR, if it does not already reflect the version dependence it requires.

Yes, I can do that. But I think Max is asking why this PR passed the CI even though I did not update `requirements.txt`, considering that it does use an unreleased feature in Terra.

mrossinek:
> Yes, I can do that. But I think Max is asking why this PR passed the CI even though I did not update `requirements.txt`, considering that it does use an unreleased feature in Terra.

I am indeed wondering that. Can you, thus, please ensure the following:
- update `requirements.txt`
- ensure there is a test which actually triggers this case

kevinsung:
The test for this change was added in the previous PR, https://github.com/Qiskit/qiskit-nature/pull/1008.

Do you want me to update `requirements.txt` now, or do you want to wait to fix the CI issue so this PR can be used to test that the CI is working properly?

mrossinek:
> The test for this change was added in the previous PR, #1008.

The test there does not seem to be actually doing this though. It constructs an operator which will have two parameters, `a` and `b`, and then assigns a value to `a`. If I understand this change correctly, the method is supposed to work correctly even if one were to provide a parameter called for example `c`, which is _unknown_ to the operator.

> Do you want me to update `requirements.txt` now, or do you want to wait to fix the CI issue so this PR can be used to test that the CI is working properly?

If my reasoning above is correct, the CI is working fine and this case is not covered correctly by the unittests.

kevinsung:
> The test there does not seem to be actually doing this though. It constructs an operator which will have two parameters, a and b, and then assigns a value to a.

The test there was actually testing this behavior because it causes the ParameterExpression containing only b to be passed an assignment dictionary only containing a. Nevertheless, I've added another test on your suggestion that is perhaps more explicit.

mrossinek:
Alright, then all that is missing is the update of the `requirements.txt` file. Please do that here, too

kevinsung:
Do you want me to update `requirements.txt` now, or do you want to wait to fix the CI issue so this PR can be used to test that the CI is working properly?

mrossinek:
As explicitly said: please update it now.

We are continuing to test Qiskit Nature against Qiskit Terra main to catch potential breakage early. This is an intentional choice of the application developers team to do this for the packages within the Qiskit organization. Thus, CI is indeed working as intended.

The reason I originally expected this to fail would have been the `neko` tests, but this simply does not catch this particular change which is not an issue we can fix here (nor wait to be fixed).

manoelmarques:
Currently it is `qiskit-terra>=0.22.*`  so it should just  bump to `qiskit-terra>=0.23.*`

requirements.txt
```diff
@@ -1,4 +1,4 @@
-qiskit-terra>=0.22.*
+git+https://github.com/Qiskit/qiskit-terra@main#egg=qiskit-terra
```
mrossinek, line 1:
```suggestion
qiskit-terra>=0.23.*
```

This is the correct way to update it. This will ensure that the CI works as expected _except_ for the Neko test case.
We had a similar situation [here](https://github.com/Qiskit/qiskit-nature/pull/856#issuecomment-1263707765) where we bumped to `0.22.*` one day before `0.22.0rc1` got released.

I would like to see the CI pass with `0.23.*` specified and have only Neko fail. If that happens, I would be okay with merging this (or we wait until we are closer to `0.23.0rc1` being released, but the PR is then effectively ready)

PR #1027, opened 2023-01-12
Add `HexagonalLattice` Class

### LINKED ISSUE #956: Add a `HexagonalLattice`
### What should we add?

A new lattice should be added to the `qiskit_nature.second_q.hamiltonians.lattices` module, allowing the simple construction of [hexagonal lattices](https://en.wikipedia.org/wiki/Hexagonal_lattice).

Based on this class, we could then also add support for the quick construction of [heavy-hex lattices](https://research.ibm.com/blog/heavy-hex-lattice) which would be of great use for simulations on IBM Quantum devices.
I am open to discuss whether we want a `HeavyHexLattice` subclass of the `HexagonalLattice` or whether we want to a `heavy: bool = True` keyword argument for the lattice class proposed by this feature request.

(This suggestion was extracted from #43)

javabster:
### Summary
fixes #956


qiskit_nature/second_q/hamiltonians/lattices/hexagonal_lattice.py
```diff
@@ -0,0 +1,73 @@
...
+        self._edge_parameter = edge_parameter
+        self._onsite_parameter = onsite_parameter
+
+        graph = generators.hexagonal_lattice_graph(rows, cols, multigraph=False)
+
+        edges = graph.weighted_edge_list()
+
+        # Add edge weights
+        for idx, _ in enumerate(edges):
+            graph.update_edge_by_index(idx, self._edge_parameter)
```
mtreinish, line 49: It would be slightly more performant to do this as:

```suggestion
        # Add edge weights
        for idx in graph.edge_indices():
            graph.update_edge_by_index(idx, self._edge_parameter)
```

as the iterator only has to convert a single int instead of a tuple and None at each step. You probably wouldn't really see a difference in practice though except at really large graphs though.

It might also be worth benchmarking this against using `range(graph.num_edges())` and `range(graph.num_nodes())` (for the loop below). While assuming a contiguous range of indices for nodes and edges in rustworkx isn't always a good idea (because there can be holes on removals) using this from the fresh output of the generator function should work fine as there will be no removals. This might be potentially even faster because the biggest overhead in rustworkx comes from the rust->python conversion and with any of the other methods listed we do that on each step in the iterator for these loops.

javabster: yeah just doing some quick benchmarking it does seem to be faster, especially getting to larger lattices:

using `graph.edge_indices()` & `graph.node_indices()`:
- `HexagonalLattice(10,10)`: 4.668223667999996s
- `HexagonalLattice(20,20)`: 18.730463289s
- `HexagonalLattice(25,25)`: 28.666557584000003s

using `range(graph.num_edges())` & `graph.node_indices()`:
- `HexagonalLattice(10,10)`: 4.253540048s
- `HexagonalLattice(20,20)`: 16.220453660000004s
- `HexagonalLattice(25,25)`: 24.83831334300001s

using `range(graph.num_edges())` & `range(graph.num_nodes())`:
- `HexagonalLattice(10,10)`: 3.9827502070000005
- `HexagonalLattice(20,20)`: 15.120913722000004
- `HexagonalLattice(25,25)`: 23.106167259000003

test/second_q/hamiltonians/lattices/test_hexagonal_lattice.py
```diff
@@ -0,0 +1,133 @@
...
+                (8, 9),
+            ]
+
+            for idx1, idx2 in neg_indices:
+                target_matrix[idx1, idx2] = 0 - 1.42j
+
+            for idx3, idx4 in pos_indices:
+                target_matrix[idx3, idx4] = 0 + 1.42j
+
+            assert_array_equal(hexa.to_adjacency_matrix(weighted=True), target_matrix)
```
mrossinek, line 133:
```suggestion
        with self.subTest("Check the adjacency matrix."):
            target_matrix = np.zeros((10, 10), dtype=complex)

            indices = [
                (0, 1),
                (0, 3),
                (1, 2),
                (2, 5),
                (3, 4),
                (4, 5),
                (4, 7),
                (5, 6),
                (6, 9),
                (7, 8),
                (8, 9),
            ]

            for idx1, idx2 in indices:
                target_matrix[idx1, idx2] = 0 + 1.42j

            target_matrix -= target_matrix.T

            np.fill_diagonal(target_matrix, 1.0)

            assert_array_equal(hexa.to_adjacency_matrix(weighted=True), target_matrix)
```
This is just a little bit shorter and it avoids storing the redundant information in the `neg_` and `pos_indices`

test/second_q/hamiltonians/lattices/test_hexagonal_lattice.py
```diff
@@ -0,0 +1,133 @@
...
+        cols = 2
+        edge_parameter = 0 + 1.42j
+        onsite_parameter = 1.0
+
+        hexa = HexagonalLattice(rows, cols, edge_parameter, onsite_parameter)
+
+        with self.subTest("Check the graph."):
+            target_graph = PyGraph(multigraph=False)
+            target_graph.add_nodes_from(range(10))
+            weighted_edge_list = [
```
mrossinek, line 36: Actually, we can shorten this a bit further if you extract `weighted_edge_list` out of this subtest and define it below `onsite_parameter`.
This will allow you to use `target_set = set(weighted_edge_list)` in the next subtest and you can replace the definition of `indices` in the last subtest with:
```python
indices = [(a, b) for a, b, _ in weighted_edge_list]
```

test/second_q/hamiltonians/lattices/test_hexagonal_lattice.py
```diff
@@ -0,0 +1,133 @@
...
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Test for HyperCubicLattice."""
```
mrossinek, line 13:
```suggestion
"""Test for HexgonalLattice."""
```

qiskit_nature/second_q/hamiltonians/lattices/hexagonal_lattice.py
```diff
@@ -0,0 +1,71 @@
...
+        self,
+        rows: int,
+        cols: int,
+        edge_parameter: complex = 1.0,
+        onsite_parameter: complex = 0.0,
+    ) -> None:
+        """
+        Args:
+            rows: Length of the x direction.
+            cols: Length of the y direction.
```
mrossinek, line 32: Rather than `length` we could say `number of hexagons along the x direction` (etc.)

mrossinek (commented):
Besides these comments, please also add a release note to this PR

woodsp-ibm:
Do we want to include this in the Lattice Models tutorial - since that tutorial shows the existing models it seems to me it would be good to include this there too.

mrossinek:
Yesterday I found out that [`networkx` supports both](https://github.com/networkx/networkx/blob/main/networkx/generators/lattice.py#L269): periodic boundary conditions and the computation of the real-space coordinates of a regular hexagonal lattice structure (something which we like to use to make the graphs look predictable; see for example [here](https://github.com/Qiskit/qiskit-nature/blob/main/qiskit_nature/second_q/hamiltonians/lattices/hyper_cubic_lattice.py#L266)).

@mtreinish is rustworkx open to supporting similar functionality? At least periodic boundary conditions would seem useful to me.

If yes, I suggest we could track the addition of similar functionality there. Since I moved this PR to 0.7.0 we could even postpone until that is available. @javabster wold you be interested in contributing this to rustworkx, assuming they want it :smile:

mtreinish:
Yeah, both of those seem totally valid to add to rustworkx. If you don't mind can you open an enhancement issue:

https://github.com/Qiskit/rustworkx/issues/new?assignees=&labels=type%3A+enhancement&template=ENHANCEMENT_REQUEST.md

with the details and we can target it for the 0.13.0 release (which we don't have a target date for yet). @javabster it'd be great if you're interested in contributing the functionality to rustworkx too. I can give you a hand if you need any help getting started on it.

javabster:
issue opened: https://github.com/Qiskit/rustworkx/issues/803

mrossinek:
For various bandwidth reasons I would like to restrict the scope of this PR as follows:

- we **neglect** periodicity for the time being and instead add support for this in the future (separate issue to be opened)
- we **need** to add a computation of the default node positions (to ensure consistent and nice visualization; see for example the networkx implementation of this)
- in the future, rustworkx might do this last point directly at which point we can remove this from this class. Thus, this computation should be done in a private method which we can remove without API breakage

mrossinek (approved):
Let's also get this one merged. I added a comment [here](https://github.com/qiskit-community/qiskit-nature/issues/956#issuecomment-1704732925) to keep track of possible future extensions.

I also added a simple release note in a similar style to the one of the recently merged Kagome lattice.

Thank you, Abby, for contributing to Qiskit Nature and being so patient with this lengthy (and sometimes stalled) review process!

Issue #1020, opened 2023-01-05
Replace `PauliSumOp` with `SparsePauliOp`


mrossinek:
### What should we add?

We need to replace all public API occurrences of `PauliSumOp` with `SparsePauliOp`.

- in cases where the type is a function *input*, for the transition period we need to replace it with the union of the two and handle it accordingly internally

- in cases where the type is an *output* we also need to make it the union. However, we cannot simply return whatever we want. I suggest that we add a setting to `qiskit_nature.settings` which can toggle between the two cases. In the next release, we still default it to return `PauliSumOp` but have it raise a `DeprecationWarning` that the default will change to `SparsePauliOp`. This would be similar to how we handled the `dict_aux_operators` transition.

Anthony-Gandon:
This change goes in parallel with a transition from the z2symmetries of opflow for PauliSumOp to the z2symmetries of quantum_info for SparsePauliOp Ideally, we would make the two transitions simultaneously.
Similarly, the qubit_converter heavily relies on the PauliSumOp but will be replaced by new mappers which will be based on SparsePauliOp.
Last thing, I wanted to mention that multiple tests use the == / assertEqual for operators which work for PauliSumOp but not for SparsePauliOp because of the Pauli's ordering:

```
from qiskit.quantum_info import SparsePauliOp

qubit_op = SparsePauliOp.from_list(
    [
        ("IIXZ", -1.0537076071291125), # 1
        ("IZXY", 0.393983679438514), # 2
        ("ZIYY", -0.39398367943851387), # 3
        ("ZZIZ", -0.01123658523318205), # 4
        ("XXYI", 0.1812888082114961), # 5
    ]
)
qubit_op2 = SparsePauliOp.from_list(
    [
        ("IIXZ", -1.0537076071291125), # 1
        ("ZIYY", -0.39398367943851387), # 3
        ("IZXY", 0.393983679438514), # 2
        ("ZZIZ", -0.01123658523318205), # 4
        ("XXYI", 0.1812888082114961), # 5
    ]
)
print(qubit_op.equiv(qubit_op2)) # True
print(qubit_op.__eq__(qubit_op2)) # False
```

I believe this will lead to additional changes

mrossinek:
Thanks, Anthony! @manoelmarques did indeed start looking into this and made the same observations. We put this on hold until the work on #967 is completed :+1:

PR #1031, opened 2023-01-13
Implementing the `TaperedQubitMapper`

### LINKED ISSUE #974: Implement a `TaperedQubitMapper`

Anthony-Gandon:
### Summary

Fix #974 
Implements a TaperedQubitMapper implementing the Z2symmetry reduction and providing the map_clifford() and symmetry_reduce_clifford() required in QEOM for the two-step tapering

### Details and comments

The TaperedQubitMapper can be created from a Z2symmetry object and a mapper

```
mapper = ParityMapper(num_particles=None)
z2_sym = Z2Symmetries([], [], [], None)
tapered_qubit_mapper = TaperedQubitMapper(mapper, z2symmetries=z2_sym)
qubit_op = tapered_qubit_mapper.map(self.h2_op).primitive
```

or directly from a problem

```
driver = PySCFDriver()
self.driver_result = driver.run()
mapper = JordanWignerMapper()
tapered_qubit_mapper = TaperedQubitMapper.from_problem(mapper, self.driver_result)
qubit_op = tapered_qubit_mapper.map(self.h2_op).primitive
```


qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -275,19 +354,27 @@ def solve(
...
 
+        untap_main_op = untap_main_op_sumop.primitive
+        untap_aux_ops = {key: op.primitive for key, op in untap_aux_ops_sumop.items()}
+
         # 2. Run ground state calculation with fully tapered custom auxiliary operators
         # Note that the solve() method includes the `second_q' auxiliary operators
         if isinstance(self.qubit_converter, QubitConverter):
-            tap_aux_operators = self.qubit_converter.symmetry_reduce_clifford(untap_aux_ops)
+            tap_aux_operators_sumop = self.qubit_converter.symmetry_reduce_clifford(
+                untap_aux_ops_sumop
```
mrossinek, line 368: Given the above, why do you continue to use the sumop variants here?

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -503,11 +588,6 @@ def _build_commutator_routine(
 
         eom_operators = {"q": q_mat_op, "w": w_mat_op, "m": m_mat_op, "v": v_mat_op}
 
-        if not z2_symmetries.is_empty():
-            for index_op, eom_op in eom_operators.items():
-                if eom_op is not None and len(eom_op) > 0:
-                    eom_operators[index_op] = z2_symmetries.taper_clifford(eom_op)
```
mrossinek, line 509: Why can we remove this?

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+
+from .qubit_mapper import QubitMapper, _ListOrDict
+
+if TYPE_CHECKING:
+    from qiskit_nature.second_q.problems import ElectronicStructureProblem
+
+logger = logging.getLogger(__name__)
+
+
+class TaperedQubitMapper(QubitMapper):  # pylint: disable=missing-class-docstring
```
mrossinek, line 35:
```suggestion
class TaperedQubitMapper(QubitMapper):
```

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+    """The wrapper around qubit mappers implementing the logic to reduce the size of a problem (operator)
+    based on mathematical `Z2Symmetries` that can be automatically detected in the operator.
```
mrossinek, line 37:
```suggestion
    based on mathematical ``Z2Symmetries`` that can be automatically detected in the operator.
```

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+        z2symmetries: Symmetries to use to reduce the Pauli operators.
+    """
+
+    def __init__(
+        self,
+        mapper: QubitMapper,
+        z2symmetries: Z2Symmetries = Z2Symmetries([], [], []),
+    ):
+        """
+
```
mrossinek, line 53:
```suggestion
```

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+                operators.
+        """
+        super().__init__()
+        self.mapper: QubitMapper = mapper
+        self.z2symmetries = z2symmetries
+
+    @classmethod
+    def from_problem(
+        cls, mapper: QubitMapper, problem: "ElectronicStructureProblem"
+    ) -> "TaperedQubitMapper":
```
mrossinek, line 67: I wonder if we can find a better name. It is just a little bit confusing to see this method name and then have a `QubitMapper` as the first argument..

Could also just be me to find it odd :thinking:

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+
+        Args:
+            mapper: ``QubitMapper`` object implementing the mapping of second quantized operators to
+                Pauli operators.
+            problem: A class encoding a problem to be solved.
+
+        Return:
+            A ``TaperedQubitMapper`` with pre-built symmetry specifications.
+        """
+
```
mrossinek, line 82:
```suggestion
```

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+        tapering_values = problem.symmetry_sector_locator(z2_symmetries, mapper)
+        z2_symmetries.tapering_values = tapering_values
+        return TaperedQubitMapper(mapper, z2_symmetries)
+
+    def _map_clifford_single(self, second_q_op: SparseLabelOp) -> SparsePauliOp:
+        mapped_op = self.mapper.map(second_q_op).primitive
+        converted_op = self.z2symmetries.convert_clifford(mapped_op)
+        return converted_op
+
+    def _symmetry_reduce_clifford_single(
```
mrossinek, line 95: Maybe we should match the naming of the `Z2Symmetries` method and call this more along the lines of `taper_clifford`. The same applies to the public method here

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+        self, converted_op: SparsePauliOp, check_commutes: bool = True
```
mrossinek, line 96:
```suggestion
        self, converted_op: SparsePauliOp, *, check_commutes: bool = True
```
This makes `check_commutes` a keyword-only argument

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+    ) -> None | SparsePauliOp:
+
+        if self.z2symmetries.is_empty() or self.z2symmetries.tapering_values is None:
+            return converted_op
+        elif check_commutes:
```
mrossinek, line 101:
```suggestion
        if check_commutes:
```
No need to `elif` since you `return` above. I am a bit surprised that `pylint` did not catch this :thinking:

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+
+        tapered_op = self.z2symmetries.taper_clifford(converted_op)
+        cast(SparsePauliOp, tapered_op)
+        return tapered_op
+
+    def _map_single(self, second_q_op: SparseLabelOp) -> PauliSumOp | None:
+        converted_op = self._map_clifford_single(second_q_op)
+        tapered_op = self._symmetry_reduce_clifford_single(converted_op)
+
+        returned_op = PauliSumOp(tapered_op) if isinstance(tapered_op, SparsePauliOp) else None
```
mrossinek, line 117: We should look into updating the API to support returning either. This will be done as part of #1020 though

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+        return returned_op
+
+    def map_clifford(
+        self,
+        second_q_ops: SparseLabelOp | ListOrDictType[SparseLabelOp],
+    ) -> PauliSumOp | ListOrDictType[PauliSumOp]:
+        """Maps a second quantized operator or a list, dict of second quantized operators based on
+        the current mapper.
```
mrossinek, line 125: This could be explained better. I think you should highlight the 2 steps taken here:
1. using the internal mapper, the operators are mapped
2. then, using the internal Z2Symmetries, `convert_clifford` is applied

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+        returned_ops: PauliSumOp | ListOrDictType[PauliSumOp]
+        returned_ops = qubit_ops.unwrap(wrapped_type)
+
+        return returned_ops
+
+    def symmetry_reduce_clifford(
+        self,
+        pauli_ops: PauliSumOp | ListOrDictType[PauliSumOp],
+        check_commutes: bool = True,
+        suppress_none: bool = False,
```
mrossinek, line 154:
```suggestion
        *,
        check_commutes: bool = True,
        suppress_none: bool = False,
```

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+        qubit_ops: _ListOrDict = _ListOrDict()
+        for name, second_q_op in iter(wrapped_second_q_ops):
+            qubit_ops[name] = PauliSumOp(self._map_clifford_single(second_q_op))
+
+        returned_ops: PauliSumOp | ListOrDictType[PauliSumOp]
+        returned_ops = qubit_ops.unwrap(wrapped_type)
+
+        return returned_ops
+
+    def symmetry_reduce_clifford(
```
mrossinek, line 150:
```suggestion
    def taper_clifford(
```
Just to be explicit and re-iterate on my previous proposal

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,197 @@
...
+
+        returned_ops: PauliSumOp | ListOrDictType[PauliSumOp] = qubit_ops.unwrap(
+            wrapped_type, suppress_none=suppress_none
+        )
+
+        return returned_ops
+
+    @staticmethod
+    def _check_commutes(sq_paulis: list[Pauli], qubit_op: SparsePauliOp) -> bool:
+        # commutes = []
```
mrossinek, line 193:
```suggestion
```

mrossinek (commented):
Alright this is my initial review. There is a lot in here but I think this is already in good shape :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -74,6 +76,82 @@ class EvaluationRule(Enum):
     DIAG = "diag"
 
 
+def _commutator(op_a: SparsePauliOp, op_b: SparsePauliOp) -> SparsePauliOp:
```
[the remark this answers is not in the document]
ikkoham: I'm fine with both.

 For opflow we have had `commutator`. https://github.com/Qiskit/qiskit-terra/blob/2eee56616d50a9e26756f855ef4aa0135920ad78/qiskit/opflow/utils.py
It would be nice to have it for `LinearOp `. i.e. `commutator(op1': LinearOp, op2: LinearOp) -> LinearOp`

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -74,6 +76,82 @@ class EvaluationRule(Enum):
...
+        AB - BA.
+
+    Args:
+        op_a: Operator A.
+        op_b: Operator B.
+
+    Returns:
+        The computed commutator.
+    """
+    return (op_a @ op_b - op_b @ op_a).simplify()
```
ikkoham, line 93: Is it ok that simplify method does not take an argument, for example `atol=0`?

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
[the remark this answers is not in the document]
mrossinek: Yes the opflow utility is exactly what we mimicked in Nature here: https://github.com/Qiskit/qiskit-nature/blob/ef1d99ecd735daa20e5603eb0b13fb9e25a614df/qiskit_nature/second_q/operators/commutators.py

I agree that it would be nice to have this directly for `LinearOp`. That would mean it should go somewhere into `quantum_info` :+1: I will open an issue shortly to track this over in Terra

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: Ah yes indeed we should force 0 instead of the default value

test/second_q/mappers/test_tapered_qubit_mapper.py
```diff
@@ -0,0 +1,469 @@
...
+            tapered_qubit_mapper = TaperedQubitMapper.from_problem(mapper, self.driver_result)
+            qubit_op = tapered_qubit_mapper.map(self.h2_op).primitive
+            self.assertTrue(qubit_op.equiv(TestTaperedQubitMapper.REF_H2_JW_TAPERED))
+
+        with self.subTest("Tapered Qubit Mapper from empty symmetry"):
+            mapper = JordanWignerMapper()
+            z2_sym = Z2Symmetries([], [], [], None)
+            tapered_qubit_mapper = TaperedQubitMapper(mapper, z2symmetries=z2_sym)
+            qubit_op = tapered_qubit_mapper.map(self.h2_op).primitive
+            self.assertTrue(qubit_op.equiv(TestTaperedQubitMapper.REF_H2_JW))
```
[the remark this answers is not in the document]
Anthony-Gandon: Indeed it's redundant

test/second_q/mappers/test_tapered_qubit_mapper.py
```diff
@@ -0,0 +1,469 @@
...
+            jw_tqm = TaperedQubitMapper(self.jw_mapper)
+            jw_op_h2 = jw_tqm.map_clifford(self.h2_op)
+            self.assertTrue(jw_op_h2.primitive.equiv(TestTaperedQubitMapper.REF_H2_JW))
+
+            jw_tqm = TaperedQubitMapper.from_problem(self.jw_mapper, self.driver_result)
+            jw_op_h2 = jw_tqm.map_clifford(self.h2_op)
+            self.assertTrue(jw_op_h2.primitive.equiv(TestTaperedQubitMapper.REF_H2_JW_CLIF))
+            # Compose with symmetry cliffords if z2 not empty even if Num_particles is empty
+            jw_tqm.tapering_values = None
+            self.assertTrue(jw_op_h2.primitive.equiv(TestTaperedQubitMapper.REF_H2_JW_CLIF))
```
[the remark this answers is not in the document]
Anthony-Gandon: This tests checks that indeed the first step of the z2symmetry reduction does not depend on the tapering values. So the resulting operator is the same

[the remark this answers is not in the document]
Anthony-Gandon: Ah I forgot a line reevaluating the mapped operator

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -89,12 +90,15 @@ def __init__(
                 later on anyways.
         """
 
-        self._mapper: QubitMapper = mapper
-        self._two_qubit_reduction: bool = (
-            two_qubit_reduction  # Setter does reset on particle number
-        )
+        self._mapper: QubitMapper = None
+        self._inner_mapper: QubitMapper = None
```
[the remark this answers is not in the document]
Anthony-Gandon: Sure

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
mrossinek: Thanks, @ikkoham! I did not noticed this before and also just opened a PR to fix this in the routines in Nature: https://github.com/Qiskit/qiskit-nature/pull/1045

test/second_q/algorithms/excited_state_solvers/test_excited_states_solvers.py
```diff
@@ -118,6 +119,20 @@ def test_solve_with_vqe_mes(self, converter: QubitConverter):
...
+        ["PM_TQR", ParityMapper(num_particles=(1, 1))],
+    )
+    def test_solve_with_vqe_mes_mapper(self, mapper: QubitMapper):
+        """Test QEOM with VQEUCCFactory and various QubitMapper"""
+        estimator = Estimator()
+        solver = VQEUCCFactory(estimator, UCCSD(), SLSQP())
+        gsc = GroundStateEigensolver(mapper, solver)
+        esc = QEOM(gsc, estimator, "sd")
+        results = esc.solve(self.electronic_structure_problem)
+        self._assert_energies(results.computed_energies, self.reference_energies)
```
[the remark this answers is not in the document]
Anthony-Gandon: Done

test/second_q/algorithms/ground_state_solvers/test_groundstate_eigensolver_mapper.py
```diff
@@ -0,0 +1,334 @@
...
+from qiskit_nature.second_q.drivers import PySCFDriver
+from qiskit_nature.second_q.mappers import JordanWignerMapper, ParityMapper
+from qiskit_nature.second_q.mappers import TaperedQubitMapper
+from qiskit_nature.second_q.hamiltonians import ElectronicEnergy
+from qiskit_nature.second_q.transformers import FreezeCoreTransformer
+from qiskit_nature.second_q.algorithms.initial_points import MP2InitialPoint
+
+
+@unittest.skipIf(not _optionals.HAS_PYSCF, "pyscf not available.")
+class TestGroundStateEigensolverMapper(QiskitNatureTestCase):
```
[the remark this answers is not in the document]
Anthony-Gandon: Exactly

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: All public method have a PauliSumOp interface. I changed some notations to make it appear clearly. In this case, `get_qubit_operators()` used to return `untapped_aux_ops` but I renamed it to `untapped_aux_ops_sumop` to make the typing clear. The public method `symmetry_reduce_clifford()` also has a PauliSumOp interface

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: These lines are basically reimplementing `symmetry_reduce_clifford()`. They are contained inside a private method so I opted for a SparsePauliOp output without the symmetry reduction. Later on, they are converted to PauliSumOp and `symmetry_reduce_clifford()` is applied just before we pass them to `estimate_observables()`

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -639,16 +719,31 @@ def _build_qeom_pseudoeigenvalue_problem(
...
         # 1. Build all EOM operators to evaluate on the ground state
-        tap_eom_matrix_ops = self._build_all_eom_operators(
+        untap_eom_matrix_ops = self._build_all_eom_operators(
             untap_operator,
             expansion_basis_data,
         )
 
+        untap_eom_matrix_ops_sumop = {
+            key: PauliSumOp(op) for key, op in untap_eom_matrix_ops.items()
+        }
```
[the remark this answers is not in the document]
Anthony-Gandon: Yes precisely

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
woodsp-ibm: When I look at the sample code in the main conversation it looks like two different ways to create this - the constructor, which takes a mapper and Z2Symmetries, and then this which takes a mapper and problem. I ended up wondering whether perhaps we could/should invert the relationship and have the problem support say a get_tapered_mapper(mapper) method. The logic makes use of calls across to problem, and by placing it there, should other problems be able to do this in the future they can add their own equivalents rather than having to somehow grow out some central location handling different problems. Maybe the build out of second_q_ops can then be managed/cached otherwise its building operators here, and mapping them just to find symmetries - the build out and mapping will then happen again in the main path when the mapper created here gets used right? Maybe there is some other integration which can do this more on the fly with the main operator build/out processing so we only do that once. Assuming I'm not missing something here.

qiskit_nature/second_q/circuit/library/initial_states/hartree_fock.py
```diff
@@ -222,6 +226,7 @@ def hartree_fock_bitstring_mapped(
     Returns:
         The bitstring representing the mapped state of the Hartree-Fock state as array of bools.
     """
+    # TODO: Remove match convert argument
```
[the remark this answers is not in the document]
Anthony-Gandon: My understanding is that this argument match_convert was tailored for the symmetry sector locator. We have another way to separate tapering from mapping in the TaperedQubitMapper (we exposed the two methods) so I believe we don't need that anymore.

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_eigensolver.py
```diff
@@ -98,7 +98,7 @@ def get_qubit_operators(
             aux_ops = self._qubit_converter.convert_match(aux_second_q_ops)
         else:
             main_operator = self._qubit_converter.map(main_second_q_op)
-            aux_ops = self._qubit_converter.map(aux_second_q_ops)
+            aux_ops = self._qubit_converter.map(aux_second_q_ops, suppress_none=True)
```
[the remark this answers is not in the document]
Anthony-Gandon: Now that the mapper can implement z2symmetries, it is possible that map() returns None in its dictionary. Additionally, measurements on None operator fail so this is the new way to suppress operators that do not commute with the symmetry (when check_commutes = True).
Even before the second_q stack, the convert_match() method from the qubit_convert was doing that under the hood

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -193,13 +269,12 @@ def get_qubit_operators(
...
         # 1. Convert the main operator (hamiltonian) to a Qubit Operator and apply two qubit reduction
-
         if isinstance(self.qubit_converter, QubitConverter):
             self.qubit_converter.force_match(num_particles=num_particles)
             main_op = self.qubit_converter.convert_only(main_operator, num_particles=num_particles)
+        elif isinstance(self.qubit_converter, TaperedQubitMapper):
+            main_op = self.qubit_converter.map_clifford(main_operator)
         else:
-            if isinstance(self.qubit_converter, ParityMapper):
-                self.qubit_converter.num_particles = num_particles
```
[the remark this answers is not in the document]
Anthony-Gandon: After some thinking, the only use case is: a poorly defined ParityMapper with the wrong particle number used on a QEOM problem providing the "true" number of particles. We can keep it, I went too quickly on that maybe

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
Anthony-Gandon: Yes this is indeed a good point. from_problem() basically uses the problem to obtain operators to obtain symmetries. The get_tapered_mapper(mapper) would reduce back and forth calls
I'll just mention a caveat I just thought of. For how it translates to code, I imagine something along the lines of
```
driver = PySCFDriver()
self.driver_result = driver.run()
mapper = ParityMapper(num_particles = self.driver_result.num_particles)
tapered_qubit_mapper = self.driver_result.get_tapered_mapper(mapper)
qubit_op = tapered_qubit_mapper.map(self.h2_op).primitive

```
I felt like - 1) Driver 2) Mapper 3) TaperedQubitMapper.from_problem() 4) TaperedQubitMapper.map() - was more linear than  - 1) Driver 2) Mapper 3) Driver.get_tapered_mapper() 4) TaperedQubitMapper.map() -

mrossinek: Yeh basically this shifts the back and forth from inside the code to the workflow the user would have to deal with.
However, I still think that Steve has several good points here, especially considering the potential future integration with other problems.

Thus I would also vote for the `Problem.get_tapered_mapper(mapper)` design rather than the current approach. It does indeed feel more natural. This should also not be too surprising to users since the `symmetry_sector_locator` has always been on the problems (and this is effectively what enables them to support Z2Symmetry reductions). Thus, keeping this logic together on one object seems reasonable.

Anthony-Gandon:
My last commit is a proposal for the `get_tapered_mapper` method following Steve's comment.
It is not 100% fixed and decided, remaining questions/tasks include:

- [ ] Add abstract method `get_tapered_mapper ` in Base Problem?
- [ ] Decide on the typing of the `symmetry_sector_locator()`. It is a public method which has an argument Z2Symmetries. I think we need to transition to the quantum_info symmetries using `Old_Symmetries | New_Symmetries`. Because they have the same name, I believe we have to rename one of the two so I made a proposition.

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -84,7 +84,10 @@ def unwrap(self, wrapped_type: type, suppress_none: bool = False) -> Dict | Iter
             else:
                 return [op for _, op in iter(self)]
         if wrapped_type == dict:
-            return dict(iter(self))
+            if suppress_none:
+                return {key: op for key, op in iter(self) if op is not None}
+            else:
+                return dict(iter(self))
```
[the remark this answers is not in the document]
mrossinek: After an offline discussion, I realized that I was indeed wrong here. In the old scenario where a `list` was used for the operators, it was necessary to inject `None` for non-commuting operators such that the order of the operators does not change and their hard-coded index remained correct.
However, now with the `dict` approach, this is no longer necessary.

Later we added the `suppress_none` option to avoid having `None` as part of the output in the `list`. We can indeed (and actually must) do the same for the `dict` case now.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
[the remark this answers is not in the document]
mrossinek: https://github.com/Qiskit/qiskit-terra/issues/9490

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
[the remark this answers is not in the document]
mrossinek: I think we need to decide on what our expectation for the mappers are (i.e. whether or not we expect our algorithms to potentially overwrite the information in them).
I would like to avoid hard-coding cases like this as much as possible. Let us remove it and if it becomes an issue down the road, we can either re-add it or improve the documentation on this aspect.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -209,6 +284,8 @@ def get_qubit_operators(
         if self.solver.supports_aux_operators():
             if isinstance(self.qubit_converter, QubitConverter):
                 aux_ops = self.qubit_converter.convert_match(aux_second_q_operators)
+            elif isinstance(self.qubit_converter, TaperedQubitMapper):
+                aux_ops = self.qubit_converter.map_clifford(aux_second_q_operators)
             else:
                 aux_ops = self.qubit_converter.map(aux_second_q_operators)
```
mrossinek, line 290: I see this same structure in multiple places in this file. Could you extract this into a private method?

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -639,16 +719,31 @@ def _build_qeom_pseudoeigenvalue_problem(
...
+        if isinstance(self.qubit_converter, QubitConverter):
+            tap_eom_matrix_ops_sumop = self.qubit_converter.symmetry_reduce_clifford(
+                untap_eom_matrix_ops_sumop
+            )
+        elif isinstance(self.qubit_converter, TaperedQubitMapper):
+            tap_eom_matrix_ops_sumop = self.qubit_converter.taper_clifford(
+                untap_eom_matrix_ops_sumop, suppress_none=True
+            )
+        else:
+            tap_eom_matrix_ops_sumop = untap_eom_matrix_ops_sumop
```
mrossinek, line 740: Just like before, this same logic also appears multiple times. Extracting this into a separate private method would be good, too

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,177 @@
...
+            A qubit operator in the form of a PauliSumOp, or list (resp. dict) thereof if a list
+            (resp. dict) of second quantized operators was supplied.
+        """
+        wrapped_type = type(pauli_ops)
+
+        if issubclass(wrapped_type, PauliSumOp):
+            pauli_ops = [pauli_ops]
+            suppress_none = False
+
+        wrapped_pauli_ops: _ListOrDict[PauliSumOp] = _ListOrDict(pauli_ops)
```
mrossinek, line 154: I wonder whether we can also de-duplicate this boilerplate code via a `_ListOrDict.wrap` method which returns both `wrapped_type` and `wrapped_ops`.

The question would be how to deal with `suppress_none` but I think this should be manageable, too.

qiskit_nature/second_q/problems/base_problem.py
```diff
@@ -20,7 +20,9 @@
 import numpy as np
 from qiskit.algorithms.eigensolvers import EigensolverResult
 from qiskit.algorithms.minimum_eigensolvers import MinimumEigensolverResult
-from qiskit.opflow import Z2Symmetries
+from qiskit.opflow.primitive_ops import Z2Symmetries
+from qiskit.quantum_info.analysis.z2_symmetries import Z2Symmetries as Z2SparseSymmetries
```
mrossinek, line 24:
```suggestion
from qiskit.opflow.primitive_ops import Z2Symmetries as OpflowZ2Symmetries
from qiskit.quantum_info.analysis.z2_symmetries import Z2Symmetries
```
I would rather rename the import which is supposed to go away in the future.
Could you please update this accordingly elsewhere, too?

qiskit_nature/second_q/problems/electronic_structure_problem.py
```diff
@@ -312,11 +312,32 @@ def symmetry_sector_locator(
...
+    ) -> List[int]:
         # Finding all the symmetries using the find_Z2_symmetries:
         taper_coeff: List[int] = []
         for sym in z2_symmetries.symmetries:
             coeff = -1 if np.logical_xor.reduce(np.logical_and(sym.z, hf_str)) else 1
             taper_coeff.append(coeff)
 
         return taper_coeff
+
+    def get_tapered_mapper(self, mapper: QubitMapper) -> TaperedQubitMapper:
```
mrossinek, line 326: It should be possible to define this directly on `BaseProblem`, right?

releasenotes/notes/add-get-tapered-mapper-for-problems-33292452930141d1.yaml
```diff
@@ -0,0 +1,13 @@
+---
+features:
+  - |
+    Adds the method 
+    :meth:`~qiskit_nature.second_q.problems.electronic_structure_problem.get_tapered_mapper()` to
```
mrossinek, line 5:
```suggestion
    :meth:`~qiskit_nature.second_q.problems.BaseProblem.get_tapered_mapper()` to
```

releasenotes/notes/add-get-tapered-mapper-for-problems-33292452930141d1.yaml
```diff
@@ -0,0 +1,13 @@
...
+    transform a :class:`~qiskit_nature.second_q.mappers.QubitMapper` instance into a 
+    :class:`~qiskit_nature.second_q.mappers.TaperedQubitMapper` based on the properties of the current
+    problem.
+  - |
+    Update API for the method
+    :meth:`~qiskit_nature.second_q.problems.electronic_structure_problem.symmetry_sector_locator()`
```
mrossinek, line 11:
```suggestion
    :meth:`~qiskit_nature.second_q.problems.BaseProblem.symmetry_sector_locator()`
```

qiskit_nature/second_q/circuit/library/ansatzes/ucc.py
```diff
@@ -316,9 +316,6 @@ def operators(self):  # pylint: disable=invalid-overridden-method
                         excitation_ops, suppress_none=False
                     )
                 else:
-                    # TODO: Issue #974 sketches the construction of a Tapered Qubit Mapper which would
-                    # implement the logic of the symmetries. Here, there should be a check for a Tapered
-                    # Qubit Mapper and a similar logic that used above.
                     operators = self.qubit_converter.map(excitation_ops)
```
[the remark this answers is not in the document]
Anthony-Gandon: Yes exactly

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -90,11 +91,11 @@ def __init__(
         """
 
         self._mapper: QubitMapper = mapper
-        self._two_qubit_reduction: bool = (
-            two_qubit_reduction  # Setter does reset on particle number
-        )
+        self._two_qubit_reduction: bool = two_qubit_reduction
+        # Setter does reset on particle number
```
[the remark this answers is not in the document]
Anthony-Gandon: Not relevant anymore indeed

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -90,11 +91,11 @@ def __init__(
...
         self._z2symmetry_reduction: Optional[Union[str, List[int]]] = None
-        self.z2symmetry_reduction = z2symmetry_reduction  # Setter does validation
+        self.z2symmetry_reduction = z2symmetry_reduction
+        # Setter does validation
```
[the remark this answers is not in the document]
Anthony-Gandon: Sure

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -128,7 +131,7 @@ def _map_single(self, second_q_op: SparseLabelOp) -> PauliSumOp:
     def map(
         self,
         second_q_ops: SparseLabelOp | ListOrDictType[SparseLabelOp],
-        suppress_none: bool = None,
+        suppress_none: bool = False,
```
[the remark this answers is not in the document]
mrossinek: Oh sorry I misread. We can leave it as is then :+1:

qiskit_nature/second_q/problems/base_problem.py
Anthony-Gandon: sure

qiskit_nature/second_q/problems/electronic_structure_problem.py
Anthony-Gandon: Done

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -157,6 +158,10 @@ def num_particles(self) -> Optional[Tuple[int, int]]:
         """
         if isinstance(self._mapper, ParityMapper):
             return self._mapper.num_particles
+        elif isinstance(self._mapper, TaperedQubitMapper) and isinstance(
+            self._mapper.mapper, ParityMapper
+        ):
+            return self._mapper.mapper.num_particles
```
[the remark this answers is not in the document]
Anthony-Gandon: I was going for the full retrocompatibilty but it is true that we don't need it on the QubitConverter. I'm fine with defending against TaperedQubitMapper in QubitConverter

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
Anthony-Gandon: I'm thinking of a wrap(xx, suppress_none) which would write onto a private attribute suppress_none of the class _ListOrDict. 
Additionally, I think the suppress_none for single operators is actually not needed. We can see in the unwrap() method that True/False would give the same result when wrapped_type is not List/Dict. I feel like it is similar to a comment above, i.e. something that was kept but the underlying logic has changed. I'll check that

Anthony-Gandon: Similarly, we may want wrapped_type to be an attribute of _ListOrDict. I'm not sure though that we want to complexify this class.

qiskit_nature/second_q/mappers/__init__.py
```diff
@@ -93,6 +93,7 @@
 from .spin_mapper import SpinMapper
 from .vibrational_mapper import VibrationalMapper
 from .interleaved_qubit_mapper import InterleavedQubitMapper
+from .tapered_qubit_mapper import TaperedQubitMapper
```
mrossinek, line 96: Just a quick comment: having fixed the merge conflict here just now I noticed that the docstring in this file is not being updated to include the `TaperedQubitMapper`.
Please add a new section to include that.

PS: I will do another full review later today :+1:

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -87,15 +88,18 @@ def __init__(
                 results which can occur when operator terms are not consistently ordered.
                 This is disabled by default, because in practice the Pauli-terms will be grouped
                 later on anyways.
-        """
 
+        Raises:
+            ValueError: If the mapper is a ``TaperedQubitMapper``.
+        """
+        if isinstance(mapper, TaperedQubitMapper):
+            raise ValueError(f"{type(mapper)} is not supported by the converter.")
```
mrossinek, line 96:
```suggestion
        if isinstance(mapper, TaperedQubitMapper):
            raise ValueError(
                "The TaperedQubitMapper is not supported by the QubitConverter. If you want to using tapering "
                "please either use the tapering built directly into the QubitConverter (see its documentation) "
                "or use the TaperedQubitMapper standalone (recommended)."
            )

```
I think this will be a bit more helpful for the user. I did not pay attention to line wrapping while writing this in the suggestion on Github so re-wrapping may be necessary.

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -236,8 +240,9 @@ def convert_only(
         Returns:
             PauliSumOp qubit operator
         """
-        if isinstance(self._mapper, ParityMapper) and num_particles is not None:
-            self._mapper.num_particles = num_particles
+        if num_particles is not None:
+            if isinstance(self._mapper, ParityMapper):
+                self._mapper.num_particles = num_particles
```
mrossinek, line 245:
```suggestion
        if num_particles is not None and isinstance(self._mapper, ParityMapper):
            self._mapper.num_particles = num_particles
```

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -263,8 +268,9 @@ def force_match(
         Raises:
             ValueError: If format of Z2Symmetry tapering values is invalid
         """
-        if isinstance(self._mapper, ParityMapper) and num_particles is not None:
-            self._mapper.num_particles = num_particles
+        if num_particles is not None:
+            if isinstance(self._mapper, ParityMapper):
+                self._mapper.num_particles = num_particles
```
mrossinek, line 273:
```suggestion
        if num_particles is not None and isinstance(self._mapper, ParityMapper):
            self._mapper.num_particles = num_particles
```

releasenotes/notes/add-get-tapered-mapper-for-problems-33292452930141d1.yaml
```diff
@@ -0,0 +1,13 @@
...
+    Adds the method 
+    :meth:`~qiskit_nature.second_q.problems.BaseProblem.get_tapered_mapper()` to
+    transform a :class:`~qiskit_nature.second_q.mappers.QubitMapper` instance into a 
+    :class:`~qiskit_nature.second_q.mappers.TaperedQubitMapper` based on the properties of the current
+    problem.
+  - |
+    Update API for the method
+    :meth:`~qiskit_nature.second_q.problems.BaseProblem.symmetry_sector_locator()`
+    to accept ``Z2Symmetries`` as defined in the quantum_info module as well as the ``Z2Symmetries``
+    defined in the opflow module.
```
mrossinek, line 13: This also needs to mention the addition of the `TaperedQubitMapper` itself. Ideally, that should also show an example of how this is used (you can also reuse the `get_tapered_mapper` feature for that example.

Anthony-Gandon:
> I did extracted the open tasks which I deferred to the future into #1049. @Anthony-Gandon if you could double check whether I missed anything there, that would be great!

I went through all the comments and also think that those are the task that we can complete on a separate PR.

qiskit_nature/second_q/mappers/__init__.py
```diff
@@ -78,6 +78,16 @@
    :nosignatures:
 
    QubitConverter
+
+Tapered Qubit Mapper
```
mrossinek, line 82: Can we move this *above* the `QubitConverter` section?

Also, it would be great if we can add one or two sentences to provide a little bit more context (similar to what is done for the `InterleavedQubitMapper` above. Just so that it is easier to see without navigating to the next page

qiskit_nature/second_q/mappers/qubit_mapper.py
```diff
@@ -147,18 +150,15 @@ def map(
 
         if issubclass(wrapped_type, SparseLabelOp):
             second_q_ops = [second_q_ops]
-            suppress_none = False
```
mrossinek, line 150: Are we truely able to remove this? What if map a single operator which turns out not to commute? What will be returned then? If `None`, the return type hint needs to also include `| None`

qiskit_nature/second_q/mappers/tapered_qubit_mapper.py
```diff
@@ -0,0 +1,185 @@
...
+    ):
+        """
+        Args:
+            mapper: ``QubitMapper`` object implementing the mapping of second quantized operators to
+                Pauli operators.
+            z2symmetries: ``Z2Symmetries`` object defining the symmetries to use to reduce the Pauli
+                operators.
+
+        Raises:
+            ValueError: If the input mapper is also a ``TaperedQubitMapper``.
```
mrossinek, line 60:
```suggestion
            ValueError: If the input mapper is already a ``TaperedQubitMapper``.
```

releasenotes/notes/add-get-tapered-mapper-for-problems-33292452930141d1.yaml
```diff
@@ -0,0 +1,49 @@
+---
+features:
+  - |
+    Adds a Tapered Qubit Mapper class. :class:`~qiskit_nature.second_q.mappers.TaperedQubitMapper` is to
+    be used as a wrapper of standard :class:`~qiskit_nature.second_q.mappers.QubitMapper` that can
```
mrossinek, line 5:
```suggestion
    be used as a wrapper of another standard :class:`~qiskit_nature.second_q.mappers.QubitMapper` that can
```

releasenotes/notes/add-get-tapered-mapper-for-problems-33292452930141d1.yaml
```diff
@@ -0,0 +1,49 @@
...
+      h2_op, _ = electronic_structure_problem.second_q_ops()
+      mapper = JordanWignerMapper()
+      tapered_qubit_mapper = electronic_structure_problem.get_tapered_mapper(mapper)
+      qubit_op = tapered_qubit_mapper.map(h2_op)
+
+  - |
+    Updates API for the method
+    :meth:`~qiskit_nature.second_q.problems.BaseProblem.symmetry_sector_locator()`
+    to accept ``Z2Symmetries`` as defined in the quantum_info module as well as the ``Z2Symmetries``
+    defined in the opflow module.
```
mrossinek, line 49:
```suggestion
    to accept the new :class:`~qiskit.quantum_info.analysis.z2_symmetries.Z2Symmetries` from ``quantum_info``
    as well as the legacy the :class:`~qiskit.opflow.Z2Symmetries` from the ``opflow`` module.
```
Having these links work seems like a nice addition

mrossinek (commented):
Alright these are my final comments :+1:

mrossinek:
I pushed another cleanup commit around `suppress_none`.. Having the default `suppress_none=True` is desirable because this aligns the `TaperedQubitMapper` more with the "normal" ones (`None` cannot be returned by default).
Also, the overwriting `TaperedQubitMapper.map` was actually redundant because `_map_single` actually does the correct logic (it simply avoids the commutativity check which I think is fine). That is really only required for our initial states anyways.

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_eigensolver.py
```diff
@@ -121,8 +119,8 @@ def get_qubit_operators(
                         name_aux,
                     )
                 # The custom op overrides the default op if the key is already taken.
-                aux_ops[name_aux] = converted_aux_op
-
+                if converted_aux_op is not None:
+                    aux_ops[name_aux] = converted_aux_op
```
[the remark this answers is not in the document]
mrossinek: This also currently only changes the ground-state case and not the excited-states one.
So I really think this change needs to be reverted

mrossinek (approved):
Thank you, Anthony, for all of your hard work here! It is greatly appreciated! :slightly_smiling_face: :tada:

Issue #879, opened 2022-10-14
Mappers should return SparsePauliOp instead of PauliSumOp


kevinsung:
### What is the expected enhancement?

See title.

mrossinek:
Thanks for opening this issue. As discussed offline, there is multiple aspects that need to happen here which would be good to describe. Otherwise this issue as such is not that useful.

For context: with the move to the primitives in `qiskit.algorithms` as part of the 0.22 release of Qiskit Terra, we started to phase out the usage of the `qiskit.opflow` module in favor of the `qiskit.quantum_info`. As such, an effort was made to only rely on `PauliSumOp` as the final `opflow` object still being used in the public API.
The reason for not transitioning fully to `quantum_info` is the implementation of the `Z2Symmetries` which has not been ported.
More specifically, there are 2 parts here:
- the general `Z2Symmetries`
- the more specific `TwoQubitReduction`
The latter is very specific to the `ParityMapper` here in Nature so I would suggest to migrate this back to Nature.
The former (afaik) is also only being used by Qiskit Nature, although technically it can also be used elsewhere, if one is able to determine the correct sector of the Hilbert space, in which the desired state lies. A discussion will need to be held, on whether to migrate the `Z2Symmetries` to `quantum_info` or to Nature.

All that said, the importance of these qubit reduction techniques is going to diminish over time, since the removal of a handful of qubits will have less of an impact the larger the total number of qubits (i.e. relatively speaking, the reduction may not be worth the classical overhead). This is however still a long-term aspect to figure out, and for the foreseeable future an implementation of the `Z2Symmetries` is highly desirable.

<hr/>

Once the above has been figured out, Nature will be able to transition to the usage of `SparsePauliOp`s.

mrossinek:
This was just implemented as part of #1050. On the latest main branch (which will eventually be released as 0.6.0), you can now enable this feature already by doing the following:
```python
from qiskit_nature import settings

settings.use_pauli_sum_op = False
```
Afterwards, your mappers will produce `SparsePauliOp` instances instead of `PauliSumOp` ones.

PR #1034, opened 2023-01-18
Refactor tutorial to use primitives

manoelmarques:


docs/tutorials/12_deuteron_binding_energy.ipynb
```diff
@@ -37,15 +37,15 @@
...
+    "from qiskit.primitives import Estimator\n",
+    "from qiskit.algorithms.minimum_eigensolvers import VQE\n",
+    "from qiskit.algorithms.observables_evaluator import estimate_observables\n",
     "from qiskit.algorithms.optimizers import COBYLA, SLSQP\n",
     "from qiskit.circuit import QuantumCircuit, Parameter\n",
     "from qiskit.circuit.library import TwoLocal\n",
-    "from qiskit.opflow import I, X, Y, Z, PauliSumOp\n",
-    "from qiskit.opflow.expectations import PauliExpectation\n",
-    "from qiskit.utils import QuantumInstance, algorithm_globals\n",
+    "from qiskit.opflow import PauliSumOp\n",
```
mrossinek, line 46: Are we able to also already get rid of `PauliSumOp` here?

docs/tutorials/12_deuteron_binding_energy.ipynb
```diff
@@ -488,99 +466,53 @@
    "id": "ad7cb72e",
    "metadata": {},
    "source": [
-    "The results obtained above will be used as our reference values. Thus, we can use them to see if the statevector and the qasm simulators give us good results. In the code cell below we ran the `VQE` algorithm for each pair of `ansatz` and Hamiltonian, using the statevector simulator and the `SLSQP` optimizer. "
+    "The results obtained above will be used as our reference values. Thus, we can use them to see if the Estimator give us good results. In the code cell below we ran the `VQE` algorithm for each pair of `ansatz` and Hamiltonian, using the Estimator and the `SLSQP` optimizer. "
```
mrossinek, line 469:
```suggestion
    "The results obtained above will be used as our reference values. Thus, we can use them to see if the Estimator gives us good results. In the code cell below we ran the `VQE` algorithm for each pair of `ansatz` and Hamiltonian, using the Estimator and the `SLSQP` optimizer. "
```

mrossinek (commented):
Thank you! Just two minor suggestions :+1:

docs/tutorials/12_deuteron_binding_energy.ipynb
manoelmarques: Yes, I remove it from the tutorial. The convert method still returns PauliSumOp but I extract the SparsePauliOp from it.

mrossinek: Yeh I suppose this is fine for now. Once we resolve #879 we can also change this piece of code to use the `JordanWignerMapper` directly rather than wrap it with a `QubitConverter` :+1:
This should happen prior to v0.6 anyways

mrossinek (approved):

mrossinek:
In order to republish the docs, does this need to be backported, @manoelmarques?

PR #1037, opened 2023-01-20
Add missing QEOMResult docs

### LINKED ISSUE #1019: QEOMResult needs docs and other minor issues
In regards of the newly updated [QEOMResult](https://github.com/Qiskit/qiskit-nature/blob/00d85e3fcca3c405dbe8fd8027784ad2fbca3971/qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py#L916) 

1. What was all getter/setters before in the result many have been changed to public instance vars but there are no docs for these - consider using `Attributes` to document these. With getter/setters the type could be different - maybe the None aspect is really invalid so when this is documented, although the type now says | None, maybe this should not be exposed. I don't think setting any of these to None is really valid right, we would expect them all to be set. 

2. A public instance field `eigenvalues` has been added yet the parent EigensolverResult has this already and I think that should be used given its the same name.

3. In terms of some of the old setters (W, Q...) being deprecated. If code is still using these - presumably some other QEOM implementation building a result, it will not set S or H so will the result be valid - well S and H will end up being None so it will look like the older result. If these new fields are not being read then it will not be an issue I guess - it would only be the case where its being setup the old way but read by new code that now expects S and H to be set. I guess there is no easy way around this. Maybe we should add some documentation that explains things - though most users I think are going to just be reading the result not  creating one. In terms of the deprecation messages maybe the docs will help as they are at present for the setters they say something like `The W matrix is now computed from the S matrix` which may imply you should be setting S now but the messages could be more prescriptive e.g. `You should now set the S matrix from which W will be computed`. But maybe things will be clearer overall if this transition over to H and S is mentioned in some docs.

Anthony-Gandon:
### Summary

Fix #1019

- This PR adds the missing docstring of the QEOMResultClass.
- It also deprecates the M, Q, W, V std setters which had been forgotten.

### Details and comments

- [x] Remaining is the issue with the typing None for public attributes in the `QEOMResult`.


qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -954,14 +954,31 @@ def _build_qeom_result(
...
+            set of basis operators spanning the linear qEOM subspace.
+        eigenstates: Approximated excited states of the hamiltonian. Note that these are set to an empty
+            list as they cannot be provided as circuits by the qEOM algorithm but are required for the
+            results.
+        h_matrix: Matrix representing the Hamiltonian in the qEOM subspace.
+        s_matrix: Matrix representing the geometry of the qEOM subspace.
+        h_matrix_std: 2 by 2 matrix of the summed standard deviation of all elements of the matrices
+            M, Q and their conjugates.
+        s_matrix_std: 2 by 2 matrix of the summed standard deviation of all elements of the matrices
+            V, W and their conjugates.
```
mrossinek, line 972: Referring to m, q, v, and w here does not work since these are not explained anywhere else. I would suggest we rather just refer to H and S, respectively, and properly define these somewhere here (probably in the few lines above explaining `h_matrix` and `s_matrix`.

releasenotes/notes/update-qeomresult-docstring-and-pending-deprecate-mqwv-matrix-std-setters-5bc2d84a941dca72.yaml
```diff
@@ -0,0 +1,11 @@
+---
+features:
+  - |
+    Completed the missing documentation for the new attributes of the class
+    :class:`~qiskit_nature.second_q.algorithms.excited_state_solvers.qeom.QEOMResult`.
+
```
mrossinek, line 6:
```suggestion
```
Since the previous changes were not released yet, this note is not necessary

mrossinek (commented):
Some initial comments :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -954,14 +954,31 @@ def _build_qeom_result(
...
+    """The results class for the qEOM algorithm.
+
+    Attributes:
+        ground_state_raw_result: The raw results of the ground state eigensolver.
+        excitation_energies: The excitation energies approximated by the qEOM algorithm.
+        expansion_coefficients: The expansion coefficients matrix of the excitation operators onto the
+            set of basis operators spanning the linear qEOM subspace.
+        eigenstates: Approximated excited states of the hamiltonian. Note that these are set to an empty
+            list as they cannot be provided as circuits by the qEOM algorithm but are required for the
+            results.
```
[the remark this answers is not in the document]
Anthony-Gandon: I thought that the EigenstateResult.from_result() was requiring an attribute eigenstate from EigensolveResult. I was not aware that this behaviour was changed. (QEOM cannot produce the excited circuits but gives a proxy on how to evaluate observables on them)

releasenotes/notes/update-qeomresult-docstring-and-pending-deprecate-mqwv-matrix-std-setters-5bc2d84a941dca72.yaml
Anthony-Gandon: Done

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
Anthony-Gandon: Done

mrossinek (dismissed):
LGTM, thanks!

Since @woodsp-ibm raised the original issue, I will await a review from him before merging :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -939,24 +942,41 @@ def _build_qeom_result(
...
         return result
 
 
 class QEOMResult(EigensolverResult):
-    """The results class for the QEOM algorithm."""
+    """The results class for the qEOM algorithm.
+
+    For more details about the definitions, please refer to https://arxiv.org/abs/1910.12890.
+
+    Attributes:
```
woodsp-ibm, line 956: Can we add types to these fields - it goes in brackets after the name before the : (examples can be found elsewhere in nature). I think I would avoid putting None here - I don't think its really a valid value as such (unless it is in which case we should have it there) more that this is just as aspect of how the class is initialized one field at a time.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -268,6 +270,7 @@ def solve(
         Args:
             problem: A class encoding a problem to be solved.
             aux_operators: Additional auxiliary operators to evaluate.
+
         Returns:
             An interpreted :class:`~.EigenstateResult`. For more information see also
             :meth:`~.BaseProblem.interpret`.
```
woodsp-ibm, line 276: Now QEOMResult, it seems, does not appear in the docs. I think the only place where we expose this to an end user is via the raw_result field of the EigenstateResult that is returned here from this solve method. So I thinking that we can expand this returns text to state that fact and link the QEOMResult class here so that it can be navigated to from here directly in the context where its relevant, rather than try to put it elsewhere like other classes where we have them documented and exposed via the init.

Anthony-Gandon:
It seems that I got the linking to the QEOMResult wrong in the qeom.solve() because I cannot seem to make it clickable in the docs. Any clue on how to fix it?

mrossinek:
> It seems that I got the linking to the QEOMResult wrong in the qeom.solve() because I cannot seem to make it clickable in the docs. Any clue on how to fix it?

I think it is probably just that type in the link which I suggested to correct above just now

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -1030,7 +1033,28 @@ def _build_qeom_result(
 
 
 class QEOMResult(EigensolverResult):
-    """The results class for the QEOM algorithm."""
+    """The results class for the qEOM algorithm.
+
+    For more details about the definitions, please refer to https://arxiv.org/abs/1910.12890.
+
```
mrossinek, line 1039: We could add a sentence here stating that the following attributes exist on this class. But I think it would also be fine as is.

mrossinek (dismissed):
A minor comment but I would be fine with merging this as is, too. Thanks, Anthony!

releasenotes/notes/update-qeomresult-docstring-and-pending-deprecate-mqwv-matrix-std-setters-5bc2d84a941dca72.yaml
```diff
@@ -0,0 +1,6 @@
+---
+deprecations:
```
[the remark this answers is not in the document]
Anthony-Gandon: Sure, done

releasenotes/notes/pending-deprecate-mqwv-matrix-setter-1cd9e2dfb68d1a85.yaml
```diff
@@ -1,7 +1,6 @@
 ---
 deprecations:
   - |
-    The M, Q, W, V matrix setters from
+    The M, Q, W, V matrix setters and M, Q, W, V matrix standard deviation setters from
     :class:`~qiskit_nature.second_q.algorithms.excited_state_solvers.qeom.QEOMResult`
```
woodsp-ibm, line 5:
```suggestion
    :class:`~qiskit_nature.second_q.algorithms.excited_state_solvers.qeom.QEOMResult`
```

woodsp-ibm: I think that typo prevents it coming out as a link

Anthony-Gandon: oops sry about that

woodsp-ibm (approved):
Thanks for fixing things @Anthony-Gandon - LGTM

PR #1038, opened 2023-01-20
Add `KagomeLattice` class

### LINKED ISSUE #957: Add a `KagomeLattice`
### What should we add?

A new lattice should be added to the `qiskit_nature.second_q.hamiltonians.lattices` module, allowing the simple construction of [Kagome lattices](https://en.wikipedia.org/wiki/Trihexagonal_tiling#Kagome_lattice).

(This suggestion was extracted from #43)

Unathi-Skosana:
### Summary
Fixes #957


qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,305 @@
...
+            rows: Length of the x direction.
+            cols: Length of the y direction.
+            edge_parameter: Weight on all the edges, specified as a single value
+                Defaults to 1.0,
+            onsite_parameter: Weight on the self-loops, which are edges connecting a node to itself.
+                Defaults to 0.0.
+            boundary_condition: Boundary condition for the lattice.
+                The available boundary conditions are:
+                BoundaryCondition.OPEN, BoundaryCondition.PERIODIC.
+                Defaults to BoundaryCondition.OPEN.
```
mrossinek, line 226:
```suggestion
                ``BoundaryCondition.OPEN``, ``BoundaryCondition.PERIODIC``.
                Defaults to ``BoundaryCondition.OPEN``.
```
This typesets them as verbatim

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,305 @@
...
+        self.cols = cols
+        self.size = (rows, cols)
+        self.dim = 2
+        self.boundary_condition = boundary_condition
+        self.num_sites_per_cell = 3
+        self.cell_positions = np.array([[0, 0], [1, 0], [1 / 2, np.sqrt(3) / 2]])
+        self.basis = np.array([[2, 0], [1, np.sqrt(3)]])
+
+        self.edge_parameter = edge_parameter
+        self.onsite_parameter = onsite_parameter
```
mrossinek, line 241: These cannot be simple attributes unless setting them also has the intended effects on the underlying graph. In the other classes all of these are private attributes for that reason and then we only expose the relevant ones as "read-only" `@property` attributes

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,305 @@
...
+
+        Args:
+            self_loop: Draw self-loops in the lattice. Defaults to False.
+            style : Styles for rustworkx.visualization.mpl_draw.
+                Please see
+                https://qiskit.org/documentation/rustworkx/stubs/rustworkx.visualization.mpl_draw.html#rustworkx.visualization.mpl_draw
+                for details.
+        """
+        graph = self.graph
+        num_sites_per_cell = self.num_sites_per_cell
```
mrossinek, line 280: What exactly does this do? I see that this is set to 3 in the `__init__`. Can this ever take another value? If not, I suggest hard-coding it differently.

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,305 @@
...
+        if style.pos is None:
+            style.pos = {}
+            for cell_idx in range(np.prod(size)):
+                cell_coord = np.array(divmod(cell_idx, size[0])[::-1])
+
+                for i in range(3):
+                    node_i = num_sites_per_cell * cell_idx + i
+                    style.pos[node_i] = (
+                        np.dot(cell_coord, self.basis) + self.cell_positions[i]
+                    ).tolist()
```
mrossinek, line 297: Except for this section, this method is identical to `HyperCubicLattice.draw_without_boundary`. Can we avoid this code duplication by creating a private method which is called when `if style.pos is None`. It could look something like this:
```python
def _style_pos(self, ...) -> dict:
    # do logic
    # return dict
    
# and then it is called like:
if style.pos is None:
    style.pos = self._style_pos()
```
The same refactoring can be done to the `TriangularLattice` class

test/second_q/hamiltonians/lattices/test_kagome_lattice.py
```diff
@@ -0,0 +1,317 @@
...
+        cols = 2
+        num_sites_per_cell = 3
+        edge_parameter = 1 - 1j
+        onsite_parameter = 0.0
+        boundary_condition = BoundaryCondition.OPEN
+        kagome = KagomeLattice(rows, cols, edge_parameter, onsite_parameter, boundary_condition)
+        with self.subTest("Check the graph."):
+            target_graph = PyGraph(multigraph=False)
+            target_graph.add_nodes_from(range(num_sites_per_cell * 6))
+            weighted_list = [
```
mrossinek, line 39: Since you are reusing almost identical graphs here, it should be possible to avoid duplicating all of these weight lists. See for example [this comment](https://github.com/Qiskit/qiskit-nature/pull/1027#discussion_r1083727357) as inspiration.

mrossinek (commented):
Generally this is already looking quite good. I have some initial comments/questions but will need to do a more thorough review later this week/early next week

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
Unathi-Skosana: This is used to appropriately enumerate node labels belonging to a unit cell given the cell index (and appropriately compute the number of nodes from `size`). It is fixed among instances; this is also true for `dim`, `cell_positions` and `basis`. All four could be assigned as class constants:

```python
class KagomeLattice(Lattice):
    ...
    # Dimension of lattice
    _dim = 2

    # Number of sites in a unit cell
    _num_sites_per_cell = 3

    # Translation vectors in each direction
    _basis = ...
    
    # Relative positions (relative to site 0) of sites in a unit cell
    _cell_positions = ...

    def __init__(self, ...):
    
    ...
```

Is this what you had in mind?

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,305 @@
...
+        rows: int,
+        cols: int,
+        edge_parameter: complex = 1.0,
+        onsite_parameter: complex = 0.0,
+        boundary_condition: BoundaryCondition = BoundaryCondition.OPEN,
+    ) -> None:
+        """
+        Args:
+            rows: Length of the x direction.
+            cols: Length of the y direction.
```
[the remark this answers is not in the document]
Unathi-Skosana: Yes, each length is the number of unit cells in the respective direction.

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
Unathi-Skosana: Since there would be a lot of resemblance between `_default_position` and `_style_pos`, we can add a keyboard argument `with_boundaries` to `_default_position` like so:

```python
def _default_position(self, with_boundaries: bool = True) -> Dict[int, List[float]]:
    ...
    if with_boundaries and boundary_condition == BoundaryCondition.PERIODIC:
        width = 0.2
        ...
```

and reuse it in `draw_without_boundary` with `with_boundaries` set to `False`?

mrossinek (commented):
I think this is shaping up quite well :+1: I have some more comments and I am noticing a few more general aspects that we should improve in our Lattice classes... But I will gather those in a separate issue tomorrow.

mrossinek:
As a side-note: you may have noticed that I assigned this to the 0.7 milestone. The reason for that is that we need to fully focus on the refactoring of the `QubitMapper` classes for the 0.6 release, as well as the transition from `PauliSumOp` to `SparsePauliOp`.

That is to say: I will try to keep reviewing this as we go, but I have to prioritize the other PRs higher. But this PR is already in a great state and we will soon have it in a state where it is ready-to-go. But even once that is the case, we may have to hold off with actually merging it until 0.6 is released (because the plans for 0.7 are better aligned with this feature which would give it more visibility than it would obtain when "thrown in with the mappers").

(just so you are aware why the review process is a bit slow :upside_down_face:)

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,411 @@
...
+                cell_b_0 = num_sites_per_cell * cell_b_idx
+                list_of_edges.append((cell_a_1, cell_b_0, edge_parameter))
+
+            # one cell north f not at the north boundary
+            if y != cols - 1:
+                cell_b_idx = self._coordinate_to_index(np.array([x, y]) + np.array([0, 1]))
+                cell_b_0 = num_sites_per_cell * cell_b_idx
+                list_of_edges.append((cell_a_2, cell_b_0, edge_parameter))
+
+            # one cell west and north if not at west north boundary
```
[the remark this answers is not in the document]
Unathi-Skosana: annotated kagome

Not quite; north east neighboring cells do not have direct connections. See the above diagram.

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,417 @@
...
+            list[tuple[int, int, complex]]: list of weighted edges that cross the boundaries.
+        """
+        list_of_edges = []
+        size = self._size
+        edge_parameter = self._edge_parameter
+        num_sites_per_cell = self._num_sites_per_cell
+        boundary_condition = self._boundary_condition
+        rows, cols = size
+        # add edges when the boundary condition is periodic.
+        if boundary_condition == BoundaryCondition.PERIODIC:
```
mrossinek, line 167: In the case of the `SquareLattice` we support either a single `BoundaryCondition` value (for none or full periodicity) or a pair to enable only-x or only-y perioidicity.
Is this something you consider adding or is this out-of-scope for you at this time? The latter is fine, I just wanted to ask :+1:

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,417 @@
...
+        else:
+            raise ValueError(
+                f"Invalid `boundary condition` {boundary_condition} is given."
+                "`boundary condition` must be " + " or ".join(str(bc) for bc in BoundaryCondition)
+            )
+        return list_of_edges
+
+    def _default_position(self, with_boundaries: bool = True) -> dict[int, list[float]]:
+        """Return a dictionary of default positions for visualization of a two-dimensional lattice.
+
```
mrossinek, line 230: This is missing the documentation of the `with_boundaries` argument.

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,417 @@
...
+        elif boundary_condition == BoundaryCondition.OPEN:
+            pass
+        else:
+            raise ValueError(
+                f"Invalid `boundary condition` {boundary_condition} is given."
+                "`boundary condition` must be " + " or ".join(str(bc) for bc in BoundaryCondition)
+            )
+        return list_of_edges
+
+    def _default_position(self, with_boundaries: bool = True) -> dict[int, list[float]]:
```
mrossinek, line 228:
```suggestion
    def _default_position(self, *, with_boundaries: bool = True) -> dict[int, list[float]]:
```
This makes this a keyword-only argument which I think is appropriate here

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,417 @@
...
+        Returns:
+            dict[int, list[float]] : The keys are the labels of lattice points,
+                and the values are two-dimensional coordinates.
+        """
+        size = self._size
+        boundary_condition = self._boundary_condition
+        num_sites_per_cell = self._num_sites_per_cell
+        pos = {}
+        width = 0.0
+        if with_boundaries and boundary_condition == BoundaryCondition.PERIODIC:
```
mrossinek, line 240: Actually, seeing this: is this really needed at all? Why can we not infer this directly from `self._boundary_condition` like it is done in the `HyperCubicLattice`?

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,417 @@
...
+            # the positions are shifted so that the edges between boundaries can be seen
+            # for the periodic cases.
+            cell_coord = np.array(divmod(cell_idx, size[0])[::-1])
+
+            for i in range(num_sites_per_cell):
+                node_i = num_sites_per_cell * cell_idx + i
+                pos[node_i] = (np.dot(cell_coord, basis) + cell_positions[i]).tolist()
+        return pos
+
+    def __init__(
```
mrossinek, line 283: Please move the definition of the `__init__` above all of the other private methods (but not the class variables).

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,417 @@
...
+        Args:
+            rows: Number of unit cells in the x direction.
+            cols: Number of unit cells in the y direction.
+            edge_parameter: Weight on all the edges, specified as a single value
+                Defaults to 1.0,
+            onsite_parameter: Weight on the self-loops, which are edges connecting a node to itself.
+                Defaults to 0.0.
+            boundary_condition: Boundary condition for the lattice.
+                The available boundary conditions are:
+                ``BoundaryCondition.OPEN``, ``BoundaryCondition.PERIODIC``.
```
mrossinek, line 301:
```suggestion
                :attr:`.BoundaryCondition.OPEN`, :attr:`.BoundaryCondition.PERIODIC`.
```
I think this should make these clickable links in the documentation. Please do the same elsewhere

mrossinek (commented):
I finally had time to give this PR another pass. Overall I think this is already looking great!
I left a few (nitpicky) comments but I think it should not take much to wrap this up. Great work so far already! :+1:

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,417 @@
...
+        # add self-loops
+        self_loop_list = self._self_loops()
+        graph.add_edges_from(self_loop_list)
+
+        # add edges that cross the boundaries
+        boundary_edge_list = self._boundary_edges()
+        graph.add_edges_from(boundary_edge_list)
+
+        # a list of edges that depend on the boundary condition
+        self.boundary_edges = [(edge[0], edge[1]) for edge in boundary_edge_list]
```
woodsp-ibm, line 327: This ` self.boundary_edges` looks like a public instance variable - is that intended - do we expect this to be part of a public API? Elsewhere property getters are used to expose private instance vars as readonly. I can see this used in the draw without boundary method - if this is the intent perhaps it should just be a private instance var.

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
```diff
@@ -0,0 +1,417 @@
...
+        super().__init__(graph)
+        # default position
+        self.pos = self._default_position()
```
woodsp-ibm, line 330: This appears also to be a public instance var - in this case I do not see it being used at all though

mrossinek: `pos` being public is dictated by the `Lattice` base class

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
mrossinek: Yes, this one I am also not sure about. I think we should make it internal.
If (in the future) we find the need for it to be public we can always add property getter/setter methods around it.

In the draw method we can easily just use the private attribute since its within the same class

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
woodsp-ibm: Ah ok, I thought I skimmed the base class but I cannot what I looked at. In looking now I see it and also that it's not documented as a public attribute at all there despite being so. Not directly an issue for this PR but it ought to be corrected I guess since its part of the public API.

qiskit_nature/second_q/hamiltonians/lattices/kagome_lattice.py
Unathi-Skosana: For the former option, how should we handle the boundary that go diagonally (as in the last if-block inside `_bulk_edges`) when the boundary condition value enables either x or y periodicity only?

mrossinek: When you have only x or y, there should not be any diagonal periodicity. Think of it like a band going either along the x or y direction.
Once you have periodicity in both, that's where the "diagonal" case arises (because instead you would only get a "plus-shape" of the the x- and y-band overlapping).

So in short:
- periodic only in x: only horizontal periodicity
- periodic only in y: only vertical periodicity
- periodic in both: both + diagonal

mrossinek (approved):
I think this is good to go.

I updated the PR with the latest changes from `main` and resolved the last linter issues.

Thank you for your contribution to Qiskit Nature and for your patience throughout this lengthy (and stalled) review process!

Issue #1039, opened 2023-01-20
Deprecate `QubitMapper.allows_two_qubit_reduction`


mrossinek:
### What should we add?

Once #1018 is merged, the attribute `QubitMapper.allows_two_qubit_reduction` is no longer needed and should be deprecated.
When doing this, we need to carefully double check any implications this might have for the public API.

PR #1044, opened 2023-01-25
Various improvements to the `QubitMapper` classes

mrossinek:
### Summary

This PR adds the following improvements to the `QubitMapper` classes:

- [x] makes `suppress_none` a keyword-only argument everywhere (this matches how it was in the `QubitConverter`; this does not need a release note since it is an unreleased change)
- [x] deprecates the `QubitMapper.allows_two_qubit_reduction` attribute (#1039)
  - [x] needs to be added to the release note
- [x] renames the `nmodes` argument to `register_length` everywhere
  - [x] needs to be added to the release note

### Details and comments

The reason for that last change is in preparation for #831. I do plan to include allowing to overwrite `register_length` as part of a call to `QubitMapper.map` in this PR. This already works mostly but I still need to check the mappers which do not use the `QubitMapper.mode_based_mapping`.

EDIT: #831 is now fully implemented in this PR, too.

Finally, I would also like to raise the question of whether we need to keep the `FermionicMapper`, `SpinMapper`, and `VibrationalMapper` classes. These do not really add any benefit except typing but those are not even enforced at runtime.
I would thus suggest any of the following options:
1. remove these classes alltogether (respecting the deprecation period ofcourse..)
2. actually use these classes to enforce types at runtime (in the sense of an `isinstance` check in their `_map_single` methods and corresponding `super()._map_single(...)` calls in their subclasses
3. at least hide these from the public API since they honestly just add noise to the documentation page of the mappers since these are abstract and cannot be used by the users directly anyways.

EDIT: I opted to keep these classes but removed them from the public documentation page (since these are not actually usable by an end-user and add unnecessary noise). I updated the classes to type the public `.map` rather than private `._map_single` method such that an IDE can actually aid a user with these type hints.


woodsp-ibm:
I am not sure about not having FermionicMapper etc as part of docs. If say I look at ParityMapper in the current resultant docs I can see some method, that pretty much as a end user of the mapper I should not care about. But I don't, unlike before, see any `map` method, the one and only thing as an end user I normally would use. I cannot navigate up the class structure to see what it may inherit since the FermionicMapper link there is not operational, although maybe that class won't show map there either. If I happen to know its QubitMapper (now I would be able to go up once again from FermionicMapper were that included) I can go there, though then I see it more as a generic map method taking a SparseLabelOp. Maybe that common base method is all we would get now anyway. For any classes etc that are exposed to an end user, or developer to extend, we have normally included them. Maybe its not a lot better in regards of how map method gets exposed with including it  (not sure)  but at least one can navigate up through the parents in the docs.

mrossinek:
Thanks for raising this! I think this is actually a symptom of the underlying issue that we do not document inherited members in our classes. I have fixed this for the `mappers` module in the last commit but I think we should revisit this choice for all docs in a separate PR.
In fact, Qiskit Terra also already differentiates between two cases where classes should document the inherited members and where they should not. For the ad-hoc fix here, I mirrored that (actually the inverse because our default is the opposite). We should revisit this in a separate step and align our docs with that of Terra. I will open a separate issue for this shortly.

woodsp-ibm (approved):

PR #1045, opened 2023-01-27
fix: enforce zero absolute tolerance in commutator routines

mrossinek:
### Summary

Thanks to @ikkoham for spotting this [here](https://github.com/Qiskit/qiskit-nature/pull/1031#discussion_r1088531480)!

### DIFF
--- a/releasenotes/notes/fix-commutator-simplification-tolerance-09c72ce1adc0bbf0.yaml
+++ b/releasenotes/notes/fix-commutator-simplification-tolerance-09c72ce1adc0bbf0.yaml
@@ -1,0 +1,8 @@
+---
+fixes:
+  - |
+    The commutator methods like :meth:`~qiskit_nature.second_q.operators.commutators.commutator`,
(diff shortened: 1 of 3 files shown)


releasenotes/notes/fix-commutator-simplification-tolerance-09c72ce1adc0bbf0.yaml
```diff
@@ -0,0 +1,8 @@
+---
+fixes:
+  - |
+    The commutator methods like :meth:`~qiskit_nature.second_q.operators.commutators.commutator`,
```
woodsp-ibm, line 4:
```suggestion
    The commutator methods :meth:`~qiskit_nature.second_q.operators.commutators.commutator`,
```
As this is a comprehensive list of all 3 affected methods I think `like` could be removed as to me it would imply this is an example and there are others like this which were changed which were not listed here.

woodsp-ibm (approved):

PR #1046, opened 2023-01-27
feat: implement the `InterleavedQubitMapper`

### LINKED ISSUE #918: Support interleaved qubit ordering
### What should we add?

Currently, all methods which map from spin orbitals to qubits assume a block-ordered arrangement of the qubit register. That means that the first half of the qubits encodes the alpha- and second half the beta-spin orbitals.

It can make sense to specifically want to use an _interleaved_ ordering where odd indices encode alpha- and even indices encode the beta-spin registers. Supporting this via a global configuration setting would be ideal since this can ensure consistent behavior across the various components affected by this (mappers, circuits, etc.)

Alternatively, it may be possible to centralize this configuration via the mappers (since our initial state and ansatz circuits also use mappers). If doing this, the 2-qubit reduction in the case of the `ParityMapper` would also need to be able to handle this, giving another argument for extracting this qubit reduction functionality out of Terra and back into Nature (see also [this comment](https://github.com/Qiskit/qiskit-nature/issues/879#issuecomment-1278531697)).

mrossinek:
### Summary

This implements an `InterleavedQubitMapper` which enables the generation of interleaved-ordered qubit operators.
It being a wrapper to another `QubitMapper`, means it can be straight forwardly used across the stack.

### Details and comments

Closes #918

### DIFF
--- a/qiskit_nature/second_q/mappers/interleaved_qubit_mapper.py
+++ b/qiskit_nature/second_q/mappers/interleaved_qubit_mapper.py
@@ -1,0 +1,106 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2023.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""A QubitMapper wrapper to transform from block-ordered to interleaved qubits."""
+
+from __future__ import annotations
+
+from qiskit.opflow import PauliSumOp
+from qiskit.quantum_info import SparsePauliOp
+
+from qiskit_nature.second_q.operators import FermionicOp
+
+from .fermionic_mapper import FermionicMapper
+
+
+class InterleavedQubitMapper(FermionicMapper):
+    """A ``FermionicMapper`` wrapper returning interleaved-ordered operators.
+
+    This class is intended to be used with fermionic systems. Furthermore, it is designed to work
+    with ``FermionicMapper`` classes which map fermionic operators to the qubit space by site
+    (unlike for example the :class:`~qiskit_nature.second_q.mappers.BravyiKitaevSuperFastMapper`
+    which maps by interactions).
+
+    .. warning::
+
+       The mapper will _not_ perform any assertions on the wrapped ``FermionicMapper``. Thus,
+       wrapping a :class:`~qiskit_nature.second_q.mappers.BravyiKitaevSuperFastMapper` is valid code
+       which will indeed produce qubit operators for you. You will just not be able to interpret the
+       order of the qubits in the same way.
+
+    For site-based mappers, Qiskit Nature always arranges the qubits corresponding to the alpha-spin
+    and beta-spin components in a blocked fashion. I.e. the first half of the qubit register
+    corresponds to the alpha-spin components, and the second half to the beta-spin one, like so:
+
+    .. code-block::
+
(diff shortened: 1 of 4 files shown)


qiskit_nature/second_q/mappers/interleaved_qubit_mapper.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        Args:
+            mapper: the actual ``QubitMapper`` mapping second-quantization operators to qubit
+                operators.
+        """
+        super().__init__(allows_two_qubit_reduction=mapper.allows_two_qubit_reduction)
+        self.mapper = mapper
+
+    def _map_single(self, second_q_op: SparseLabelOp) -> PauliSumOp:
+        blocked_op = self.mapper._map_single(second_q_op).primitive
```
woodsp-ibm, line 75: Note sure what happened to my comment - I know I wrote something like the following before!

Although this is intended for fermionic systems here its general enough that this order alteration could be done with any operator. Would would want do a size check (size of labels i,e number of qubits) to ensure the operator has an even size and if not raise an error. Otherwise I think the code will fail with some index error and maybe more obscure for a user to figure what went wrong.

mrossinek: I think you commented on the commit itself [here](https://github.com/mrossinek/qiskit-nature/commit/7a31a4ee1ac3daa342b688a74c28628e80dca396#r98322174). Anyways, I suppose we can add a check like that. Will do this next week :+1: 

Should we also be adding more checks like whether or not the `mapper` provided to `__init__` is a `FermionicMapper`? That should also deal with operators other than `FermionicOp` being used here.

woodsp-ibm: Since this is just a changing order in the operator it could be done for any system right. Not sure why one would use it outside of a fermionic system and even there BKSF does interactions. So while we could restrict it, if you think thats desirable, things won't fail. If you use this mapper you might need to account for the fact if you have some custom circuits you built you used - U(VCC) would get taken care of automatically, as you, note via the mapper use.

mrossinek: Indeed, this could be done for any kind, but the naming of this class specifically shows the intend of going from block-ordered fermionic spin orbitals to interleaved ordered ones. Thus, I think it is important to do this restriction because we do not _want_ to guarantee any behavior in the other cases.

To make this more clear, I updated the class to subclass `FermionicMapper` directly

woodsp-ibm: Do you want to check and deny its use with BKSF - while that might map an even number of orbitals to even number of qubits, since its mapping is done on interactions, not sites, that may not occur. One could check the qubit size resultant from the mapping, rather than the number or orbitals. That would mean an explicit check to BKSF was not needed - and reordering if the size was even would do no harm I think.

mrossinek: Yeh I had not thought of that. I had hoped to do the check before the mapping in order to avoid the unnecessary mapping computation if it is bound to fail.

I suppose we can either:
- check the operator length on the qubit space
- expose `.mapper` via a property getter/setter and disallow BKSF in there

What do you think?

woodsp-ibm: A third option would be to check the self.mapper type at the same time as the existing orbital check there. You could change the mapper property to potentially detect things earlier, but since BKSF scales by interactions its uncommon for it to ever be used. I think we should note either way this interleaved mapper does not support BKSF in the docstring somewhere.

mrossinek: The point is that this check is done _for every operator_ which seems like an annoying overhead to have to re-check the same mapper every time.

woodsp-ibm (approved):

PR #1050, opened 2023-02-03
Replace PauliSumOp with SparsePauliOp

manoelmarques:
### Summary
Fixes Qiskit/qiskit-nature#1020


mrossinek:
It will be easiest to deal with this *after* #1044 is merged

mrossinek (commented):
Some initial comments (I need to do a more thorough review later).
Also some observations:

- [x] we need to add a release note
- [ ] we should check for any more usage of `opflow` in the unittests and replace it wherever possible

qiskit_nature/second_q/circuit/library/initial_states/hartree_fock.py
```diff
@@ -254,10 +254,13 @@ def hartree_fock_bitstring_mapped(
     else:
         qubit_op = qubit_converter.map(bitstr_op)
 
+    if not isinstance(qubit_op, SparsePauliOp):
+        qubit_op = qubit_op.primitive
```
woodsp-ibm, line 258: This code I imagine could stay here even after PauliSumOp, as a possible return type from mapping, gets removed. We may even easliy forget to remove it as-is down the road. Would it be better though to invert the test

```
    if isinstance(qubit_op, PauliSumOp):
        qubit_op = qubit_op.primitive
```
since we are much more likely to search and remove PauliSumOp logic and indeed this code would fail on the import, like it would elsewhere, when PauliSumOp gets removed so we have to take care of this.

qiskit_nature/second_q/mappers/interleaved_qubit_mapper.py
```diff
@@ -17,9 +17,11 @@
...
 from qiskit.quantum_info import SparsePauliOp
 
+from qiskit_nature import settings
 from qiskit_nature.second_q.operators import FermionicOp
 
 from .fermionic_mapper import FermionicMapper
+from .qubit_mapper import QubitOperator
 
 
 class InterleavedQubitMapper(FermionicMapper):
```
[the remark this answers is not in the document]
woodsp-ibm: Even though its new, and does not have the compatibility constraints of the existing mappers, done this way all the mappers are consistent for an end user even if they happen to swap out some existing mapper for one of these in existing code it should still work. The extra logic handling the switch seems simple/common enough.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom_vibrational_ops_builder.py
```diff
@@ -36,7 +37,7 @@ def build_vibrational_ops(
     ],
     qubit_converter: QubitConverter | QubitMapper,
 ) -> Tuple[
-    Dict[str, PauliSumOp],
+    Dict[str, PauliSumOp | SparsePauliOp],
```
woodsp-ibm, line 40: I'll comment on this example but I could have picked on others. In the qubit mapper it defines, and throughout the mappers uses, a type QubitOperator which is the same Union. What is the thinking when PauliSumOp goes way - that gets reduced to just effectively a synonym for SparsePauliOp or would we replace it everywhere with SparsePauliOp so the type definition goes away. If the former maybe we should consider using QubitOperator in such places such as these.

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_solver.py
```diff
@@ -19,13 +19,14 @@
...
 from qiskit.quantum_info.operators.base_operator import BaseOperator
+from qiskit.quantum_info import SparsePauliOp
 
 from qiskit_nature.second_q.operators import SparseLabelOp
 from qiskit_nature.second_q.mappers import QubitConverter, QubitMapper
 from qiskit_nature.second_q.problems import BaseProblem
 from qiskit_nature.second_q.problems import EigenstateResult
 
-QubitOperator = Union[BaseOperator, PauliSumOp]
+QubitOperator = Union[BaseOperator, PauliSumOp, SparsePauliOp]
```
woodsp-ibm, line 29: I guess this QubitOperator type is the different than the one of the same name in QubitMapper. I had not seen this when I made the prior comment about that type and explicit Union of what it defines elsewhere

qiskit_nature/second_q/mappers/linear_mapper.py
```diff
@@ -22,16 +22,21 @@
 import numpy as np
 
 from qiskit.opflow import PauliSumOp
-from qiskit.quantum_info.operators import Pauli, SparsePauliOp
+from qiskit.quantum_info import SparsePauliOp
+from qiskit.quantum_info.operators import Pauli
```
woodsp-ibm, line 26: I'm curious over the split here of one being imported from quantum_info now directly and the other from operators. The import will work either way for both and if one prefers the shorter then

> from qiskit.quantum_info import Pauli, SparsePauliOp

Same it true elsewhere where Operator can be imported likewise. Its no biggie since either works, but since it got changed....

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_solver.py
mrossinek: Actually since `BaseOperator` is already in there, adding `SparsePauliOp` should not be necessary in the first place :thinking: 

But using this `QubitOperator` in the mappers module will not work. So I suppose the path forward will be that in the future:
- the `mappers` module replaces its `QubitOperator` union type with _only_ `SparsePauliOp`
- whereas the `algorithms` module will use _only_ `BaseOperator`

qiskit_nature/second_q/mappers/linear_mapper.py
mrossinek: Yeh I also wondered what triggered this. It might have been some automatic editor change or it was an artifact of a macro used to iterate over all files which needed changing. In any case, I agree that this can be reverted although it really should not matter :sweat_smile:

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom_vibrational_ops_builder.py
mrossinek: I suppose in the cases where we don't include `BaseOperator` we could also import the type alias from the `mappers` stack :+1: 

Note that this is coupled to the thread below (in case you are reading this first :sweat_smile:)

qiskit_nature/second_q/circuit/library/initial_states/hartree_fock.py
mrossinek: Yeh that is a good point :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/excited_states_eigensolver.py
```diff
@@ -63,8 +64,8 @@ def solver(self, solver: Union[Eigensolver, EigensolverFactory]) -> None:
     def get_qubit_operators(
```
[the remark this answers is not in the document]
woodsp-ibm: It would not harm to add a message to each place, right, if they are now using the older types that will replaced - in fact arguably we should since for a user it may end up being quite helpful as it should signal more exactly the place(s) they need to change their code. Maybe in conjunction with the setting they end up seeing more deprecation messages - presently there is nothing around setting right, which I seem commented on here elsewhere - but I think we should have them in places like this too.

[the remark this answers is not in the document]
mrossinek: Agreed :+1: So it would be good if as part of this PR we add deprecation warnings to both:
- the newly added setting (commented elsewhere already, indeed)
- all public API places where `PauliSumOp` is an input option

qiskit_nature/second_q/algorithms/ground_state_solvers/ground_state_solver.py
woodsp-ibm: We should be careful with a QubitOperator type. This ends up getting exposed to the end user in the API ref and if they mean different things in different places its likely to lead to confusion

mrossinek:
I have applied the following updates to this PR:
- removed the `QubitOperator` alias in favor of being explicit everywhere (in doing so I also removed the `BaseOperator` type hint which was inconsistently used and thus not really supported (and especially not tested; does this warrant mention in the release note?)
- I updated type hints to the "future-style" in files touched by this PR
- all usage of `PauliSumOp` is now explicitly findable by grepping for `PauliSumOp`. It is merely used for type hints and otherwise all "real logic" uses `SparsePauliOp` now
- deprecation messages for `PauliSumOp` usage are in place thanks to @manoelmarques :+1:

qiskit_nature/second_q/algorithms/excited_states_solvers/excited_states_eigensolver.py
```diff
@@ -85,6 +85,13 @@ def get_qubit_operators(
 
         if aux_operators is not None:
             for name_aux, aux_op in aux_operators.items():
+                if isinstance(aux_op, PauliSumOp):
+                    warn_deprecated(
+                        "0.6.0",
+                        old_type=DeprecatedType.ARGUMENT,
+                        old_name="PauliSumOp",
+                        new_name="SparsePauliOp",
+                    )
```
woodsp-ibm, line 94: This deprecation warning comes out as

>DeprecationWarning: The PauliSumOp argument is deprecated as of version 0.6.0 and will be removed no sooner than 3 months after the release. Instead use the SparsePauliOp argument.

Basically we are deprecating passing objects of PauliSumOp type as aux_operators and instead requiring them to be of SparsePauliOp type. The message is more making me look to an argument named PauliSumOp, which it doesn't have

If I replace `argument` by `type`, in the original message, (not that we have a DeprecatedType enum to do that at present) it comes closer (as below) but lacks aux_operators argument to correlate the warning

DeprecationWarning: The PauliSumOp type is deprecated as of version 0.6.0 and will be removed no sooner than 3 months after the release. Instead use the SparsePauliOp type.

releasenotes/notes/paulisumop-replace-d1ae54d1854b301b.yaml
```diff
@@ -0,0 +1,12 @@
+---
+deprecates:
+  - |
+    Replaced all public API occurrences of :class:`~qiskit.opflow.PauliSumOp` with :class:`~qiskit.quantum_info.SparsePauliOp`:
+
+    1. in cases where the type is a function *input*, support for ``SparsePauliOp`` objects was added
+       alongside the use of ``PauliSumOp`` objects. When providing the latter as input, a deprecation
+       warning is raised.
+    2. in cases where the type is an *output* the return type now dependson the new setting
```
woodsp-ibm, line 9:
```suggestion
    2. in cases where the type is an *output* the return type now depends on the new setting
```

woodsp-ibm (approved):

PR #1053, opened 2023-02-07
Install sparse on gettext tox

manoelmarques:
### Summary
No need to backport to stable because the fix has been applied there already


woodsp-ibm (approved):

mrossinek (approved):
Thanks!

PR #1057, opened 2023-02-10
fix: printing debye-dipole-moment

wltr-tlrc:
### Summary
I found a bug when printing the debye-dipole-moment. This fix has been discussed with @mrossinek.

### DIFF
--- a/qiskit_nature/second_q/problems/electronic_structure_result.py
+++ b/qiskit_nature/second_q/problems/electronic_structure_result.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2020, 2022.
+# (C) Copyright IBM 2020, 2023.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
--- a/qiskit_nature/second_q/problems/electronic_structure_result.py
+++ b/qiskit_nature/second_q/problems/electronic_structure_result.py
@@ -164,11 +164,7 @@ class ElectronicStructureResult(EigenstateResult):
         tdm = self.total_dipole_moment
         if tdm is None:
             return None
-        tdmd = []
-        for dip in tdm:
-            dipmd = dip / DEBYE
-            tdmd.append(dipmd)
-        return tdmd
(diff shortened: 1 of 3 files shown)


qiskit_nature/second_q/problems/electronic_structure_result.py
```diff
@@ -162,7 +162,13 @@ def total_dipole_moment(self) -> Optional[List[float]]:
     def total_dipole_moment_in_debye(self) -> Optional[List[float]]:
         """Returns total dipole of moment in Debye"""
         tdm = self.total_dipole_moment
-        return [dip / DEBYE if dip is not None else None for dip in tdm]
+        if tdm is None:
```
woodsp-ibm, line 165: From what I can see the change is to handle the condition where the dipole is not computed at all and ends up as None which would have failed before. When we do compute a dipole, if Z2 symmetry reduction is also being used, one or more of the X,Y and Z components may not commute with the symmetry of the main operator, against whose ground state we would compute the dipole moment, and in this case that aspect is set at None. The list comprehension that was there dealt with any of those individual parts being None, so None was placed into the result. I think the list comprehension above could just be preceded by the overall None check as below. 
```
tdm = self.total_dipole_moment
if tdm is None:
   return None
return [dip / DEBYE if dip is not None else None for dip in tdm]
```

mrossinek: Sure we can do that, too. I will add this as a suggestion to simplify applying the change

qiskit_nature/second_q/problems/electronic_structure_result.py
```diff
@@ -162,7 +162,13 @@ def total_dipole_moment(self) -> Optional[List[float]]:
...
+            return None
+        tdmd = []
+        for dip in tdm:
+            dipmd = dip / DEBYE
+            tdmd.append(dipmd)
+        return tdmd
```
mrossinek, line 171:
```suggestion
        return [dip / DEBYE if dip is not None else None for dip in tdm]
```

mrossinek (approved):
Thank you, @wltr-tlrc, for your first contribution to Qiskit! :partying_face:

woodsp-ibm (approved):
LGTM. @wltr-tlrc Thanks as well for taking the time to do a contribution for a fix!

PR #1062, opened 2023-02-15
Fix tutorial thumbnails

### LINKED ISSUE #1061: Tutorial thumbnails are not showing correctly
For a couple of the tutorials (ground and excited state solvers), as per the image below, which was captured from the current published docs, there is no thumbnail just what looks like a scrollbar with arrows at the end with the text thumbnail above it.

https://qiskit.org/documentation/nature/tutorials/index.html

image

woodsp-ibm:
### Summary
Fixes #1061

### Details and comments
Image name in conf.py (was h2...) needs to match case of the image file (H2...) as server is case sensitive. See more info in #1061


mrossinek (approved):
Thanks for finding this!

Issue #933, opened 2022-11-03
Improve `FCIDump` parser


mrossinek:
### What is the expected enhancement?

Our current implementation of the `FCIDump` parser runs into **serious** bottlenecks for larger systems. This is mostly due to the fact of how we are dealing with loading the permutational symmetries. Instead of tracking the index tuples which we encounter and later on populating the symmetric variations, we construct the full list of potential index tuples ahead of time, and then "cross off" the ones that we did indeed find. Considering that the index tuples scale as `N^4`, this is obviously horrendously expensive for large `N`.

This issue suggests that the implementation be revisited and improved as to remove said bottleneck :+1:

lucarighetti3:
I would be happy to work on this!

PR #1072, opened 2023-02-22
Adds symmetry-reduced two-body electronic integral utilities

mrossinek:
### Summary

This adds utility classes to automatically handle symmetry-reduced two-body electronic integrals.
This will enable calculations with lower memory footprints and will in the future also allow to solve #933 fairly elegantly.

~For now, this is a WIP which still needs to be battle-tested. The classes themselves are fully functional but the integration into the stack still poses some questions here and there. If we need to proceed with this PR faster, we may need to fallback to expanding to the fully expanded integrals when running into errors.~

### Details and comments

- [x] the actual symmetry classes:
  - [x] `S1Integrals`: the fully expanded case but stored in chemistry notation and useful for conversions
  - [x] `S4Integrals`: for 4-fold symmetric integrals
  - [x] `S8Integrals`: for 8-fold symmetric integrals (cannot be used for beta-alpha integrals in the general case)
- [x] integration with `ElectronicEnergy` in the MO basis (mostly working; more unittests needed)
- [x] integration with `transformers` module
  - [x] `BasisTransformer`: planned as part of this PR
  - [x] `ActiveSpaceTransformer`: not planned for now, needs to be caught
  - [x] `FreezeCoreTransformer`: not planned for now, needs to be caught
- [x] reno

### DIFF
--- a/qiskit_nature/second_q/operators/symmetric_two_body.py
+++ b/qiskit_nature/second_q/operators/symmetric_two_body.py
@@ -299,7 +299,7 @@ class S8Integrals(SymmetricTwoBodyIntegrals):
     This class is a utility subclass of the central :class:`.Tensor` used for storing n-dimensional
     arrays. This particular one holds 2-body electronic integrals when contracted to 8-fold
     symmetry. This means, that the full 4-dimensional integrals are contracted to a 1-dimensional
-    representation of length :math:`M * (M + 1) / 2` where math:`M` is the number of orbital pairs
+    representation of length :math:`M * (M + 1) / 2` where :math:`M` is the number of orbital pairs
     given by :math:`M = N * (N + 1) / 2` where :math:`N` is the number of orbitals.
     """
 
--- a/qiskit_nature/second_q/operators/symmetric_two_body.py
+++ b/qiskit_nature/second_q/operators/symmetric_two_body.py
@@ -1,0 +1,768 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2022, 2023.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+# pylint: disable=invalid-name
+
+"""
+Symmetric 2-body electronic integrals (:mod:`qiskit_nature.second_q.operators.symmetric_two_body`)
+==================================================================================================
+
+.. currentmodule:: qiskit_nature.second_q.operators.symmetric_two_body
+
+This module provides utilities to deal with symmetry-reduced 2-body electronic integrals.
+
+Container classes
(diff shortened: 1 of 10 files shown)


qiskit_nature/second_q/operators/symmetric_two_body.py
```diff
@@ -0,0 +1,751 @@
...
+
+   SymmetricTwoBodyIntegrals
+   S1Integrals
+   S4Integrals
+   S8Integrals
+
+Expansion methods
+-----------------
+
+These methods can be used to expand from lower to higher symmetries.
```
woodsp-ibm, line 39: I would have considered 8 fold symmetry higher than 4 fold - in that it has more symmetries. To me these are reducing the symmetries which expands the memory. Except for the first one which is taking a full matrix and creating an S1 from it.
I am not going to comment on the reduce since thats the opposite here, except to say the expand does not seem to be an inverse of reduce its more an any to S1 well not quite any just S4 and S8 (full to S1 is covered by expand.) I guess on reduce side we don't need the opposite of expand since any of these are perceived as a full array by and end user - and maybe some .array property does just that?

mrossinek: Yes I see what you mean. I liked `expand` since this clearly states the expansion from lower to higher memory requirements. I thought `reduce` would be a good mnemonic to the opposite operation.

But you are right that "lower symmetry" correlates with "higher memory consumption".

And yes, it is not exactly the inverse. Expand really always goes up to `S1` whereas reduce tries its best. There are standalone methods for each conversion which a user can use to their liking if they want to achieve a specific symmetry.

So to summarize:
- `expand`: any to S1
- `reduce`: any to as low memory consumption as possible (highest symmetry) [I think you mixed this up above]

Do you think I can rephrase the docstring to talk about memory consumption or would this be too confusing?
Or do you have some better naming suggestions?

mrossinek: I pushed a suggested rewording just now. Let me know what you think :+1:

qiskit_nature/second_q/operators/symmetric_two_body.py
```diff
@@ -0,0 +1,768 @@
...
+                        yield value, tuple(full_idx)
+
+
+class S8Integrals(SymmetricTwoBodyIntegrals):
+    """A container for 8-fold symmetric 2-body electronic integrals in chemist ordering.
+
+    This class is a utility subclass of the central :class:`.Tensor` used for storing n-dimensional
+    arrays. This particular one holds 2-body electronic integrals when contracted to 8-fold
+    symmetry. This means, that the full 4-dimensional integrals are contracted to a 1-dimensional
+    representation of length :math:`M * (M + 1) / 2` where math:`M` is the number of orbital pairs
```
woodsp-ibm, line 302:
```suggestion
    representation of length :math:`M * (M + 1) / 2` where :math:`M` is the number of orbital pairs
```

woodsp-ibm (approved):

qiskit_nature/second_q/operators/symmetric_two_body.py
mrossinek: Thanks for spotting this! :+1:

PR #1073, opened 2023-02-22
`QubitConverter` deprecation

mrossinek:
### Summary

This PR resolves the remaining open points revolving around #967.

### Details and comments

I am already opening this now but I plan to tackle the renaming of `qubit_converter -> qubit_mapper` in this PR, too.


qiskit_nature/second_q/problems/base_problem.py
```diff
@@ -77,6 +77,15 @@ def second_q_ops(self) -> tuple[SparseLabelOp, dict[str, SparseLabelOp]]:
 
         return main_op, aux_ops
 
+    @deprecate_function(
+        "0.6.0",
+        additional_msg=(
+            ". This function is deprecated because it has been removed from the public API. It is "
```
woodsp-ibm, line 83: > This function is deprecated because it has been removed from the public API.

We can't have already removed it right - its deprecated as it will be removed.

mrossinek: Yeh I phrased that poorly.. It is not actually removed yet but will be in the future :+1:

qiskit_nature/second_q/mappers/qubit_converter.py
```diff
@@ -26,6 +26,7 @@
 from qiskit.opflow.primitive_ops import Z2Symmetries
 
 from qiskit_nature import QiskitNatureError, settings
+from qiskit_nature.deprecation import warn_deprecated, DeprecatedType
```
woodsp-ibm, line 29: As this whole class is deprecated can we note it in the docs i.e this line below (I cannot comment there - had to pick a changed line)

>"""A converter from Second-Quantized to Qubit Operators.

add DEPRECATED so it shows up in the API ref. Otherwise there is no indication there at all its deprecated.

> """DEPRECATED A converter from Second-Quantized to Qubit Operators.

PR #1074, opened 2023-02-22
Fix prior release note formatting

woodsp-ibm:
### Summary
Fixes formatting of a prior release note. Adds a space so `:class:` is parsed correctly


mrossinek (approved):
Thank you! :+1:

PR #1077, opened 2023-02-24
Fix: generate fermionic excitations

### LINKED ISSUE #1071: ```generate_fermionic_excitations``` produces incorrect output with ```perserve_spin=False``` and unequal alpha and beta spins.
### Environment

- **Qiskit Nature version**: 0.5.0
- **Python version**: 3.8.15
- **Operating system**: Debian 9.13

### What is happening?

The standard ordering for N spatial orbitals should be that the first N orbitals correspond to spin up and the second N orbitals correspond to spin down. When ```perserve_spin=False``` is used and an unequal number of alpha and beta spins is specified via ```num_particles```, ```generate_fermionic_excitations``` produces incorrect output/output that is inconsistent with the standard ordering of the spin orbitals.

```python
>>> from qiskit_nature.second_q.circuit.library.ansatzes.utils import generate_fermionic_excitations

>>> generate_fermionic_excitations(num_excitations=1, num_spatial_orbitals=3, num_particles=(2,0), preserve_spin=False)
[((0,), (1,)), ((0,), (2,)), ((0,), (4,)), ((0,), (5,)), ((3,), (1,)), ((3,), (2,)), ((3,), (4,)), ((3,), (5,))]

```

### What should happen?

Instead of generating excitations from the "0" and "3" orbitals, it should instead generate excitations from the "0" and "1" orbitals. Cf. output when ```preserve_spin=True```:

```python
>>> from qiskit_nature.second_q.circuit.library.ansatzes.utils import generate_fermionic_excitations

>>> generate_fermionic_excitations(num_excitations=1, num_spatial_orbitals=3, num_particles=(2,0), preserve_spin=True)
[((0,), (2,)), ((1,), (2,))]
```

```preserve_spin=False``` should generate output with the same occupied orbitals "0" and "1" but excitations to more unoccupied orbitals.

### Any suggestions?

This seems to be an issue with how ```generate_fermionic_excitations``` handles ```preserve_spin=False``` https://github.com/Qiskit/qiskit-nature/blob/e8befeb518f53069a5a784a22968c1b3e1abd09b/qiskit_nature/second_q/circuit/library/ansatzes/utils/fermionic_excitation_generator.py#L198-L227
This interleaving business only works if the number of alpha and beta spins specified in ```num_particles``` are equal. This code treats ```num_particles=(2,0)``` and ```num_particles=(1,1)``` exactly the same as it only involves ```sum(num_particles)```

### PR DESCRIPTION
**Pull request by LSticher**
### Summary

Fixes #1071. 

### Details and comments

I added some tests to ```test_fermionic_excitation_generator``` that check behaviour with uneven alpha and beta spins, since that case didn't really seem to get much coverage in the tests (including tests to catch the bug in #1071).

### CODE REVIEW

File: releasenotes/notes/fix-generate-fermionic-excitations-309b771ceb14761f.yaml [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+fixes:
+  - |
+    Fixes output of :func:`generate_fermionic_excitations` when called with ``perserve_spin=False``
```
mrossinek, line 4:
```suggestion
    Fixes output of :func:`.generate_fermionic_excitations` when called with ``perserve_spin=False``
```

mrossinek (commented):
Thanks a lot, this looks like a great PR! :tada: 

Besides the one formatting update I suggest below (this ensures the link is properly clickable), you also need to update the copyright header in some files (hence the CI still fails).

We actually have a small script to automatize this for us, so you can simply run `make copyright` from a UNIX shell in the root of your cloned repo :+1: 

Other than that, this LGTM :slightly_smiling_face:

mrossinek (approved):
Thank you for your contribution to Qiskit Nature! :tada:

Issue #1036, opened 2023-01-19
Moving away from targeting black box design


mrossinek:
## TL;DR

I propose that we:
- phase out the `(Minimum)EigensolverFactory` classes (with their replacement being better documentation/education of users)
- no longer enforce the re-usability of algorithm instances (and their subcomponents)

<hr/>

Qiskit Nature provides the glue between quantum computational concepts and classical computational science. As such, there is quite a large variety in the background and knowledge of our users. This does not always make code design decisions easy and in the past we have added a lot of tools which wrap other components in order to achieve a workflow that goes from a `problem` places it into an `algorithm` to obtain a `result`. While this overall workflow is good and is also reflected in other Qiskit application modules, we have added a few components that are coupled too tightly making more complex workflows not always easy to realize. Some examples are reflected by:
- #701 
- #967 
- https://github.com/Qiskit/qiskit-nature/issues/854#issuecomment-1324745418

This issue gathers some general thoughts on this topic.

1. We should avoid tight wrapping of components (e.g. `QubitConverter` wrapping `QubitMapper` instances) and requiring this wrapping throughout the stack.

This is a general observation which @woodsp-ibm has also brought up in the past (mostly offline) but I would like to re-iterate on this and stress this as an important point. Resolving #967 will largely deal with this, but I am explicitly writing it down here for documentation purposes.

2. We should move away from providing black box components.

Quantum computing is simply not in a state yet, where "it just works", especially when looking at chemistry applications. More concretely, what I am referring to here is the `VQEUCCFactory` (and its siblings). The design of this concept is not scalable to the various options of algorithm and ansatz variants (see e.g. #854). Furthermore, the default provided by that class is only useful for classical simulation purposes where UCCSD is feasible for smaller systems. For now, a real quantum device will not be able to handle a full UCCSD approach.
Thus, the class itself does not provide a real use-case which is aligned with Qiskit focusing on real device applications. Furthermore, the usefulness of the class was largely decreased when we switched to primitives, because compared to the old workflow in which a single import was sufficient, one now needs to import the ansatz (e.g. `UCCSD`), the optimizer (e.g. `SLSQP`) and the primitive (e.g. `Estimator`) to plug them into the factory, at which point most of the work is already done.
And finally, since problems are now fully defined _before_ being passed to `.solve(...)`, the argument of building out the ansatz at runtime is also no longer relevant.

[...]

PR #1084, opened 2023-03-02
Deprecate the `(Minimum)EigensolverFactory` classes

### LINKED ISSUE #854: Add an `AdaptVQEUCCFactory`
### What should we add?

With the `AdaptVQE` algorithm being migrated to Terra (https://github.com/Qiskit/qiskit-terra/pull/7930), we should add a factory class to Nature which can leverage this algorithm in combination with the at-runtime constructed UCC ansatz. The same could be done for the UVCC case.

mrossinek:
### Summary

This PR deprecates the various `(Minimum)EigensolverFactory` classes.
In their place we will have detailed How-To guides as well as generally improved documentation around the intricacies which the Factory classes used to hide from the user.

### Details and comments

- [x] add How-To guides
- [x] deprecate Factory classes
- [x] update all docstrings to:
    - [x] *not* refer to the `Factory` classes
    - [x] detail intricacies which these classes used to handle (and point to the new How-To guides)
- [x] add a release note

Closes #854, #1036

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,15 +85,16 @@ estimator = Estimator()
 # setup the ansatz for VQE
 from qiskit_nature.second_q.circuit.library import HartreeFock, UCCSD
 
-ansatz = UCCSD()
-ansatz.num_particles = problem.num_particles
-ansatz.num_spatial_orbitals = problem.num_spatial_orbitals
-ansatz.qubit_mapper = mapper
-initial_state = HartreeFock()
-initial_state.num_particles = problem.num_particles
-initial_state.num_spatial_orbitals = problem.num_spatial_orbitals
-initial_state.qubit_mapper = mapper
-ansatz.initial_state = initial_state
+ansatz = UCCSD(
+    problem.num_spatial_orbitals,
+    problem.num_particles,
+    mapper,
+    initial_state=HartreeFock(
+        problem.num_spatial_orbitals,
+        problem.num_particles,
+        mapper,
+    ),
+)
 
 # set up our actual VQE instance
 from qiskit.algorithms.minimum_eigensolvers import VQE
--- a/README.md
+++ b/README.md
@@ -99,6 +99,9 @@ ansatz.initial_state = initial_state
 from qiskit.algorithms.minimum_eigensolvers import VQE
 
 vqe = VQE(estimator, ansatz, optimizer)
+# ensure that the optimizer starts in the all-zero state which corresponds to
+# the Hartree-Fock starting point
+vqe.initial_point = [0] * ansatz.num_parameters
 
 # prepare the ground-state solver and run it
 from qiskit_nature.second_q.algorithms import GroundStateEigensolver
@@ -113,18 +116,17 @@ H<sub>2</sub>, where the two atoms are configured to be at a distance of 0.735 a
 input specification is processed by the PySCF driver. This driver produces an `ElectronicStructureProblem`
 which gathers all the problem information required by Qiskit Nature.
 The second-quantized operators contained in that problem can be mapped to qubit operators with a
-`QubitConverter`. Here, we chose the parity mapping in combination with a 2-qubit reduction, which
-is a precision-preserving optimization removing two qubits; a reduction in complexity that is particularly
+`QubitMapper`. Here, we chose the `ParityMapper` which automatically removes 2 qubits due to inherit
+symmetries when the `num_particles` are provided to it; a reduction in complexity that is particularly
 advantageous for NISQ computers.
 
 For actually finding the ground state solution, the Variational Quantum Eigensolver (VQE) algorithm is used.
-Its main three components, the estimator primitive, wavefunciton ansatz (`UCCSD`), and optimizer, are passed
-to the `VQEUCCFactory`, a utility of Qiskit Nature simplifying the setup of the `VQE` algorithm and its
-components. This factory also ensures consistent settings for the ansatzes initial state and the optimizers
-initial point.
+Its main three components are the estimator primitive, wavefunciton ansatz (`UCCSD`), and optimizer.
+The `UCCSD` component is the only one provided directly by Qiskit Nature and it is usually paired with the
+`HartreeFock` initial state and an all-zero initial point for the optimizer.
 
-The entire problem is then solved using a `GroundStateEigensolver` which wraps both, the `QubitConverter`
-and `VQEUCCFactory`. Since an `ElectronicStructureProblem` is provided to it (which was the output of the
+The entire problem is then solved using a `GroundStateEigensolver` which wraps both, the `ParityMapper`
+and `VQE`. Since an `ElectronicStructureProblem` is provided to it (which was the output of the
 `PySCFDriver`) it also returns an `ElectronicStructureResult`.
 
 ### Further examples
--- a/README.md
+++ b/README.md
@@ -67,12 +67,10 @@ driver = PySCFDriver(
 )
 problem = driver.run()
 
-# setup the mapper and qubit converter
+# setup the qubit mapper
 from qiskit_nature.second_q.mappers import ParityMapper
-from qiskit_nature.second_q.mappers import QubitConverter
 
-mapper = ParityMapper()
-converter = QubitConverter(mapper=mapper, two_qubit_reduction=True)
+mapper = ParityMapper(num_particles=problem.num_particles)
 
 # setup the classical optimizer for the VQE
 from qiskit.algorithms.optimizers import L_BFGS_B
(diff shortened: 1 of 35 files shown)


README.md
```diff
@@ -85,19 +83,30 @@ from qiskit.primitives import Estimator
...
-vqe_factory = VQEUCCFactory(estimator, ansatz, optimizer)
+ansatz.num_particles = problem.num_particles
+ansatz.num_spatial_orbitals = problem.num_spatial_orbitals
+ansatz.qubit_mapper = mapper
+initial_state = HartreeFock()
+initial_state.num_particles = problem.num_particles
+initial_state.num_spatial_orbitals = problem.num_spatial_orbitals
+initial_state.qubit_mapper = mapper
+ansatz.initial_state = initial_state
+
```
woodsp-ibm, line 97:
```
num_particles = problem.num_particles
num_spatial_orbitals = problem.num_spatial_orbitals

hf_state = HartreeFock(num_spatial_orbitals, num_particles, mapper)
ansatz = UCCSD(num_spatial_orbitals, num_particles, mapper, initial_state=hf_state)
```
Could we do the HF and UCCSD via their constructors. Since we know all the information from the problem, while we can use the setters most code I see from users does it with the constructor.

docs/howtos/adapt_vqe.rst
```diff
@@ -8,48 +8,62 @@ This tutorial outlines how the algorithm can be used.
 
 1. We obtain an ``ElectronicStructureProblem`` which we want to solve:
```
woodsp-ibm, line 9: This is just a comment - we can maybe improve them later (or now) - we should be able to have these all as class refs so they are clickable links not just highlighted. I see in places in the new howtos you do have class refs. Perhaps they should be used more generally, although some may be doable eg QubitMapper I think most should, and with intersphinx we can do it for VQE etc too right.

docs/howtos/adapt_vqe.rst
```diff
@@ -8,48 +8,62 @@ This tutorial outlines how the algorithm can be used.
...
->>> ansatz.qubit_converter = converter
->>> initial_state = HartreeFock()
->>> initial_state.num_particles = problem.num_particles
->>> initial_state.num_spatial_orbitals = problem.num_spatial_orbitals
->>> initial_state.qubit_converter = converter
->>> ansatz.initial_state = initial_state
+.. testcode::
+
+    from qiskit_nature.second_q.circuit.library import UCCSD, HartreeFock
+    ansatz = UCCSD()
```
woodsp-ibm, line 29: Similar comment to that of the readme in terms of HF+UCCSD creation

docs/howtos/numpy_eigensolver.rst
```diff
@@ -0,0 +1,51 @@
...
+.. testcode::
+
+    from qiskit_nature.second_q.algorithms import ExcitedStatesEigensolver
+    solver = ExcitedStatesEigensolver(mapper, algo)
+
+5. We solve the problem:
+
+.. testcode::
+
+    result = solver.solve(problem)
```
woodsp-ibm, line 51: I am thinking for code like this, which should run fast, can we not do testoutput here so we can keep an eye on the howto that it continues to work as expected. I thought about commenting on the AdaptVQE one but am not sure how long that takes to run. If its relatively quick though...

qiskit_nature/second_q/algorithms/excited_states_solvers/eigensolver_factories/numpy_eigensolver_factory.py
```diff
@@ -18,12 +18,19 @@
...
 from .eigensolver_factory import EigensolverFactory
 
 
 class NumPyEigensolverFactory(EigensolverFactory):
-    """A factory to construct a NumPyEigensolver."""
+    """DEPRECATED A factory to construct a NumPyEigensolver.
+
+    .. warning::
+
+        This class is deprecated! Please see :ref:`this guide <how-to-numpy>` for how to replace
```
woodsp-ibm, line 31: I was wondering about migration guide that shows the existing factory way and the new way for an identical problem. It could point to the howto for more explanation for the new way of course.

woodsp-ibm:
The tutorials, that use a Factory class, need revising too right.

qiskit_nature/second_q/algorithms/excited_states_solvers/eigensolver_factories/numpy_eigensolver_factory.py
mrossinek: I will write the migration guide for the 0.5 to 0.6 transition in a separate PR. That will also need to do some minor restructuring of the migration guides page in order to separate the 0.4->0.5 transition guides.

docs/howtos/adapt_vqe.rst
mrossinek: Sorry I overlooked it in the existing how-to guide and only updated the newly added ones. Will do this rn

docs/howtos/numpy_eigensolver.rst
mrossinek: Alright I added assertions against the total energies up to 4 decimal places. Two comments:
1. I cannot test the UVCC case because I do not actually have a valid problem loaded to test there
2. I chose to actually include the output in the rendering because otherwise I think it can look confusing with the `print` statement included but no output actually shown

docs/howtos/numpy_eigensolver.rst
```diff
@@ -0,0 +1,58 @@
+.. _how-to-numpy:
+
+How to use the ``NumPyEigensolver``
+===================================
+
+In order to ensure a physically meaningful excited states of a hamiltonian are found when using the
+:class:`~qiskit.algorithms.minimum_eigensolvers.NumPyEigensolver` one needs to set the
```
woodsp-ibm, line 7:
```suggestion
:class:`~qiskit.algorithms.eigensolvers.NumPyEigensolver` one needs to set the
```

docs/howtos/numpy_eigensolver.rst
```diff
@@ -0,0 +1,58 @@
...
+:attr:`~qiskit.algorithms.minimum_eigensolvers.NumPyEigensolver.filter_criterion` attribute
```
woodsp-ibm, line 8:
```suggestion
:attr:`~qiskit.algorithms.eigensolvers.NumPyEigensolver.filter_criterion` attribute
```

docs/howtos/numpy_eigensolver.rst
```diff
@@ -0,0 +1,58 @@
...
+    from qiskit_nature.second_q.algorithms import ExcitedStatesEigensolver
+    solver = ExcitedStatesEigensolver(mapper, algo)
+
+5. We solve the problem:
+
+.. testcode::
+
+    result = solver.solve(problem)
+
+    print(f"Total ground state energy = {result.total_energies[0]:.4f}")
```
woodsp-ibm, line 54: I think it might have been nice to have shown the excited state values here too since this is about excited states.

docs/howtos/vqe_ucc.rst
```diff
@@ -0,0 +1,102 @@
+.. _how-to-vqe-ucc:
+
+How to use a UCC-like ansatz with a ``VQE``
+===========================================
+
+When using a :class:`~qiskit_nature.second_q.circuit.library.UCC`-style ansatz with a
+:class:`~qiskit.algorithms.minimum_eigensolvers.VQE` one needs to pay particular attention to the
+:attr:`~qiskit.algorithms.minimum_eigensolvers.VQE.initial_point` attribute which indicates from
+which set of initial parameters the optimization routine should start.
+By default, the algorithm will start from a *random* initial point. In this how to we show how one
```
woodsp-ibm, line 10:
```suggestion
By default, VQE will start from a *random* initial point. In this how to we show how one
```
I am wondering if saying "the algorithm" is a bit vague here, given this also uses algorithms from nature - maybe explictly stating  "VQE" is clearer.

docs/howtos/vqe_uvcc.rst
```diff
@@ -0,0 +1,74 @@
+.. _how-to-vqe-uvcc:
+
+How to use a UVCC-like ansatz with a ``VQE``
+============================================
+
+When using a :class:`~qiskit_nature.second_q.circuit.library.UVCC`-style ansatz with a
+:class:`~qiskit.algorithms.minimum_eigensolvers.VQE` one needs to pay particular attention to the
+:attr:`~qiskit.algorithms.minimum_eigensolvers.VQE.initial_point` attribute which indicates from
+which set of initial parameters the optimization routine should start.
+By default, the algorithm will start from a *random* initial point. In this how to we show how one
```
woodsp-ibm, line 10:
```suggestion
By default, VQE will start from a *random* initial point. In this how to we show how one
```
Same change suggestion as above

docs/howtos/vqe_uvcc.rst
```diff
@@ -0,0 +1,74 @@
...
+
+.. testcode::
+
+    from qiskit_nature.second_q.algorithms.initial_points import VSCFInitialPoint
+    initial_point = VSCFInitialPoint()
+    initial_point.ansatz = ansatz
+    initial_point.problem = problem
+    vqe.initial_point = initial_point.to_numpy_array()
+
+Just like in the UCC-ansatz case, this is mostly useful when building more code on top of the
```
woodsp-ibm, line 73: As this `UCC-ansatz case` references the other howto could we make it link so its easy to jump there

woodsp-ibm (approved):

Issue #795, opened 2022-08-18
Implement a `MixedOperator`


mrossinek:
### What should we add?

We would like to add support for more complex models which couple various types of particles (e.g. fermions with bosons).
Among other things, this finds application in lattice gauge theory which is a planned addition to the lattice framework (see e.g. #556 which is the simplest LGT case).

A rough list of requirements which are needed for this to work are:
- a `MixedOperator` class which can hold multiple kinds of operators (i.e. `FermionicOp`, `SpinOp`, etc.)
  - the above class should allow storage of interaction coefficients between different operator kinds (in addition to the operator-internal coefficients ofcourse)
- an extension/addition of the `QubitConverter` to handle a `MixedOperator`
  - this requires multiple `QubitMapper` instances for all operator types contained in the mixed op
  - this should handle delegation of the various qubit registers: e.g. the fermionic operators get mapped to qubits `q1 .. qN` and the spin operators get mapped to qubits `q(N+1) ... q(N+M)`
  - purely fermionic operators need to be tensored with all-identity operators on the qubit-registers assigned to the spin-subsystem
  - purely spin operators are tensored with all-identity terms on the fermionic-subsystem (etc.)
  - mixed operators are mapped individually and then tensored together in the qubit-space taking into account the potential mixing coefficients stored in the `MixedOperator`

> :warning: Disclaimer: this is mostly an assortment of some thoughts. A first step for tackling this issue would be to come with a design on how to achieve this, given the rough outline of requirements presented above. I am happy to answer questions with more details :+1:

Anthony-Gandon:
I would like to participate implementing this MixedOp

PR #1085, opened 2023-03-03
Bosonic operators

ftroisi:


qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,619 @@
...
+
+class BosonicOp(SparseLabelOp):
+    r"""N-mode Bosonic operator.
+
+    A ``BosonicOp`` represents a weighted sum of bosonic creation/annihilation operator terms.
+    These terms are encoded as sparse labels, which are strings consisting of a space-separated list of
+    expressions. Each expression must look like :code:`[+-]_<index>`, where the :code:`<index>` is a
+    non-negative integer representing the index of the bosonic mode where the ``+`` (creation) or
+    ``-`` (annihilation) operation is to be performed. The value of :code:`index` is bound by the
+    number of spin orbitals (``num_spin_orbitals``) of the operator (Note: since Python indices are
```
mrossinek, line 41: We need to replace all references to `number of spin orbitals` to `number of modes` (or yet a more descriptive name; I am open to suggestions).

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,619 @@
...
+            operators will result in a new operator with the maximum number of spin orbitals of any
+            of the involved operators.
+
+    .. note::
+
+        A BosonicOp can contain :class:`qiskit.circuit.ParameterExpression` objects as coefficients.
+        However, a BosonicOp containing parameters does not support the following methods:
+
+        - ``is_hermitian``
+        - ``to_matrix``
```
mrossinek, line 159:
```suggestion
```

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,619 @@
...
+        sparse_mat = csc_matrix(
+            (csc_data, (csc_row, csc_col)),
+            shape=(dimension, dimension),
+            dtype=complex,
+        )
+
+        if sparse:
+            return sparse_mat
+        else:
+            return sparse_mat.toarray()
```
mrossinek, line 409:
```suggestion
```
This looks like left-over code from the `to_matrix` method.

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,619 @@
...
+            The commutation relation between two bosonic operator is:
+            [b_i, b_j^\dagger]_- = \delta_ij => b_i * b_j^\dagger = b_j^\dagger * b_i + \delta_ij
+
+        .. note::
+
+            This method implements the transformation of an operator to the normal ordered operator.
+            The transformation is calculated by considering all commutation relations between the
+            operators.
+            For example, for the case :math:`\\colon c_0 c_0^\\dagger\\colon` where :math:`c_0`
+            is an annihilation operator, this method returns :math:`1 - c_0^\\dagger c_0` due to
```
mrossinek, line 440:
```suggestion
            is an annihilation operator, this method returns :math:`1 + c_0^\\dagger c_0` due to
```

mrossinek (commented):
A few initial observations. We will also need to add a release note :+1:

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+    def _map_single(
+        self, second_q_op: BosonicOp, *, register_length: int | None = None
+    ) -> SparsePauliOp | PauliSumOp:
+        # If register_length was passed, override the one present in the BosonicOp
+        if register_length is None:
+            register_length = second_q_op.num_modes
+
+        qubit_register_length = register_length * (self.truncation + 1)
+        # Create a Pauli operator, which we will fill in this method
+        # TODO: Check if we need to normal/index_order
```
mrossinek, line 59: I do not think we should do either of those. The math works correctly regardless of the term
ordering, so should the mapping.

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+        for terms, coeff in second_q_op.terms():
+            # Then loop over each term (terms -> List[Tuple[string, int]])
+            bos_op_to_pauli_op = SparsePauliOp(["I" * qubit_register_length], coeffs=[1.])
+            for op, idx in terms:
+                if op not in ('+', '-'):
+                    break
+                pauli_expansion: list[SparsePauliOp] = []
+                # Now we are dealing with a single bosonic operator. We have to perform the linear mapper
+                for n_k in range(self.truncation):
+                    prefactor = np.sqrt(n_k + 1) / 4
```
mrossinek, line 71:
```suggestion
                    prefactor = np.sqrt(n_k + 1) / 4.0
```
I like to be explicit about float division

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+                    # Define the actual index in the qubit register. It is given n_k plus the shift due
```
mrossinek, line 72:
```suggestion
                    # Define the actual index in the qubit register. It is given by n_k plus the shift due
```

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+                # Add the Pauli expansion for a single n_k to map of the bosonic operator
+                bos_op_to_pauli_op = reduce(operator.add, pauli_expansion).compose(bos_op_to_pauli_op)
+            # Add the map of the single boson op (e.g. +_0) to the map of the full bosonic operator
+            pauli_op.append(coeff * reduce(operator.add, bos_op_to_pauli_op.simplify()))
+
+        # return the lookup table for the transformed XYZI operators
+        bos_op_encoding = reduce(operator.add, pauli_op)
+        return bos_op_encoding
+
+    def get_ij_pauli_matrix(self, register_index: int, register_length: int):
```
mrossinek, line 93: Can we make this a private utility instead of having it be part of the public API?
And just like with `QubitMapper.pauli_table`, I think this method might benefit from an `lru_cache`
decorator applied to it. I also don't think that it needs to be an instance method and can instead
become a `classmethod`.

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+        """This method builds the Qiskit Pauli operators of the operators XX, YY, XY and YX"""
```
mrossinek, line 94: Even though I want it private, I think having a proper docstring will be useful down the road.
I.e. please add `Args:` and `Returns:` sections.

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,54 @@
...
+        defined in this class.
+        Then, the user calls BosonicLinearMapper.map(). Since map is not defined in BosonicLinearMapper,
+        the call is actually to the map method in this class.
+        QubitMapper.map() calls QubitMapper.map() thank to the super() class.
+        Inside QubitMapper.map() there is a call to _map_single. This method is present in QubitMapper
+        class, but it is overloaded in BosonicLinearMapper.
+        Thus, the call is to BosonicLinearMapper._map_single()
+    """
+
+    def __init__(self, truncation: int) -> None:
```
mrossinek, line 41: This method is missing a docstring.

Also: the public attribute `truncation` needs to be documented in an `Attributes:` section of the
class-level docstring. A good example for what that looks like is
[here](https://qiskit.org/ecosystem/nature/stubs/qiskit_nature.second_q.problems.ElectronicStructureProblem.html):

> The following attributes can be read and updated once the ElectronicStructureProblem object has been constructed.

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+                    else:
+                        tmp_op += + 1j*SparsePauliOp(x_y) - 1j*SparsePauliOp(y_x)
+                    pauli_expansion.append(prefactor * tmp_op)
+                # Add the Pauli expansion for a single n_k to map of the bosonic operator
+                bos_op_to_pauli_op = reduce(operator.add, pauli_expansion).compose(bos_op_to_pauli_op)
+            # Add the map of the single boson op (e.g. +_0) to the map of the full bosonic operator
+            pauli_op.append(coeff * reduce(operator.add, bos_op_to_pauli_op.simplify()))
+
+        # return the lookup table for the transformed XYZI operators
+        bos_op_encoding = reduce(operator.add, pauli_op)
```
mrossinek, line 90: I wonder whether we also need to support `qiskit_nature.settings.use_pauli_sum_op` in this new
mapper 🤔 I will get back to that..

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,560 @@
...
+    In principle, you can also add `BosonicOp` and integers, but the only valid case is the
+    addition of `0 + BosonicOp`. This makes the `sum` operation from the example above possible
+    and it is useful in the following scenario:
+
+    .. code-block:: python
+
+        boson = 0
+        for i in some_iterable:
+            # some processing
+            boson += BosonicOp(somedata)
```
mrossinek, line 138:
```suggestion
```
Since starting out from the `FermionicOp` to make this class, I updated that class' docstring and
removed this section. See #1112 for more details.

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,560 @@
...
+    @classmethod
+    def from_polynomial_tensor(cls, tensor: PolynomialTensor) -> BosonicOp:
+        cls._validate_polynomial_tensor_key(tensor.keys())
+
+        data: dict[str, _TCoeff] = {}
+
+        for key in tensor:
+            if key == "":
+                # TODO: deal with complexity
+                data[""] = cast(float, tensor[key])
```
mrossinek, line 255:
```suggestion
                data[""] = tensor[key]
```
See also #1144

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,560 @@
...
+
+    def normal_order(self) -> BosonicOp:
+        """Convert to the equivalent operator in normal order.
+
+        The normal order for bosons is defined
+        [here](https://en.wikipedia.org/wiki/Normal_order#Bosons).
+
+        Returns a new operator (the original operator is not modified).
+
+        .. reminder:
```
mrossinek, line 360: This is not a valid admonition and, thus, this block is actually not rendered in the docs. You can
find a list of valid admonitions
[here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#directives).

I suggest simply using `note` again.

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,560 @@
...
+
+            The commutation relation between two bosonic operator is:
+            [b_i, b_j^\\dagger]_- = \\delta_ij => b_i * b_j^\\dagger = b_j^\\dagger * b_i + \\delta_ij
```
mrossinek, line 363: This will also need to be rendered using `.. math::` otherwise it is not very legible.

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,560 @@
...
+            ValueError: Operator contains parameters.
+        """
+        if self.is_parameterized():
+            raise ValueError("is_hermitian is not supported for operators containing parameters.")
+        atol = self.atol if atol is None else atol
+        diff = (self - self.adjoint()).simplify(atol=atol)
+        return all(np.isclose(coeff, 0.0, atol=atol) for coeff in diff.values())
+
+    def simplify(self, atol: float | None = None) -> BosonicOp:
+        # TODO: currently simplify() DOES NOT commute with index_order()
```
mrossinek, line 496: Ah, I recall that we had this issue. I will need to dig a bit deeper into this. I would be willing
to leave this as is (for now) if we add a `.. warning::` to the docstrings of both, `simplify` and
`index_order`, pointing this out.

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,560 @@
...
+        atol = self.atol if atol is None else atol
+
+        # First, we normal-order the operator
+        ordered_op = self.normal_order()
+
+        data = defaultdict(complex)  # type: dict[str, _TCoeff]
+        # TODO: use parallel_map to make this more efficient (?)
```
mrossinek, line 503: FYI for another reviewer: this TODO is copied over from `FermionicOp` and out-of-scope for this PR

test/second_q/mappers/test_bosonic_linear_mapper.py
```diff
@@ -0,0 +1,177 @@
...
+        finally:
+            settings.use_pauli_sum_op = aux
+
+    @data(
+        (bos_op1, ref_qubit_op1_tr2),
+        (bos_op2, ref_qubit_op2_tr2),
+        (bos_op3, ref_qubit_op3_tr2),
+        (bos_op4, ref_qubit_op4_tr2),
+        (bos_op5, ref_qubit_op5_tr2),
+        # (bos_op6, ref_qubit_op6_tr2),
```
mrossinek, line 156: Why is this test case disabled?

test/second_q/mappers/test_bosonic_linear_mapper.py
```diff
@@ -0,0 +1,177 @@
+# This code is part of Qiskit.
+#
+# (C) Copyright IBM 2021, 2023.
```
mrossinek, line 3:
```suggestion
# (C) Copyright IBM 2023.
```

mrossinek (commented):
Overall I think this PR is already in pretty good shape. I left quite a few comments, some of which
are nitpicks, others are a bit more broad. In general, I'd like to add these notes:

- [ ] we need to add a release note (see [here](https://github.com/Qiskit/qiskit-nature/blob/main/CONTRIBUTING.md#adding-a-new-release-note))
- [ ] you need to run the code formatter: `make black` (from the repo root)
- [ ] you need to add `bosonicop` as a spell exception to `.pylintdict`
- [ ] the docstrings are mostly not formatted properly. This can be quite tricky when just getting
      started so I suggest this in our offline-scheduled meeting. I left some pointers in my
      comments below already though.

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,560 @@
...
+                if occs[-1] + 4 < len(new_label):
+                    new_label = (new_label[0:occs[-1]] + new_label[(occs[-1] + 4):]).strip()
+                else:
+                    new_label = (new_label[0:occs[-1]]).strip()
+            # Then, update the operator_mapper dictionary
+            # Creation operator (+) creates a new particle at that index, so we increase the value in
+            # the dict by one. Annihilation operator (-) does the opposite
+            operator_mapper[idx] += 1 if operator == "+" else -1
+
+        return new_label, coeff
```
mrossinek, line 560: We just had a meeting talking about this, but I wanted to put it up here, too.
<hr/>
We noticed that the commutation is not properly handled in this method. The simplest test for this is the following:

```python
from qiskit_nature.second_q.operators import BosonicOp
from qiskit_nature.second_q.operators.commutators import commutator
print(commutator(BosonicOp({"+_0": 1}) , BosonicOp({"-_0": 1})))
```
This currently yields `-1` but we expect this to be `+1`. So some more investigation on this method needs to be done. Adding unittests for the various `commutator` outcomes is a good idea, too.

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+        is the truncation of the mode (meaning the number of states used in the expansion of the mode,
+        or equivalently the state at which the maximum excitation can take place).
+        The mode |k> is then mapped to the occupation number vector
+        |0_{nk_^max}, 0_{nk_^max - 1}, ..., 0_{n_k + 1}, 1_{n_k}, 0_{n_k - 1}, ..., 0_{0_k}>
+    """
+
+    def _map_single(
+        self, second_q_op: BosonicOp, *, register_length: int | None = None
+    ) -> SparsePauliOp | PauliSumOp:
+        # If register_length was passed, override the one present in the BosonicOp
```
ElePT, line 53: Suggestion to be discussed later (Note: add warning for no use of PauliSumOp)

```suggestion
    ) -> SparsePauliOp:
        """Maps a :class:`~qiskit_nature.second_q.operators.SparseLabelOp`
        to a``SparsePauliOp``.

        Args:
            second_q_op: the ``SparseLabelOp`` to be mapped.
            register_length: when provided, this will be used to overwrite the ``register_length``
                attribute of the operator being mapped. This is possible because the
                ``register_length`` is considered a lower bound in a ``SparseLabelOp``.

        Returns:
            The qubit operator corresponding to the problem-Hamiltonian in the qubit space.
        """
```

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
ElePT, line 59:
```suggestion
```

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
ElePT, line 94: Template suggestion to address comments above (needs to import `from functools import lru_cache` at the top)

```suggestion
    @classmethod
    @lru_cache(maxsize=32)
    def _get_ij_pauli_matrix(register_index: int, register_length: int):
        """This method builds the Qiskit Pauli operators of the operators XX, YY, XY and YX
               Args:
                   ...
               Returns:
                   ...
        """
```

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+"""The Linear Mapper for Bosons."""
+
+from __future__ import annotations
+import operator
+
+from functools import reduce
+
+import numpy as np
+
+from qiskit.opflow import PauliSumOp
```
ElePT, line 22: Suggestion to be discussed later:

```suggestion
```

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,54 @@
...
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Bosonic Mapper."""
+
+from __future__ import annotations
+
+from qiskit.opflow import PauliSumOp
```
ElePT, line 17: Suggestion to be discussed later:

```suggestion
```

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Workflow:
+        First, the user initializes an instance of BosonicLinearMapper, which call the __init__ method
+        defined in this class.
+        Then, the user calls BosonicLinearMapper.map(). Since map is not defined in BosonicLinearMapper,
+        the call is actually to the map method in this class.
+        QubitMapper.map() calls QubitMapper.map() thank to the super() class.
+        Inside QubitMapper.map() there is a call to _map_single. This method is present in QubitMapper
+        class, but it is overloaded in BosonicLinearMapper.
+        Thus, the call is to BosonicLinearMapper._map_single()
```
ElePT, line 38: While I see that this explanation can be useful, I think that the workflow can be inferred from the code, and it does not look consistent with the rest of our documentation, so I would opt to remove it in this case. I am adding though the `attributes` field that Max mentions below (we can discuss it later):

```suggestion
     The following attributes can be read and updated once the ``BosonicMapper`` object
     has been constructed.

        Attributes:
            truncation (int): INSERT DEFINITION HERE. Mention that the min. value is 1.
```

qiskit_nature/second_q/mappers/bosonic_mapper.py
ElePT, line 41:
```suggestion
    def __init__(self, truncation: int) -> None:
        """
        Args:
            truncation: an int .... (insert definition here)
        """
```

qiskit_nature/second_q/operators/bosonic_op.py
ElePT, line 360:
```suggestion
        .. note:
```

qiskit_nature/second_q/operators/bosonic_op.py
ElePT, line 363:
```suggestion
            .. math::
                   [b_i, b_j^\dagger]_- = \delta_{ij} \Rightarrow b_i * b_j^\dagger = b_j^\dagger * b_i + \delta_{ij}
```

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,560 @@
...
+            [b_i, b_j^\\dagger]_- = \\delta_ij => b_i * b_j^\\dagger = b_j^\\dagger * b_i + \\delta_ij
+
+        .. note::
+
+            This method implements the transformation of an operator to the normal ordered operator.
+            The transformation is calculated by considering all commutation relations between the
+            operators.
+            For example, for the case :math:`\\colon b_0 b_0^\\dagger\\colon` where :math:`c_0`
+            is an annihilation operator, this method returns :math:`1 + b_0^\\dagger b_0` due to
+            commutation relations.
```
ElePT, line 372: (Suggestion to be discussed)
```suggestion
            For example, for the case :math:`\colon b_0 b_0^\dagger\colon` where :math:`c_0`
            is an annihilation operator, this method returns :math:`1 + b_0^\dagger b_0` due to
            commutation relations.
```

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,560 @@
...
+                    # No need to multiply the coefficient by -1 due to commutation relations
+
+        new_label = " ".join(f"{term[0]}_{term[1]}" for term in terms)
+        ordered_op += self._new_instance({new_label: coeff})
+        return ordered_op
+
+    def index_order(self) -> BosonicOp:
+        """Convert to the equivalent operator with the terms of each label ordered by index.
+
+        Returns a new operator (the original operator is not modified).
```
ElePT, line 431: Add `Returns:` to docstring

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
ElePT: Discussed offline:  We have decided not to support it, but check for the value of that setting and raise a warning if it is True saying that we are ignoring it

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -0,0 +1,560 @@
...
+
+        .. note::
+
+            This method implements the transformation of an operator to the normal ordered operator.
+            The transformation is calculated by considering all commutation relations between the
+            operators.
+            For example, for the case :math:`\\colon b_0 b_0^\\dagger\\colon` where :math:`c_0`
+            is an annihilation operator, this method returns :math:`1 + b_0^\\dagger b_0` due to
+            commutation relations.
+            See the reference: https://en.wikipedia.org/wiki/Normal_order#Multiple_bosons.
```
ElePT, line 373: Add hyperlink instead of exposing link.

ElePT (commented):
Here are a few more comments in preparation for our meeting (no need to address them before).

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    The Linear boson-to-qubit mapping.
+    This mapper implements the formula in Section II.C of: https://arxiv.org/abs/2108.04258
+
+    b_k^\\dagger -> sum(sqrt(n_k + 1) \\sigma_(n_k)^+\\sigma_(n_k + 1)^-, from n_k = 0 to n_k^max + 1)
+    where n_k^max is the truncation order (defined by the user)
+    In the following, we explicit the operators \\sigma^+ and \\sigma^- with the Pauli matrices:
+    \\sigma_(n_k)^+ := S_j^+ = 0.5 * (X_j + i*Y_j) ; \\sigma_(n_k)^- := S_j^- = 0.5 * (X_j - i*Y_j)
+
+    The length of the qubit register is: BosonicOp.num_mode * (BosonicLinearMapper.truncation + 1)
```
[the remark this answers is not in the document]
ftroisi: Applied

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
ftroisi: Solved

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Suggestion applied

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

test/second_q/mappers/test_bosonic_linear_mapper.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: As we discussed during the meeting, the behavior is correct because 
`commutator(BosonicOp({"-_0": 1}) , BosonicOp({"+_0": 1})) = 1`, as in a commutator the order of the operators matter

ftroisi: I added tests to the bosonic operator

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+from qiskit.quantum_info import Pauli, SparsePauliOp
+
+from qiskit_nature.second_q.operators import BosonicOp
+from .bosonic_mapper import BosonicMapper
+
+
+class BosonicLinearMapper(BosonicMapper):
+    """
+    The Linear boson-to-qubit mapping.
+    This mapper implements the formula in Section II.C of: https://arxiv.org/abs/2108.04258
```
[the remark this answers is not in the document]
ftroisi: Solved

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,115 @@
...
+
+    b_k^\\dagger -> sum(sqrt(n_k + 1) \\sigma_(n_k)^+\\sigma_(n_k + 1)^-, from n_k = 0 to n_k^max + 1)
+    where n_k^max is the truncation order (defined by the user)
+    In the following, we explicit the operators \\sigma^+ and \\sigma^- with the Pauli matrices:
+    \\sigma_(n_k)^+ := S_j^+ = 0.5 * (X_j + i*Y_j) ; \\sigma_(n_k)^- := S_j^- = 0.5 * (X_j - i*Y_j)
```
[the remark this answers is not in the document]
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved (and also added a proper docstring for this method)

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,165 @@
...
+       \\sigma_{n_k}^- := S_j^- = 0.5 * (X_j - i*Y_j)
+
+   The length of the qubit register is: ``BosonicOp.num_modes * (BosonicLinearMapper.truncation + 1)``
+
+   References:
+       [1] A. Miessen et al., Quantum algorithms for quantum dynamics: A performance study on the
+       spin-boson model, Phys. Rev. Research 3, 043212.
+       https://link.aps.org/doi/10.1103/PhysRevResearch.3.043212
+
+    """
```
ElePT, line 63: I am not sure if all you needed to add was the `r` at the beginning of the string, but just in case I also adjusted the indentation to match other classes. I have not been able to `make html` locally so we will have to see the formula visualizations in the CI artifacts (once `make html` runs in CI).

```suggestion
    r"""The Linear boson-to-qubit mapping.
    
    This mapper generates a linear encoding of the Bosonic operator :math:`b_k^\\dagger, b_k` to qubit
    operators (linear combinations of pauli strings).
    In this linear encoding each bosonic mode is represented via :math:`n_k^{max} + 1` qubits, where
    :math:`n_k^{max}` is the truncation of the mode (meaning the number of states used in the expansion of
    the mode, or equivalently the state at which the maximum excitation can take place).
    The mode :math:`|k\rangle` is then mapped to the occupation number vector
    :math:`|0_{n_k^{max}}, 0_{n_k^{max} - 1}, ..., 0_{n_k + 1}, 1_{n_k}, 0_{n_k - 1}, ..., 0_{0_k}\rangle`

    It implements the formula in Section II.C of Reference [1]:

    .. math::
        b_k^\\dagger = \\sum(\\sqrt{n_k + 1} \\sigma_{n_k}^+\\sigma_{n_k + 1}^-)

    from :math:`n_k = 0` to :math:`n_k^max + 1`, where :math:`n_k^max` is the truncation order (defined by the user).
    In the following implementation, we explicit the operators :math:`\\sigma^+` and :math:`\\sigma^-`
    with the Pauli matrices:

    .. math::
        \\sigma_{n_k}^+ := S_j^+ = 0.5 * (X_j + i*Y_j)\\
        \\sigma_{n_k}^- := S_j^- = 0.5 * (X_j - i*Y_j)

    The length of the qubit register is: ``BosonicOp.num_modes * (BosonicLinearMapper.truncation + 1)``

    References:
        [1] A. Miessen et al., Quantum algorithms for quantum dynamics: A performance study on the
        spin-boson model, Phys. Rev. Research 3, 043212.
        https://link.aps.org/doi/10.1103/PhysRevResearch.3.043212
    """
```

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
```diff
@@ -0,0 +1,24 @@
+---
+features:
+  - |
+    Introduced a new feature that implements the bosonic operators ``BosonicOp``.
+    Its functionalities are analogous to the ``FermionicOp`` It should be used to represent a bosonic
+    operator, so if one wants to represent the boson number operator it should do for example::
+
+      from qiskit_nature.second_q.operators import BosonicOp
+      bosonic_op = BosonicOp({'+_0 -_0': 1}, num_modes=1)
```
ElePT, line 9: For the code block to look like python code, you should follow this syntax: (the empty line between `..code-block::python` and the snippet is key)

```suggestion
    operator, so if one wants to represent the boson number operator it should do for example:
    
    .. code-block:: python
    
      from qiskit_nature.second_q.operators import BosonicOp
      bosonic_op = BosonicOp({'+_0 -_0': 1}, num_modes=1)
```

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - |
+    In order to use the bosonic operator in a quantum circuit, this featrue also introduces the
+    bosonic linear mapper, which allows to map the BosonicOp to the qubit space. This mapper is based
+    on the paper: A. Miessen et al., Quantum algorithms for quantum dynamics: A performance study on the
+    spin-boson model, Phys. Rev. Research 3, 043212.
+    To use this mapper on can for example::
+
+      from qiskit_nature.second_q.mappers import BosonicLinearMapper
+      mapper = BosonicLinearMapper(truncation=1)
+      qubit_op = mapper.map(bos_op)
```
ElePT: Similarly to the comment above:

```suggestion
    To use this mapper on can for example:

    .. code-block:: python
    
      from qiskit_nature.second_q.mappers import BosonicLinearMapper
      mapper = BosonicLinearMapper(truncation=1)
      qubit_op = mapper.map(bos_op)
```

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
```diff
@@ -0,0 +1,24 @@
+---
+features:
+  - |
+    Introduced a new feature that implements the bosonic operators ``BosonicOp``.
```
ElePT: sphinx uses the following syntax to link class names to the corresponding API ref:

```suggestion
    Introduced a new feature that implements the bosonic operators :class:`.BosonicOp`.
```

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
```diff
@@ -0,0 +1,24 @@
...
+    Its functionalities are analogous to the ``FermionicOp`` It should be used to represent a bosonic
```
ElePT, line 5: can you reproduce the syntax from above for FermionicOp too?

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
```diff
@@ -0,0 +1,24 @@
...
+    operator, so if one wants to represent the boson number operator it should do for example::
+
+      from qiskit_nature.second_q.operators import BosonicOp
+      bosonic_op = BosonicOp({'+_0 -_0': 1}, num_modes=1)
+
+    The main differences with respect to the fermionic operator are that this class uses the commutator
+    relations instead of the anti-commutator (due to the nature of the particles) and that the simplify
+    method ``bosonic_op.simplify()`` does not commute with the index order ``bosonic_op.index_order()``
```
ElePT, line 13:
```suggestion
    method :meth:`.BosonicOp.simplify` does not commute with the index order :meth:`.BosonicOp.index_order`
```

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+  - |
+    In order to use the bosonic operator in a quantum circuit, this featrue also introduces the
+    bosonic linear mapper, which allows to map the BosonicOp to the qubit space. This mapper is based
+    on the paper: A. Miessen et al., Quantum algorithms for quantum dynamics: A performance study on the
+    spin-boson model, Phys. Rev. Research 3, 043212.
```
ElePT, line 19: I think that the reno can just directly have a link to the paper :)

```suggestion
    on `this paper <https://link.aps.org/doi/10.1103/PhysRevResearch.3.043212>`_
```

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,165 @@
...
+        Returns:
+            The qubit operator corresponding to the problem-Hamiltonian in the qubit space.
+        """
+        if settings.use_pauli_sum_op:
+            warnings.warn(
+                DeprecationWarning(
+                    "As of version 0.6.0 the PauliSumOp is deprecated. We are ignoring this setting"
+                    "for this function. Thus the returned object will have type SparsePauliOp"
+                )
+            )
```
ElePT, line 85: I am not sure which error/warning we should raise here, but it is not necessarily a deprecation one, we are just warning that the setting should not be there. Sth like this (you will have to remove `import warnings` from the top or lint will complain):

```suggestion
            raise NotImplementedError("Setting not supported for this class. Please disable settings.use_pauli_sum_op.")
```

ElePT (commented):
Thank you Francesco! A few comments:

1. Because we are not using `PauliSumOp`at all, I think that we can remove all the parts of the tests that reference it and just keep one unit test to check that the error/warning is raised.
2. There was a recently merged PR to fix the interleaved mapper that has added 2 new abstract classes to the `SparseLabelOp` (which as you know, the `BosonicOp` inherits from). To avoid lint complaints, you will have to implement `from_terms` and `_permute_term` in the bosonic operator class too. You can check out the implementations for the fermionic and spin ops in the following commit: https://github.com/Qiskit/qiskit-nature/commit/92bc524c5ef949c7bad8ba59c88a2c7d583dd9e8. If you have any question, let us know.
3. You can normally check the results of the docs build (`make html`) as part of the CI artifacts. However, it looks like this build has failed (see https://github.com/Qiskit/qiskit-nature/actions/runs/4992640870/jobs/8940774582?pr=1085#step:14:886). This might also be the reason why you did not manage to build it locally. I have added a suggestion that I think will fix the build.

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
ftroisi: Solved

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
ftroisi: Solved

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
ftroisi: Solved

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
ftroisi: Solved

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
ftroisi: Solved

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
ftroisi: Solved

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
ftroisi: Solved

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
```diff
@@ -0,0 +1,28 @@
+---
+features:
+  - |
+    Introduced a new feature that implements the bosonic operator :class:`.BosonicOp`.
+    Its functionalities are analogous to the :class:`.FermioniOp`. It should be used to represent a
```
mrossinek, line 5:
```suggestion
    Its functionalities are analogous to the :class:`.FermioniOp`, but for commuting bosonic particles. It should be used to represent a
```

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
```diff
@@ -0,0 +1,28 @@
...
+    bosonic operator, so if one wants to represent the boson number operator it should do for example:
+
+    .. code-block:: python
+
+      from qiskit_nature.second_q.operators import BosonicOp
+      bosonic_op = BosonicOp({'+_0 -_0': 1}, num_modes=1)
+
+    The main differences with respect to the fermionic operator are that this class uses the commutator
```
mrossinek, line 13: Maybe rephrase this a little based on my previous suggestion

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
```diff
@@ -0,0 +1,28 @@
...
+    relations instead of the anti-commutator (due to the nature of the particles) and that the simplify
+    method :meth:`.BosonicOp.simplify` does not commute with the index order method
+    :meth:`.BosonicOp.index_order`
+
+  - |
+    In order to use the bosonic operator in a quantum circuit, this feature also introduces the
```
mrossinek, line 19:
```suggestion
    In order to use the bosonic operator for quantum applications, this feature also introduces the
```

test/second_q/mappers/test_bosonic_linear_mapper.py
```diff
@@ -0,0 +1,332 @@
...
+    # Using: truncation = 2
+    ref_qubit_op6_tr2 = SparsePauliOp(
+        ["III", "IZZ", "IZI", "IIZ", "ZZI", "ZII"], coeffs=[0.75, -0.25, -0.25, -0.25, -0.5, 0.5]
+    )
+
+    bos_op7 = BosonicOp({"+_0 -_1": 1})
+    bos_op8 = BosonicOp({"-_1 +_0": 1})
+    # Using: truncation = 1
+    ref_qubit_op7_8_tr1 = SparsePauliOp(
+        [
```
mrossinek, line 95: Can we make the formatting here a bit less cumbersome?
For example we can format it manually and then guard it with comments to disallow `black` from changing it.
See [here](https://github.com/Qiskit/qiskit-nature/blob/780547c1d7f759821f9f024f0ddf20817ada74ee/qiskit_nature/constants.py#L42-L56) for an example on how to do this.

Alternatively we could also extract these larger operators into a file and load them from there. See for example [here](https://github.com/Qiskit/qiskit-nature/blob/780547c1d7f759821f9f024f0ddf20817ada74ee/test/second_q/hamiltonians/resources/electronic_energy_op.json) and how it is loaded [here](https://github.com/Qiskit/qiskit-nature/blob/780547c1d7f759821f9f024f0ddf20817ada74ee/test/second_q/hamiltonians/test_electronic_energy.py#L42)

mrossinek (commented):
Alright this is looking great already!

I have a few more comments but other than that I think this is good to go :+1:

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,167 @@
...
+
+from qiskit.quantum_info import Pauli, SparsePauliOp
+from qiskit_nature import settings
+
+from qiskit_nature.second_q.operators import BosonicOp
+from .bosonic_mapper import BosonicMapper
+
+
+class BosonicLinearMapper(BosonicMapper):
+    r"""The Linear boson-to-qubit mapping.
```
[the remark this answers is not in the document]
ftroisi: Solved (I formatted the superscripts, they should be all right now)

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,167 @@
...
+
+    This mapper generates a linear encoding of the Bosonic operator :math:`b_k^\\dagger, b_k` to qubit
+    operators (linear combinations of pauli strings).
+    In this linear encoding each bosonic mode is represented via :math:`n_k^{max} + 1` qubits, where
+    :math:`n_k^{max}` is the truncation of the mode (meaning the number of states used in the expansion
+    of the mode, or equivalently the state at which the maximum excitation can take place).
+    The mode :math:`|k\rangle` is then mapped to the occupation number vector
```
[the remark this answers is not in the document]
ftroisi: Solved

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    Mapper of Bosonic Operator to Qubit Operator
+
+    The following attributes can be read and updated once the ``BosonicMapper`` object
+    has been constructed.
+
+    Attributes:
+        truncation (int): defines the excitation space of the k-th bosonic state. Together with the
+        number of modes required to represent the bosonic operator, it defines the minimum length of the
+        qubit register. The minimum value is 1.
```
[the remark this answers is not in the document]
ftroisi: Solved

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+
+    def __init__(self, truncation: int) -> None:
+        """
+        Args:
+            truncation (int): defines the excitation space of the k-th bosonic state. Together with the
+            number of modes required to represent the bosonic operator, it defines the minimum length of
+            the qubit register. The minimum value is 1.
```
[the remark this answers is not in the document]
ftroisi: Solved

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,59 @@
...
+    Attributes:
+        truncation (int): defines the excitation space of the k-th bosonic state. Together with the
+        number of modes required to represent the bosonic operator, it defines the minimum length of the
+        qubit register. The minimum value is 1.
+    """
+
+    def __init__(self, truncation: int) -> None:
+        """
+        Args:
+            truncation (int): defines the excitation space of the k-th bosonic state. Together with the
```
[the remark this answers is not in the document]
ftroisi: Solved

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
ftroisi: Solved

test/second_q/mappers/test_bosonic_linear_mapper.py
ftroisi: How about it now? I went for the first suggestion

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
ftroisi: How about now?

test/second_q/mappers/test_bosonic_linear_mapper.py
```diff
@@ -0,0 +1,188 @@
...
+
+    bos_op6 = BosonicOp({"-_0 +_0": 1})
+    # Using: truncation = 1
+    ref_qubit_op6_tr1 = SparsePauliOp(["II", "ZZ", "ZI", "IZ"], coeffs=[0.25, -0.25, +0.25, -0.25])
+    # Using: truncation = 2
+    ref_qubit_op6_tr2 = SparsePauliOp(
+        ["III", "IZZ", "IZI", "IIZ", "ZZI", "ZII"], coeffs=[0.75, -0.25, -0.25, -0.25, -0.5, 0.5]
+    )
+
+    # fmt: off
```
mrossinek, line 91: Can you please limit the format disabling to really just the few lines where its needed?

releasenotes/notes/bosonic-operator-and-mapper-45bfde873f092681.yaml
mrossinek: Yes, I think it is better now :+1:

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,171 @@
...
+        \\sigma_{n_k}^+ := S_j^+ = 0.5 * (X_j + i*Y_j)\\
+        \\sigma_{n_k}^- := S_j^- = 0.5 * (X_j - i*Y_j)
+
+    The length of the qubit register is: ``BosonicOp.num_modes * (BosonicLinearMapper.truncation + 1)``
+
+    .. note::
+        Since this mapper uses a truncation in the representation of the Bosonic operators, the
+        commutation relation after the mapping differ from the standard ones. Please refer to
+        Section 4, equation 22 of Reference [2] for more details
+
```
mrossinek, line 60: I noticed just now that this docstring does not have a code example on how to instantiate a mapper. Maybe we can add one (just like in the release note perhaps).

<hr/>

Our other mappers do not have this either which should be improved upon in the future.

test/second_q/mappers/test_bosonic_linear_mapper.py
ftroisi: Done

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,180 @@
...
+        Raises:
+            NotImplementedError: when `qiskit_nature.settings.use_pauli_sum_op` is set to `True`.
+            This value is deprecated and, thus, not supported by this new implementation.
+            Set it to `False` in order to use this mapper.
+        """
+        if settings.use_pauli_sum_op:
+            raise NotImplementedError(
+                "Setting not supported for this class. Please disable settings.use_pauli_sum_op."
+            )
```
woodsp-ibm, line 101: Since this is new code I am wondering if it might just be better to ignore the setting (perhaps log a warning if we feel that a use might try and switch this to using PauliSumOp). The BosonicMapper already states in the map(), which is overriding that of the qubitmapper, that only SparsePauliOp types are returned.

It was really done to allow someone with existing code to transition over to the SparsePauliOp with the existing code in Nature so on new code... though this. I am expecting that we would be changing the default value over to False as well anyway in the next release, before it gets removed entirely as an option later - more as a comment to  @mrossinek

mrossinek: I explicitly decided against ignoring this in view of #795 being planned after this. When we silently just return a `SparsePauliOp` while the rest of the stack might still expect a `PauliSumOp`, this can lead to some odd edge cases.
Thus the decision to raise an error here explicitly.

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,58 @@
...
+                number of modes required to represent the bosonic operator, it defines the minimum length
+                of the qubit register. The minimum value is 1.
+        """
+        super().__init__()
+        if truncation < 1:
+            raise QiskitNatureError(
+                "Truncation for bosonic linear mapper must be at least 1. "
+                + f"Detected value: {truncation}"
+            )
+        self.truncation = truncation
```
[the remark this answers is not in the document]
mrossinek: A default cannot be set here since "all possible" is an infinite number. Thus, this is a required setting.
And choosing a default on our side would be ill-advised because this can have serious implications on the quality of the result.

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
woodsp-ibm: Ok, with combining this with an existing operator to form a MixedOperator I can see that. For the next release, as I mentioned, I imagine we would be changing the default value for that setting, such by default things do not fail here if you simply use this alone. We might want to change this text `Please disable settings.use_pauli_sum_op` to be a bit clearer what disable means, ie simply state it must be set False.

qiskit_nature/second_q/mappers/bosonic_mapper.py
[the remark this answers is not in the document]
woodsp-ibm: Ok, with the name stating `truncation (int): defines the excitation space of the k-th bosonic state.` as in it defines something the name truncation still seems a bit weird to me. It feels to me like this is analogous to defining an active space for an electronic structure problem - the goal there being to limit things too, though there at least things are finite to start with.

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
mrossinek: Yes we can be more explicit in that warning. Maybe something along the lines of:
```
This class does not support the deprecated `settings.use_pauli_sum_op` value. Please set `settings.use_pauli_sum_op = False` in order to use this mapper.
```

qiskit_nature/second_q/mappers/bosonic_mapper.py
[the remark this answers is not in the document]
mrossinek: No, this is not related to active spaces in electronic structure. Let me try to explain in as non-technical terms as possible:

Bosonic particles can theoretically be excited an infinite number of times. This is in contrast to fermions where `+_0 +_0` is an invalid operation because a fermion can only be excited once.
Consequently, this means that for bosons `+_0 +_0 +_0 ...` is valid and, thus, one can have states such as `|0>`, `|1>`, `|2>` etc.
In order to map these states onto a qubit register, this linear mapper encodes these in a linear fashion:
```
|0> -> |10000...>
|1> -> |01000...>
|2> -> |00100...>
...
```
As you can see, theoretically we would need an infinite number of qubits to encode all possible bosonic states. That is where this `truncation` comes in, as it specifies the maximum excitation that we allow to be represented. So for example for `truncation=1`, we get:
```
|0> -> |10>
|1> -> |01>
|2> -> can no longer be represented
```
For `truncation=2` we will get out 3 qubits and so on.
This is essentially what the `LinearMapper` does for `SpinOp`, too.

In the future, we should also look into implementing the equivalent of the `LogarithmicMapper`, where one encodes for example:
```
|0> -> |00>
|1> -> |10>
|2> -> |01>
|3> -> |11>
```

Now, the difference to the `LinearMapper` is that in those cases, the `SpinOp` carries the truncation information in the form of its `.spin` attribute. But for the bosonic case, the bosonic operator itself is not the right place to store this, as this is merely an artifact of the mapping procedure.

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
ftroisi: I changed the error message to the one suggested by @mrossinek

qiskit_nature/second_q/mappers/bosonic_mapper.py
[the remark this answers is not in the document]
woodsp-ibm: Thanks for the explanation - you still feel truncation is the best name? The field currently says `defines the excitation space` and you explain it as `max excitation` yes it may truncate but the name truncate on it own does not really say what it is just truncating `max_excitation`, `excitation_limit` or something like that seem to be in name more what description says. I guess excitation_truncation, a bit more verbose though. And presently this field is just in the base class right with the parameter docstring being generic about limiting things but that base class is not documented. Can see how it looks after inheritence change I guess.

And maybe this got "lost" in all this discussion, from the first comment above.

>This will only do this check in the constructor here and if you set the attribute directly it won't get done. I would think a better choice, in this case, would be a property setter/getter where the value is checked by the setter. If the setter is also used by the constructor then no matter how the value is set it would always be checked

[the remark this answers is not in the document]
mrossinek: Yes, the property setter is a good idea.

Regarding the name: I really don't see it as a big issue. If we want to make it more verbose, we should probably stick to how this is called in the literature. The main reference here is [Miessen et al.](https://journals.aps.org/prresearch/abstract/10.1103/PhysRevResearch.3.043212) where they talk phrase it as

> The excitation space of the k-th bosonic state will be truncated at a maximum occupation number $n_k^{max}$

That's where the `truncation` name came from originally. But I would be fine with `max_occupation`.
My original explanation kind of misused "excitation" but `+_0` really is a particle creation and thus `|n>` is really not the n-th excited state but rather a bosonic state with n particles in it.

<hr/>

TL;DR: I suggest we rename `truncation -> max_occupation` and that we implement it as a property setter :+1:

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,67 @@
...
+class BosonicMapper(QubitMapper):
+    """Mapper of Bosonic Operator to Qubit Operator
+
+    The following attributes can be read and updated once the ``BosonicMapper`` object
+    has been constructed.
+
+    Attributes:
+        max_occupation (int): defines the excitation space of the k-th bosonic state. Together with the
+            number of modes required to represent the bosonic operator, it defines the minimum length of
+            the qubit register. The minimum value is 1.
```
ElePT, line 34: This description is only necessary for attributes, because we have decided to make `max_occupation` a property, we can now remove it (sorry for the back and forth):

```suggestion
```

ElePT (commented):
Thanks Francesco! Just a few more steps. We were looking for a specific way of defining properties and property setters, see specific comments in the code.

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,67 @@
...
+
+        Raises:
+            QiskitNatureError: if max_occupation < 1
+        """
+        if max_occupation < 1:
+            raise QiskitNatureError(
+                "Max occupation for bosonic linear mapper must be at least 1. "
+                + f"Detected value: {max_occupation}"
+            )
+        self.max_occupation = max_occupation
```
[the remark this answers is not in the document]
ftroisi: Done for all three suggestions :)
Thanks a lot

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,182 @@
...
+    (defined by the user).
+    In the following implementation, we explicit the operators :math:`\\sigma^+` and :math:`\\sigma^-`
+    with the Pauli matrices:
+
+    .. math::
+        \\sigma_{n_k}^+ := S_j^+ = 0.5 * (X_j + i*Y_j)\\
+        \\sigma_{n_k}^- := S_j^- = 0.5 * (X_j - i*Y_j)
+
+    The length of the qubit register is:
+    ``BosonicOp.num_modes * (BosonicLinearMapper.max_occupation + 1)``
```
ElePT, line 55: Inline code does not work well with spaces, so maybe we could make this look better if it were a single code block (might have to run black again)

```suggestion
    .. code-block:: python
    
     BosonicOp.num_modes * (BosonicLinearMapper.max_occupation + 1)
```

ftroisi: Done :)

ElePT (commented):
Sorry, there was a missing empty line in my code block suggestion from before and it didn't render properly. New suggestion should fix that (also added an empty line between imports and code in the example below because I think it looks a bit better)

Screenshot 2023-05-22 at 16 04 17

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,186 @@
...
+     BosonicOp.num_modes * (BosonicLinearMapper.max_occupation + 1)
+
+    To use this mapper one can for example:
+
+    .. code-block:: python
+
+      from qiskit_nature.second_q.mappers import BosonicLinearMapper
+      from qiskit_nature.second_q.operators import BosonicOp
+      mapper = BosonicLinearMapper(max_occupation=1)
+      qubit_op = mapper.map(BosonicOp({'+_0 -_0': 1}, num_modes=1))
```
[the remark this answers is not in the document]
ftroisi: done!

ElePT (approved):
Max will probably give a last round of review, but in terms of docs, I have nothing to say other than good job. Thanks a lot!

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,188 @@
...
+
+    from :math:`n_k = 0` to :math:`n_k^{max} + 1` where :math:`n_k^{max}` is the maximum occupation
+    (defined by the user).
+    In the following implementation, we explicit the operators :math:`\\sigma^+` and :math:`\\sigma^-`
+    with the Pauli matrices:
+
+    .. math::
+        \\sigma_{n_k}^+ := S_j^+ = 0.5 * (X_j + i*Y_j)
+
+        \\sigma_{n_k}^- := S_j^- = 0.5 * (X_j - i*Y_j)
```
mrossinek, line 53: A minor detail: in the rendered version of this I did not immediately recognize `i` as the imaginary number but instead assumed it was another index like `j`. I am not sure how to improve this.

Maybe we can do either or both of the following:
- change the font of `i`
- remove the `*` between `i` and `Y_j`

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,188 @@
...
+      from qiskit_nature.second_q.mappers import BosonicLinearMapper
+      from qiskit_nature.second_q.operators import BosonicOp
+
+      mapper = BosonicLinearMapper(max_occupation=1)
+      qubit_op = mapper.map(BosonicOp({'+_0 -_0': 1}, num_modes=1))
+
+    .. note::
+        Since this mapper uses a truncation in the representation of the Bosonic operators, the
+        commutation relation after the mapping differ from the standard ones. Please refer to
+        Section 4, equation 22 of Reference [2] for more details
```
mrossinek, line 74:
```suggestion
        Since this mapper truncates the maximum occupation of a bosonic state as represented in the
        qubit register, the commutation relation after the mapping differ from the standard ones.
        Please refer to Section 4, equation 22 of Reference [2] for more details
```

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,188 @@
...
+                attribute of the operator being mapped. This is possible because the
+                ``register_length`` is considered a lower bound in a ``SparseLabelOp``.
+
+        Returns:
+            The qubit operator corresponding to the problem-Hamiltonian in the qubit space.
+
+        Raises:
+            NotImplementedError: when `qiskit_nature.settings.use_pauli_sum_op` is set to `True`.
+            This value is deprecated and, thus, not supported by this new implementation.
+            Set it to `False` in order to use this mapper.
```
mrossinek, line 103:
```suggestion
                This value is deprecated and, thus, not supported by this new implementation.
                Set it to `False` in order to use this mapper.
```

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,188 @@
...
+            # Add the map of the single boson op (e.g. +_0) to the map of the full bosonic operator
+            pauli_op.append(coeff * reduce(operator.add, bos_op_to_pauli_op.simplify()))
+
+        # return the lookup table for the transformed XYZI operators
+        bos_op_encoding = reduce(operator.add, pauli_op)
+        return bos_op_encoding
+
+    @classmethod
+    @lru_cache(maxsize=32)
+    def _get_ij_pauli_matrix(cls, register_index: int, register_length: int):
```
mrossinek, line 154: This is missing a return type hint. I suppose it should be something along the lines of the following:
```suggestion
    def _get_ij_pauli_matrix(cls, register_index: int, register_length: int) -> tuple[Pauli, Pauli, Pauli, Pauli]:
```

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,188 @@
...
+        """This method builds the Qiskit Pauli operators of the operators XX, YY, XY and YX
+
+        Args:
+            register_index: the index of the qubit register where the mapped operator should be placed.
+            register_length: the length of the qubit register.
+
+        Returns:
+            Four Pauli operators that represent XX, YY, XY and YX at the specified index in the
+            current qubit register.
```
mrossinek, line 163:
```suggestion
            Four Pauli operators that represent XX, XY, YX and YY at the specified index in the
            current qubit register.
```

qiskit_nature/second_q/mappers/bosonic_linear_mapper.py
```diff
@@ -0,0 +1,188 @@
...
+                prefix_zeros + [1, 1] + suffix_zeros,
+            )
+        )
+        y_x = Pauli(
+            (
+                prefix_zeros + [1, 0] + suffix_zeros,
+                prefix_zeros + [1, 1] + suffix_zeros,
+            )
+        )
+        y_y = Pauli((prefix_zeros + [1, 1] + suffix_zeros, prefix_zeros + [1, 1] + suffix_zeros))
```
mrossinek, line 187:
```suggestion
        y_y = Pauli(
            (
                prefix_zeros + [1, 1] + suffix_zeros,
                prefix_zeros + [1, 1] + suffix_zeros,
            )
        )
```
Just to make it a bit easier to follow the code and see the difference to above

mrossinek (commented):
Okay, some last final nitpicky comments.
Once these changes are applied I'm confident we can merge this PR :+1:

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,62 @@
...
+            max_occupation: defines the excitation space of the k-th bosonic state. Together with the
+                number of modes required to represent the bosonic operator, it defines the minimum length
+                of the qubit register. The minimum value is 1.
+        """
+        super().__init__()
+        self.max_occupation = max_occupation
+
+    @property
+    def max_occupation(self) -> int:
+        """The max. occupation for the bosonic linear mapper"""
```
mrossinek, line 44: I found one final little detail:
screenshot_1684823986

Fixing it like this:
```suggestion
        """The maximum occupation of any bosonic state."""
```

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,62 @@
...
+        return self._max_occupation
+
+    @max_occupation.setter
+    def max_occupation(self, max_occupation: int) -> None:
+        if max_occupation < 1:
+            raise ValueError(
+                "Max occupation for bosonic linear mapper must be at least 1. "
```
mrossinek, line 51:
```suggestion
                "The maximum occupation must be at least 1. "
```
Just because this generalizes to any bosonic mapper and not just the linear one

mrossinek (approved):
Assuming I did not mess up my prior suggestions, CI should pass on this...

<hr/>

Thanks a lot, Francesco, for your hard work on this great contribution to Qiskit Nature! :tada: 
Thank you, Elena, for your support in reviewing and helping Francesco out on the home stretch of this PR.

qiskit_nature/second_q/mappers/bosonic_mapper.py
```diff
@@ -0,0 +1,62 @@
...
+        """The maximum occupation of any bosonic state."""
+        return self._max_occupation
+
+    @max_occupation.setter
+    def max_occupation(self, max_occupation: int) -> None:
+        if max_occupation < 1:
+            raise ValueError(
+                "The maximum occupation must be at least 1. "
+                + f"Detected value: {max_occupation}"
+            )
```
mrossinek, line 53:
```suggestion
            raise ValueError(
                f"The maximum occupation must be at least 1, and not {max_occupation}."
            )
```

:black_heart:

mrossinek (approved):

PR #1092, opened 2023-03-10
Clean up deprecation warnings in CI

mrossinek:
The qEOM algorithm used to trigger a lot of warnings when finding the ground-state solution to start from. This commit avoids that by filtering out expected operator names unless the user explicitly overwrote them.

### Summary

The CI is currently littered with a lot of (deprecation) warnings. In this PR I would like to see if this situation can be improved upon.


mrossinek:
There are still some warnings left but the situation has overall improved quite a lot already. I consider this to be good enough for the time being. In the future we should go over the unittests and remove/duplicate the tests which currently depend on `QubitConverter` and the `(Minimum)EigensolverFactory` classes with their replacements, respectively.

This was done to some degree during the refactoring but could be improved further.

woodsp-ibm (approved):

PR #1097, opened 2023-03-16
The example runs locally

julenl:
### Summary
Made a simple edit to a line to avoid confusion.

### Details and comments
Since multiple previous version, the code snippet has been updated, but not the paragraph below. The examples provided do not run on a quantum computer, in any of the cases/versions. Instead, they run run locally on an quantum computer simulator (Aer if I'm not wrong). I added the *virtual* for emphasizing that it runs on a "non-hardware" quantum computer. Alternatively, it could also say something like:

    The program above uses quantum (computing) algorithms to calculate...

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ algorithm = GroundStateEigensolver(mapper, vqe)
 electronic_structure_result = algorithm.solve(problem)
 print(electronic_structure_result)
 ```
-The program above uses a quantum computer to calculate the ground state energy of molecular Hydrogen,
+The program above uses a virtual quantum computer simulator to calculate the ground state energy of molecular Hydrogen,
 H<sub>2</sub>, where the two atoms are configured to be at a distance of 0.735 angstroms. The molecular
 input specification is processed by the PySCF driver. This driver produces an `ElectronicStructureProblem`
 which gathers all the problem information required by Qiskit Nature.


README.md
```diff
@@ -112,7 +112,7 @@ algorithm = GroundStateEigensolver(mapper, vqe)
 electronic_structure_result = algorithm.solve(problem)
 print(electronic_structure_result)
 ```
-The program above uses a quantum computer to calculate the ground state energy of molecular Hydrogen,
+The program above uses a virtual quantum computer simulator to calculate the ground state energy of molecular Hydrogen,
```
woodsp-ibm, line 115:
```suggestion
The program above computes the ground state energy of molecular Hydrogen,
```
The text there, which includes `uses a quantum computer`, has been around a long time. In the past the backend was arguably always more evident, see here https://github.com/Qiskit/qiskit-aqua#creating-your-first-chemistry-programming-experiment-in-qiskit . Qiskit is now using primitives, and in this case no it's not using Aer simulator, rather its done using the reference Estimator from qiskit.primitives which is implemented using the quantum_info module in Terra. Nowadays, pointing out that this is quantum, which was done from when this sample code first came about, I would argue is less needed. And rather than worry about where its computed, which can simply be changed by changing the Estimator, I would more say that if any change is needed here that I would more just drop that now from the text altogether as per my suggestion here.

julenl:
This might not be the best place to ask the question, but I'll do it anyway, since it is related to the topic.

As the *backend* concept keeps getting pushed deeper and deeper below wrappers and more wrappers, we kind of forget the fact that this code is supposed to run on quantum hardware at some point... Right?

Couldn't we provide a tutorial to run a "Hello World!" (H<sub>2</sub>) on quantum hardware? Even if it is just for symbolic reasons...

woodsp-ibm:
> Couldn't we provide a tutorial to run a "Hello World!" (H2) on quantum hardware? Even if it is just for symbolic reasons...

In the [past we used to have a commented out block in tutorials](https://github.com/Qiskit/qiskit-tutorials/blob/stable/0.12.x/qiskit/advanced/aqua/chemistry/dissociation_profile_of_molecule.ipynb) (code has changed a lot since then!). Thing is that this can not reasonably be tested under CI, so if things change it may break if its not noticed, so it was dropped. 

In practice nowadays one would simply choose a primitive - Estimator - from the provider, Qiskit IBM Runtime, and configure use that. They follow the base interface but configuration and what it supports etc, can vary. Now there are tutorials for runtime
https://qiskit.org/documentation/partners/qiskit_ibm_runtime/ including a chemistry one https://qiskit.org/documentation/partners/qiskit_ibm_runtime/tutorials/vqe_with_estimator.html

Also there is work elsewhere on documenting primitives in general and choosing a primitive e.g. Qiskit/qiskit-tutorials#1385
Whether we can tie things together/link to these from the applications repo we will have to see, The use/selection of primitives to run the code applies equally to all.

PR #1109, opened 2023-03-31
Various small updates for docs and docs CI

woodsp-ibm:
### Summary

- Removes dependency on sphinx-autodoc-typehints as per Qiskit/qiskit-terra#9705
- Adds default thumbnail config as that changed in nbsphinx 0.9 see Qiskit/qiskit-metapackage#1707
- Updates tox.ini gettext section used for translatable strings due to ipykernel not being installed by default now see Qiskit/qiskit-machine-learning#589

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,10 +6,9 @@ pylatexenc>=1.4
 stestr>=2.0.0
 ddt>=1.2.0,!=1.4.0,!=1.4.3
 reno>=3.4.0
-Sphinx>=1.8.3,!=3.1.0,!=5.2.0,!=5.2.0.post0
+Sphinx>=5.0
 sphinx-design>=0.2.0
 sphinx-gallery
-sphinx-autodoc-typehints<1.14.0
 sphinxcontrib-spelling!=7.3.1
 jupyter-sphinx
 discover
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.7]
+        python-version: [3.8]
     steps:
       - name: Print Concurrency Group
         env:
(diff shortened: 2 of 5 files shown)


woodsp-ibm:
Looking at the html ouput on my local machine, built with `make html`, it seems that parameters for the SparseLabelOp init(and subclasses) now have types, whereas before, and as can be seen on the [currently published SparseLabelOp](https://qiskit.org/documentation/nature/stubs/qiskit_nature.second_q.operators.SparseLabelOp.html) it did not have. Hence the CI failure now for _TCoeff I imagine

image

I say above "I imagine" as the error seems to point towards `terms` but I do not see tcoeff or _TCoeff there

>qiskit_nature/second_q/operators/sparse_label_op.py:docstring of qiskit_nature.second_q.operators.sparse_label_op.SparseLabelOp.terms:1: : Spell check: tcoeff: _TCoeff.

woodsp-ibm:
I marked this onhold as I have temporarily disabled spell check so it creates the artifacts so I can see how its building here as compared to my local machine. That check needs to go back before this ever gets merged.

woodsp-ibm:
The terms return type locally for me is like this

image

whereas from CI here its 

image

which at least mentions TCoeff as per the error message. I will have to check, e.g library versions, as to why this is different.

requirements-dev.txt
```diff
@@ -6,10 +6,9 @@ pylatexenc>=1.4
 stestr>=2.0.0
 ddt>=1.2.0,!=1.4.0,!=1.4.3
 reno>=3.4.0
-Sphinx>=1.8.3,!=3.1.0,!=5.2.0,!=5.2.0.post0
+Sphinx>=5.0
```
mrossinek, line 9: This removes the penalty against `5.2.0` and `5.2.0.post0`. I don't think that was intentional was it?

mrossinek (commented):
Looking at your previous comments I still think keeping 3.8 for the CI job is good (I also noted your offline message regarding required CI jobs needing updates accordingly).

requirements-dev.txt
woodsp-ibm: I did this based on simplifying it down to what Terra it updated to. I think at the time we had issues with those versions. I do not know if that still is the case. Since its not what we pull into CI any longer I figured why maintain an older list of exceptions here that I do not know if we hit problems with any more or not. If someone has that old version and finds a problem in building locally they can simply update it, It does not break our CI here anymore (since we install 5.3.0), which is more why these restrictions were added.

PR #1120, opened 2023-04-11
Fall back to numpy arrays in `VibrationalIntegrals`

mrossinek:
### Summary

This extends the `VibrationalIntegrals` class to fall back to using numpy arrays when the `sparse` package is not installed.

### Details and comments

This is required if we want to support Python 3.11 before `sparse` does so.


woodsp-ibm (approved):
LGTM. 

One thing I wonder, to go a bit beyond the logged warning. is that perhaps (separate issue/PR) is that we could update the [Nature Getting Started](https://qiskit.org/documentation/nature/getting_started.html) docs page with the optional installs like `sparse` and `opt_einsum` to note them and what the benefits are of installing them. For now it just talks about drivers, which probably were the only optionals at the time when it was originally done.

PR #1122, opened 2023-04-12
Extract code example from README into docs

### LINKED ISSUE #1099: Add version number for README code examples
### What should we add?

While answering [this question](https://stackoverflow.com/questions/75755628/modulenotfounderror-no-module-named-qiskit-chemistry/75766605#75766605) on StackOverflow, I realized that the README code examples didn't work with qiskit-nature 0.5.2. I thought there were mistakes in the code, but it turns out the code example was only updated yesterday by https://github.com/Qiskit/qiskit-nature/pull/1084 to work with `main` branch or the future `0.6.0` release. 

I personally like the README code example because it contains all the code necessary for a standard workflow within one code block. But I want to suggest to add a note to users that this code example works with the `main` branch, if people want to see the corresponding code examples for stable releases, go to https://github.com/Qiskit/qiskit-nature/tree/stable/0.5

mrossinek:
### Summary

We have been receiving several comments/questions/issues regarding the code example from the README not working as expected. This is due to it being aligned with the development and not the published code. In order to resolve this, I am moving the code example into the documentations "getting started" page. In this way, the README can simply link there which ensures that the snippet is aligned with the published (stable) code.

### Details and comments

In order to still be able to test the proper functioning of this code, I am using doctest and have introduced the new `formatting_precision` attribute on the result objects.

Fixes #1099


woodsp-ibm:
Since this affects the readme, and the link will not have the code until we republish the docs, after this upcoming version has been released, probably we should hold this PR until just before when we do the release so as to minimize the time the updated readme is out there before the docs are published with this new version that does include the code.

docs/getting_started.rst
```diff
@@ -110,6 +110,129 @@ and how to install the dependent library/program that it requires.
...
+automatically removes 2 qubits due to inherit symmetries when the `num_particles`
+are provided to it; a reduction in complexity that is particularly advantageous
+for NISQ computers.
+
+For actually finding the ground state solution, the Variational Quantum
+Eigensolver (:class:`~qiskit.algorithms.minimum_eigensolvers.VQE`) algorithm is
+used. Its main three components are the estimator primitive
+(:class:`~qiskit.primitives.Estimator`), wavefunction ansatz
+(:class:`~qiskit_nature.second_q.circuit.library.UCCSD`), and optimizer
+(:class:`~qiskit.algorithms.optimiziers.L_BFGS_B`).
```
woodsp-ibm, line 221:
```suggestion
(:class:`~qiskit.algorithms.optimizers.L_BFGS_B`).
```

woodsp-ibm (approved):
Approved so its ready to go when the hold is removed.

PR #1135, opened 2023-04-20
Fix lint after opflow deprecation

woodsp-ibm:
### Summary

Qiskit/qiskit-terra#9176 deprecated opflow, which included StarAlebraMixin used by SecondQuantizedOp. To add the deprecated decorator for the class an empty init() method was added in order to add a decorator.

The above then gave rise to this in the nightly scheduled builds here in Nature

```
************* Module qiskit_nature.operators.second_quantization.second_quantized_op
qiskit_nature/operators/second_quantization/second_quantized_op.py:39:19: E1121: Too many positional arguments for constructor call (too-many-function-args)
```

which I have simply supressed in the code. The code, like opflow, is deprecated and will be removed in the future,

### DIFF
--- a/qiskit_nature/operators/second_quantization/second_quantized_op.py
+++ b/qiskit_nature/operators/second_quantization/second_quantized_op.py
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021, 2022.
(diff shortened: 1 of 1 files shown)


mrossinek (approved):
Thanks!

PR #1140, opened 2023-04-25
Update intersphinx mappings

woodsp-ibm:
### Summary

With the change to publishing many of the qiskit docs in `ecosystem` this updates the mapping url for `rustworkx` accordingly. 

Since this is now the correct url, I also marked it for stable backport, for any docs updates we do, though I imagine the redirect on the former url will be in place for some time.


mrossinek (approved):
Thanks for catching this!

PR #1144, opened 2023-04-26
Remove unneeded casts

mrossinek:
### Summary

This removes some unneeded casts from the operator classes.

### Details and comments

Complexity should not pose a real problem any longer because we do not do any explicit down- or up-conversions in the stack any longer. This can be seen since `Tensor` and `PolynomialTensor` explicitly allow complex values. I do not recall why I added these casts in the first place, but I think they can simply be removed.


mrossinek:
If approved, should we backport this to stable, too?

woodsp-ibm (approved):

PR #1145, opened 2023-04-28
Pin ipython

woodsp-ibm:
### Summary

ipython 8.13 dropped python 3.8 support. See Qiskit/qiskit-optimization#500


mrossinek (approved):
Thanks :+1:

PR #1146, opened 2023-04-30
Use a coherent name for the problem

julenl:
The class defining the electronic structure problem, is defined on the sample code as `es_problem`. Thereby, for the sake of coherence, either the variable name in the code should be updated to `problem` in all cases or we just update this string here.

### Summary
Update `problem` to `es_problem`

### Details and comments
Update `problem` to `es_problem` on the comment summarizing the steps for the calculation, for sake of coherence.

### DIFF
--- a/docs/tutorials/03_ground_state_solvers.ipynb
+++ b/docs/tutorials/03_ground_state_solvers.ipynb
@@ -178,7 +178,7 @@
    "metadata": {},
    "source": [
     "This will now take of the entire workflow:\n",
-    "1. generating the second-quantized operators stored in our `problem`\n",
+    "1. generating the second-quantized operators stored in our `es_problem`\n",
     "2. mapping (and potentially reducing) the operators in the qubit space\n",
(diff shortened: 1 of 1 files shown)


docs/tutorials/03_ground_state_solvers.ipynb
mrossinek, line 1: If you want to change this, I would actually rather remove the verbatim formatting from the word "problem" completely. The reason for that being is that we actually have `es_problem` and `vib_problem` being used in the tutorial, and the explanations given here applies to either of those cases. So rather than further limiting the scope of this explanation, we could simply refer to it as a "problem" without formatting it as a variable (which is also done elsewhere in this tutorial).

julenl (commented):
Looks good.

docs/tutorials/03_ground_state_solvers.ipynb
mrossinek: Sorry I must have overlooked your reply here.
If you change it to "(here referred to as `es_problem`)" then its good :+1:

mrossinek (approved):

PR #1149, opened 2023-05-02
Remove outdated comment from README

mrossinek:
### Summary

I just noticed this now outdated comment in the README so I am removing it here.
I am not yet removing the `.mailmap` file itself because I am not sure whether this is still being used in some workflow. I will check with the Terra folks and once they remove it, so can we.


woodsp-ibm (approved):
Guess we need to do this in the other repos as well.

mrossinek:
> I am not yet removing the `.mailmap` file itself because I am not sure whether this is still being used in some workflow. I will check with the Terra folks and once they remove it, so can we.

Actually the `.mailmap` file serves another purpose since `git log` can actually make use of that information: https://git-scm.com/docs/git-log#Documentation/git-log.txt---no-mailmap

Thus this file will not be removed.

PR #1153, opened 2023-05-03
Enable sparse optional for 3.11

woodsp-ibm:
### Summary

Now that the sparse dependence numba has a 3.11 supported release this removes the Python version limit from the optional sparse install.


mrossinek (approved):
Thanks a lot!

PR #1158, opened 2023-05-05
Fix the `InterleavedQubitMapper` logic

### LINKED ISSUE #542: Permute indices of SparseLabelOp subclasses
### What should we add?

FermionicOp should support relabeling its indices according to a permutation. This is useful for performing fermionic swaps virtually.

### LINKED ISSUE #954: Construct SparseLabelOp subclasses from list of tuples
### What should we add?

The new version of FermionicOp does not support this yet; see #493 . It should be implemented as a class method.

### LINKED ISSUE #1157: The `InterleavedQubitMapper` does not yield the expected benefits
### Environment

- **Qiskit Nature version**: 0.6.0 and later
- **Python version**: any
- **Operating system**: any

### What is happening?

#1046 added the new `InterleavedQubitMapper` which changes from a block- to an interleaved ordering of the qubits (w.r.t. the fermionic spin orbitals these qubits have been mapped from). I only just now realized that (while the code is theoretically correct) it does not provide the actually intended outcome. What should really happen, is that the _fermionic spin orbitals_ are reordered **before** they get mapped into the qubit space.
This actually makes sense because on the qubit level we should not have the notion of spin orbitals any more. The reason this makes is a difference can best be seen when taking a paired double excitation:
```
Fermionic Operator
number spin orbitals=8, number terms=2
  1j * ( +_0 +_4 -_2 -_6 )
+ (-0-1j) * ( +_6 +_2 -_4 -_0 )
```
When mapping this with the Jordan-Wigner mapping we obtain:
```
-0.125 * IYZYIXZY
- 0.125 * IXZYIYZY
- 0.125 * IXZXIXZY
+ 0.125 * IYZXIYZY
- 0.125 * IXZYIXZX
+ 0.125 * IYZYIYZX
+ 0.125 * IYZXIXZX
+ 0.125 * IXZXIYZX
```
where we see various `Z` terms appearing as expected.
However, when re-arranging the spin orbitals in an interleaved fashion, we would expect these `Z` terms to vanish [1]:
```
0.125 * IIYYIIXY
- 0.125 * IIXXIIXY
- 0.125 * IIXYIIYY
- 0.125 * IIYXIIYY
+ 0.125 * IIXYIIXX
+ 0.125 * IIYXIIXX
+ 0.125 * IIYYIIYX
- 0.125 * IIXXIIYX
```
This is how _immense_ savings can be achieved in terms of circuit depths when using an interleaved ordering.

```python
from qiskit_nature.second_q.circuit.library import PUCCD
from qiskit_nature.second_q.mappers import InterleavedQubitMapper, JordanWignerMapper

mapper = InterleavedQubitMapper(JordanWignerMapper())
p = PUCCD(4, (2, 2), mapper)
for op in p.operators:
    print(op)
```
Observe that one still has many `Z` terms in the resulting qubit operators although all of these should vanish:
```
-0.125 * IYZYIXZY
- 0.125 * IXZYIYZY
- 0.125 * IXZXIXZY
+ 0.125 * IYZXIYZY
- 0.125 * IXZYIXZX
+ 0.125 * IYZYIYZX
+ 0.125 * IYZXIXZX
+ 0.125 * IXZXIYZX
-0.125 * YZZYXZZY
- 0.125 * XZZYYZZY
- 0.125 * XZZXXZZY
+ 0.125 * YZZXYZZY
- 0.125 * XZZYXZZX
+ 0.125 * YZZYYZZX
+ 0.125 * YZZXXZZX
+ 0.125 * XZZXYZZX
-0.125 * IYYIIXYI
- 0.125 * IXYIIYYI
- 0.125 * IXXIIXYI
+ 0.125 * IYXIIYYI
- 0.125 * IXYIIXXI
+ 0.125 * IYYIIYXI
+ 0.125 * IYXIIXXI
+ 0.125 * IXXIIYXI
-0.125 * YZYIXZYI
- 0.125 * XZYIYZYI
- 0.125 * XZXIXZYI
+ 0.125 * YZXIYZYI
- 0.125 * XZYIXZXI
+ 0.125 * YZYIYZXI
+ 0.125 * YZXIXZXI
+ 0.125 * XZXIYZXI
```

### What should happen?

All `Z` terms should vanish like so:
```
0.125 * IIYYIIXY
- 0.125 * IIXXIIXY
- 0.125 * IIXYIIYY
- 0.125 * IIYXIIYY
+ 0.125 * IIXYIIXX
+ 0.125 * IIYXIIXX
+ 0.125 * IIYYIIYX
- 0.125 * IIXXIIYX
0.125 * YYIIIIXY
- 0.125 * XXIIIIXY
- 0.125 * XYIIIIYY
- 0.125 * YXIIIIYY
+ 0.125 * XYIIIIXX
+ 0.125 * YXIIIIXX
+ 0.125 * YYIIIIYX
- 0.125 * XXIIIIYX
0.125 * IIYYXYII
- 0.125 * IIXXXYII
- 0.125 * IIXYYYII
- 0.125 * IIYXYYII
+ 0.125 * IIXYXXII
+ 0.125 * IIYXXXII
+ 0.125 * IIYYYXII
- 0.125 * IIXXYXII
0.125 * YYIIXYII
- 0.125 * XXIIXYII
- 0.125 * XYIIYYII
- 0.125 * YXIIYYII
+ 0.125 * XYIIXXII
+ 0.125 * YXIIXXII
+ 0.125 * YYIIYXII
- 0.125 * XXIIYXII
```

### Any suggestions?

I have a WIP fix locally which performs the permutation on the second-quantization level instead of on the qubit level. I will open a PR to address this soon.

mrossinek:
### Summary

Fixes #1157 

### Details and comments

In doing so I also had to implement the following features:
Closes #954 
Closes #542


mrossinek:
CI failure is unrelated and due to the removal of 3.7 support in Terra's development branch. A separate PR will fix that so I am marking this as ready for review.

qiskit_nature/second_q/mappers/interleaved_qubit_mapper.py
```diff
@@ -38,6 +38,17 @@ class InterleavedQubitMapper(FermionicMapper):
...
 
+    .. warning::
+
+        The builtin two-qubit reduction of the :class:`.ParityMapper` will also not provide correct
+        results when combined with this mapper. Again, this is not asserted so be aware of this
+        pitfall.
+
+    .. note::
+
+        Given the above, you may instead want to use a :class:`.TaperedQubitMapper` to remove
```
woodsp-ibm, line 49: In the output html we have two warnings ahead of this note. `Given the above` seems ambiguous in this context to me - the BKSF warning is basically irrelevant in this regard since BKSF maps to qubits based on interactions, not sites, thus the qubit operator was always different in that regard. Does it in fact make any difference now - I guess it depends on how it builds up the interactions as to whether the re-ordering  makes any difference in the outcome. As it was before, acting on the qubit operator it would have switched things, maybe it still alters the outcome in some way.

image

Maybe just adding a second paragraph to the second warning that suggests if they want reduction to use the TaperedQubitMapper instead. And of course, whether this is worth noting too - where the number of qubits reduced is based on the symmetries it finds rather than always being two based on knowledge of the parity mapper block-mode behavior when its two-qubit reduction is chosen.

qiskit_nature/second_q/mappers/parity_mapper.py
```diff
@@ -40,6 +40,12 @@ class ParityMapper(FermionicMapper):
     (middle and last qubit) because the spin orbitals are ordered in two spin sectors
     (block spin order). Based on the provided number of particles this allows the automatic selection
     of the correct symmetry sector.
+
+    .. warning::
+
+       This additional two-qubit reduction will **not** work in conjunction with the
```
woodsp-ibm, line 46: Should we qualify `will not work` here. It will reduce it by two qubits, won't it. just that the resulting operator is likely not going to be "correct" - so using it the answer is likely to be wrong and hence will not work is more in that sense. It seemingly seems to work in that the code still runs, just that the computation it does is not likely to result in a correct answer.

mrossinek:
> Not sure if this is what you wanted or not. I think it comes out ok so I'll approve things anyway and leave it to you...

It was not intentional, but I agree that it looks okay :+1:

PR #1159, opened 2023-05-05
Terra main vs stable install switch

woodsp-ibm:
### Summary

As Terra main has dropped Python 3.7 support this enables a switch of the install between main or stable and its set at present not to use main (i,e, pip install latest stable)


mrossinek (approved):
Thanks a lot, Steve! :+1:

PR #1162, opened 2023-05-09
Fix language toggle for docs

Eric-Arellano:
### Summary

They were broken because:

1. The template expects `language_label`, not `version_label`
2. jQuery is no longer included by default in Sphinx 6

This also makes some other docs improvements:

* Pins qiskit_sphinx_theme to 1.11 for better stability.
* Removes some outdated template overrides.
* Fixes docs build on macOS by turning off Tox's develop mode
* Adds a `docs-clean` command


javabster:
> the table on the equivalent of [this page](https://qiskit.org/ecosystem/nature/apidocs/qiskit_nature.second_q.html#module-qiskit_nature.second_q) has become a scrollable thing again

I don't think we ever implemented the table scroll changes to make it not scrollable. I believe it was because Steve wanted nature to be consistent with terra but I can't recall exactly. I'm not sure why it's different in the artifact compared to the live page but it seems like theres something broken with the live page generally (e.g. the background colours for the table are also broken). I think the CI artifact is showing the intended behaviour

javabster:
personally I also hate that the tables are scrollable and would like to fix this for all projects at the sphinx theme level, but for terra I know there are much longer tables and I'm not sure they would want the tables to be expanded by default.

mrossinek (approved):
Thanks for the feedback everyone, let's leave this as is then for the time being :+1:

Hopefully the table changes can be done on the theme level at some point in the future :crossed_fingers:

javabster:
fyi @mrossinek I opened an issue to track this scroll styling issue in the sphinx theme: https://github.com/Qiskit/qiskit_sphinx_theme/issues/325

PR #1163, opened 2023-05-09
Turn off CI for forks

Eric-Arellano:
### Summary

Most people agree that GitHub made a mistake by having forks run CI both on the fork and in the upstream repository. It wastes computing resources (which has a real environmental cost!).

### Details and comments

If people really want to run CI in their fork, they can always modify their fork's config to revert this change.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,6 +31,7 @@ concurrency:
 
 jobs:
   Checks:
+    if: github.repository_owner == 'Qiskit'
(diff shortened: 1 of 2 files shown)


mrossinek (approved):
Thanks, Eric! I agree, this is a good idea :+1:

Eric-Arellano:
Thanks! I don't have write access so will need someone to hit merge.

woodsp-ibm (approved):

PR #1166, opened 2023-05-12
Remove unused custom CSS files and Sphinx directives

Eric-Arellano:
No one was using the custom directives `galleryitem`, `customgalleryitem`, `includenodoc`, `customcarditem`. 

After removing this, it's safe to remove `sphinx-gallery` because it's not used. That allows removing `gallery.css`, which is meant to override `sphinx-gallery`'s default styling.

This also removes two CSS files:

* `custom.css`: There is no `.toggle` element anymore
* `style.css`: its 3 rules are holdovers from the Read The Docs theme

`CustomCalloutItemDirective` is used though. It's used in a few other projects too, so I'm likely going to move this to `qiskit_sphinx_theme` in its 1.13 release so that you don't have to copy and paste the code. Tracked by https://github.com/Qiskit/qiskit_sphinx_theme/issues/323.


docs/custom_directives.py
```diff
@@ -3,12 +3,6 @@
 from docutils import nodes
 import re
 import os
-import sphinx_gallery
-
-try:
-    FileNotFoundError
-except NameError:
-    FileNotFoundError = IOError
```
Eric-Arellano, line 11: Leftover Python 2 code, I believe.

woodsp-ibm: Although this is no different than the other repos in that regard. Guess you noticed it here.

Eric-Arellano: > Guess you noticed it here.

Haha precisely. My process has been to go one project at a time cleaning up these CSS files, tracked by this checklist at https://github.com/Qiskit/qiskit_sphinx_theme/issues/262#issuecomment-1544476510. The further along I get in these update-every-single-project-style changes, I usually notice a slightly better way to do things halfway through. Like, activate `qiskit_sphinx_theme` rather than `sphinxcontrib.jquery`. If that better way truly matters, I'll go back and update the original PRs. Otherwise, I leave it for a follow-up PR.

> not that it matters too much I guess.

Yeah, it shouldn't matter. The goal is to get rid of all of this duplication. Tracked by https://github.com/Qiskit/qiskit_sphinx_theme/issues/323.

mrossinek (approved):
Thanks a lot Eric! :+1: 

Do you want this merged now or await some more changes aligned with the other repos?

PR #1175, opened 2023-05-22
Improve docs

### LINKED ISSUE #1055: Document inherited members by default
### What should we add?

Some time in the past, we decided to _not_ document inherited members of our classes by default. The main reason were the excessive build times for the documentation caused by (in particular) the dependencies on `QuantumCircuit`.

The choice of not documenting inherited members has always been a bit problematic and recently resurfaced [here](https://github.com/Qiskit/qiskit-nature/pull/1044#issuecomment-1419735637). For now, I fixed this in that PR specifically for the `mappers` module but we should strive towards a package-wide fix of this.

Qiskit Terra actually introduced significant improvements related to this as part of https://github.com/Qiskit/qiskit-terra/pull/6865 which we should propagate as much as possible to Qiskit Nature. This will ensure consistency across the various Qiskit components.

woodsp-ibm:
### Summary

Similar to Qiskit/qiskit-machine-learning#635 
Resolves #1055

This changes things here, which used to have a couple of templates where inherited methods was the choice and no inheritence the default, over to inherited methods are shown by default and it has to be selected to suppress them - which presently is just done for circuit library classes inheriting from QuantumCircuit. This is also how Qiskit does things.

It fails locally with make spell for me, as I imagine it pulls stuff in that was not before. But I will let it go through CI here once before changing anything in that regard assuming it fails likewise here.

Locally I had deleted the with_inherited rst template and added a the no_inherited one - but it ended up as a rename and change of contents which is fine.

### DIFF
--- a/docs/_templates/autosummary/class.rst
+++ b/docs/_templates/autosummary/class.rst
@@ -21,11 +21,9 @@
    .. autosummary::
       :toctree: ../stubs/
    {% for item in all_attributes %}
-      {%- if item not in inherited_members %}
-        {%- if not item.startswith('_') %}
-            ~{{ name }}.{{ item }}
-        {%- endif -%}
-      {%- endif %}
+      {%- if not item.startswith('_') %}
+          ~{{ name }}.{{ item }}
+      {%- endif -%}
    {%- endfor %}
    {% endif %}
    {% endblock %}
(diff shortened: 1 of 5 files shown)


woodsp-ibm:
FYI. This has an interplay with #1085 where this part in the init file, the `:template:...` line below would be removed, as showing inherited members is now the default.
```
BosonicOp Mappers
+++++++++++++++++++

.. autosummary::
   :toctree: ../stubs/
   :template: autosummary/class_with_inherited_members.rst
   :nosignatures:
   
   BosonicLinearMapper
```

mrossinek (approved):
Thanks a lot, Steve!

I resolved the interplay after #1085 got merged.

In hindsight, maybe merging in the other order would have been preferable as now this backport will have a minor conflict, but its too late for that now :sweat_smile:

PR #1177, opened 2023-05-23
Upgrade to qiskit_sphinx_theme 1.12

Eric-Arellano:
The 1.12 theme release is focused on stability, including CSS fixes. See https://github.com/Qiskit/qiskit_sphinx_theme/releases/tag/1.12.0rc1 for the changelog.

It allows us to de-duplicate custom_directives.py, versionutils.py, and _static/no_image.png.

This PR also:

* Fixes incomplete `pyproject.toml` entry for how to build this distribution. It is highly recommended that all Python projects have this.
* Enables `-j auto` with Sphinx to have parallelism. This should result in a faster docs build.


pyproject.toml
```diff
@@ -1,5 +1,6 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools"]
```
Eric-Arellano, line 2: setuptools' docs used to incorrectly say you should include `wheel`. They've corrected this.

tox.ini
```diff
@@ -79,9 +79,3 @@ deps =
 commands =
   sphinx-build -W -T --keep-going -b gettext docs/ docs/_build/gettext {posargs}
   sphinx-intl -c docs/conf.py update -p docs/_build/gettext -l en -d docs/locale
-
-[pycodestyle]
```
Eric-Arellano, line 83: Unused in this repo

requirements-dev.txt
```diff
@@ -7,13 +7,13 @@ stestr>=2.0.0
 ddt>=1.2.0,!=1.4.0,!=1.4.3
 reno>=3.4.0
 Sphinx>=5.0
-sphinx-design>=0.2.0
+sphinx-design~=0.4.1
```
woodsp-ibm, line 10: What is the rationale on this. I get it with the sphinx theme but its less than clear to me doing it with this.

Eric-Arellano: In general, I think it's helpful to pin docs because it improves code stability. For example, it reduces the risk of CI breaking overnight. But I can revert this to >=0.4.1 if you prefer.

Imo, the ideal is using loose requirements in the requirements.txt file, but then using a lockfile to pin every transitive dependency to an exact version. That way, you only make changes to deps when you intentionally want to. It substantially reduces the risk of things breaking overnight. But you also get the newest versions of deps when you regenerate the lockfile. If you're interested, I'm happy to set this up for you by using `pip-compile`, which is a popular tool to generate ~lockfiles.

woodsp-ibm: >=0.2.0 also works - or it least it always has.

We generally get more breakage from lint or mypy changing than anything else. If its pinned its likely to stay that way. Yes it sometimes CI breaks but at least it makes us keep up to date with the latest - or we temporarily pin it while things are changed if its more involved.

mrossinek (approved):
Thanks, Eric!

woodsp-ibm (approved):

PR #1178, opened 2023-05-25
[WIP] Temp change to get output html zip

woodsp-ibm:
### Summary

**THIS IS JUST FOR DEBUGGING**

 I have temporarily disabled Warnings being treated as errors in order to have it create the doc artifacts here so I can see what is coming out from CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,13 +62,13 @@ test_ci:
 
 spell:
 	python -m pylint -rn --disable=all --enable=spelling --spelling-dict=en_US --spelling-private-dict-file=.pylintdict --ignore=gauopen qiskit_nature test tools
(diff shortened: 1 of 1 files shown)


woodsp-ibm:
The artifact here shows the generated is being all lower case now with hypens replacing non-alpha chars so links do not match. Also as it now lowercase's `Table of Contents` to `table-of-contents` it triggers a css style to hide such a section. In the end, for reasons that we already have a TOC in the right sidebar and that refs to title text are discouraged (fragile around any edits to title text and keeping all refs uptodate with that) #1180 sorts things out better. As this was just to see the outcome of CI to guide things forward it has served its purpose so closing.

PR #1180, opened 2023-05-25
Remove manual Table of Contents in favor of autogenerated Page Table of Contents

Eric-Arellano:
This is a better fix for the problem described in https://github.com/Qiskit/qiskit-nature/pull/1179.

@woodsp-ibm and I realized that it's redundant for us to have a manual Table of Contents section in the docs. qiskit_sphinx_theme will already auto-generate a page table of contents. It's duplicative to have our own manual section too.

image

We suspect that we picked up this habit from old docs that did it before qiskit_sphinx_theme used Pytorch to autogenerate the page table of contents. It's now an out-of-date practice that has a bad user experience & results in the issue in https://github.com/Qiskit/qiskit-nature/pull/1179.

### DIFF
--- a/docs/migration/0.5_b_solving_problems.rst
+++ b/docs/migration/0.5_b_solving_problems.rst
@@ -17,17 +17,6 @@ suggest that you read the `corresponding
 resources <https://qiskit.org/documentation/apidoc/primitives.html>`__
 of the Qiskit Terra documentation instead.
 
-Table of Contents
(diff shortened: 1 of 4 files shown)


mrossinek (approved):
Thanks for fixing this!

woodsp-ibm:
Was not quick enough to remove the backport label. We still have notebooks there right

PR #1185, opened 2023-05-30
Add project urls to setup

woodsp-ibm:
### Summary

Adds project urls to setup so that PyPI project page has more than just a `Homepage` url. This adds additional project urls comparable to qiskit-terra i.e. `Bug Tracker`, `Documentation` and `Source Code`.

As per Qiskit/qiskit-optimization#523

Could be backported, if desired, to add these urls for next bug fix release or wait until next minor release.


mrossinek (approved):
Thanks!

PR #1193, opened 2023-06-05
Fix CI not running due to qiskit-community migration

Eric-Arellano:
This repository was recently moved from `qiskit` to `qiskit-community`, but CI was not updated.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ concurrency:
 
 jobs:
   Checks:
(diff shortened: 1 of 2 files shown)


woodsp-ibm (approved):

mrossinek (approved):
Thanks, Eric! :+1:

mrossinek:
`automerge` is not going to work here because CI will not run at all with the current setup. That is what this PR is fixing after all.
Thus I am forcefully merging this now without the checks having run

PR #1199, opened 2023-06-07
fix: the excited state solvers tutorial

### LINKED ISSUE #1197: ExcitedStatesEigensolver does not return excited states
### Environment

- **Qiskit Nature version**: 0.6.2
- **Python version**:  3.10.6
- **Operating system**: Linux (WSL)

### What is happening?

Hi guys, I was running [this tutorial](https://qiskit.org/ecosystem/nature/tutorials/04_excited_states_solvers.html) on excited state solvers on my local machine and found that the ExcitedStatesEigensolver was not actually returning any excited states. When I ran the tutorial on the IBM cloud, the same error was produced. In fact that tutorial page itself does not have any excited states produced by the numpy eigensolver. 

Upon checking the Japanese (JA) version of the [same tutorial](https://qiskit.org/ecosystem/nature/locale/ja_JP/tutorials/04_excited_states_solvers.html), it seems that the tutorial used to be in fact working.

You can check the tutorial itself, or try running the code from the tutorial and the problem should reproduce itself.

### What should happen?

The ExcitedStatesEigensolver should produce the energies for the excited states of H2. 

### Any suggestions?

_No response_

mrossinek:
### Summary

Closes #1197


woodsp-ibm (approved):
Looks fine to me. I will note that this tutorial is now dependent on the minor tweak for the import in the init file for it at the algorithms level (EvaluationRule). We may not want to republish docs after this is merged to stable until a bug fix release is done otherwise people trying it will have the import error.

mrossinek:
> Looks fine to me. I will note that this tutorial is now dependent on the minor tweak for the import in the init file for it at the algorithms level (EvaluationRule). We may not want to republish docs after this is merged to stable until a bug fix release is done otherwise people trying it will have the import error.

Ah yes that is true. What do you think is best here?
- release 0.6.3
- do a different PR to only fix the tutorial based on the currently stable import and withhold this PR from backporting

PR #1202, opened 2023-06-08
fix: typing compatibility with rustworkx 0.13.0

mrossinek:
### Summary

Makes mypy happy due to the new typing support of rustworkx 0.13.0
So far, this is only a partial typing support which mypy does not seem to pick up on, despite https://github.com/Qiskit/rustworkx/blob/58fdd933cd81c576a3cc02c8a6b7b4e4646ea9c3/rustworkx/py.typed#L1

### Details and comments

The constraint for Python 3.7 is due to https://github.com/Qiskit/rustworkx/issues/893

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -217,7 +217,14 @@ jobs:
           source "$CONDA/etc/profile.d/conda.sh"
           conda activate psi4env
           make mypy
-        if: ${{ !cancelled() }}
+        if: ${{ !cancelled() && matrix.python-version != 3.7 }}
+        shell: bash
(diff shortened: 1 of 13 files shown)


.github/workflows/main.yml
```diff
@@ -217,7 +217,14 @@ jobs:
           source "$CONDA/etc/profile.d/conda.sh"
           conda activate psi4env
           make mypy
```
woodsp-ibm, line 219: I imagine you could have done a simple if else around the make mypy to do it one way or another depending on whether python version was 3.7 or not - akin to what is done in line 176 above. But this does the job too.

PR #1204, opened 2023-06-08
Update urls following repo move

woodsp-ibm:
### Summary

Following the recent move of this repo from `Qiskit` organization to `qiskit-community` this updates all the urls to the new location. The file `theme-variables.jinja` is quite out of date and being unused the conclusion was to remove this.

See qiskit-community/qiskit-optimization#529 for similar

### Details and comments

I will note that there are refs to issues in error messages raised in the code and in reno yaml files. I did not change these as I did not think it worth it and if you ever see these the redirect will work. It can be done though if desired. I did change the one place in the code, an init file, where it is published out as docs.


woodsp-ibm:
I will note that the formatting for that `link_str` went the other way in ML from bracketed over 3 lines to a long single line. Then in Nature it wanted to do it the other way! Admittedly in ML its a bit longer and maybe it prefers a single line since either way its too long but better too long as a single line.

mrossinek (approved):
Thanks a lot, Steve! :+1:

Issue #1208, opened 2023-06-13
`BosonicOp.simplify` behaves incorrectly


mrossinek:
### Environment

- **Qiskit Nature version**: `main` @ c616370a8e2e8c5667ad7fce2b532a893ec81baf
- **Python version**: Any
- **Operating system**: Any

### What is happening?

The `BosonicOp.simplify` method makes a wrong assumption and therefore results in wrongly simplified operators.

### How can we reproduce the issue?

```python
>>> from qiskit_nature.second_q.operators import BosonicOp
>>> op = BosonicOp({'+_0 -_0 -_1 +_0': 1})
>>> op.simplify()
BosonicOp({'+_0 -_1': (2+0j)}, num_modes=2, )
```

### What should happen?

The operator in the example above cannot be simplified further.

### Any suggestions?

The problem is the a wrong assumption built into the `BosonicOp._simplify_label` method.
It results from how we simplify a number operator which in general should be:

$$
\hat{a}^\dagger \hat{a} | n \rangle = n | n \rangle
$$

In the fermionic case, $n$ can be either $0$ or $1$, but in the bosonic case it is indeed general. That is why the simplification we took over from the fermionic case is wrong.
A good example to show this is also the following:

$$
\hat{a}^\dagger \hat{a} \hat{a}^\dagger \hat{a} | n \rangle = n^2 | n \rangle
$$

In our operators, since we do not know what state $|n\rangle$ we apply them to, we cannot simplify this further. Thus, we have two options:

1. do not perform any simplification other than getting rid of zero-coefficient terms
2. implement a construct to specifically handle number operators (e.g. using parameters) and treat these separately all the way through the mapping down to the qubit level

Option 2 is a lot more involved and has many unclear aspects to it. Thus, we will go with option 1.
If at some point in the future we come up with a plan on how to realize option 2, then we can revisit this. But I do not consider this a high priority.

ftroisi:
Yes I'm happy to work on it!

ftroisi:
@mrossinek There's something we haven't really considered about the commutator.
## Issue
So currently the `simplify()` method for `BosonicOp` first performs `normal_order()` and then simplifies the operator.

If we remove the `normal_order()`, then we break the test for the commutator. In fact, currently we have:
```
>>> from qiskit_nature.second_q.operators import BosonicOp
>>> from qiskit_nature.second_q.operators.commutators import commutator
[...]

PR #1210, opened 2023-06-14
Bosonic op bugfix

ftroisi:
### Summary
This PR fixes a bug regarding the bosonic operator. The bug is described in Issue #1208

### DIFF
--- a/qiskit_nature/second_q/operators/bosonic_op.py
+++ b/qiskit_nature/second_q/operators/bosonic_op.py
@@ -489,7 +489,7 @@ class BosonicOp(SparseLabelOp):
         if self.is_parameterized():
             raise ValueError("is_hermitian is not supported for operators containing parameters.")
         atol = self.atol if atol is None else atol
-        diff = (self - self.adjoint()).simplify(atol=atol)
+        diff = (self - self.adjoint()).index_order().simplify(atol=atol)
         return all(np.isclose(coeff, 0.0, atol=atol) for coeff in diff.values())
 
     def simplify(self, atol: float | None = None) -> BosonicOp:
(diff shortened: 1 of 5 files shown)


ftroisi:
@mrossinek little explanation for the unit tests of the commutators. In the subtests with a simplification tolerance I had to change `FermionicOp({"+_0": 1e-10}` to `FermionicOp({"+_0": 1e-08}`. This is due to the fact that now we first perform a `normal_order()`, and then `simplify(atol=0)`.
However, also `normal_order()` simplifies if the coefficient is smaller than the default tolerance (which is 1e-08). I think we should add an argument to the `normal_order()` method to set a tolerance (just like we do for the simplify). then, in the commutators we can do 
```python
(...).normal_order(atol=0).simplify(atol=0)
```

mrossinek:
I would actually argue that `normal_order` (and by extension `index_order`) should _not_ perform any simplification of close-to-zero terms. This makes the methods more atomic and also avoids one more iteration over all terms. Could you include that change as part of this PR?

ftroisi:
Yeah you're right, it makes more sense that they don't do any simplification of those terms

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -432,15 +432,11 @@ def index_order(self) -> BosonicOp:
...
             You can use this method to achieve the most aggressive simplification of an operator
-            without changing the operation order per index. :meth:`simplify` does *not* reorder the
-            terms and, thus, cannot deduce ``-_0 +_1`` and ``+_1 -_0 +_0 -_0`` to be
-            identical labels. Calling this method will reorder the latter label to
-            ``-_0 +_0 -_0 +_1``, after which :meth:`simplify` will be able to correctly collapse
+            without changing the operation order per index. :meth:`simplify` can only reduce terms
+            with identical label, but does *not* reorder them. Thus, cannot deduce ``-_0 +_1`` and
+            ``+_1 -_0`` to be identical labels. Calling this method will reorder the latter label to
+            ``-_0 +_1``, after which :meth:`simplify` will be able to correctly collapse
             these two labels into one.
```
mrossinek, line 439: This example now lacks a bit. Maybe we can rewrite it as follows:
```
            You can use this method to achieve the most aggressive simplification of an operator
            without changing the operation order per index. :meth:`simplify` does *not* reorder the
            terms and, thus, cannot deduce ``-_0 +_1`` and ``+_1 -_0`` to be identical labels.
            Calling this method will reorder the latter label to ``-_0 +_1``, correctly collapsing
            these two labels into one.
```

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -489,7 +485,7 @@ def is_hermitian(self, atol: float | None = None) -> bool:
         if self.is_parameterized():
             raise ValueError("is_hermitian is not supported for operators containing parameters.")
         atol = self.atol if atol is None else atol
-        diff = (self - self.adjoint()).simplify(atol=atol)
+        diff = (self - self.adjoint()).index_order().simplify(atol=atol)
```
mrossinek, line 488:
```suggestion
        diff = (self - self.adjoint()).normal_order().simplify(atol=atol)
```
At least that is what we do in `FermionicOp`. Does this cause any problems?

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -501,76 +497,27 @@ def simplify(self, atol: float | None = None) -> BosonicOp:
...
-        atol = self.atol if atol is None else atol
+        .. note::
 
-        # First, we normal-order the operator
-        ordered_op = self.normal_order()
+            You can use this method to an operator which has two or more identical labels, as it
+            can only reduce terms but it does not perform any reordering. Thus, it cannot deduce
+            ``-_0 +_1`` and ``+_1 -_0``.
+            Another use is to simplify terms whose coefficients are close to zero, up to the
+            specified numerical tolerance.
```
mrossinek, line 506: I don't think this is needed. The simplification of identical labels happens automatically because dictionary keys are unique (and we take care to sum the coefficients).

Thus, the `BosonicOp.simplify` method really only removes close-to-zero (based on `atol`) entries.
It still differs slightly from `.chop` because that will chop real and imaginary phase components individually.
Maybe this is worthwhile explaining in a note here.

qiskit_nature/second_q/operators/bosonic_op.py
```diff
@@ -501,76 +497,27 @@ def simplify(self, atol: float | None = None) -> BosonicOp:
...
-        for label, coeff in ordered_op.items():
-            label, coeff = ordered_op._simplify_label(label, coeff)
-            data[label] += coeff
+        .. note::
+
+            :meth:`simplify` is not allowed to simplify the labels ``+_0 -_0`` or ``-_0 +_0``. The
+            former corresponds to the boson number operator, and when it is applied to a state yields
+            the number of bosons in that state ``n``, not zero. Similarly, the latter yields
+            :math:`1 + n`. As a consequence, the label ``+_0 -_0 -_1 +_0`` will remain untouched by
+            this method.
```
mrossinek, line 514:
```suggestion
            the number of bosons in that state :math:`n`. Similarly, the latter yields
            :math:`1 + n`. As a consequence, the label ``+_0 -_0 -_1 +_0`` will remain untouched by
            this method. This is in contrast to how :meth:`.FermionicOp.simplify` works,
            because it exploits that :math:`n` can be either :math:`0` or :math:`1`.
```

test/second_q/operators/test_bosonic_op.py
```diff
@@ -198,25 +198,25 @@ def test_index_order(self):
             self.assertEqual(bos_op, targ)
 
         with self.subTest("index order + simplify"):
-            orig = BosonicOp({"+_1 -_0 +_0 -_0": 1, "-_0 +_1": 2})
+            orig = BosonicOp({"+_1 -_0": 1, "-_0 +_1": 2})
             bos_op = orig.index_order().simplify()
```
mrossinek, line 202:
```suggestion
            bos_op = orig.index_order()
```
This is actually no longer needed

test/second_q/operators/test_bosonic_op.py
```diff
@@ -198,25 +198,25 @@ def test_index_order(self):
...
+            targ = BosonicOp({"+_0 -_0": 1, "-_0 +_0": 1}, num_modes=1)
             self.assertEqual(simplified_op, targ)
 
         with self.subTest("simplify complex"):
-            bos_op = BosonicOp({"+_0": 1, "+_0 -_0 +_0": 1j}, num_modes=1)
+            bos_op = BosonicOp({"+_0 -_0": 1}) + BosonicOp({"+_0 -_0": 1j})
             simplified_op = bos_op.simplify()
-            targ = BosonicOp({"+_0": (1 + 2j)}, num_modes=1)
+            targ = BosonicOp({"+_0 -_0": (1 + 1j)}, num_modes=1)
             self.assertEqual(simplified_op, targ)
```
mrossinek, line 220: These tests no longer really make sense, because in the first case nothing changes (although we could ofcourse test for that) and in the second case the addition already does the correct logic.

test/second_q/operators/test_bosonic_op.py
```diff
@@ -228,18 +228,21 @@ def test_simplify(self):
         with self.subTest("simplify parameters"):
             bos_op = BosonicOp({"+_0 -_0": self.a, "+_0 -_0 +_0": 1j})
             simplified_op = bos_op.simplify()
-            targ = BosonicOp({"+_0 -_0": self.a, "+_0": 2j})
+            targ = BosonicOp({"+_0 -_0": self.a, "+_0 -_0 +_0": 1j})
```
mrossinek, line 231: Compared to my previous comment, this test still makes sense because it asserts that we do not accidentally simplify the `+_0 -_0 +_0` again, guarding against this bug reappearing.

mrossinek (commented):
I think this has everything we need. Just a few nitpicks.

I do not think we need a reno for this since this is fixing a yet unreleased feature. And you are updating the existing reno for the `BosonicOp` accordingly :+1:

test/second_q/operators/test_bosonic_op.py
ftroisi: I would remove only the second one. As you say, I think that it's useful to test that nothing changes

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: Solved

qiskit_nature/second_q/operators/bosonic_op.py
ftroisi: How about now?

mrossinek (approved):
All good, thanks a lot for the quick fix! :+1:

PR #1212, opened 2023-06-22
Update CODEOWNERS

mrossinek:
### Summary

This adds @ftroisi as a new code owner of the `operators` and `mappers` modules. Welcome aboard! :tada: 

### Details and comments

I am also updating the comments slightly (inline with how Terra has them) and added exceptions to some other folders which we should handle more generally.


woodsp-ibm (approved):
@ftroisi Welcome to Nature!

PR #1213, opened 2023-06-28
Multi-Tensor splitting and stacking utilities

mrossinek:
### Summary

In a recent project I had to implement a lot of array stacking and splitting.
This use case cannot be handle by our `.apply` utilities of `PolynomialTensor` and `ElectronicIntegrals` because:
- in the stacking case the numpy routines expect the arrays to be stacked provided as a list rather than separate positional arguments
- in the splitting case we return multiple tensors instead of one

The latter case would have been possible to implement in a "multi-apply" setting (which I plan to add in the future, too) but splitting arrays is fairly common so I think it deserves its own utility.

### DIFF
--- a/qiskit_nature/second_q/operators/electronic_integrals.py
+++ b/qiskit_nature/second_q/operators/electronic_integrals.py
@@ -469,7 +469,7 @@ class ElectronicIntegrals(LinearMixin):
                 operands. This function must take a sequence of numpy (or sparse) arrays as its
                 first argument. You should use :code:`functools.partial` if you need to provide
                 keyword arguments (e.g. :code:`partial(np.stack, axis=-1)`). Common methods to use
-                here are :meth:`np.hstack` and :meth:`np.vstack`.
+                here are :meth:`numpy.hstack` and :meth:`numpy.vstack`.
             operands: a sequence of ``ElectronicIntegrals`` instances on which to operate.
             validate: when set to False, no validation will be performed. Disable this setting with
                 care!
@@ -521,7 +521,7 @@ class ElectronicIntegrals(LinearMixin):
                 sparse) array as its first input followed by a sequence of indices to split on.
                 You should use :code:`functools.partial` if you need to provide keyword arguments
                 (e.g. :code:`partial(np.split, axis=-1)`). Common methods to use here are
-                :meth:`np.hsplit` and :meth:`np.vsplit`.
+                :meth:`numpy.hsplit` and :meth:`numpy.vsplit`.
             indices_or_sections: a single index or sequence of indices to split on.
             validate: when set to False, no validation will be performed. Disable this setting with
                 care!
--- a/qiskit_nature/second_q/operators/electronic_integrals.py
+++ b/qiskit_nature/second_q/operators/electronic_integrals.py
@@ -463,7 +463,7 @@ class ElectronicIntegrals(LinearMixin):
                 operands. This function must take a sequence of numpy (or sparse) arrays as its
                 first argument. You should use :code:`functools.partial` if you need to provide
                 keyword arguments (e.g. :code:`partial(np.stack, axis=-1)`). Common methods to use
-                here are ``np.hstack`` and ``np.vstack``.
+                here are :meth:`np.hstack` and :meth:`np.vstack`.
             operands: a sequence of ``ElectronicIntegrals`` instances on which to operate.
             validate: when set to False, no validation will be performed. Disable this setting with
                 care!
@@ -509,7 +509,7 @@ class ElectronicIntegrals(LinearMixin):
                 sparse) array as its first input followed by a sequence of indices to split on.
                 You should use :code:`functools.partial` if you need to provide keyword arguments
                 (e.g. :code:`partial(np.split, axis=-1)`). Common methods to use here are
-                ``np.hsplit`` and ``np.vsplit``.
+                :meth:`np.hsplit` and :meth:`np.vsplit`.
             indices_or_sections: a single index or sequence of indices to split on.
             validate: when set to False, no validation will be performed. Disable this setting with
                 care!
--- a/qiskit_nature/second_q/operators/electronic_integrals.py
+++ b/qiskit_nature/second_q/operators/electronic_integrals.py
@@ -436,6 +436,60 @@ class ElectronicIntegrals(LinearMixin):
         return cls(alpha, beta, beta_alpha, validate=validate)
 
     @classmethod
+    def stack(
+        cls,
+        function: Callable[..., np.ndarray | SparseArray | Number],
+        operands: Sequence[PolynomialTensor],
+        *,
+        validate: bool = True,
+    ) -> PolynomialTensor:
+        """TODO."""
+        alpha = PolynomialTensor.stack(function, [op.alpha for op in operands], validate=validate)
+
+        beta: PolynomialTensor = None
+        if any(not op.beta.is_empty() for op in operands):
+            # If any beta-entry is non-empty, we have to perform this computation.
+            # Empty tensors will be populated with their alpha-terms automatically.
+            beta = PolynomialTensor.stack(
+                function,
+                [op.alpha if op.beta.is_empty() else op.beta for op in operands],
+                validate=validate,
+            )
+
+        beta_alpha: PolynomialTensor = None
+        if all(not op.beta_alpha.is_empty() for op in operands):
+            # We can only perform this operation, when all beta_alpha tensors are non-empty.
+            beta_alpha = PolynomialTensor.stack(
+                function, [op.beta_alpha for op in operands], validate=validate
+            )
+        return cls(alpha, beta, beta_alpha, validate=validate)
+
+    def split(
+        self,
+        function: Callable[..., np.ndarray | SparseArray | Number],
+        indices_or_sections: int | Sequence[int],
+        *,
+        validate: bool = True,
+    ) -> list[PolynomialTensor]:
+        """TODO."""
+        alphas = self.alpha.split(function, indices_or_sections, validate=validate)
+
+        if self.beta.is_empty():
+            betas = [None] * len(alphas)
+        else:
+            betas = self.beta.split(function, indices_or_sections, validate=validate)
+
+        if self.beta_alpha.is_empty():
+            beta_alphas = [None] * len(alphas)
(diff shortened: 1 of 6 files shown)


mrossinek:
I rebased this on top of #1214 for the time being. Once that is merged (and CI fixed accordingly except for Neko) I will rebase back on `main` :+1:

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -435,6 +435,104 @@ def apply(
...
+        if self.beta.is_empty():
+            betas = [None] * len(alphas)
+        else:
+            betas = self.beta.split(function, indices_or_sections, validate=validate)
+
+        beta_alphas: list[PolynomialTensor | None]
+        if self.beta_alpha.is_empty():
+            beta_alphas = [None] * len(alphas)
+        else:
+            beta_alphas = self.beta_alpha.split(function, indices_or_sections, validate=validate)
```
ftroisi, line 529: According to the definitions of `alpha`, `beta` and `beta_alpha` I imagine that the former two typically have the same length, but that's not the case the `beta_alpha`. If so, why are the same `indices_or_sections` used to split all three PolynomialTensors? Is it possible to have a case where we ask to spilt the Tensor in an out-of-bound index?

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -435,6 +435,104 @@ def apply(
...
+        """Exposes the :meth:`qiskit_nature.second_q.operators.PolynomialTensor.split` method.
+
+        This behaves identical to the ``split`` implementation of the ``PolynomialTensor``, applied
+        to the :attr:`alpha`, :attr:`beta`, and :attr:`beta_alpha` attributes of the provided
+        ``ElectronicIntegrals`` operands.
+
+        This method is special, because it handles the scenario in which any operand has a non-empty
+        :attr:`beta` attribute, in which case the empty-beta attributes of any other operands will
+        be filled with :attr:`alpha` attributes of those operands.
+        The same applies to the :attr:`beta_alpha` attributes.
```
ftroisi, line 504: Perhaps some unit tests should be added for this method. I know that this method heavily relies on the same one in the PolynomicTensor (which is tested), but still it would be good to test it. Especially due to this special behavior

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -435,6 +435,104 @@ def apply(
...
+        """Exposes the :meth:`qiskit_nature.second_q.operators.PolynomialTensor.stack` method.
+
+        This behaves identical to the ``stack`` implementation of the ``PolynomialTensor``, applied
+        to the :attr:`alpha`, :attr:`beta`, and :attr:`beta_alpha` attributes of the provided
+        ``ElectronicIntegrals`` operands.
+
+        This method is special, because it handles the scenario in which any operand has a non-empty
+        :attr:`beta` attribute, in which case the empty-beta attributes of any other operands will
+        be filled with :attr:`alpha` attributes of those operands.
+        The same applies to the :attr:`beta_alpha` attributes.
```
ftroisi, line 455: See comment about unit tests in split method

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -435,6 +435,104 @@ def apply(
...
+
+        Args:
+            function: the stacking function to apply to the internal arrays of the provided
+                operands. This function must take a sequence of numpy (or sparse) arrays as its
+                first argument. You should use :code:`functools.partial` if you need to provide
+                keyword arguments (e.g. :code:`partial(np.stack, axis=-1)`)
+            operands: a sequence of ``ElectronicIntegrals`` instances on which to operate.
+            validate: when set to False, no validation will be performed.
```
ftroisi, line 463: Please add the warning present in PolynomialTensors: `Disable this setting with care!`

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -435,6 +435,104 @@ def apply(
...
+        be filled with :attr:`alpha` attributes of those operands.
+        The same applies to the :attr:`beta_alpha` attributes.
+
+        Args:
+            function: the splitting function to use. This function must take a single numpy (or
+                sparse) array as its first input followed by a sequence of indices to split on.
+                You should use :code:`functools.partial` if you need to provide keyword arguments
+                (e.g. :code:`partial(np.split, axis=-1)`)
+            indices_or_sections: a single index or sequence of indices to split on.
+            validate: when set to False, no validation will be performed.
```
ftroisi, line 512: Please add the warning present in PolynomialTensors: `Disable this setting with care!`

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -435,6 +435,104 @@ def apply(
...
+        ``ElectronicIntegrals`` operands.
+
+        This method is special, because it handles the scenario in which any operand has a non-empty
+        :attr:`beta` attribute, in which case the empty-beta attributes of any other operands will
+        be filled with :attr:`alpha` attributes of those operands.
+        The same applies to the :attr:`beta_alpha` attributes.
+
+        Args:
+            function: the splitting function to use. This function must take a single numpy (or
+                sparse) array as its first input followed by a sequence of indices to split on.
```
ftroisi, line 508: Perhaps we should explicit about the two numpy function that will be most commonly used (`hsplit`, `vsplit`).
This applies also to the stack method

ftroisi (commented):
Overall it looks good to me, but I'm not an expert on PolynomialTensors so a second pair of eyes would be good

qiskit_nature/second_q/operators/electronic_integrals.py
mrossinek: Why do you say that the lengths will not be the same? All these matrices will have an "edge length" of the number of molecular orbitals. The `alpha` and `beta` PolynomialTensors will usually have 1-body and 2-body terms (typically stored in `+-` and `++--`) while `beta_alpha` only has a 2-body matrix.
However, (at least how I use these methods) I am mostly interested in the manipulation of MO coefficients which only have 1-body terms in `alpha` and `beta` and no mixing terms present.

Nonetheless, the length should always match.

It will be up to the user to ensure they split on valid indices there is nothing we can do in this regard here.

qiskit_nature/second_q/operators/electronic_integrals.py
mrossinek: Yeh I was kind of in a hurry. I will see if I can add some unittests similar to how it is done in the `einsum` routine.

qiskit_nature/second_q/operators/polynomial_tensor.py
```diff
@@ -626,6 +626,112 @@ def apply(
...
+            import numpy as np
+            from qiskit_nature.second_q.opertors import PolynomialTensor
+            rand_a = np.random.random((2, 2))
+            rand_b = np.random.random((2, 2))
+            a = PolynomialTensor({"+-": rand_a})
+            b = PolynomialTensor({"+": np.random.random(2), "+-": rand_b})
+
+            # np.hstack
+            ab_hstack = PolynomialTensor.stack(np.hstack, [a, b], validate=False)
+            print(ab_hstack == PolynomialTensor({"+-": np.hstack([a, b], validate=False)}))  # True
```
woodsp-ibm, line 652: How much is `validate` actually valid to use in these scenarios. For instance this will fail if validate is True. The only way I can see it passing validate is if you stacked here 2 arrays that individually themselves would fail. I note all the examples  here set it false, as do the unittests. Should we rather default it to false on the stack method or remove the parameter all together so internally tensors are created with validation off?

woodsp-ibm: Fair enough - but then would it be better to say something a little different about `validate` in stack and split, than in the constructor? Maybe for split one is more likely going from something that would not validate to something that will. But stating something like the sentiment you expressed above in the stack method, maybe even as a Note to highlight things could be helpful.  By default its likely to fail as per that example - we may intend for this to be the case to ensure a user really knows what they are doing, but some text to that effect may help. For split I guess it usually ought to pass right.

qiskit_nature/second_q/operators/electronic_integrals.py
ftroisi: Ok I think I got the point. Should be alright then!

qiskit_nature/second_q/operators/polynomial_tensor.py
mrossinek: Splitting is just as likely to lead to mismatching shapes. I will add a note to both of the docstrings.

qiskit_nature/second_q/operators/electronic_integrals.py
woodsp-ibm: In the output from the documentation built here these unfortunately did not end up as links - just highlighted - since the url for say hsplit is https://numpy.org/doc/stable/reference/generated/numpy.hsplit.html I wonder whether here if its says `numpy.hsplit` whether that resolves fine

qiskit_nature/second_q/operators/electronic_integrals.py
```diff
@@ -435,6 +438,116 @@ def apply(
...
+            When stacking arrays this will likely lead to array shapes which would fail the shape
+            validation check. This is considered an advanced use case which is why the user is left
+            to disable this check themselves, to ensure they know what they are doing.
+
+        Args:
+            function: the stacking function to apply to the internal arrays of the provided
+                operands. This function must take a sequence of numpy (or sparse) arrays as its
+                first argument. You should use :code:`functools.partial` if you need to provide
+                keyword arguments (e.g. :code:`partial(np.stack, axis=-1)`). Common methods to use
+                here are :meth:`numpy.hstack` and :meth:`numpy.vstack`.
```
woodsp-ibm, line 472: Still not working - I think it was my bad, sorry - I said `meth` but for these they should be `func` I tried locally in PolyTensor just adding a couple of func refs and it comes out fine  so should here - I did try with np as well but it does not seem to resolve that in this context so ``:func:`numpy.hstack` `` etc and it should all be good.

woodsp-ibm (approved):

PR #1222, opened 2023-08-13
remove deploy documentation to /documentation

1ucian0:
Follow up on https://github.com/qiskit-community/qiskit-nature/pull/1103

The documentation is correctly deployed in https://qiskit.org/ecosystem/nature/ and the https://qiskit.org/documentation/nature redirect is working. So, no need to upload to /documentation anymore.

PR #1223, opened 2023-08-14
Drop Python 3.7 support

woodsp-ibm:
### Summary

CI nightly has been failing now for a while under 3.7 around mypy and rustworkx - the latter got updated to 0.13.1 and the constraint we had no longer worked. While I could have amended constraint, as Qiskit has now dropped support for 3.7 this drops it from Nature too, rendering the rustworkx constraint mute (which I removed), and means CI should now pass.

### Details and comments

A reno notes the upgrade needed for the user.

- [x] Once this passes we should amend the CI branch rules accordingly.

This should also allow #1222 to go forwards once merged.

I had meant to look for sys version checks and forgot (now done) - CI flagged a line no longer being tested and it turns out it was a version check. So I just removed this and what was needed for 3.7. This was protocols.py - I mused over changing the eigenstate result not to be a string - but there are other places where this could be done so Iet it be. I had noticed as the string was missing a period - which I fixed; I let that be the extent of the change otherwise.

### DIFF
--- a/.github/actions/install-psi4/action.yml
+++ b/.github/actions/install-psi4/action.yml
@@ -1,6 +1,6 @@
 # This code is part of Qiskit.
 #
-# (C) Copyright IBM 2021, 2022.
+# (C) Copyright IBM 2021, 2023.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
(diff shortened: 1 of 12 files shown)


mrossinek (approved):
Thanks a lot, Steve! :+1:

PR #1226, opened 2023-08-28
Update file headers in regards of Qiskit text

woodsp-ibm:
### Summary

Similar change as per qiskit-community/qiskit-finance#289

Changes all headers from `This code is part of Qiskit` to `This code is part of a Qiskit project` (via search replace in all project files) and the code that checks for this (which the search/replace in the project files did anyway).

Copyright dates were amended accordingly. (`make copyright` fixes the dates as needed - but I guess it does not check github action/workflow files and knowing that from doing Finance I changed them manually during the search/replace).


qiskit_nature/second_q/problems/base_problem.py
```diff
@@ -9,7 +9,7 @@
 # Any modifications or derivative works of this code must retain this
 # copyright notice, and modified files need to carry a notice indicating
 # that they have been altered from the originals.
-# This code is part of Qiskit.
+# This code is part of a Qiskit project.
```
ftroisi, line 12: This looks like a duplicate, we already have at line 1

ftroisi (commented):
Just a very minor thing

qiskit_nature/second_q/problems/base_problem.py
woodsp-ibm: Good eyes - I had not noticed when I did the change (search/replace) - it was rather a lot of files though! It may as well be removed.

mrossinek (approved):
Thanks for taking care of this, Steve! :+1:

PR #1227, opened 2023-08-28
Remove PyQuante2 from CI and optionals

woodsp-ibm:
### Summary

I noticed that PyQuante2 was still being installed by CI. As the driver was removed, and is no longer in the codebase, this install is no longer required. Updated the workflow so as to remove this.

Also removed the entry from the `optionals.py` file as that's not needed either. 

I left the `pyquante` entry in the custom dict since we have docs that mention this for when it was removed so we probably need to preserve it there for spelling to pass.


mrossinek (approved):
Good catch, thanks a lot! :+1:

PR #1229, opened 2023-08-31
Switch from qiskit-terra to qiskit

woodsp-ibm:
### Summary

Similar to qiskit-community/qiskit-machine-learning#680

Switches over to `qiskit` from `qiskit-terra` 

* requirements.txt updated
* CI updated for install main dependencies
* CONTRIBUTING.MD links etc updated. Plus I removed/simplified some bits which are no longer correct and may have led to confusion down the line.

Removed Neko integration/workflow

I also updated a couple of places that referred to Terra (I was searching for Terra) in all the files.  I noted an occurence in adapt_vqe.rst but figured that will be changed to use qiskit_algorithms and sorted at that point.


CONTRIBUTING.md
```diff
@@ -244,10 +225,8 @@ particular will be located at `docs/_build/html/release_notes.html`
 Please see the [Installing Qiskit Nature from
 Source](https://github.com/qiskit-community/qiskit-nature#installation)
 
-Note: Nature depends on Terra, and has optional dependence on Aer and IBM Q Provider, so
-these should be installed too. The main branch of Nature is kept working with those other element
-main branches so these should be installed from source too following the instructions at 
-the same location
+Note: Nature depends on Qiskit, and has optional dependence on Aer so
+should be installed too.
```
mrossinek, line 229:
```suggestion
Note: Nature depends on Qiskit, and has optional dependence on Aer so
these should be installed too.
```

We could be even more specific and say that Qiskit has to be installed and Aer may be, but I think that is not necessary.

CONTRIBUTING.md
```diff
@@ -265,12 +244,12 @@ The test suite can be run from a command line or via your IDE. You can run `make
 run all unit tests. Another way to run the test suite is to use
 [**tox**](https://tox.readthedocs.io/en/latest/#). For more information about using tox please
 refer to
-[Terra CONTRIBUTING](https://github.com/Qiskit/qiskit-terra/blob/main/CONTRIBUTING.md#test)
+[Qiskit CONTRIBUTING](https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#test)
 Test section. However please note Nature does not have any
-[online tests](https://github.com/Qiskit/qiskit-terra/blob/main/CONTRIBUTING.md#online-tests)
+[online tests](https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#online-tests)
```
mrossinek, line 249: This link does not exist any more. I am not even sure what this was originally referring to.

CONTRIBUTING.md
```diff
@@ -265,12 +244,12 @@ The test suite can be run from a command line or via your IDE. You can run `make
...
 nor does it have
 [test skip
- options](https://github.com/Qiskit/qiskit-terra/blob/main/CONTRIBUTING.md#test-skip-options).    
+ options](https://github.com/Qiskit/qiskit/blob/main/CONTRIBUTING.md#test-skip-options).
```
mrossinek, line 252: And this is said to not exist, but we actually do have slow tests now which can be triggered to run (or not) based on the state of an environment variable.

mrossinek (commented):
A few observations on the updates to `CONTRIBUTING.md` but otherwise this LGTM

CONTRIBUTING.md
woodsp-ibm: We had the same sort of text in all the repos and I left Aer there since its often used in tutorials and unit tests. I did not check explicitly as my main goal was just to switch to to Qiskit but with the repos moving to community I cleaned up the contributing where it seemed outdated but did not want to spend a bunch of time.

Since Nature explicitly depends on Qiskit installing Nature will install that by default. Hence the only optional install is Aer that you would need to do separately which is why it was worded like that. I had toyed with dropping the while text there but I figured using the Aer primitives people of do so I left it - and its in our requirements dev too.

CONTRIBUTING.md
woodsp-ibm: I admit I did not check over these links - it is the same across the app repos as the contributing came from a common source. Its likely that since this was originally done, based on what Terra had, that these not longer exist. I believe this was around if the ibm provider was used - some tutorials had an optional section that you could enable. I do not recall what might have been in terra long ago. We could leave this for now - rather than make it different from the other app repos. I really do not know if/how this file might change - I just wanted to change any refs from Terra and removed some text that was not how things are done in order to remove any confusion that may have caused going forwards,

CONTRIBUTING.md
mrossinek: All of that is okay, but without some word (like `these` from my suggestion) this sentence does not work, does it?
```
and has optional dependence on Aer so should be installed too.
```
This is the current (sub-)sentence.

CONTRIBUTING.md
```diff
@@ -244,10 +225,8 @@ particular will be located at `docs/_build/html/release_notes.html`
 Please see the [Installing Qiskit Nature from
 Source](https://github.com/qiskit-community/qiskit-nature#installation)
 
-Note: Nature depends on Terra, and has optional dependence on Aer and IBM Q Provider, so
-these should be installed too. The main branch of Nature is kept working with those other element
-main branches so these should be installed from source too following the instructions at 
-the same location
+Note: Nature depends on Qiskit, and has optional dependence on Aer so
```
woodsp-ibm, line 228:
```suggestion
Note: Nature depends on Qiskit, and has an optional dependence on Aer so
```
Guess it was getting later in the day when I did this this is what I had elsewhere in Opt for instance, I guess I missed adding the an here

> Note: Optimization depends on Qiskit, and has an optional dependence on Aer, so should be installed too.

mrossinek: Would it not need to be `so that should/could be installed too`?

mrossinek: Okay sorry, I stressed the wrong part here. The missing `"that"` is what is throwing me off.

woodsp-ibm: Ah ok `that` or `it` etc - I guess the lack of it did not - I read that as implied but it did have these there before. Ok I updated the suggestion above. Hopefully that works.

mrossinek (approved):

Issue #1091, opened 2023-03-07
PySCF-based tests segfault on Mac OS with Python 3.8


mrossinek:
### Environment

- **Qiskit Nature version**: `main` @ https://github.com/Qiskit/qiskit-nature/commit/ec47efea9856b6e4e597ac8b7c55f020ae4b10cc
- **Python version**: 3.8
- **Operating system**: MacOS

### What is happening?

Three tests starting failing in CI: https://github.com/Qiskit/qiskit-nature/commit/ec47efea9856b6e4e597ac8b7c55f020ae4b10cc

### How can we reproduce the issue?

Run the tests from `test/algorithms/excited_state_solvers/test_excited_states_solvers.py`

### What should happen?

The tests should pass.

mrossinek:
It seems like this issue is related to PySCF because when it is not installed, the CI works fine. This will need to be investigated further in the future.

mrossinek:
It appears as though Psi4-based tests are affected in the same way. For now, we will fully disable 3.8 on MacOS.

rryoung98:
Just to add, I'm running python 3.9 with Mac M1 Ventura 3.1

```  × Running setup.py install for pyscf did not run successfully.
  │ exit code: 1
  ╰─> [8 lines of output]
      scipy>1.1.0 may crash when calling scipy.linalg.eigh. (Issues https://github.com/scipy/scipy/issues/15362 https://github.com/scipy/scipy/issues/16151)
      running install
      /Users/rickyyoung/opt/anaconda3/envs/qiskit/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
        warnings.warn(
      running build
      running build_ext
      cmake -S/private/var/folders/z0/w8wxtq691l55jcqndh2hzmfh0000gn/T/pip-install-h8rl5j7b/pyscf_51ed2725a8494bd0afa6490860b7313d/pyscf/lib -Bbuild/temp.macosx-11.1-arm64-cpython-39
      error: command 'cmake' failed: No such file or directory
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: legacy-install-failure

× Encountered error while trying to install package.
╰─> pyscf
```

woodsp-ibm:
@rryoung98 See #689 in which I also linked an issue in PySCF to build M1 wheels which is still open

[...]

PR #1238, opened 2023-09-04
ci: test minimal Python version on Mac OS

mrossinek:
### Summary

I would like to make sure that we run our unittests for the lowest and highest supported Python versions on Mac OS and Windows, respectively.
We recently changed the tested Python version on Mac OS due to #1091. I would like to see if this is still needed.

### Details and comments

Since #1163 I cannot test this in my fork, hence the draft PR.


mrossinek:
Seems like this is working again, so I am removing the draft status :+1:

PR #1239, opened 2023-09-04
feat: improve the implementation of the AngularMomentum operator

mrossinek:
The AngularMomentum operator is actually just the $S^2$ operator. However, in the code prior to this commit this was really not easy to follow.

Instead, this commit changes the implementation to rely on the sub-operators which the $S^2$ operator consists of. More specifically, it adds generator functions for the $S^+$ and $S^-$ ladder operators as well as the $S^x$, $S^y$, and $S^z$ operators. All of these may also be useful to users who want to evaluate these separately.

Unittests are added to assert the relations of all of these operators.

### DIFF
--- a/docs/apidocs/qiskit_nature.second_q.properties.s_operators.rst
+++ b/docs/apidocs/qiskit_nature.second_q.properties.s_operators.rst
@@ -1,0 +1,28 @@
+﻿
+
+s_operators
+=============================================
+
+.. automodule:: qiskit_nature.second_q.properties.s_operators
+
+   
+   
+   .. rubric:: Functions
+
+   .. autosummary::
+   
+      s_minus_operator
+      s_plus_operator
+      s_x_operator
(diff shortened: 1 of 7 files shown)


woodsp-ibm (approved):

docs/apidocs/qiskit_nature.second_q.properties.s_operators.rst
```diff
@@ -0,0 +1,28 @@
+﻿
+
+s_operators
+=============================================
+
+.. automodule:: qiskit_nature.second_q.properties.s_operators
+
+
```
woodsp-ibm, line 8: This file has a lot of blank lines - not that it really matters, but...

mrossinek: I think some others do and am willing to ignore this since it has no effect in any case

PR #1240, opened 2023-09-04
feat: improve the `SparseLabelOp.register_length` handling

### LINKED ISSUE #1190: Improve handling of `SparseLabelOp.register_length`
### What should we add?

Currently, the `SparseLabelOp.register_length` property dictates that it can be `int | None`.
The idea is that operator instances do not have a fixed length, but instead treat this as a lower bound.
A user may set a specific value for this length, or it is inferred from the maximum index.

However, this latter point is currently a bit problematic, because this is done implicitly during the `_validate_keys` call, a side effect that is in an unnatural place and which does _not_ occur, when `copy=False` or `validate=False` is used during construction.

Furthermore, the `register_length` attribute of operators in used in various places where it being `None` will actually cause problems. Here are some examples which (to my surprize) `mypy` does not inform us about:
- https://github.com/Qiskit/qiskit-nature/blob/c80eb0e7e50cd7c351e1a8cc72a26ad296d006db/qiskit_nature/second_q/mappers/linear_mapper.py#L56 (similar in the `LogarithmicMapper`)
- https://github.com/Qiskit/qiskit-nature/blob/c80eb0e7e50cd7c351e1a8cc72a26ad296d006db/qiskit_nature/second_q/operators/sparse_label_op.py#L447 (this one won't outright break, but might give false positives)
- https://github.com/Qiskit/qiskit-nature/pull/1189#discussion_r1212752232

As a way forward, I propose the following:
- remove `None` from the possible return value of `SparseLabelOp.register_length`
- extract the lower bound setting from `_validate_keys` into a separate utility method (not ideal because we need to loop over the labels again. Maybe we can find a more performant solution here)
- during the property getter of `register_length`, if the underlying value currently is `None`, run the utility over the labels to infer the actual lower bound (or we do this during `__init__` but I would prefer the lazy way). Once gotten, we should store this lower bound in the internal attribute to avoid rerunning this lower bound finding multiple times.

mrossinek:
### Summary

Closes #1190

### Details and comments

I chose *not* to cache the value which gets inferred during the `register_length` property getter.
The main reason is that this should not be too much of a performance issue (and a user can set it if they wish to improve performance) and invalidating the cache would become annoying to deal with and litter the code unnecessarily.


releasenotes/notes/improve-register-length-handling-cb55f644da136828.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    The :attr:`.SparseLabelOp.register_length` attribute (and by extension that of its subclasses, too)
+    can no longer take ``None`` as its value. However, the attributes which this one might rely on
+    (e.g. :attr:`.FermionicOp.num_spin_orbitals` or :attr:`.BosonicOp.num_modes`) can remain ``None``.
+    This ensures that the lower-bound behavior works as intended.
```
woodsp-ibm, line 7: I see the Attributes: docstring for these in the two classes for some reason do not have the type in them - now it would be int | None I should think. It would be good to add the typehint.

Do you imagine any cases where users would need to be wary with existing code - if they relied on getting a value back before with None now I imagine code could have problems. We could add an upgrade section and call this aspect out to say users might need to upgrade code using these attributes to now cater to None?

mrossinek: Actually, I do not think the previous code every truely return `None` :thinking: 
Unless they used `copy=False` but then other cases would likely have been broken for them.

I will update the type hints though

PR #1241, opened 2023-09-04
feat: implements the PUCCSD ansatz

mrossinek:
### Summary

Adds a new convenience subclass of the `UCC` ansatz. Namely, the
spin-symmetry-adapted ansatz, `PUCCSD`, which includes single and double
excitations while always pairing the excitations such that both, the number of
particles and the total spin, will be preserved.


qiskit_nature/second_q/circuit/library/ansatzes/puccsd.py
```diff
@@ -0,0 +1,172 @@
...
+    ) -> None:
+        # pylint: disable=unused-argument
+        """
+
+        Args:
+            num_spatial_orbitals: The number of spatial orbitals.
+            num_particles: The tuple of the number of alpha- and beta-spin particles.
+            qubit_mapper: The :class:`~qiskit_nature.second_q.mappers.QubitMapper` or
+                :class:`~qiskit_nature.second_q.mappers.QubitConverter` instance (use of the latter
+                is deprecated) which takes care of mapping to a qubit operator.
```
woodsp-ibm, line 68: With the typehint and the types listed in the opposite order when looking at the html, with the typehints standing out more I had to do a double take on this as at first it seemed the wrong way around with the latter reference.

image

I think I would almost argue with a new class just to document this for the QubitMapper - it may well work at present for QubitConvertor too while that code remains in the parent, but for "new code", to introduce with already deprecated  aspects...

mrossinek: Ah this was a copy-paste mistake. I will remove `QubitConverter` :+1:

woodsp-ibm (approved):

woodsp-ibm:
@Mergifyio requeue

woodsp-ibm:
@mergifyio refresh

PR #1243, opened 2023-09-04
feat: use PySCF unfolding routines for performance

mrossinek:
### Summary

The symmetry folding and unfolding routines listed below could become severe
bottlenecks for larger system sizes. Now, when PySCF is installed, its routine will be
used which is significantly faster.

- `unfold_s4_to_s1`
- `unfold_s8_to_s1`
- `unfold_s8_to_s4`
- `fold_s1_to_s4`
- `fold_s1_to_s8`
- `fold_s4_to_s8`


mrossinek:
Something is going on when the integrals come from an unrestricted-spin calculation. I will need to dig into this a little bit, but this may take some time.

Issue #1224, opened 2023-08-22
Fermi-Hubbard Model block/interleaved inconsistency


ialsina:
### Environment

- **Qiskit Nature version**: 343d2901
- **Python version**: 3.9.16
- **Operating system**: RHEL 8.8

### What is happening?

Since the introduction of `InterleavedQubitMapper` (#1046) the `FermiHubbardModel` class doesn't behave consistently with this feature, producing an interleaved-encoded hamiltonian when used together with `JordanWignerMapper`, and a block-encoded hamiltonian when used together with `InterleavedQubitMapper`.

### How can we reproduce the issue?

```

>>> from qiskit_nature import settings
>>> from qiskit_nature.second_q.hamiltonians import FermiHubbardModel
>>> from qiskit_nature.second_q.hamiltonians.lattices import LineLattice
>>> from qiskit_nature.second_q.mappers import JordanWignerMapper, InterleavedQubitMapper
>>> settings.use_pauli_sum_op = False

>>> lattice = LineLattice(2)
>>> model = FermiHubbardModel(
    lattice.uniform_parameters(
        uniform_interaction=-1.,
        uniform_onsite_potential=0.
    ),
    onsite_interaction=1.
)
>>> hamiltonian = model.second_q_op().simplify()
>>> print(hamiltonian)

FermionicOp({'+_0 -_2': (-1+0j), '-_0 +_2': (1+0j), '+_1 -_3': (-1+0j), '-_1 +_3': (1+0j), '+_0 -_0 +_1 -_1': (1+0j), '+_2 -_2 +_3 -_3': (1+0j)}, num_spin_orbitals=4, )

>>> mapper = JordanWignerMapper()
>>> print(mapper.map(hamiltonian)) # Block encoding

SparsePauliOp(['IYZY', 'IXZX', 'YZYI', 'XZXI', 'IIII', 'IIZI', 'IIIZ', 'IIZZ', 'ZIII', 'IZII', 'ZZII'],
              coeffs=[-0.5 +0.j, -0.5 +0.j, -0.5 +0.j, -0.5 +0.j,  0.5 +0.j, -0.25+0.j,
 -0.25+0.j,  0.25+0.j, -0.25+0.j, -0.25+0.j,  0.25+0.j])

>>> interleaved_mapper = InterleavedQubitMapper(mapper)
>>> print(interleaved_mapper.map(hamiltonian)) # Interleaved encoding

[...]

PR #1245, opened 2023-09-05
Unify orbital ordering of `FermiHubbardModel` with `ElectronicEnergy`…

ialsina:
### Summary

This PR addresses #1224. This commit changes the orbital ordering of the `FermiHubbardModel` to make it consistent with the block-ordering of the `ElectronicEnergy` class.

### Details and comments

With this change, `JordanWignerMapper` produces block encoding when mapping the second quantization hamiltonian of `FermiHubbardModel`. `InterleavedQubitMapper` acting on `JordanWignerMapper` produces interleaved encoding. See full description in the original issue.


mrossinek (changes requested):
While I think that this change is - in principle - good as it is, you will need to address the following things before we can merge this:

- [ ] fix the currently existing unittests which should fix the CI, too
- [ ] add a release note (see `CONTRIBUTING.md` on how to do that)

And as a bonus:

- [ ] It would be nice if you could also add a new unittest which asserts that using `InterleavedQubitMapper(JordanWignerMapper())` is actually reproducing the original behavior.

Issue #1169, opened 2023-05-16
`utils.test_linalg` fails when `use_symmetry_reduced_integrals` is set True


woodsp-ibm:
In looking to remedy the amount of deprecation messages in the unit test logs I had also changed the settings over to the defaults I would expect us to move to in the next version. So I had updated the settings file, e.g.

```
        self._use_symmetry_reduced_integrals: bool = True
```
This however causes `utils.test_linalg` to fail (works fine when set False), as follows:

```
  File "test/utils/test_linalg.py", line 104, in test_double_factorized_error_threshold_max_vecs
    diag_coulomb_mats, orbital_rotations = double_factorized(
  File "qiskit_nature/utils/linalg.py", line 312, in double_factorized
    reshaped_tensor = np.reshape(two_body_tensor, (n_modes**2, n_modes**2))
  File "<__array_function__ internals>", line 5, in reshape
  File "qiskit_nature/second_q/operators/symmetric_two_body.py", line 159, in __array_function__
    return super().__array_function__(func, types, args, kwargs)
  File "qiskit_nature/second_q/operators/tensor.py", line 179, in __array_function__
    ret = func(*new_args, **kwargs)
  File "<__array_function__ internals>", line 5, in reshape
  File "python3.9/site-packages/numpy/core/fromnumeric.py", line 298, in reshape
    return _wrapfunc(a, 'reshape', newshape, order=order)
  File "python3.9/site-packages/numpy/core/fromnumeric.py", line 57, in _wrapfunc
    return bound(*args, **kwds)
ValueError: cannot reshape array of size 231 into shape (36,36)
```

PR #1142, opened 2023-04-26
Remove the deprecated code stack from prior to version 0.5

mrossinek:
### Summary

This is THE PR which removes all of the legacy code from how the stack prior to version 0.5 used to work.

### Details and comments

- [x] Reno with upgrade section for this removal

PR #1248, opened 2023-09-06
refactor: remove all deprecated code

### LINKED ISSUE #1196: Switch to using new deprecation utils provided by Qiskit Terra
### What should we add?

Qiskit Terra recently added new deprecation utility decorators which are more powerful than ours (for example they also add deprecation hints to the docstrings).

We should switch to using those and deprecate our utils which will become obsolete by this move.

@Eric-Arellano you voiced an interest but I cannot assign you to this without a comment or tag. So here you go :wink:

mrossinek:
### Summary

This commit removes all deprecated code. Even though this is done a bit earlier than originally announced under the Qiskit deprecation policy, we are taking this step now since this project has become part of a community effort.
As a direct consequence, new codeowners will soon join the team behind Qiskit Nature. Rather than handing over a codebase that includes pieces of code which need attention in the near future (and even continued support until then) we are removing these parts early.

As a consequence, if you still require parts of this code, we suggest you stick to Qiskit Nature 0.6 until you can transition based on the provided migration guides.

This commit also removes the deprecation utilities which are superseded by the ones provided by Qiskit directly (see qiskit.utils.deprecation). Thus, this PR closes #1196

### Details and comments

- [x] add a release note with a prelude section to explain the above


mrossinek:
This also closes #1169

mrossinek:
Since there is too much to go through in detail in this PR, here are a few noteworthy details:

- I removed the deprecation utils (see prior comments)
- I removed all occurrences of `opflow` (this includes `PauliSumOp`)
- I removed all factories and their usage
- I removed the `QubitConverter` and all its usage
- I removed the `runtime` module
- I removed all the `settings` attributes which become unused as a result of all these removals
- please have a look at the suggested `prelude` in the release note

woodsp-ibm:
What do you want to do about the AdaptVQE howto `docs/howtos/adapt_vqe.rst` It refers to settings that this removes. I get that it will also need to be updated around qiskit_algorithms but this PR removes some of what it refers to, so if we want it in sync with this then maybe update it here - and again for the algorithms change later.

mrossinek:
> What do you want to do about the AdaptVQE howto `docs/howtos/adapt_vqe.rst` It refers to settings that this removes. I get that it will also need to be updated around qiskit_algorithms but this PR removes some of what it refers to, so if we want it in sync with this then maybe update it here - and again for the algorithms change later.

Nice catch! I did not think to look through the tutorials and how-tos for additional usage of settings or other removed aspects. Will do that now :+1:

mrossinek:
I have pushed two additional changes:

- removing all further usage of setting attributes which have been removed. Grepping for `settings` now only shows results for `optimize_einsum` and in the migration guides (where they should remain)

- I removed the release note which was added in #1142 because it is now obsolete

qiskit_nature/settings.py
```diff
@@ -14,67 +14,21 @@
...
-
-    pass
-
 
 class QiskitNatureSettings:
     """Global settings for Qiskit Nature."""
 
     def __init__(self) -> None:
         self._optimize_einsum = True
+        self._use_symmetry_reduced_integrals: bool = True
```
woodsp-ibm, line 23: Is there any reason to keep this now, along with the setter/getter? Its not used anywhere (and not covered by testing) and any user setting it will make no difference. If it was a question well we did not deprecate this or anything - even if a user did set this in their code the code would not fail after such a removal - it would just monkey patch that attribute here. Reading it would but then seems highly unlikely anyone would have use this and its a simply fix to anyones code had they done it.

woodsp-ibm (approved):

PR #1251, opened 2023-09-07
Add modularity example, tutorial, re-add readme test

ElePT:
### Summary
Improves readme with a new code example using primitives and re-adds test for readme code examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ estimator = Estimator(options={"shots": int(1e4), "seed": 42})
 # define ansatz and initial point
 ansatz = EfficientSU2(qubit_op.num_qubits)
 num_params = ansatz.num_parameters
-initial_point = [2.95891852, 1.46946667, 3.34553262, 
+initial_point = 2 * np.pi * np.random.random(num_params)
                  1.96029912, 1.47424821, 5.75416389,
                  3.0753125,  4.2122225,  0.16412572, 
                  4.63598071, 1.99559646, 1.73394948,
--- a/README.md
+++ b/README.md
@@ -81,9 +81,6 @@ from qiskit.circuit.library import EfficientSU2
 from scipy.optimize import minimize
 import numpy as np
 
-# set seed for reproducibility
-np.random.seed(42)
-
 driver = PySCFDriver(
     atom="H 0 0 0; H 0 0 0.735",
     basis="sto3g",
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -81,6 +81,9 @@ from qiskit.circuit.library import EfficientSU2
 from scipy.optimize import minimize
 import numpy as np
 
+# set seed for reproducibility
+np.random.seed(42)
```
woodsp-ibm, line 85: This is really the legacy way and not the way to do things nowadays. I would do one of two things. The simplest would be to set some fixed initial point and not generate a random one. The other, if you want to stick with this sort of thing, would be to initialize an rng here with a seed
```
rng = np.random.default_rng(42)
```
and use this rng later to get the random values from.

README.md
```diff
@@ -113,7 +110,17 @@ estimator = Estimator(options={"shots": int(1e4), "seed": 42})
 # define ansatz and initial point
 ansatz = EfficientSU2(qubit_op.num_qubits)
 num_params = ansatz.num_parameters
-initial_point = 2 * np.pi * np.random.random(num_params)
+initial_point = [2.95891852, 1.46946667, 3.34553262,
```
woodsp-ibm, line 113: I don't know - technically its not invalid but it seems to takeover the example. How about

`initial_point = [0.1] * ansatz.num_parameters`

Yes you can make it more random, but maybe that's just as good for the sample

mrossinek:
```suggestion
initial_point = 2 * np.pi * np.random.random(num_params)
```

mrossinek: Oops, sorry I thought this was a different PR... Nevermind :sweat_smile:

woodsp-ibm: Seems like its in an odd state now. I would go with with some simple init point as I described above - one could use all zeros but I tend to avoid that as things used to get stuck there in the past.

woodsp-ibm:
This is a more general comment - I can see a tutorial using the blocks at different levels, I am just not convinced about this being the main sample on the readme. So why....

Well Nature takes some form of problem from the user and translates it to something that can be solved on quantum. Here you start with one of the drivers with a molecular spec, but as youe tutorial points out the problem can be built from different inputs, eg from a QCSchema from a third party code etc. Now the translation to something to be solved, in the app stacks also comes with some interpret to take the quantum result and "interpret" that to form the answer to the original problem given.

In the case you show the problem we solve on quantum using an estimator, but is only part of the original problem. What is computed is the electronic energy part only - this has confused people in the past when doing things more manually like this. 

The getting started sample shows it using the GroundStateSolver - its results is interpreted since thats part of its workflow. The next level down, and people often do this, is just as here where you get the operators, but then they are passed into one of the algorithms Nature knows, that it would use in conjunction with a GroundStateSolver - say VQE. That gives the electronic ground state part but the result object from VQE can be used with the problem to interpret it and get the complete ground state energy etc. If you use your own custom algorithm, as it done, if you want the same answer you have to more do the interpret yourself and know more about whats going on.

And with the GroundStateSolver, and even with VQE if you pass all the operators from the problem other aspects like spin, dipole etc are computed which can help determine whether the state you got to was good as well. You can do that with the Estimator too of course but you end up needing to know/understand more to build the answer. Add transformers and it becomes more as these can extract other energies so what you are computing the min eigenvalue for is whatever part of the original problem remains.

I don't know whether one can show in your tutorial the GroundStateSolver way - or refer to it. Then perhaps used VQE and interpret the result, that to me would be the next level down. And finally go all the way to a custom algo. If you wanted the outcome interpreted you could show creating a mineigsolverresult from it (unless we wanted to find some way to have a mechanism to interpret just the estimated value passed in simply).

So for me the sample more shows the extent of the modularity we have where if you solve it your own way you have to deal with interpreting the result too. Its really to that end I am just not sure of having it so front and center. Maybe say it computes the electronic part and leave it at that , maybe add in the nuclear repulsion energy which can be obtained. Maybe like the getting started having some statement there about being able to use Nature at different levels of modularity and linking to the tutorial since that has a bunch of explanation around things. It is modular you can do all these things - but the finer grain you go the more work you do yourself that we already built out somewhere. Its a choice, you may get additional flexibility/capability doing this, for sure its usually more work and requires you to know more about what you are doing and whats going on.

ElePT:
Let's take the discussion offline, I will close this PR and open one with just the tutorial, and Max has added a new disclaimer for the tutorials section to help add context #1255.

PR #1252, opened 2023-09-07
Switch from `qiskit.algorithms` to `qiskit_algorithms`

### LINKED ISSUE #1250: Migrate from `qiskit.algorithms` to `qiskit_algorithms`
### What should we add?

The latest release of Qiskit has deprecated the `qiskit.algorithms` module.
The entire module has been extracted in its state at the time into the new [`qiskit-algorithms` repository](https://github.com/qiskit-community/qiskit-algorithms).

We should switch all of our use of `qiskit.algorithms` to the new location and add `qiskit-algorithms` as a required dependency.

When doing so, we need to be mindful of `isinstance` checks as these would break.
In order to preserve backwards-compatibility with `qiskit.algorithms` we could add some string-comparison based on class names as a temporary solution. Once `qiskit.algorithms` is actually gone, these can be removed again.

mrossinek:
### Summary

This replaces all usage of the deprecated `qiskit.algorithms` module with the new `qiskit_algorithms` one provided by the [corresponding community project](https://github.com/qiskit-community/qiskit-algorithms).

Closes #1250 

### Details and comments

Right now, this will still fail CI because I need to handle cases where `isinstance` checks are going to fail. I am using CI to tell me where this occurs.
What remains to be done is:
- [x] fix the `isinstance` check cases
- [x] update all unittests to use `qiskit_algorithms`
- [x] update all docs - tutorials, howtos, getting started sample etc


mrossinek:
I don't understand why `mypy` is failing here. Will need to dig into this tomorrow...

qiskit_nature/second_q/algorithms/__init__.py
```diff
@@ -78,7 +78,7 @@
 
 Initial Points
 ++++++++++++++
-When using variational algorithms such as the :class:`~qiskit.algorithms.minimum_eigensolvers.VQE`
+When using variational algorithms such as the :class:`~qiskit_algorithms.minimum_eigensolvers.VQE`
```
woodsp-ibm, line 81:
```suggestion
When using variational algorithms such as the :class:`~qiskit_algorithms.VQE`
```
For any docs references they are published from `qiskit_algorithms` directly as they have the convenience imports back at that level too in qiskit_algorithms. So that in qiskit_algorithms refers to the new primitive based VQE (the old QI one does not exist there so there was no name conflict in that regard any more)

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
```diff
@@ -24,10 +24,10 @@
...
 from scipy import linalg
 
-from qiskit.algorithms.eigensolvers import EigensolverResult
-from qiskit.algorithms.list_or_dict import ListOrDict as ListOrDictType
-from qiskit.algorithms.minimum_eigensolvers import MinimumEigensolver
-from qiskit.algorithms.observables_evaluator import estimate_observables
+from qiskit_algorithms.eigensolvers import EigensolverResult
+from qiskit_algorithms.list_or_dict import ListOrDict as ListOrDictType
+from qiskit_algorithms.minimum_eigensolvers import MinimumEigensolver
+from qiskit_algorithms.observables_evaluator import estimate_observables
```
woodsp-ibm, line 30: I will just comment here since this is a bit different - the eigensolvers and minimum_eigensolvers, yes drop them, The others, well I can see estimate_observables in importable direct from qiskit_algorithms  rather than from the module, but I do not see that for ListOrDict and actually neither are in the algorithm docs from what I can see.

qiskit_nature/second_q/algorithms/excited_states_solvers/excited_states_eigensolver.py
```diff
@@ -16,7 +16,7 @@
 
 import logging
 
-from qiskit.algorithms.eigensolvers import Eigensolver
+from qiskit_algorithms.eigensolvers import Eigensolver
```
woodsp-ibm, line 19:
```suggestion
from qiskit_algorithms import Eigensolver
```
While you can import from the folders, as you would have to do in qiskit.algorithms since it still had the QI based ones imported as conveniences at that level, in qiskit_algorithms this can be dropped. I would argue its better to use the suggestion as algos got reorganized in the past and that always worked. And if you look it up in the docs its at that place too.

qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
mrossinek: I think I will leave the `estimate_observables` for now, too. The others I will drop in accordance with your other comments, too

mrossinek:
CI of commit c07c3e00c868be91c1e4a19036af7791f15aa36c has successfully passed. This indicates that, even though the `qiskit_nature` module has been fully switched over to use `qiskit_algorithms`, the deprecated module `qiskit.algorithms` is still supported (since the `tests` have not been switched over yet).

In the next commits, I will update the docs and tests to use the new `qiskit_algorithms`. We will not test the working support of `qiskit.algorithms` going forward.

mrossinek:
I have addressed the previous comment and also updated all occurrences of `algorithm_globals`. See df0137c5c982682e0469f0840af61093ac3826b6 for more details on the latter.

woodsp-ibm:
One other thing - this also uses qiskit.utils validation eg like this in places `from qiskit.utils.validation import validate_min` this should get changed to `from qiskit_algorithms.utils.validation import validate_min` as the validation.py file, originally done in Aqua for algorithms will be deprecated/removed from Qiskit.

woodsp-ibm (approved):
LGTM thx!

Issue #1258, opened 2023-09-24
from_polynomial_tensor constructor creates 0d array instead of scalar for constant coefficient


grossardt:
### Environment

- **Qiskit Nature version**: 0.7.0
- **Python version**: 3.10.12
- **Operating system**: Linux

### What is happening?

`from_polynomial_tensor` constructor for

- `BosonicOp`
- `FermionicOp`
- `SpinOp`
- `VibrationalOp`

produces constant part (empty string key) with coefficient a 0d numpy array (should be number, as for the rest of coefficients)

### How can we reproduce the issue?

```
from qiskit_nature.second_q.operators import BosonicOp, PolynomialTensor
import numpy as np
data = {'': 1, '+': np.array([1,0])}
t = PolynomialTensor(data)
b = BosonicOp.from_polynomial_tensor(t)
print(b._data)
```

Output:

```
{'': array(1), '+_0': 1}
```

### What should happen?

Output should be:

```
{'': 1, '+_0': 1}
```

### Any suggestions?

In line 246 of `bosonic_op.py` (and equivalent lines for `fermionic_op.py`, `spin_op.py`, `vibrational_op.py`):

```
data[""] = tensor[key]
```

cast rhs to scalar, e.g. by applying sum method to 0d tensor:

```
data[""] = tensor[key].sum()
```

PR #1259, opened 2023-09-24
Fix issue1258

grossardt:
### Summary

Fixes issue #1258 

For BosonicOp, FermionicOp, SpinOp, VibrationalOp, changed the line that takes care of the empty string coefficient (constant part of operator) in `from_polynomial_tensor` constructor to assign a scalar number rather than 0d array.

### Details and comments

PolynomialTensor stores all coefficients as tensors, including the empty string contribution, which is consistent. The SparseLabelOps have a scalar number coefficient per label string. When constructed from PolynomialTensor, coefficients are assigned from the scalar elements of the corresponding tensor. However, the constant term (empty string) is dealt with separately, assigning the 0d tensor:

```
data[""] = tensor[key]
```

To convert this also to scalar, this is replaced by:

```
data[""] = tensor[key].sum()
```

### DIFF
--- a/releasenotes/notes/fix-from-polynomial-tensor-constant-9df7f63be4a4c819.yaml
+++ b/releasenotes/notes/fix-from-polynomial-tensor-constant-9df7f63be4a4c819.yaml
@@ -1,6 +1,7 @@
 ---
 fixes:
   - |
-    When using :meth:`.SparseLabelOp.from_polynomial_tensor` constructor for one of :class:`.BosonicOp`, :class:`.FermionicOp`,
-    :class:`.SpinOp`, or :class:`.VibrationalOp`, the coefficient for the constant term was a 0d Tensor object
-    rather than a number. This has been fixed.
+    When using :meth:`.SparseLabelOp.from_polynomial_tensor` constructor for one of
+    :class:`.BosonicOp`, :class:`.FermionicOp`, :class:`.SpinOp`, or :class:`.VibrationalOp`, the
+    coefficient for the constant term was a 0d Tensor object rather than a number. This has been
+    fixed.
--- a/releasenotes/notes/fix-from-polynomial-tensor-constant-9df7f63be4a4c819.yaml
+++ b/releasenotes/notes/fix-from-polynomial-tensor-constant-9df7f63be4a4c819.yaml
@@ -1,6 +1,6 @@
 ---
 fixes:
   - |
-    When using `from_polynomial_tensor` constructor for one of `BosonicOp`, `FermionicOp`,
-    `SpinOp`, or `VibrationalOp`, the coefficient for the constant term was a 0d Tensor object
+    When using :meth:`.SparseLabelOp.from_polynomial_tensor` constructor for one of :class:`.BosonicOp`, :class:`.FermionicOp`,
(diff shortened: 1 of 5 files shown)


mrossinek (commented):
Interesting approach, thanks for fixing this!

Before merging, please also [add a release note](https://github.com/qiskit-community/qiskit-nature/blob/main/CONTRIBUTING.md#release-notes) which explains the bug and states that it has been fixed. Please focus on this being an end-user centric document, so try to keep it from being too technical.

grossardt:
Sorry, I misinterpreted the CONTRIBUTING.md, assuming that fixes like this don't count as 'end user facing' and therefore don't need release notes. I added a short release note now, I hope this is appropriate in length and detail.

woodsp-ibm:
Maybe this might logically be better to convert to scalar https://numpy.org/doc/stable/reference/generated/numpy.ndarray.item.html ? While sum works you did have to comment to explain what its really doing.

releasenotes/notes/fix-from-polynomial-tensor-constant-9df7f63be4a4c819.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    When using `from_polynomial_tensor` constructor for one of `BosonicOp`, `FermionicOp`,
+    `SpinOp`, or `VibrationalOp`, the coefficient for the constant term was a 0d Tensor object
```
mrossinek, line 5:
```suggestion
    When using :meth:`SparseLabelOp.from_polynomial_tensor` constructor for one of :class:`.BosonicOp`, :class:`.FermionicOp`,
    :class:`.SpinOp`, or :class:`.VibrationalOp`, the coefficient for the constant term was a 0d Tensor object
```
I suggest we make these names proper links as shown above.

If we do not, we would still need to fix the verbatim markup to the RST style rather than Markdown one (i.e. use double backticks).

mrossinek (commented):
I prefer the `.item()` method for converting to scalar. I actually did not know about this, so thanks a lot, Steve, for pointing this out! :+1:

releasenotes/notes/fix-from-polynomial-tensor-constant-9df7f63be4a4c819.yaml
grossardt: You van tell that I am new at this ;). Thank you!

grossardt: PS: not sure if the 100 char line length limit applies to release notes, but fixed that just in case.

mrossinek (approved):
Thank you for your contribution! :+1:

releasenotes/notes/fix-from-polynomial-tensor-constant-9df7f63be4a4c819.yaml
mrossinek: > PS: not sure if the 100 char line length limit applies to release notes, but fixed that just in case.

No, this limit only applies to Python files. But its good to keep these files legible in any case :+1:

PR #1261, opened 2023-09-27
docs: improve the documentation of SparseLabelOp.simplify

### LINKED ISSUE #1260: FermionicOp simplify fails to do its task
### Environment

- **Qiskit Nature version**: qiskit-nature==0.5.2
- **Python version**: 3.10.10
- **Operating system**: macos

### What is happening?

Calling simplify on a FermionicOp doesn't yield a fully simplified representation of the operator.
Calling normal order first seems to be a work around for this issue, but it makes simplifying SparseLabelOp in a generic manner impossible.

Simplify for Fermionic op should likely make use of normal ordering to perform simplification. The current implementation doesn't work.
This is similar to issue #901, but it's not that simplify is not agressive enough, but that it fail entirely to simplify the fermionic operator.

```
from qiskit_nature.second_q.hamiltonians import FermiHubbardModel
from qiskit_nature.second_q.hamiltonians.lattices import LineLattice,BoundaryCondition
from qiskit_nature.second_q.operators import commutators

#Create a hamiltonian, Hubbard model because that's what i have on hand.
U = 4
mu = 2
t=1
L=2
line_lattice = LineLattice(num_nodes=L, boundary_condition=BoundaryCondition.OPEN)
Hamiltonian = FermiHubbardModel(line_lattice.uniform_parameters(uniform_interaction=t,uniform_onsite_potential=-mu),onsite_interaction=U).second_q_op()

#secondary operator
C0 = FermionicOp({ "+_0": 1 },    num_spin_orbitals=4)

#create a more complicated fermionic operator
F = commutators.commutator(Hamiltonian,commutators.commutator(Hamiltonian,C0))
print(len(F)) #prints 106
print(len(F.simplify())) #prints 106
print(len(F.normal_order())) #prints 6
```

### What should happen?

simplify should manage to reduce the number of label in F as much or more than normal_order

### Any suggestions?

aplying normal ordering could be part of simplify's implementation.

mrossinek:
### Summary

Closes #1260

### DIFF
--- a/qiskit_nature/second_q/operators/fermionic_op.py
+++ b/qiskit_nature/second_q/operators/fermionic_op.py
@@ -497,6 +497,35 @@ class FermionicOp(SparseLabelOp):
         return all(np.isclose(coeff, 0.0, atol=atol) for coeff in diff.values())
 
     def simplify(self, atol: float | None = None) -> FermionicOp:
+        """Simplify the operator.
+
+        The simplifications implemented by this method should be:
+        - to eliminate terms whose coefficients are close (w.r.t. ``atol``) to 0.
+        - to combine the coefficients which correspond to equivalent terms
+
+        .. note::
+
+            :meth:`simplify` should be used to simplify terms whose coefficients are close to zero,
+            up to the specified numerical tolerance. It still differs slightly from :meth:`chop`
+            because that will chop real and imaginary part components individually.
+
+        .. note::
+
+           The meaning of "equivalence" between multiple terms depends on the specific operator
+           subclass. As a restriction this method is required to preserve the order of appearance of
+           the different components within a term. This avoids some possibly unexpected edge cases.
+           However, this also means that some equivalencies cannot be detected. Check for other
+           methods of a specific subclass which may affect the order of terms and can allow for
+           further simplifications to be implemented. For example, check out :meth:`index_order`.
+
+        This method returns a new operator (the original operator is not modified).
+
+        Args:
+            atol: Absolute numerical tolerance. The default behavior is to use ``self.atol``.
+
+        Returns:
+            The simplified operator.
+        """
         atol = self.atol if atol is None else atol
 
         data = defaultdict(complex)  # type: dict[str, _TCoeff]
--- a/qiskit_nature/second_q/operators/fermionic_op.py
+++ b/qiskit_nature/second_q/operators/fermionic_op.py
@@ -501,7 +501,7 @@ class FermionicOp(SparseLabelOp):
 
         The simplifications implemented by this method should be:
         - to eliminate terms whose coefficients are close (w.r.t. ``atol``) to 0.
-        - to combine the coefficients which correspond to equivalent terms
+        - to combine the coefficients which correspond to equivalent terms (see also the note below)
 
         .. note::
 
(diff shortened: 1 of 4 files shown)


woodsp-ibm:
In FermionicOp there is a note in `index_order` method that pertains to simplify. Would it be better in conjunction with this to more have a note in `simplify` so if you look at the subclass `simplify` docs you have an idea of the other methods that can be called first to potentially allow simplify to further reduce things. At present, both it and VibrationalOp inherit the SparseLabelOp simplify docstring that was updated here - so it would mean adding something for them - which BosonicOp already seems to do as it has further details there pertinent to that sub-class.

mrossinek:
I tried to find a solution which does not require duplicating the `simplify` docstring into the subclasses in order to append to it.
If you feel this solution is not sufficient, we can do that though.

woodsp-ibm:
I certainly would not suggest duplicating the entire thing - of course the args bit one has too and a sentence describing it. This is pretty much what BosonicOp.simplify() has plus some extra specific notes for that operator - it does not inherit the parent class docstring unlike FermionicOp or VibrationalOp. It is a bit extra work to do this for simplify but it could save the user going around other methods in the class to see if they would help - like index_order for example may do. I'll leave it to you - I'll approve this since in improves what it states - anything further like I was suggesting can always be done later since its "just docs".

woodsp-ibm (approved):

mrossinek:
I actually had overlooked that `BosonicOp.simplify` already has a custom docstring. I will take another look on how to improve things :+1:

qiskit_nature/second_q/operators/fermionic_op.py
```diff
@@ -497,6 +497,35 @@ def is_hermitian(self, atol: float | None = None) -> bool:
         return all(np.isclose(coeff, 0.0, atol=atol) for coeff in diff.values())
 
     def simplify(self, atol: float | None = None) -> FermionicOp:
+        """Simplify the operator.
+
+        The simplifications implemented by this method should be:
+        - to eliminate terms whose coefficients are close (w.r.t. ``atol``) to 0.
+        - to combine the coefficients which correspond to equivalent terms
```
ftroisi, line 504: The simplify method for the `FermioniOp` does a bit more than that (at least compared with the one in the `BosonicOp`). Have a look at the closed issue #1208 
I don't know if wanted the have this included in the concept of equivalence, but perhaps it's worth specifying it better

mrossinek: Well that is because the meaning of "equivalence" is different in the `FermionicOp` where we can simplify terms like `+_0 -_0 +_0` to become `+_0`.

Or do you mean something more than that? I tried to use the example from the `index_order` docstring further down to clarify this.

ftroisi: Ok I see
Yeah I guess we can include it in the meaning of "equivalence" for different operators. However, I think that if one reads the docstring for `FermionicOp.simplify` does not understand immediately that `+_0 -_0 +_0 becomes `+_0`. Like it's not a trivial thing

mrossinek: Let me try to make that more clear

ftroisi: Looks better now, thanks!

ftroisi (approved):
Looks fine to me! Thanks for updating these docs

PR #1265, opened 2023-10-12
Used trusted publisher for PyPi

woodsp-ibm:
### Summary

This PR follows Qiskit/rustworkx#1001 to update the release CI workflow to use PyPI's trusted publisher mechanism.


mtreinish (approved):
This LGTM, I'm planning to hopefully migrate the pypi account on Wednesday, so as long as you aren't planning any releases before then we can merge this whenever (it only will work after the migration though).

mrossinek (approved):

PR #1266, opened 2023-11-01
Update CI PSI4 install

woodsp-ibm:
### Summary

Nightly tests seem to sometimes be failing around PSI4 with Windows and Python 3.8. CI is using Windows 2019 (not latest) and an older 1.6 PSI4. Newer versions of PSI4 are out there which also now support 3.11 so it can be installed there now too. I think part of using WIndows-2019 was that we used to build Aer and that was done in that platform.

Anyway this does two things, first changes to `windows-latest` instead of `windows-2019` (presently latest is windows-2022) and it removes the PSI4 1.6 specific version install to install whatever is the stable release, as well as installing it across the board (i.e. remove platform/version restrictions)

I will note the Windows and the default (Ubuntu) sections of the PSI4 install action now are the same, but I left them in separate blocks for now. I am unsure if the openmp updates are needed for Mac, but it seemed that could be done anyway so I left them. I want to see how this change fairs before thinking about possibly collapsing the blocks - though it does no harm to leave them as-is in case things need differentiating still (or in the future) - which is really why I did not already do this.

**_Update_**: In the end the openmp updates that were needed for the version Mac 3.8 was using is now not needed. The conda install was taken from https://psicode.org/installs/v182/ where its the same for all platforms. This works fine and now 3.11 CI includes PSI4 (for all platforms) where it did not before. Currently its installing 1.8.2 across the board versus 1.7 for ubuntu and 1.6 for windows & mac as things were. I simplified the psi4 install action to remove the if/then based on platform since the contents ended up all the same.

### DIFF
--- a/.github/actions/install-psi4/action.yml
+++ b/.github/actions/install-psi4/action.yml
@@ -31,10 +31,10 @@ runs:
         elif [[ "${{ inputs.os }}" == "macos-latest" ]]; then
           echo 'installs psi4 stable release'
           conda install -y psi4 python=${{ inputs.python-version }} -c conda-forge/label/libint_dev -c conda-forge
-          echo 'update intel-openmp'
-          conda update -y -c anaconda intel-openmp
-          echo 'update llvm-openmp'
-          conda update -y -c conda-forge llvm-openmp
+#          echo 'update intel-openmp'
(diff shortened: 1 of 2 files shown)


mrossinek (approved):
Thank you Steve!

Issue #1269, opened 2023-11-11
Inconsistent handling of class attributes in docs and missing attribute descriptions


grossardt:
I noticed that class attributes are not documented very well, take e.g. [`FermionicOp`](https://qiskit.org/ecosystem/nature/stubs/qiskit_nature.second_q.operators.FermionicOp.html#qiskit_nature.second_q.operators.FermionicOp):

- `num_spin_orbitals` is listed in the class docstring under an `Attributes:` header. In the html docs this is then displayed directly under the previous headline 'Iteration' without any indication that this is now a listing of attributes.
- further down in the docs there is a section 'Attributes' that lists `atol`, `rtol`, and `register_length` which seem to be gathered by Sphinx from the `@property` decorated methods of the parent classes. However, the docstrings of these attributes are not shown.

I am not completely sure if this is a bug or somehow intentional. Also not sure if it can just be fixed via the Sphinx configuration, but my feeling is that the docstrings themselves should be amended?

Comparing with the qiskit docs, attributes typically seem to be defined directly after the class docstring with a type declaration and followed by their docstring (and mostly without assigning a value), e.g. for the `FermionicOp` we would have something like:

```
class FermionicOp(SparseLabelOp):
    r"""N-mode Fermionic operator.
    ...
    """

    num_spin_orbitals: int
    """The number of spin orbitals"""

    atol: float
    """Absolute numerical tolerance used by the class methods"""
```

mrossinek:
Thanks for pointing this out! Indeed, the handling of `@property` attributes and publicly directly accessible attributes is a bit inconsistent, especially when it comes to how they are displayed in the docs.
In the past, we used to only have `@property` attributes but over the past few releases we have started using pubic attributes directly, if the `@property` getter/setter methods are trivial.

What you see now, is likely also different to before, because the Sphinx theme which we use for our docs has been changed significantly. I will bring this issue to the team's attention who is working on the theme because I believe they would be interested in improving this :+1:

Issue #582, opened 2022-02-25
JKMN / Neven / Ternary tree qubit mapping


Qottmann:
### What should we add?

It would be very useful to have the ternary tree qubit mapping (sometimes also referred to as the JKMN or Neven mapping after the authors) from this paper [Optimal fermion-to-qubit mapping via ternary trees with applications to reduced quantum states learning](https://arxiv.org/abs/1910.10746).

Qottmann:
Actually, [this repository](https://github.com/matteoacrossi/adapt_ic-povm/) has it implemented in qiskit. The relevant file is [here](https://github.com/matteoacrossi/adapt_ic-povm/blob/main/neven.patch) with all the diffs, should be just a matter of copy-paste as far as I can tell :)

matteoacrossi:
Hi @Qottmann, I'm one of the authors of the repository you cited. We wrote the code for a older version of qiskit, and I'm interested in the development of this Issue.

woodsp-ibm:
Hi @matteoacrossi as you may have noticed the code has been refactored/changed since the original chemistry in Aqua. You can see the mappers are standalone objects here, and the second quantization FermionicOp is rather different than the similarly named object from the past. Things were to me more modular so like adding a new type mapper should just be a new mapper object now unlike in the past. Maybe you have already looked at the code here and know this.

In regards of looking at this contributions are welcomed, please take a look at the [CONTRIBUTING.md](https://github.com/Qiskit/qiskit-nature/blob/main/CONTRIBUTING.md) file which has some info for nature, also more generally https://qiskit.org/documentation/contributing_to_qiskit.html which applies to Qiskit overall as these may be helpful.

 If you have any questions please feel free to ask.

grossardt:
Hi, @rabinswosti and I used a (very rudimentary) version of tenary qubit mapping recently and we were interested in implementing this properly as a mapper for qiskit-nature. What is the status on this? 
We both have very little to no experience contributing to qiskit, but would be interested in this as a small project and happy to be assigned to it, if no one else is actively working on it.

mrossinek:
You are welcome to start working on this! To my knowledge, no work is currently going on in this direction.

That said, you may have noticed that Qiskit Nature was moved into the qiskit-community organization. A more detailed announcement on what this means will come with the 0.7 release but in short this means that there will be some changes to the code-owners and that (at least currently) the bandwidth for reviewing and guiding PRs is kept to a minimum.

[...]

PR #1270, opened 2023-11-12
Add MajoranaOp class

### LINKED ISSUE #1257: Implement Majorana Operator
### What should we add?

Implement a standalone MajoranaOp (incl. a MajoranaOp.from_fermionic_op constructor).

As discussed in #582 the MajoranaOp could be the basis for the implementation of Majorana mappers such as ternary tree.
Adding this as a separate issue so the first pull request can refer to it. (Hope this is ok. I am working on this, so happy to be assigned.)

grossardt:
### Summary

Adds a new class `.second_q.operators.MajoranaOp` implementing a Majorana fermion operator, along with a `M̀ajoranaOp.from_fermionic_op` constructor method to convert a `FermionicOp` into a `MajoranaOp`. Fixes #1257 

`M̀ajoranaOp` is implemented as a `SparseLabelOp`, closely following the implementation of `FermionicOp`. The syntax for operator labels is `'_0 _1'` for $\gamma_0 \gamma_1$ etc.

#### Label syntax

The reasoning for choosing `'_0 _1'` as a label style was a compromise between compatibility (`MajoranaOp` should also be a `SparseLabelOp` with a syntax close to those of `FermionicOp` and `BosonicOp`), readability, and sparsity. Technically, a Majorana operator is defined by just a list of indices, but any non-string format would be incompatible with the other operators. I was considering something like `'$_0 $_1'` with some character $ in order to have the closest possible format to the other operators, but since there is no canonical choice for the $ symbol ($\gamma$ would be, but is problematic as a non-ASCII character, I was considering `'g_0'`, `'#_0'`, etc.) I thought, the most reasonable choice is to keep the underscore for readability and clarity that we are dealing with an index but omit the +/-/$/... completely. Since this is clearly a relevant choice and pretty subjective, I post this as a draft pull request and am looking forward to discussing this.

#### Implementation of MajoranaOp

`MajoranaOp` is largely copy&paste from `FermionicOp` with some more or less trivial adaptions. The main differences are

- there is no method for normal ordering as this makes no sense for Majorana operators
- there is the constructor method `MajoranaOp.from_fermionic_op` that converts a Fermionic operator to Majorana. Conversion follows the same convention as in openfermion, i.e. $f_k = (\gamma_{2k} + i \gamma_{2k+1})/2$ is the $k$-th annihilation operator expressed in terms of the Majoranas.
- `num_spin_orbitals` is truly meant to count the number of fermionic orbitals, hence indices run from 0 to `2 * num_spin_orbitals - 1` since every orbital with its creation/annihilation operators gets mapped to two Majorana operators.

I documented everything mostly in the exact same style as in the docstrings for `FermionicOp` - i.e. I believe that the documentation is taken care off, modulo the issue with arguments I opened in #1269 .

I also implemented tests for all methods (again mostly just copy&paste from the corresponding tests of `FermionicOp`).

#### Order of operator classes in documentation

In `__init__.py` I added `MajoranaOp` directly underneath `FermionicOp` which makes it the second operator to appear in the list in the API reference before `BosonicOp`. I guess this is debatable, since `BosonicOp` is clearly the more important operator, but then I think that it makes sense to group Majorana and Fermionic operators together. Alternatively, one could also put `BosonicOp` at the top, followed by `FermionicOp` and `MajoranaOp` which is the order in which one would most likely find them in a textbook, or simply opt for alphabetical order.

#### Changes to FermionicOp

I made two changes to `FermionicOp`:

- `FermionicOp._index_order` is turned into a classmethod so it can be called from within `MajoranaOp.index_order` as well (to avoid redundancy).
- I removed the reference to `FermionicOp.to_matrix` in the docstring since such a method does not seem to exist (unrelated to the Majorana changes)

All changes are summarized in the release note.


grossardt:
I changed `MajoranaOp.terms()` to return tuples `("+", index)` rather than `("", index)`. Mainly because I was afraid of breaking something by using the empty string label in `QubitMapper`. Also added a few more tests.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+
+        for key in keys:
+            if set(key) - allowed_chars:
+                raise QiskitNatureError(
+                    f"The key {key} is invalid. PolynomialTensor keys may only consists of `_` "
+                    "characters, for them to be expandable into a MajoranaOp."
+                )
+
+    @classmethod
+    def from_polynomial_tensor(cls, tensor: PolynomialTensor) -> MajoranaOp:
```
ftroisi, line 296: Please add a documentation string for this method.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+        if num_so is None:
+            self.num_spin_orbitals = (max_index + 1 if max_index % 2 else max_index + 2) // 2
+
+    @staticmethod
+    def _majorana_label(label: str) -> str:
+        """Converts a Fermionic label into a Majorana label."""
+        return label.replace("+", "").replace("-", "")
+
+    @classmethod
+    def _validate_polynomial_tensor_key(cls, keys: Collection[str]) -> None:
```
ftroisi, line 283: Please add a documentation string for this method.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+                if order:
+                    trms = next(trm for trm, _ in MajoranaOp({majorana_label: new_coeff}).terms())
+                    fermion_label, new_coeff = FermionicOp._index_order(trms, new_coeff)
+                    majorana_label = cls._majorana_label(fermion_label)
+                if simplify:
+                    majorana_label, new_coeff = cls._simplify_label(majorana_label, new_coeff)
+                data[majorana_label] += new_coeff
+        return cls(data, num_spin_orbitals=op.num_spin_orbitals)
+
+    def _permute_term(
```
ftroisi, line 397: Please add a documentation string for this method.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+        self, term: list[tuple[str, int]], permutation: Sequence[int]
+    ) -> list[tuple[str, int]]:
+        return [(action, permutation[index]) for action, index in term]
+
+    def compose(self, other: MajoranaOp, qargs=None, front: bool = False) -> MajoranaOp:
```
ftroisi, line 402: Please add a documentation string for this method.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+        if front:
+            return self._tensor(self, other, offset=False)
+        else:
+            return self._tensor(other, self, offset=False)
+
+    def tensor(self, other: MajoranaOp) -> MajoranaOp:
+        return self._tensor(self, other)
+
+    def expand(self, other: MajoranaOp) -> MajoranaOp:
+        return self._tensor(other, self)
```
ftroisi, line 417: Why do we need both? It should be enough to have: 
- tensor: `a_op.tensor(b_op)`
- expand: `b_op.tensor(a_op)`
right?

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+
+    @classmethod
+    def _tensor(cls, a: MajoranaOp, b: MajoranaOp, *, offset: bool = True) -> MajoranaOp:
```
ftroisi, line 420: Please add a documentation string for this method.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+                    new_data[new_label] += cf1 * cf2
+                else:
+                    new_data[new_label] = cf1 * cf2
+
+        new_op = a._new_instance(new_data, other=b)
+        if offset:
+            new_op.num_spin_orbitals = a.num_spin_orbitals + b.num_spin_orbitals
+        return new_op
+
+    def transpose(self) -> MajoranaOp:
```
ftroisi, line 437: Please add a documentation string for this method.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+                        majorana_label += " "
+                    odd_index = (i >> j) & 1
+                    index = 2 * int(term[2:]) + odd_index
+                    if odd_index:
+                        if term[0] == "-":
+                            coeff_power += 1
+                        else:
+                            coeff_power += 3
+                    majorana_label += f"_{index}"
+                new_coeff = 1j**coeff_power * coeff / 2 ** len(terms)
```
ftroisi, line 387: I'd remove the spaces (or add parenthesis) to clarify what is the power operation: `2**len(terms)`

qiskit_nature/second_q/operators/majorana_op.py
woodsp-ibm: @ftroisi Do bear in mind that methods that override a parent method we have setup to inherit the parent method docs by default so things do not need to be replicated/copied generally in such a case. Lint will complain if its truly missing a docstring overall. So for instance in the output (which you can download in the CI Summary artifacts) this method shows up as 

image

It might be valid to an explicit doctring in the derived class, if something special/specific to the subclass needed stating.  but then the parent docstring will not be included so any aspects that are relevant need to be copied.

grossardt:
@ftroisi Thanks for your comments!

> So my question is how would you map a `MajoranaOp` to thee qubit space? Are you planning to implement a mapper or perhaps to rely on the Fermioni mappers?

Yes, the whole point of this exercise is as a means to the end of implementing the ternary tree mapper, cf. #582 (which I have basically completed but need to add some reasonable tests before opening a PR).

Regarding the docstrings, as @woodsp-ibm has pointed out, there are generally no docstrings for classes that inherit and can use the parent docstring. I mainly followed 1:1 the implementation of FermionicOp here. I'm not sure if I like this either - I guess it keeps the code compact but needing to find the parent method to understand what a method is supposed to do can be a bit tedious. Nonetheless, I accepted it as a style choice and tried to be consistent.

Same goes essentially for the .tensor and .expand methods. Simply trying to be as close to FermionicOp as possible which has both these methods.

I have no strong feelings about any of those things, but would argue that if they are changed here one should possibly also discuss them for the other operators and the rest of qiskit-nature.

qiskit_nature/second_q/operators/majorana_op.py
grossardt: black doesn't allow for the form without spaces (that's probably how the spaces got there in the first place), but I added parentheses for clarification.

woodsp-ibm:
> I'm not sure if I like this either - I guess it keeps the code compact but needing to find the parent method to understand what a method is supposed to do can be a bit tedious

The methods are documented in the derived class as well. If you click on the `Details` in one of the checks that ard run by CI (you may need to click the `Show all checks` you can then click `Summary` at top left and in the right hand pane if you scroll down you will find a link to artifacts that are built by CI. The `documentation.zip` contains the html that is built so you can download and unzip that and you find a index.html that you can navigate from. The example image I posted above, was taken from such a download. You can see it its the `compose` method and that is says MajoranaOp in it - as you say you followed FermionicOp you can see that method is there in the live docs https://qiskit.org/ecosystem/nature/stubs/qiskit_nature.second_q.operators.FermionicOp.html#qiskit_nature.second_q.operators.FermionicOp.compose

grossardt:
@woodsp-ibm 
Yes, for the documentation that is clear. I was mainly talking about working with the code. Sorry, if I wasn't making that clear enough.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+
+    .. code-block:: python
+
+        from qiskit_nature.second_q.operators import FermionicOp, MajoranaOp
+        f_op = FermionicOp({"+_0 -_1": 1}, num_spin_orbitals=2)
+        m_op = MajoranaOp.from_fermionic_op(f_op)
+
+    Note that every term of the ``FermionicOp`` will result in :math:`2^n` terms in the
+    ``MajoranaOp``, where :math:`n` is the number of fermionic modes in the term. The conversion
+    uses the convention that
```
mrossinek, line 106: I had to read this twice. Can we rephrase it a little, maybe like this?
```suggestion
    Note that each ``FerminonicOp``-term consisting of :math:`n` expressions will result
    in a ``MajoranaOp``-term consisting of :math:`2^n` expressions. The conversion uses
    the convention that
```

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+        a_i^\dagger = \frac{1}{2}(\gamma_{2i} - i \gamma_{2i+1}) \,,
+
+    where :math:`a_i` and :math:`a_i^\dagger` are the Fermionic annihilation and creation operators
+    and :math:`\gamma_i` the Majorana operators.
+
+    .. note::
+
+        When creating a ``MajoranaOp`` from a ``PolynomialTensor`` using
+        :meth:`from_polynomial_tensor`, the underscore character :code:`_` is the only allowed
+        character in the keys of the ``PolynomialTensor``.
```
mrossinek, line 120: I think it would help, to add a small example here.
Maybe you can also separate this from the previous section with another heading along the lines of `Constructing from a PolynomialTensor` in the same way you titled the previous section w.r.t. the `FermionicOp`.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+
+    Adjoint
+
+    .. code-block:: python
+
+      MajoranaOp({"_0 _1": 1j}, num_spin_orbitals=2).adjoint()
+
+    .. note::
+
+        Since Majorana generators are self-adjoined, the adjoint of a ``MajoranaOp`` is the original
```
mrossinek, line 169: Why do you call them "generators" now? We should be consistent with our wording here.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+                        f"The index, {index}, from the label, {key}, exceeds twice the number of "
+                        f"spin orbitals, {num_so}."
+                    )
+
+        if num_so is None:
+            self.num_spin_orbitals = (max_index + 1 if max_index % 2 else max_index + 2) // 2
+
+    @staticmethod
+    def _majorana_label(label: str) -> str:
+        """Converts a Fermionic label into a Majorana label."""
```
mrossinek, line 279: I am not quite sure I agree with this. This does not do any proper conversion and merely replaces/removes the `+` and `-` prefixes. Is this just a utility that you use to be able to reuse `FermionicOp` implementations? If so, maybe be a bit more explicit about that fact in the docstring.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+        return label.replace("+", "").replace("-", "")
```
mrossinek, line 280: We can avoid the duplicate call to `replace` using [`str.translate`](https://docs.python.org/3/library/stdtypes.html?highlight=str%20translate#str.translate)` as we also do in other parts of `FermionicOp` and `BosonicOp`.
This case is a bit more tricky because we want to remove characters without replacement. Here is how to do that:
```suggestion
        trans = "".maketrans("", "", "+-")
        return label.translate(trans)
```

For another example check out the `FermionicOp.transpose` method.

If you happen to go with the `"g"` prefix rather than the empty string, you can use the following snippet:
```suggestion
        trans = "".maketrans("+-", "gg")
        return label.translate(trans)
```

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+
+            mat = tensor[key]
+
+            empty_string_key = ["" for _ in key]  # label format for Majorana is just '_<index>'
+            label_template = mat.label_template.format(*empty_string_key)
+
+            for value, index in mat.coord_iter():
+                data[label_template.format(*index)] = value
+
+        num_so = (tensor.register_length + 1) // 2
```
mrossinek, line 314: This is the same code as in my suggestion above but with the `+` and `// 2` operations reversed. Pick whichever one you like, but I guess we should be consistent to avoid confusion.

mrossinek (commented):
Alright, I left quite a number of comments but hopefully this gives you everything that you need :+1:

qiskit_nature/second_q/operators/majorana_op.py
ftroisi: @woodsp-ibm thanks for the explanation

qiskit_nature/second_q/operators/fermionic_op.py
```diff
@@ -539,6 +539,7 @@ def simplify(self, atol: float | None = None) -> FermionicOp:
 
         data = defaultdict(complex)  # type: dict[str, _TCoeff]
         # TODO: use parallel_map to make this more efficient (?)
+        #       (if this is done, apply equally to MajoranaOp.simplify())
```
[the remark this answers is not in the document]
grossardt: done

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+
+
+class MajoranaOp(SparseLabelOp):
+    r"""N-mode Majorana operator.
+
+    A ``MajoranaOp`` represents a weighted sum of Majorana fermion operator terms.
+    These terms are encoded as sparse labels, which are strings consisting of a space-separated list
+    of expressions. Each expression must look like :code:`_<index>`, where the :code:`<index>` is a
+    non-negative integer representing the index of the mode on which the Majorana
+    creation/annihilation operator is applied. The value of :code:`index` is bound by twice the
```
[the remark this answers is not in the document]
grossardt: Well, I would say it still makes sense, just that the (self-adjoint) creator = annihilator in the majorana case (i.e. particles are their own antiparticles), but I omitted it to avoid confusion. You're right that it's not really helpful for understanding the notation.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+    number of spin orbitals (``num_spin_orbitals``) of the operator (Note: since Python indices are
+    0-based, the maximum value an index can take is given by :code:`2 * num_spin_orbitals - 1`).
+
+    .. note::
+
+        For compatibility reasons (e.g. with mappers) :meth:`MajoranaOp.terms()` returns a list of
+        tuples of the form `("+", index)`, i.e. is compatible with the format for
```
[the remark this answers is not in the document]
grossardt: Yeah, I don't completely like it either. To be fair, it's not completely artificial, since any majorana operator is both a creator and annihilator, either + or - would be appropriate choices, but it's inconsistent with the "_index" labelling and hence confusing. I am not really sold on the "g" (or any other label) either though.
My main problem with the empty string in .terms() is the occurence in `QubitMapper.mode_based_mapping`. In line 260 of qubit_mapper.py there is a check for empty label for which I don't truly understand if it has any relevant use case. If this check can be safely omitted, then it's fine to use the empty string.
(On a side note, I feel that qubit_mapper needs an overhaul anyway, because some of the methods seem very specifically tailored to FermionicOp whereas this should be a base class which is applicable to any SparseLabelOp.)
```suggestion
        go back to tuples of the form `("", index)` compatible with the `_index` format and amend qubit_mapper accordingly when implementing the TernaryTreeMapper. (I would then assume that noting gets broken if all tests pass successfully.)
```

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+        from qiskit_nature.second_q.operators import MajoranaOp
+
+        op = MajoranaOp(
+            {
+                "_0 _1": .25j,
+                "_1 _0": -.25j,
+                "_2 _3": -.25j,
+                "_3 _2": .25j,
+            },
+            num_spin_orbitals=2,
```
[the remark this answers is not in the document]
grossardt: That's an excellent question. The main use case, as far as I am concerned, is to convert FermionicOp to MajoranaOp in order to then apply a Majorana based mapper (e.g. Ternary Tree). Then the MajoranaOp is just a different representation for electron orbitals, and the attribute still makes sense. But if there are more general use cases it probably does make more sense to change it to num_modes. I will make this change, if nobody objects.

qiskit_nature/second_q/operators/majorana_op.py
grossardt: Agreed and changed.

qiskit_nature/second_q/operators/majorana_op.py
grossardt: indeed. changed.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+
+    .. code-block:: python
+
+      MajoranaOp({"_0 _1": 1j}, num_spin_orbitals=2).adjoint()
+
+    .. note::
+
+        Since Majorana generators are self-adjoined, the adjoint of a ``MajoranaOp`` is the original
+        operator with all strings reversed, e.g. :code:`"_0 _1"` becomes :code:`"_1 _0"` in the
+        example above, and coefficients complex conjugated.
```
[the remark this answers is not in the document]
grossardt: fine with me.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+    @classmethod
+    def from_fermionic_op(
+        cls, op: FermionicOp, simplify: bool = True, order: bool = True
+    ) -> MajoranaOp:
+        """Constructs the operator from a :class:`~.FermionicOp`.
+
+        Args:
+            op: the :class:`~.FermionicOp` to convert.
+            simplify: whether to simplify the resulting operator.
+            order: whether to perform index ordering on the resulting operator.
```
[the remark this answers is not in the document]
grossardt: If you don't call `.index_order().simplify()` you will end up with really messy terms like `_0 _1 _0 _0 _1` (which would just simplify to `_0`), so effectively the default way to convert a FermionicOp to MajoranaOp would be `MajoranaOp.from_fermionic_op(op).index_order().simplify()` which is quite ugly. I would much prefer if the simplification is done by default.
I would rather ask whether there is any relevant situation in which one would *not* want to simplify. So, we might not need these arguments, but I do think that they should be true by default, i.e. just calling `.from_fermionic_op` should return a simplified `MajoranaOp`.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+    def index_order(self) -> MajoranaOp:
+        """Convert to the equivalent operator with the terms of each label ordered by index.
+
+        Returns a new operator (the original operator is not modified).
+
+        .. note::
+
+            You can use this method to achieve the most aggressive simplification.
+            :meth:`simplify` does *not* reorder the terms and, thus, cannot deduce
+            ``_0 _1 _2`` and ``_2 _0 _1 _0 _0`` to be identical labels.
```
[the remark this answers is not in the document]
grossardt: I hope the rephrasing made it clearer?

grossardt:
Made some changes regarding the more obvious comments. Will work in the rest later this week. Reverted the PR to draft until then.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+            collapse these two labels into one.
+
+        Returns:
+            The index ordered operator.
+        """
+        data = defaultdict(complex)  # type: dict[str, _TCoeff]
+        for terms, coeff in self.terms():
+            # index ordering is identical to FermionicOp, hence we call classmethod there:
+            label, coeff = FermionicOp._index_order(terms, coeff)
+            data[self._majorana_label(label)] += coeff
```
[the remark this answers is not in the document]
grossardt: That's a good point. I just noticed that BosonicOp implements the exact same method. Wouldn't it make sense to simply make this a method of the parent class `SparseLabelOp` instead? (Granted, the labels in SparseLabelOp are completely generic and hence the method can't be applied for any SparseLabelOp, but since it's a private method, I don't see much harm.)

qiskit_nature/second_q/operators/majorana_op.py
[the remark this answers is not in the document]
mrossinek: Yes, I agree that the `QubitMapper` implementation needs to be improved. We did overhaul them a while back but made some questionable choices that time around.
The whole `mode_based_mapping` method should not be part of the base class and instead be a specific subclass (e.g. `ModeBaseMapper`) which can then be subclassed by those mappers which actually use it, without having this method be artificially present on _all_ mappers.

Another thing we wanted to change was also the use of `staticmethod` since this broke the initial implementation attempt of the [Bonsai mapper](https://journals.aps.org/prxquantum/abstract/10.1103/PRXQuantum.4.030314).

For now, since your PR does not include any mapper utilties, let's stick to the empty string label and not touch the `QubitMapper` stack. I will open a separate issue in a moment to track the changes that are required in that module.

qiskit_nature/second_q/operators/majorana_op.py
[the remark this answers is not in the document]
mrossinek: I see your point. I recall that we wanted to avoid implied simplification in the electronic structure stack in the past because this caused problems when looking into https://github.com/qiskit-community/qiskit-nature/issues/48.

Maybe we can merge the `order` and `simplify` arguments though. Having both seems unnecessary. We could maybe only give the `simplify` argument (defaulting to `True`) and explain in the docstring that this will indeed perform both, `index_order` and `simplify`. Then, if a user does _not_ want to do one (or either) of those, they need to disable it and can then do things manually.

qiskit_nature/second_q/operators/majorana_op.py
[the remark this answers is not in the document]
mrossinek: I am not quite sure where you see this method in `BosonicOp`? :thinking:

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,523 @@
...
+
+        .. note::
+
+            You can use this method to achieve the most aggressive simplification.
+            :meth:`simplify` does *not* reorder the terms. For instance, using only :meth:`simplify`
+            will reduce ``_2 _0 _1 _0 _0`` to ``_2 _0 _1`` but cannot deduce these labels to be
+            identical to ``_0 _1 _2``.
+            Calling this method will reorder the latter label to
+            ``_0 _0 _0 _1 _2``, after which :meth:`simplify` will be able to correctly
+            collapse these two labels into one.
```
mrossinek, line 458:
```suggestion
            You can use this method to achieve the most aggressive simplification.
            :meth:`simplify` does *not* reorder the terms. For instance, using only :meth:`simplify`
            will reduce ``_2 _0 _1 _0 _0`` to ``_2 _0 _1`` but cannot deduce this label to be
            identical to ``_0 _1 _2``.
            Calling this method will reorder the former label to
            ``_0 _0 _0 _1 _2``, after which :meth:`simplify` will be able to correctly
            collapse these two labels into one.
```

qiskit_nature/second_q/operators/majorana_op.py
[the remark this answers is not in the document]
mrossinek: Thanks, I left one suggestion below. Please also add this as a unittest to ensure this example remains functioning :+1:

qiskit_nature/second_q/operators/majorana_op.py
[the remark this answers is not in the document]
mrossinek: Back-referencing https://github.com/qiskit-community/qiskit-nature/issues/1289.

qiskit_nature/second_q/operators/majorana_op.py
grossardt: You're right. Must have been tired when writing this.

qiskit_nature/second_q/operators/majorana_op.py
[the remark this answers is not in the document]
grossardt: Sorry, my mistake. I was under the wrong impression that BosonicOp._index_order is identical, but of course it uses bosonic commutation relations. I got rid of _majorana_label which is indeed unnecessary and keep the rest as it was (i.e. call FermionicOp._index_order)

grossardt:
That should cover all the points raised now. I hope I didn't forget any.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+
+        Returns:
+            The index ordered operator.
+        """
+        data = defaultdict(complex)  # type: dict[str, _TCoeff]
+        for terms, coeff in self.terms():
+            # index ordering is identical to FermionicOp, hence we call classmethod there:
+            label, coeff = FermionicOp._index_order(terms, coeff)
```
mrossinek, line 464: I was verifying that this works as intended which is when I ran into #1298.

In any case, I can verify that the following works:
```python
from qiskit_nature.second_q.operators import FermionicOp, MajoranaOp

fop = FermionicOp({"+_2 -_0 +_1": 1.0})

assert not fop.equiv(fop.index_order())

mop1 = MajoranaOp.from_fermionic_op(fop).index_order()
mop2 = MajoranaOp.from_fermionic_op(fop.index_order())

assert mop1.equiv(mop2)
```

That might be a nice thing to add as a unittest to ensure this continues to work.

mrossinek: The issue with the `anti_commutator` I found on a tangent to try and verify the following, which would make another great unittest:

```python
from qiskit_nature.second_q.operators import MajoranaOp

def anti_commutator(op_a, op_b):
    return (op_a @ op_b + op_b @ op_a).simplify(atol=0)

mop1 = MajoranaOp({"_0": 1})
mop2 = MajoranaOp({"_0": 1})

assert anti_commutator(mop1, mop2).equiv(MajoranaOp({"": 2}))
```

Once #1298 has been adressed, we can replace the custom `anti_commutator` method with the one from the `qiskit_nature.second_q.operators.commutators` module :+1:

mrossinek (commented):
Alright, this is looking very good!

I still have some nitpicks and a few more suggestions for unittests but coverage is otherwise very good and I think this is basically good to go.
I did find a semi-related bug (#1298) which we will need to address separately, but this should not hold up the merging of this PR :+1:

qiskit_nature/second_q/operators/majorana_op.py
grossardt: Thanks for checking this. I added these two things as unittests.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+            terms = [("", int(lbl[1:])) for lbl in label.split()]
+            yield (terms, self[label])
+
+    @classmethod
+    def from_terms(cls, terms: Sequence[tuple[list[tuple[str, int]], _TCoeff]]) -> MajoranaOp:
+        data = {" ".join(f"_{index}" for _, index in label): value for label, value in terms}
+        return cls(data)
+
+    @classmethod
+    def from_fermionic_op(cls, op: FermionicOp, simplify: bool = True) -> MajoranaOp:
```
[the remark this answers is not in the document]
grossardt: Good idea. No need to change anything in the unittests.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+        copy: bool = True,
+        validate: bool = True,
+    ) -> None:
+        """
+        Args:
+            data: the operator data, mapping string-based keys to numerical values.
+            num_modes: the number of modes on which this operator acts.
+            num_spin_orbitals: the number of spin orbitals. Providing :code:`num_spin_orbitals=n`
+                is equivalent to providing :code:`num_modes=2*n`. Ignored if ``num_modes`` is
+                provided.
```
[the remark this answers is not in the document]
grossardt: You are probably right and it's relatively pointless. The idea was to make the syntax compatible with FermionicOp. So if someone knows to initiate a FermionicOp with `FermionicOp(data, num_spin_orbitals=n)` they might expect `MajoranaOp(data, num_spin_orbitals=n)` to work the same way. But I think you are right, that it will cause more confusion than it helps. I also only noticed afterwards that VibrationalOp has a `num_modals` attribute rather than `num_spin_orbitals`, so I agree that this is unnecessary. I removed it again.

qiskit_nature/second_q/operators/majorana_op.py
```diff
@@ -0,0 +1,521 @@
...
+
+    .. note::
+
+        It is the users' responsibility, that in the above scenario, :code:`some_big_data` is not
+        changed after initialization of the ``MajoranaOp``, since the operator contents are not
+        guaranteed to remain unaffected by such changes.
+
+    **Construction from Fermionic operator**
+
+    The default way to construct a ``MajoranaOp`` is from an existing ``FermionicOp``:
```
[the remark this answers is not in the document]
grossardt: Good point. I like that.

releasenotes/notes/add-majoranaop-1cbf9d4a1d4c264e.yaml
```diff
@@ -0,0 +1,16 @@
...
+    to handle operators that are sums of tensor products of Majorana fermion operators.
+    This is needed for the implementation of the Ternary Tree Mapper, but might be useful
+    later on for other purposes as well.
+
+    Majorana operators use a string representation with underscore only, e.g. ``'_0 _1'``
+    corresponds to :math:`\gamma_0 \gamma_1` where there are twice the number of spin orbitals
+    operators satisfying :math:`\{\gamma_i,\gamma_j\} = 2 \delta_{ij}`.
+
+    Methods of :class:`~qiskit_nature.second_q.operators.MajoranaOp` are the same as for
+    :class:`~qiskit_nature.second_q.operators.FermionicOp` except for normal ordering, which is
```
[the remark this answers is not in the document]
grossardt: Better?

releasenotes/notes/add-majoranaop-1cbf9d4a1d4c264e.yaml
```diff
@@ -0,0 +1,16 @@
+---
+features:
+  - |
+    Adds a new operator class, :class:`~qiskit_nature.second_q.operators.MajoranaOp`
+    to handle operators that are sums of tensor products of Majorana fermion operators.
+    This is needed for the implementation of the Ternary Tree Mapper, but might be useful
+    later on for other purposes as well.
```
mrossinek, line 7:
```suggestion
```
I don't think this is needed as part of this release note.

releasenotes/notes/add-majoranaop-1cbf9d4a1d4c264e.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+    Majorana operators use a string representation with underscore only, e.g. ``'_0 _1'``
+    corresponds to :math:`\gamma_0 \gamma_1` where there are twice the number of spin orbitals
+    operators satisfying :math:`\{\gamma_i,\gamma_j\} = 2 \delta_{ij}`.
+
+    Methods of :class:`~qiskit_nature.second_q.operators.MajoranaOp` follow the same API as for
+    :class:`~qiskit_nature.second_q.operators.FermionicOp` except for normal ordering, which is
+    unnecessary. A Majorana operator can be created from a Fermionic operator using the
+    :meth:`~qiskit_nature.second_q.operators.MajoranaOp.from_fermionic_op`  class method.
```
mrossinek, line 16: If you want, including a small code snippet to show how this is done could be nice.

mrossinek (approved):
Relying heavily on my previous reviews, I think this LGTM :+1: 

Just two very minor things about the release note.

mrossinek (approved):
Thank you for this valuable contribution! :tada:

PR #1271, opened 2023-11-13
fix: compatibility with mypy==1.7.0

mrossinek:

### DIFF
--- a/qiskit_nature/second_q/formats/watson/watson_hamiltonian.py
+++ b/qiskit_nature/second_q/formats/watson/watson_hamiltonian.py
@@ -75,8 +75,8 @@ class WatsonHamiltonian:
                     yield value, tuple((-1) ** kinetic * (i + 1) for i in index)
         elif isinstance(array, SparseArray):
             coo = as_coo(array)
-            for value, *index in zip(coo.data, *coo.coords):
-                yield value, tuple((-1) ** kinetic * (i + 1) for i in index)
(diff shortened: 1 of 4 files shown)


qiskit_nature/second_q/formats/watson/watson_hamiltonian.py
```diff
@@ -75,8 +75,8 @@ def _iter_array(
                     yield value, tuple((-1) ** kinetic * (i + 1) for i in index)
         elif isinstance(array, SparseArray):
             coo = as_coo(array)
-            for value, *index in zip(coo.data, *coo.coords):
-                yield value, tuple((-1) ** kinetic * (i + 1) for i in index)
+            for value, *idx in zip(coo.data, *coo.coords):
```
mrossinek, line 78: Variable renamings such as this one avoid the reuse of the same variable name from an earlier context (even if it was in a different branch of an `if/else` structure) which confuses mypy.
In this particular instance it did not like the mixing of types `list` and `tuple`.

This is essentially the only change in this PR besides removing a now-unneeded type-ignore comment.

woodsp-ibm: I had to do something similar for ML at the end of last week qiskit-community/qiskit-machine-learning#713

I labelled it backportable so as to ensure CI will still run on stable - should we want/need to do any bug fix release

woodsp-ibm (approved):

PR #1276, opened 2023-11-14
Formatting error message

julenl:
An space was missing here as well.

### DIFF
--- a/qiskit_nature/second_q/circuit/library/ansatzes/ucc.py
+++ b/qiskit_nature/second_q/circuit/library/ansatzes/ucc.py
@@ -384,7 +384,7 @@ class UCC(EvolvedOperatorAnsatz):
             if any(n >= self.num_spatial_orbitals for n in self.num_particles):
                 if raise_on_failure:
                     raise ValueError(
-                        f"The number of spatial orbitals {self.num_spatial_orbitals}"
+                        f"The number of spatial orbitals {self.num_spatial_orbitals} "
                         f"must be greater than number of particles of any spin kind "
                         f"{self.num_particles}."
                     )
--- a/qiskit_nature/second_q/operators/spin_op.py
+++ b/qiskit_nature/second_q/operators/spin_op.py
@@ -243,7 +243,7 @@ class SpinOp(SparseLabelOp):
             other_spin = other.spin
             if spin != other_spin:
                 raise TypeError(
-                    f"Invalid operation between operators with different spin"
+                    f"Invalid operation between operators with different spin "
                     f"values. Found spin_1={spin} and spin_2={other_spin}."
                 )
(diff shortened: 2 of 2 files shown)


mrossinek:
Did you do a search if you find any more such cases besides this one?

julenl:
Yesterday I didn't but today, before I sent this one, I run a recursive grep on the circuit folder and only those two (the one from yesterday and this one) came up.

mrossinek:
Since you already have this search set up, why don't you run that on the entire source code rather than just the `circuit` module?

julenl:
Sure, I did that too, a short after I answered you.

```  bash
cd qiskit_nature
grep -R 'The number of spatial orbitals {' -R
second_q/circuit/library/initial_states/hartree_fock.py:      f"The number of spatial orbitals {self.num_spatial_orbitals} "
second_q/circuit/library/ansatzes/ucc.py    f"The number of spatial orbitals {self.num_spatial_orbitals}"
```
Only those two, by that syntax.

mrossinek:
Ah I see. I had assumed you would have done a search for:
- a non-whitespace character
- followed by the end of a string (i.e. `'` or `"`)
- followed by a new line
- followed by the continuation of that string (i.e. optional whitespace at the begging of a line and then `'` or `"`)
- followed by another non-whitespace character

That would probably require a little bit of regex magic so if that is not something you are familiar with, we can also leave it at that.

julenl:
With this snippet:

``` python
for entry in os.walk('qiskit_nature'):
    for filename in entry[2]:
        file = entry[0] + os.sep + filename
        if file.endswith('.py'):
            with open(file, 'r') as fh:
                text = fh.read()
            mymatch = re.search('f".*[^\s]"\n\s+f"[^\s].*"', text)
            if mymatch:
                print(file)
                print('  ', mymatch.group()
```
I get:
```
qiskit_nature/second_q/hamiltonians/quadratic_hamiltonian.py
   f"QuadraticHamiltonian("
            f"hermitian_part={repr(self.hermitian_part)}, "
qiskit_nature/second_q/circuit/library/ansatzes/ucc.py
   f"The number of spatial orbitals {self.num_spatial_orbitals}"
                        f"must be greater than number of particles of any spin kind "
qiskit_nature/second_q/operators/spin_op.py
   f"Invalid operation between operators with different spin"
                    f"values. Found spin_1={spin} and spin_2={other_spin}."
qiskit_nature/second_q/operators/electronic_integrals.py
   f"The reported register_length of your beta-alpha-spin tensor, {beta_alpha_len}"
                    f", does not match the alpha-spin tensor one, {alpha_len}."
```

But only the spin_op and ucc require changes.

woodsp-ibm (approved):
LGTM

Issue #1277, opened 2023-11-14
Inconsistent behavior of `HartreeFock` and `UCC` for fully occupied orbitals


julenl:
### Environment

- **Qiskit Nature version**: '0.7.0'
- **Python version**:  3.10.12
- **Operating system**: Ubuntu

### What is happening?

Trying to run a UCC calculation on noble gas dimmers does not seem to work.

### How can we reproduce the issue?

``` python
from qiskit_nature.second_q.drivers import PySCFDriver

driver = PySCFDriver( atom='He 0 0 0; He 0 0 5.2' )
problem = driver.run()

from qiskit_nature.second_q.mappers import JordanWignerMapper
mapper = JordanWignerMapper()

from qiskit_algorithms.optimizers import SLSQP
from qiskit_algorithms import VQE
from qiskit.primitives import Estimator
from qiskit_nature.second_q.circuit.library import HartreeFock, UCCSD

ansatz = UCCSD(
    problem.num_spatial_orbitals,
    problem.num_particles,
    mapper,
    initial_state=HartreeFock(
        problem.num_spatial_orbitals,
        problem.num_particles,
        mapper,
    ),
)

print(ansatz.num_parameters)
```
Returns:
```
ValueError: The number of spatial orbitals 2must be greater than number of particles of any spin kind (2, 2).
```

### What should happen?

The ansatz should work also for fully occupied orbitals.

### Any suggestions?

I have noticed that there is some incoherence between two similar validations:
- On file qiskit_nature/second_q/circuit/library/ansatzes/ucc.py it says:
  _"The number of spatial orbitals 2 must be greater than the number of particles of any spin kind (2,2)"_
  And the condition is: `if any(n >= self.num_spatial_orbitals for n in self.num_particles):`
   
- While on file qiskit_nature/second_q/circuit/library/initial_states/hartree_fock.py it says:
  _"The number of spatial orbitals 2 must be greater than or equal to the number of particles of any spin kind (2,2)"_
  And the condition is: `if any(n > self.num_spatial_orbitals for n in self.num_particles):`

The text on the second one says "_greater than or equal_" but the condition only has a ">", and it works.
But the ucc one says "_greater than_" and the condition is ">=", so it does not work for any noble gas.

I tried removing the "=" from the ">=" for ucc, but then the operator turns for some reason into a list at some point and crashes.

mrossinek:
[...]

PR #1286, opened 2023-11-27
UCC and fully occupied MOs (Fix for issue #1277)

julenl:
### Summary
Improved the functionality of UCC a little bit, to work if only one of the spin registers is fully occupied. 

### Details and comments
The configuration check looked for whether any number of alpha or beta orbitals (length of the spin register) was greater or equals than the number of particles.
Obviously, there cannot be more particles than spin orbitals, and in order to handle this case, I updated the previous `any(n >= ...` to `any(n > ...`. 
For handling case where **both** alpha and beta are the same as the number of particles (i.e. (2, 2) for the He2 molecule), I created a new raise condition, notifying that the function is still not implemented.
Implementing this option will probably involve dealing with the empty excitation list at line 339.
For the case where **either** alpha or beta is equals to the number of particles (i.e. H-He molecule), it should just pass through.

### DIFF
--- a/releasenotes/notes/fix-UCC-and-fully-occupied-MOs-c8fdf252a0607395.yaml
+++ b/releasenotes/notes/fix-UCC-and-fully-occupied-MOs-c8fdf252a0607395.yaml
@@ -1,6 +1,5 @@
 ---
 features:
   - |
-    Allow for building UCC
-    (:class:`qiskit_nature.second_q.circuit.library.ucc`) ansatze for systems
+    Allow for building :class:`.UCC` ansatze for systems
     where at least one of the spin registers is not fully occupied.
--- a/releasenotes/notes/fix-UCC-and-fully-occupied-MOs-c8fdf252a0607395.yaml
+++ b/releasenotes/notes/fix-UCC-and-fully-occupied-MOs-c8fdf252a0607395.yaml
@@ -1,12 +1,6 @@
 ---
-prelude: >
-    The UCC ansatz doesn't allow to calculate systems with either, or both of
-    the spin registers is fully occupied.
 features:
   - |
-    Allow for building UCC ansatze for systems where at least one of the spin
-    registers is not fully occupied.
-fixes:
-  - |
-    Fixes the :class:`qiskit_nature.second_q.circuit.library.ucc` to accept
-    systems with one spin register full.
+    Allow for building UCC
+    (:class:`qiskit_nature.second_q.circuit.library.ucc`) ansatze for systems
+    where at least one of the spin registers is not fully occupied.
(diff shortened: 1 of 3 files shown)


mrossinek (commented):
Thanks, this looks like a good start. Can you please also do the following changes?

- [ ] add a unittest which asserts that the behavior works as intended
- [ ] verify whether the docstrings on the section of which errors are being raised needs to be updated
- [ ] add a release note ([see here](https://github.com/qiskit-community/qiskit-nature/blob/main/CONTRIBUTING.md#adding-a-new-release-note) for instructions)

And finally:
- [ ] please also check all of the above for the `HartreeFock` initial state and add a unittest if it already works (just to be certain that it won't break in the future)

releasenotes/notes/fix-UCC-and-fully-occupied-MOs-c8fdf252a0607395.yaml
```diff
@@ -0,0 +1,12 @@
+---
+prelude: >
+    The UCC ansatz doesn't allow to calculate systems with either, or both of
+    the spin registers is fully occupied.
+features:
+  - |
+    Allow for building UCC ansatze for systems where at least one of the spin
+    registers is not fully occupied.
+fixes:
```
woodsp-ibm, line 9: For such a fix/change we would not have a prelude. This is combined to give an high level overview of significance for the release. I do not think in this case its warranted at all. Then I think either this a fix or a new feature (the latter meaning the UCC chose not to cater to this rather). Whatever seems, better I would have just one section rather than saying similar things around both fix and feature.

julenl: Done :white_check_mark:

mrossinek (approved):
Thanks, this looks good to me :+1: 

<hr/>

On a side note: I noticed that none of the errors which `_check_configuration` (and `_check_ucc_configuration`) _might_ raise when `raise_on_failure=True` are documented.. I guess this does not show up, because these are all part of internal methods, which get called under the hood in Qiskit's `BlueprintCircuit` (which is a base-class of these circuits incl. `HartreeFock`).

We should probably at some point look into addressing this documentation-wise.

releasenotes/notes/fix-UCC-and-fully-occupied-MOs-c8fdf252a0607395.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
+  - |
+    Allow for building UCC
+    (:class:`qiskit_nature.second_q.circuit.library.ucc`) ansatze for systems
```
mrossinek, line 5: Actually, I just noticed that this probably won't show up correctly unless we change it like so:
```suggestion
    Allow for building :class:`.UCC` ansatze for systems
```

mrossinek (approved):

PR #1292, opened 2023-12-01
Fix the `AngularMomentum` (again)

### LINKED ISSUE #1291: The orbital overlap is still not handled correctly by the `AngularMomentum` operator
### Environment

- **Qiskit Nature version**: main @ fa75ce1f337672e5f4255137efba9aeed3e956ae (also part of `0.7.1`)
- **Python version**: Any
- **Operating system**: Any

### What is happening?

#1273 unfortunately has not been fixed properly.

This issue resurfaced while I was looking into some more extended tests. Specifically, using the overlap matrix which I obtained for a `(6,4)` active space of O2:
```python
ovlpab = np.asarray(
    [
        [0.987256, -0.001123, 0.00006, -0.0],
        [-0.001123, -0.987256, -0.0, -0.00006],
        [0.000019, 0.000055, -0.3195, -0.931662],
        [0.000056, -0.000019, -0.931662, 0.3195],
    ],
)
```
Basically, comparing the computed `S^2` values for different occupations obtained from PySCF's FCI module and Qiskit Nature still produce vastly different results.

```python
import numpy as np

from qiskit_algorithms.observables_evaluator import estimate_observables
from qiskit_nature.second_q.circuit.library import HartreeFock
from qiskit_nature.second_q.mappers import JordanWignerMapper
from qiskit_nature.second_q.properties import AngularMomentum, ElectronicDensity

norb = 4
nelec = (4, 2)
ovlpab = np.asarray(
    [
        [0.987256, -0.001123, 0.00006, -0.0],
        [-0.001123, -0.987256, -0.0, -0.00006],
        [0.000019, 0.000055, -0.3195, -0.931662],
        [0.000056, -0.000019, -0.931662, 0.3195],
    ],
)

hf_density = ElectronicDensity.from_particle_number(norb, nelec, include_rdm2=True)
dm1a = hf_density.alpha["+-"]
dm1b = hf_density.beta["+-"]
dm2aa = hf_density.alpha["++--"]
dm2bb = hf_density.beta["++--"]
dm2ab = hf_density.alpha_beta["++--"]

ovlpaa = np.eye(ovlpab.shape[0])
ovlpbb = np.eye(ovlpab.shape[0])
ovlpba = ovlpab.T

# manual computation just like done inside of `pyscf.fci.spin_op`
ssz = (
    np.einsum("ijkl,ij,kl->", dm2aa, ovlpaa, ovlpaa)
    - np.einsum("ijkl,ij,kl->", dm2ab, ovlpaa, ovlpbb)
    + np.einsum("ijkl,ij,kl->", dm2bb, ovlpbb, ovlpbb)
    - np.einsum("ijkl,ij,kl->", dm2ab, ovlpaa, ovlpbb)
    + np.einsum("ji,ij->", dm1a, ovlpaa)
    + np.einsum("ji,ij->", dm1b, ovlpbb)
) * 0.25

dm2abba = -dm2ab.transpose(0, 3, 2, 1)  # alpha^+ beta^+ alpha beta
dm2baab = -dm2ab.transpose(2, 1, 0, 3)  # beta^+ alpha^+ beta alpha
ssxy = (
    np.einsum("ijkl,ij,kl->", dm2abba, ovlpab, ovlpba)
    + np.einsum("ijkl,ij,kl->", dm2baab, ovlpba, ovlpab)
    + np.einsum("ji,ij->", dm1a, ovlpaa)
    + np.einsum("ji,ij->", dm1b, ovlpbb)
) * 0.5
spin_square = ssxy + ssz

s = np.sqrt(spin_square + 0.25) - 0.5
mult = s * 2 + 1
print("FCI", spin_square, mult)

ang_mom = AngularMomentum(norb, ovlpab).second_q_ops()

mapper = JordanWignerMapper()
qubit_op = mapper.map(ang_mom)

hf_state = HartreeFock(norb, nelec, mapper)

result = estimate_observables(Estimator(), hf_state, qubit_op)
new_ss = result["AngularMomentum"][0]
new_s = np.sqrt(new_ss + 0.25) - 0.5
new_mult = new_s * 2 + 1
print("Qiskit", new_ss, new_mult)
```

This results in:
```
FCI 2.050648651787 3.0335778557914086
Qiskit 2.0 3.0
```

### What should happen?

Qiskit should produce the same result as the PySCF implementation.

### Any suggestions?

Basically, there are multiple problems:

- the formula $S^2 = S^- S^+ + S^z (S^z + 1)$ assumes that there are more alpha than beta spin particles. One needs to take the symmetric version to allow either case: $S^2 = 0.5 * (S^- S^+ + S^+ S^-) + S^z S^z$

- on top of that, the logic that was added in #1275 is actually flawed. Through mostly trial and error, I found a patch last night which I admittedly did not understand completely but it made all my test cases pass :eyes: Essentially, I ended up overwriting the coefficients of the one-body terms which result from normal-ordering the $S^+ S^-$ (and v.v.) product to all be 1 (instead of the coefficient that resulted from the overlap squared).

Today, I found out why (thanks also to @AlbertoBaiardi for the extensive discussions). The problem is, that $S^+$ and $S^-$ do _NOT_ need to the include the overlap as coefficients! Rather, we need to adapt our anti-commutation relation and weight this with the overlap. This is explained quite nicely in sections 4.1 and 4.2 of [this paper](https://www.tandfonline.com/doi/full/10.1080/00268976.2022.2126802?scroll=top&needAccess=true) of which I will include the relevant screenshots below:

screenshot_1701338968
screenshot_1701338987
screenshot_1701339006

Long story short, to fix this issue we need to:
- [x] use the symmetric formula for $S^2$
- ~revert the inclusion of the overlap as coefficients inside $S^+$ and $S^-$ (and $S^x$ and $S^y$ by extension)~ WRONG: see next comment
- ~integrate the non-unity anti-commutation relation directly into the `FermionicOp`~ also wrong

mrossinek:
### Summary

Fixes #1291 

### Details and comments

See also https://github.com/qiskit-community/qiskit-nature/issues/1291#issuecomment-1836417011 for some more details.

### DIFF
--- a/qiskit_nature/second_q/properties/angular_momentum.py
+++ b/qiskit_nature/second_q/properties/angular_momentum.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import logging
 from typing import Mapping
 
 import numpy as np
@@ -22,6 +23,8 @@ import qiskit_nature  # pylint: disable=unused-import
 from qiskit_nature.second_q.operators import FermionicOp
 
 from .s_operators import s_minus_operator, s_plus_operator, s_z_operator
+
+LOGGER = logging.getLogger(__name__)
 
 
 class AngularMomentum:
@@ -49,8 +52,6 @@ class AngularMomentum:
 
     Attributes:
         num_spatial_orbitals (int): the number of spatial orbitals.
-        overlap (np.ndarray | None): the overlap-matrix between the $\alpha$- and $\beta$-spin
-            orbitals. When this is `None`, the overlap-matrix is assumed to be identity.
     """
 
     def __init__(self, num_spatial_orbitals: int, overlap: np.ndarray | None = None) -> None:
@@ -61,7 +62,42 @@ class AngularMomentum:
                 is `None`, the overlap-matrix is assumed to be identity.
         """
         self.num_spatial_orbitals = num_spatial_orbitals
+        self._overlap: np.ndarray | None = None
         self.overlap = overlap
+
+    @property
+    def overlap(self) -> np.ndarray | None:
+        r"""The overlap-matrix between the $\alpha$- and $\beta$-spin orbitals.
+
+        When this is ``None``, the overlap-matrix is assumed to be identity.
+        """
+        return self._overlap
+
+    @overlap.setter
+    def overlap(self, overlap: np.ndarray | None) -> None:
+        self._overlap = overlap
+
+        if overlap is not None:
+            norb = self.num_spatial_orbitals
+            delta = np.eye(2 * norb)
+            delta[:norb,:norb] -= overlap.T @ overlap
+            delta[norb:,norb:] -= overlap @ overlap.T
+            summed = np.einsum("ij->", np.abs(delta))
+            if not np.isclose(summed, 0.0):
+                LOGGER.warning(
+                    "The provided alpha-beta overlap matrix is NOT unitary! This can happen when "
+                    "the alpha- and beta-spin orbitals do not span the same space. To provide an "
+                    "example of what this means, consider an active space chosen from unrestricted-"
+                    "spin orbitals. Computing <S^2> within this active space may not result in the "
+                    "same <S^2> value as obtained on the single-reference starting point. More "
+                    "importantly, this implies that the inactive subspace will account for the "
+                    "difference between these two <S^2> values, possibly resulting in significant "
+                    "spin contamination in both subspaces. You should verify whether this is "
+                    "intentional/acceptable or whether your choice of active space can be improved."
+                    " As a reference, here is the summed-absolute deviation of `S^T @ S` from the "
+                    f"identity: {summed}"
+                )
+
 
     def second_q_ops(self) -> Mapping[str, FermionicOp]:
         """Returns the second quantized angular momentum operator.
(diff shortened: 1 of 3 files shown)


qiskit_nature/second_q/properties/angular_momentum.py
```diff
@@ -49,20 +52,53 @@ class AngularMomentum:
...
         r"""
         Args:
             num_spatial_orbitals: the number of spatial orbitals in the system.
             overlap: the overlap-matrix between the $\alpha$- and $\beta$-spin orbitals. When this
-                is `None`, the overlap-matrix is assumed to be identity.
+                is ``None``, the overlap-matrix is assumed to be identity.
         """
         self.num_spatial_orbitals = num_spatial_orbitals
+        self._overlap: np.ndarray | None = None
         self.overlap = overlap
```
robertodr, line 66: For my education: isn't this overridden by the `overlap` property?

robertodr: I see it now (I think! 😺) this line will call the setter and apply the initialization logic. But then is there a point to type-hint `self._overlap` a union type? Couldn't it be just `np.ndarray`?

mrossinek: The first line here makes sure that pylint does not complain about the private `_overlap` attribute not being defined inside the `__init__` (because otherwise it would be defined inside the `property` method.
I am typing it for the sake of completeness :+1:

robertodr (approved):

PR #1299, opened 2023-12-07
fix: do not use `normal_order` as part of commutator methods

### LINKED ISSUE #1298: The `commutator` methods break API by using `normal_order`
### Environment

- **Qiskit Nature version**: main @ 55e6776726a71b53feade60f889242a067c2d30c
- **Python version**: Any
- **Operating system**: Any

### What is happening?

The methods provided by the `qiskit_nature.second_q.operators.commutators` module call `.normal_order()` even though they can*NOT* because this method is not part of the `SparseLabelOp` API. I am honestly very surprised how our linters did not catch this.

I found this while testing something with the `MajoranaOp` from #1270, but we can also reproduce it with the `SpinOp` which has no `normal_order` method.

```python
from qiskit_nature.second_q.operators import SpinOp
from qiskit_nature.second_q.operators.commutators import commutator

op = SpinOp({"X_0": 1.0})

print(commutator(op, op))
```

This results in:
```
  File "/home/oss/Files/Dev/Qiskit/qiskit-nature/main/tmp-no.py", line 6, in <module>
    print(commutator(op, op))
  File "/home/oss/Files/Dev/Qiskit/qiskit-nature/main/qiskit_nature/second_q/operators/commutators.py", line 48, in commutator
    return (op_a @ op_b - op_b @ op_a).normal_order().simplify(atol=0)
AttributeError: 'SpinOp' object has no attribute 'normal_order'
```

### What should happen?

The methods should work for any `SparseLabelOp`. This bug was introduced by #1210 which attempted to fix #1208.

@ftroisi we should figure out a different way of fixing that problem with the `BosonicOp`.

### Any suggestions?

_No response_

mrossinek:
### Summary

Fixes #1298

### Details and comments

In #1208 we had discussed the inclusion of `normal_order` into the commutator methods without realizing that this breaks API. Here, I simply remove this call and update the unittests accordingly.
Basically, this means that the user needs to normal-order the result themselves if needed (which is totally fine).

### DIFF
--- a/qiskit_nature/second_q/operators/commutators.py
+++ b/qiskit_nature/second_q/operators/commutators.py
@@ -45,7 +45,7 @@ def commutator(op_a: SparseLabelOp, op_b: SparseLabelOp) -> SparseLabelOp:
     Returns:
         The computed commutator.
     """
-    return (op_a @ op_b - op_b @ op_a).normal_order().simplify(atol=0)
+    return (op_a @ op_b - op_b @ op_a).simplify(atol=0)
 
 
 def anti_commutator(op_a: SparseLabelOp, op_b: SparseLabelOp) -> SparseLabelOp:
@@ -61,7 +61,7 @@ def anti_commutator(op_a: SparseLabelOp, op_b: SparseLabelOp) -> SparseLabelOp:
     Returns:
         The computed anti-commutator.
     """
-    return (op_a @ op_b + op_b @ op_a).normal_order().simplify(atol=0)
+    return (op_a @ op_b + op_b @ op_a).simplify(atol=0)
 
 
 def double_commutator(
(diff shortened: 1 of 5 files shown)


qiskit_nature/second_q/operators/commutators.py
```diff
@@ -45,7 +45,7 @@ def commutator(op_a: SparseLabelOp, op_b: SparseLabelOp) -> SparseLabelOp:
     Returns:
         The computed commutator.
```
ftroisi, line 46: I think what you said in the description: "Basically, this means that the user needs to normal-order the result themselves if needed (which is totally fine)" should appear in the docsctring of the `commutator`/`anti_commutator`. The user might be confused otherwise

ftroisi (commented):
@mrossinek Thanks a lot for taking care of this!

ftroisi (approved):

ftroisi (approved):

robertodr (approved):

Issue #1289, opened 2023-11-29
Improvement to the `qiskit_nature.second_q.mappers` module


mrossinek:
### What should we add?

A few things in the `mappers` module could be improved. I will try to summarize some thoughts below based on gathered observations. (See also this thread: https://github.com/qiskit-community/qiskit-nature/pull/1270#discussion_r1408833996.)

- [ ] the `QubitMapper.mode_based_mapping` method should really not be part of the base class. Instead, we could implement this as a `ModeBasedMapper` class which only the mappers that actually follow this implementation derive from (i.e. `JordanWignerMapper`, `ParityMapper`, `BravyiKitaevMapper`, and `DirectMapper`). This would avoid all other mappers to have this public method without it actually working (since they don't implement the `pauli_table` classmethod.

- [ ] we should revert the use of `classmethod` for the `pauli_table` in those mappers above, because this broke the initial implementation of the [Bonsai mapper](https://journals.aps.org/prxquantum/abstract/10.1103/PRXQuantum.4.030314) which was done externally. We should see if we can still keep the caching that was added in #699 as part of this change, to avoid a performance regression (#644).

I think the above two tasks could be combined into a single PR quite nicely.

- [ ] Performance: Rather than duplicating #771 I am simply referring to it here for completeness.

grossardt:
**Re: ModeBasedMapper**

What would you do with the `FermionicMapper`/`BosonicMapper` classes that are a layer in between right now? I guess something like `JordanWignerMapper` would have to inherit from both `FermionicMapper` and `ModeBasedMapper` then?

**Re: pauli_table**

So `pauli_table` should be an instance method, do I get that right?
Regarding the caching, I think one could simply do that on the level of the actual implementation of `pauli_table` by having a helper classmethod or staticmethod `_pauli_table` with `@lru_cache` decorator.

I need to do something similar for the `TernaryTreeMapper` which has an attribute for the priority of X, Y, Z and the return value of `pauli_table` depends on this attribute. So I simply introduced a staticmethod `_pauli_table` that takes one extra argument and wrap lru_cache around that instead of the original `pauli_table` classmethod.
Although that works with a class attribute, it would make a lot more sense to implement `pauli_table` as an instance method also for the `TernaryTreeMapper` and then have the priority be a regular attribute of an instance.

mrossinek:
> **Re: ModeBasedMapper**
> 
[...]

PR #1301, opened 2023-12-12
Improve mappers, add ModeBasedMapper

grossardt:
### Summary

Adds a new class `ModeBasedMapper` which implements the basic functionality of the mode based mappers (JW, BK, parity, direct mapper) through a Pauli table. Reverts `pauli_table` to instance method and removes all caching from the parent class. As discussed in #1289.

### Performance testing

I have performed the tests as in #545 and #644 with [code and more plots available here](https://github.com/grossardt/notebooks/blob/main/qiskit-scrap/qiskit-nature-improve-mapper-performance-test.ipynb). The results look like this:
Parity_operator
Number_operators_of_length_15

Essentially, the effect of caching the Pauli table methods seems marginal. As there appear to be some fringe cases, where caching is somewhat useful (e.g. performances almost doubles for the BK and Parity mappers when mapping large numbers of length 10 number operators but not for length 5 or 15) and no real harm is being done, I implement the caching on the level of the individual `pauli_table` implementations. For this, the instance method `pauli_table` simply calls the private static method `_pauli_table` which is cached. This way, one can keep the caching for the existing mappers, but can also implement new mappers that actually use that `pauli_table` is an instance method.


woodsp-ibm:
In looking at the number op timing in your notebook, it looks like #644, in that the timing includes the operator build out. Which is not the case in the parity op one where the operator is built before the call to map() rather than being built as the operator parameter.

I do not know how closely things were looked at before. What might be interesting to know is if you put timing in the code around the pauli table buildout how long does that take in comparison to the overall mapping. I imagine that varies a lot though right since the table is built based on the operator width (register size) but the mapping depends on how many terms the operator has. 

#644 seems to talk about auxiliary operators - if these are normally relatively few terms then the overhead of building the table for each may end up being more significant. And I guess the table being cached should save some memory too.

grossardt:
> In looking at the number op timing in your notebook, it looks like #644, in that the timing includes the operator build out. Which is not the case in the parity op one where the operator is built before the call to map() rather than being built as the operator parameter.
> 
> I do not know how closely things were looked at before. What might be interesting to know is if you put timing in the code around the pauli table buildout how long does that take in comparison to the overall mapping. I imagine that varies a lot though right since the table is built based on the operator width (register size) but the mapping depends on how many terms the operator has.

Yes, right. Based on the result my suspicion is that building the Pauli table is actually not the most expensive part, but it's a good idea to look at this explicitly. I will try to find some time to do that (but possibly only after Christmas).
 
> #644 seems to talk about auxiliary operators - if these are normally relatively few terms then the overhead of building the table for each may end up being more significant. And I guess the table being cached should save some memory too.

That's also a good point. Generally I just took these two examples from the previous issues without much questioning, but I think it would be a good idea to put some thought into coming up with some realistic and meaningful benchmark in some sense.

That being said: as I wrote above, I think there is no harm done by doing the caching (on the level of the individual mapper's `pauli_table` method).

My guess is that the compose and simplify calls in `mode_based_mapping` are quite expensive. I see if I can time that as well. Possibly, improving the efficiency of simplify would do more for the efficiency of the mappers as well, as compared to caching the `pauli_table`.

The main difference regarding the caching is that `sparse_pauli_operators` is not cached any longer, which probably explains the somewhat better performance of the old version even when compared to the new version with caching. However, the composition of the creation/annihilation operators in `sparse_pauli_operators` seems to not have a large effect overall.

### On a different but related note:

I was wondering what the point is of having these two separate methods, `pauli_table` and `sparse_pauli_operators`? As I understand it, mode based mapping is supposed to capture all those cases in which the single particle Fock space operators $a_j$, $a_j^\dagger$ are each mapped to some `SparsePauliOp`. It is obvious that these Pauli operators are then $(P_j \pm Q_j)/2$ where for the common mappers (JW, Parity, BK) $P_j$ and $Q_j$ are single Pauli string operators. I don't see, however, why `ModeBasedMapper` shouldn't be more general and accept any such mapping.

Now, `pauli_table` determines these single Pauli strings $P_j$ and $Q_j$, whereas `sparse_pauli_operators` takes over the combination to $a_j$ and $a_j^\dagger$, or $(P_j \pm Q_j)/2$, respectively. Does it really make sense to keep these two steps as separate methods? I can't think of use cases that would need $P_j$ and $Q_j$ as intermediate results in this process. If anything, it is a mere convenience to prevent repetition of these ~4-6 lines of code in the implementation for the child classes. On the other hand, combining the two methods into a single one that just returns the `sparse_pauli_operators` matching the creation/annihilation operators may allow for more efficient implementations that do not need to first build the operators $P_j$ and $Q_j$.

My personal sentiment is that I would expect `pauli_table` to return what `sparse_pauli_operators` returns, namely a list of Pauli operators corresponding to all the modes in the system, and it should for every mapper separately implement the most efficient way to obtain this list, with or without the detour via calculating $P_j$ and $Q_j$.

Neither  `pauli_table` nor `sparse_pauli_operators`  is currently called anywhere outside of `mode_based_mapping` (at least within qiskit-nature), so it would probably still be possible without too much harm to change this, but maybe there is a reason for the splitting that I am missing?

grossardt:
Happy New Year everyone! To continue where we left off, I changed the timing of the number operator to not include building the operators. That doesn't change much, however, even for the total time. So building the operators seems to take negligible time compared to the mapper, but caching seems to not have much benefit.
In an attempt to come up with some more practical test, I also built versions of the Fermi-Hubbard Hamiltonian with creation/annihilation operators transformed by a random unitary. I built n operators of size n for n = 1...9, and the result again shows only marginal benefit of the caching. (But there is this marginal advantage, so I would leave the caching in as is.)

For me, the only open question is the "different note" above about whether the `pauli_table` and `sparse_pauli_operators` methods should be merged.

qiskit_nature/second_q/mappers/logarithmic_mapper.py
```diff
@@ -143,7 +143,7 @@ def _logarithmic_encoding(
             op.chop()
             spin_op_encoding.append(op)
 
-        return tuple(spin_op_encoding)
+        return (spin_op_encoding[0], spin_op_encoding[1], spin_op_encoding[2], spin_op_encoding[3])
```
mrossinek, line 146: Not sure why this is necessary

qiskit_nature/second_q/mappers/mode_based_mapper.py
```diff
@@ -0,0 +1,143 @@
...
+    def pauli_table(self, register_length: int) -> list[tuple[PauliType, PauliType]]:
+        r"""Generates a Pauli-lookup table mapping from modes to Pauli operators or pairs of Pauli
+        operators.
+
+        This table is a list of tuples :math:`(P, Q)` of two Pauli operators, corresponding to the
+        real part :math:`P` and imaginary part :math:`Q` for the respective mode index. These Pauli
+        operators are used to construct the creation and annihilation operators
+        :math:`(P \pm i Q)/2`.
+
+        The generated table is processed by :meth:`.QubitMapper.sparse_pauli_operators`.
```
mrossinek, line 49:
```suggestion
        The generated table is processed by :meth:`.sparse_pauli_operators`.
```
This is actually part of `ModeBasedMapper` now, but since this is inside the same class, the link should resolve like this as well.

releasenotes/notes/improve-mappers-b55cb0ca5fd656e4.yaml
```diff
@@ -0,0 +1,18 @@
...
+    :class:`.ParityMapper`, and :class:`.DirectMapper` are now inheriting also from the new
+    class :class:`.ModeBasedMapper` to implement the :meth:`.ModeBasedMapper.mode_based_mapping`
+    method.
+  - |
+    :meth:`.ModeBasedMapper.pauli_table` is now an instance method of all :class:`.ModeBasedMapper`
+    subclasses (previously a class method of :class:`.QubitMapper`)
+  - |
+    Caching has been removed from the parent classes :class:`.QubitMapper` and
+    :class:`.ModeBasedMapper` and is now implemented exclusively for the
+    :meth:`.ModeBasedMapper.pauli_table` methods in the child classes.
```
mrossinek, line 18: You should also note the methods which were previously part of the public API of `QubitMapper` have been removed from it. This is unlikely to affect anyone (since it will not have worked on mappers that didn't use them) but it should be mentioned.

mrossinek (approved):
A few minor details, but overall this LGTM.

---

Regarding your question about `pauli_table` and `sparse_pauli_operators`: the separation is likely historical. I don't have a strong opinion on whether this is kept as is or not. However, changing it will require proper deprecation.

Maybe the approach suggested in #1340 could inform a potential new implementation.

qiskit_nature/second_q/mappers/logarithmic_mapper.py
grossardt: I changed this because otherwise mypy fails due to this issue: https://github.com/python/mypy/issues/7509

qiskit_nature/second_q/mappers/mode_based_mapper.py
grossardt: Thanks for spotting this.

releasenotes/notes/improve-mappers-b55cb0ca5fd656e4.yaml
grossardt: Good point. Done.

robertodr (approved):

releasenotes/notes/improve-mappers-b55cb0ca5fd656e4.yaml
```diff
@@ -0,0 +1,21 @@
+---
+features:
+  - |
+    The class :class:`.second_q.mappers.ModeBasedMapper` has been added to implement mode based
+    mapping via a Pauli table (previously part of :class:`.second_q.mappers.QubitMapper`).
+upgrade:
```
woodsp-ibm, line 6:
```suggestion
```
`upgrade` is supposed to be things an end user needs to do to upgrade (alter) their code for the release e.g. things have been removed and users code needs to be adjusted. I would just remove this and make it all part of the features statement - though honestly this is supposed to be more an end user facing statement around the changes where I am not sure how much, if anything, this internal change affects them or surfaces to them in any way. I imagine code they had before using the mappers works just as it did before.

grossardt: That explanation makes sense. I have removed the point about changes in caching, and the inheritance which I believe has no end user facing consequences. The removal of public methods as pointed out by @mrossinek should obviously stay. Based on your reasoning, the info that `pauli_table` is now an instance method instead of a class method should stay, I guess, in case someone is calling it on a class in their code.

mrossinek (approved):
Thanks for your patience and hard work on this! You substantially improved the design of our mappers with this PR :+1:

PR #1313, opened 2024-01-28
Add Ternary Tree Mapper

### LINKED ISSUE #582: JKMN / Neven / Ternary tree qubit mapping

grossardt:
### Summary

Adds a new mapper class `TernaryTreeMapper` to implement the Ternary Tree based mapping algorithm (cf. [arXiv:1910.10746](https://arxiv.org/abs/1910.10746)). Fixes #582 

Depends on the two previous pull requests #1270 and #1301 .


test/second_q/mappers/test_ternary_tree_mapper.py
```diff
@@ -0,0 +1,185 @@
...
+
+""" Test Ternary Tree Mapper """
+
+import unittest
+from test import QiskitNatureTestCase
+
+import numpy as np
+from qiskit.circuit import Parameter
+from qiskit.quantum_info import SparsePauliOp
+from qiskit.algorithms.eigensolvers import NumPyEigensolver
```
woodsp-ibm, line 21:
```suggestion
from qiskit_algorithms import NumPyEigensolver
```
qiskit.algorithms has been deprecated for a while as the code was moved out of Qiskit into a separate repo. While you can import it direct from the subfolder, all the algorithms are importable from the top level, so I suggest to change it to that. It failed across the board when I updated the branch and re-ran this as things are now tested against the latest Qiskit i.e. 1,0 in which qiskit.algorithms no longer exists having finally been removed following the earlier deprecation.

grossardt: Thanks, I missed this.

grossardt:
I just wanted to get back to this after I have been quite busy for the past year. I am slowly catching up and was wondering what the status was of this pull request? The way I recall, it was finished from my side and waiting for review. Is there anything I can do still, so that this can be merged?

PR #1325, opened 2024-02-01
Add slow_test as it's no longer part of Qiskit 1.0

woodsp-ibm:
### Summary

`slow_test` used to be part of qiskit.test but the whole test folder was removed for Qiskit 1.0. This adds a copy of that slow_test decorator locally so that things continue working and of course will still work with 0.45 and 0.46 where its deprecated.

Same change as done in qiskit-community/qiskit-algorithms#141

### Details and comments

(This seems to be the only apps repo that uses slow_test. I could not find any uses in ML, Opt or Fin.)

### DIFF
--- a/.pylintdict
+++ b/.pylintdict
@@ -194,6 +194,7 @@ formatter
 fortran
 fp
 fredkin
+func
 functionalities
 für
 gauopen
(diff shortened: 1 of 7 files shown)


woodsp-ibm:
It seems that algorithms already had a custom dict entry for `func` which is why similar code there passed. Anyway I added it here.

mrossinek (approved):

PR #1328, opened 2024-02-01
Update Qiskit parallel map usage and remove TextProgressBar

### LINKED ISSUE #1314: parallel_map import from qiskit needs amending
Qiskit has removed the `tools` folder and `parallel_map` was moved to `utils`

From qiskit release notes

```
The ``qiskit.tools`` module has been removed. This module was deprecated in Qiskit 0.46.0.
All the contents from this module have been removed except for the ``qiskit.tools.parallel_map``
function which now can be used from :func:`qiskit.utils.parallel_map` instead.
```
Perhaps what to do is to try the import from qiskit.tools, while suppressing any warning that would result if 0.46 is used, and in a catch import from utils. This will maintain compatibility and also work with 1.0, for which the present code that uses that will fail. Maybe put the try catch in some common place and import parallel map from there to save replicating it around - from what I can see parallel map is used in qeom and 2 ops builders (with just comments in other places to use it)

### LINKED ISSUE #1324: TextProgressBar is deprecated in Qiskit 0.46 and removed in 1.0
QEOM uses TextProgressBar which as the title says is deprecated in Qiskit 0.46 and removed in 1.0.

Its used to log progress, if INFO logging is enabled. My take would be to simply drop this logging there. 

There is also some suppression in test_logging with a comment around TextProgressBar. I guess that suppression can stay though arguable that could be removed too if desired.

woodsp-ibm:
### Summary

Closes #1314
Closes #1324

### Details and comments

I updated the code to import the `parallel_map` into qiskit_nature.utils and expose it as `_parallel_map` to avoid creating a new public method as such. The import tries in the new location, which exists from 0.46 and above, before it goes back to the former location. The former location still exists in 0.46 but raises a deprecation warning hence the order of trying the imports to avoid this.

We could update the min requirements of Qiskit to 0.46 to avoid this and directly import/use parallel map from the new location. I chose the path of broader compatibility and did not update requirements as 0.46 is not quite yet released.

I completely removed TextProgressBar usage as TextProgressBar is removed in 1.0.

Tests all pass locally on my machine where I have the main branch with what will end up in 1.0 installed. It also works with 0.45 as I tried that too as that s currently what CI is using until a newer version gets released. **Update**: 0.46 has since been released, nightly CI is now failing.

As parallel map is internal logic that was changed, and TextProgressBar just affects what comes out in logs if enabled, neither change warrants a release note I think, so I did not include any.

### DIFF
--- a/qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
+++ b/qiskit_nature/second_q/algorithms/excited_states_solvers/qeom.py
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2020, 2023.
+# (C) Copyright IBM 2020, 2024.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
@@ -28,8 +28,6 @@ from qiskit_algorithms import EigensolverResult, MinimumEigensolver
 from qiskit_algorithms.list_or_dict import ListOrDict as ListOrDictType
 from qiskit_algorithms.observables_evaluator import estimate_observables
 from qiskit.circuit import QuantumCircuit
-from qiskit.tools import parallel_map
-from qiskit.tools.events import TextProgressBar
 from qiskit.quantum_info import SparsePauliOp
 from qiskit.primitives import BaseEstimator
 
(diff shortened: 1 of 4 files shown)


1ucian0 (approved):
LGTM

The dual support for qiskit 0.* and 1.* looks good.

PR #1341, opened 2024-02-22
Revert Python 3.12.0 specific usage

### LINKED ISSUE #1327: Add Python 3.12 support
Qiskit 0.45.1 added Python 3.12 support, and its been added to Algorithms see qiskit-community/qiskit-algorithms#108 (although at the time of writing this algorithms has yet to officially release a version supporting this).

CI will likely need to be specific to 3.12.0 until an updated testtools is released - see qiskit-community/qiskit-algorithms#113 which goes into the failure using 3.12.1

- [X] #1338 adds 3.12 support but needs to limit CI to 3.12.0 for the reason above
- [X] #1341 (When possible change `3.12.0` to `3.12` to allow newer bug fix version of Python 3.12 to be used in CI)

woodsp-ibm:
### Summary

Closes #1327

It seems that the breaking change that caused test failure in 3.12.1, that had us pin to 3.12.0, has been reverted in 3.12.2. While the testtools has been fixed to accommodate that change, it has not yet had a release with the update in it, but given the change in Python has been reverted its no longer needed presently.

### Details and comments

Will need Branch Rules updating (again!)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -149,16 +149,16 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, 3.9, '3.10', 3.11, 3.12.0]
+        python-version: [3.8, 3.9, '3.10', 3.11, 3.12]
         include:
           - os: macos-latest
             python-version: 3.8
           - os: macos-latest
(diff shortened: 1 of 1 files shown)


mrossinek (approved):
Thanks!

Issue #255, opened 2021-06-30
Reverse Jordan-Wigner transformation


molar-volume:
### What is the expected behavior?
While **Jordan-Wigner transformation** is implemented in the `mappers` framework (`JordanWignerMapper`), the reverse transformation is missing.

Please implement reversed Jordan-Wigner transformation. 
Inspiration can be drawn from https://github.com/quantumlib/OpenFermion/blob/master/src/openfermion/transforms/opconversions/reverse_jordan_wigner.py

0sophy1:
Thank you for creating the issue! We are mentors of this event -  https://github.com/qiskit-community/quantum-hackathon-korea-21  and while handling the questions, we found there is no reverse Jordan Wigner function in Qiskit.

woodsp-ibm:
Do you have a use-case/example of situations where this reverse transformation would be useful for a particular algorithm or problem being solved?

0sophy1:
it can be used for making anti-symmetric groups. also https://github.com/quantumlib/OpenFermion-FQE/blob/e4261fb7de19ae64a2a608e7c136503088976ca9/src/fqe/openfermion_utils.py this doc shows another use-case of this function in the open fermion side. In this time we've received a question because they want to reverse encode the computed qubit state into a fermionic state.

BStar14:
[@GyeonghumKim](https://github.com/GyeonghunKim) implemented mapper
[https://github.com/molar-volume/qiskit-nature/blob/reverse_jw/qiskit_nature/mappers/second_quantization/reverse_jordan_wigner_mapper.py]
(https://github.com/molar-volume/qiskit-nature/blob/reverse_jw/qiskit_nature/mappers/second_quantization/reverse_jordan_wigner_mapper.py)

and test file
[https://github.com/molar-volume/qiskit-nature/blob/reverse_jw/test/mappers/second_quantization/test_reverse_jordan_wigner_mapper.py](https://github.com/molar-volume/qiskit-nature/blob/reverse_jw/test/mappers/second_quantization/test_reverse_jordan_wigner_mapper.py)

We would improve these codes and consider making relevant examples.

mrossinek:
#702 shows that we are in the process of migrating the `FermionicOp` into a new location.
As shown in #701 this will also relocate the mappers into a new location (namely `qiskit_nature.second_quantization.operators.fermionic.mappers`).

[...]

PR #1344, opened 2024-02-25
Implements a reverse_map() method for JordanWignerMapper().

MarcoBarroca:
### Summary
This implements a reverse_map() method in the JordanWignerMapper(). It allows users to go back from qubit operators into fermionic operators. Completes issue #255.

Also included a unit test.

### Details and comments
Instead of using the implementation linked in issue #255 I modified it to use the FermionicOp object for the mapping. Seemed like it would be easier for anyone to understand the code. Both implementations are equivalent.


woodsp-ibm:
I see there are failures which are not of this change but due to a newly released pylint version - I created #1346 as this needs fixing outside of this PR to get CI running again and passing on the code that is already in main here.

MarcoBarroca:
Sure, I'll fix the mypy error by pushing the complex inside FermionicOp.

MarcoBarroca:
Included the pylint fixes in this PR. Hope that's'okay.

woodsp-ibm:
I'll let this run so we can check things anyway if you like. (Once you have a PR that has been merged here it will not longer need a maintainers approval to run - its just on the first PR)

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
```diff
@@ -41,5 +43,63 @@ def pauli_table(cls, register_length: int) -> list[tuple[Pauli, Pauli]]:
             # c_x = np.asarray([0] * register_length, dtype=bool)
             pauli_table.append((Pauli((a_z, a_x)), Pauli((b_z, b_x))))
             # TODO add Pauli 3-tuple to lookup table
-
         return pauli_table
+
+    @classmethod
+    def reverse_map(cls, qubit_op: SparsePauliOp) -> FermionicOp:
+        """Maps a qubit operator ``SparsePauliOp`` back into the second
```
woodsp-ibm, line 50: I have a question around this. This seems to take qubit operator and maps it back regardless of what it really is. I get if I this is given a JW mapped Fermionic Op as a qubit operator it will reverse that. What happens to some other operator - does it always work and produce something even if its meaningless? Is there any check possible? Should we add any note here about this?

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
```diff
@@ -41,5 +43,63 @@ def pauli_table(cls, register_length: int) -> list[tuple[Pauli, Pauli]]:
...
+                coef_term *= -1j
+            elif target_pauli_op.phase == 2:
+                coef_term *= -1
+            elif target_pauli_op.phase == 3:
+                coef_term *= 1j
+            total_fermionic_op += coef_term * term_fermionic_op
+        return total_fermionic_op.normal_order()
+
+    @staticmethod
+    def invert_pauli_terms(sparse_pauli_op: SparsePauliOp) -> SparsePauliOp:
```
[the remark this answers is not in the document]
MarcoBarroca: Right now I think I’m the only one asking for this. I don’t see why we shouldn’t have this on the public API though. If Open Fermion can do this and thinks it’s relevant why shouldn’t Qiskit Nature?

woodsp-ibm:
Sorry my bad in the suggestion, if you did not correct it was 

> :class:``JordanWignerMapper`

should have been

> :class:\`.JordanWignerMapper\` where it was a . not a second \`

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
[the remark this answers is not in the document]
woodsp-ibm: We are talking just about the `invert_pauli_terms` method right - that was all I was meaning as being private not the whole reverse map method

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
MarcoBarroca: It will work for any operator that is a SparsePauliOp and write that in terms of fermionic operators. But if a user tries to reverse an operator that was mapped with a different mapper(I.e. ParityMapper()) they will just get different ones.

It can also take any qubit operator and turn into a fermionic Op even if it doesn’t make sense, so it will always give a result.

Maybe a note making it clear that this works for JW? I wouldn’t know how to check if an qubit operator was transformed via Jordan Wigner or something else.

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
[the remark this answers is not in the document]
MarcoBarroca: Oh! my bad. Sure, I can turn that into a private method. It's just a utility for the reverse_map().

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
woodsp-ibm: A note sounds like the best way then - that says it users responsibility to provide an operator that was mapped with JW to get a meaningful outcome as this does the reverse mapping for any operator on the given assumption it was done with a JW mapping in the first place - or however you want to phrase it,.

MarcoBarroca:
@woodsp-ibm Let me know if you think the note is okay when you can.

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
```diff
@@ -41,5 +43,65 @@ def pauli_table(cls, register_length: int) -> list[tuple[Pauli, Pauli]]:
...
             # TODO add Pauli 3-tuple to lookup table
-
         return pauli_table
+
+    @classmethod
+    def reverse_map(cls, qubit_op: SparsePauliOp) -> FermionicOp:
+        """Maps a qubit operator ``SparsePauliOp`` back into the second quantized
+        operator ``FermionicOp``. While it'll provide an output for any ``SparsePauliOp``
+        operator, it should be used on operators that were created with
+        ``JordanWignerMapper`` to ensure accurate results.
```
woodsp-ibm, line 53:
```suggestion
        ``JordanWignerMapper`` to ensure a meaningful result.
```
Rather than talk about accuracy - is it more a question of whether the result is meaningful or not?

woodsp-ibm:
@mrossinek Can I get you to take a look at this too.

MarcoBarroca:
Hey, just wanted an update here. Is there anything I should do about this still?

woodsp-ibm:
The build errors with mypy are fixed by #1362, which is still open. As far as progressing this hopefully @mrossinek can take a look.

MarcoBarroca:
Saw that this repo is seeing some activity recently. Let me know if you need anything from me to get this merged.

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
```diff
@@ -44,5 +46,65 @@ def _pauli_table(register_length: int) -> list[tuple[PauliType, PauliType]]:
...
+            total_fermionic_op += coef_term * term_fermionic_op
+        return total_fermionic_op.normal_order()
+
+    @staticmethod
+    def __invert_pauli_terms(sparse_pauli_op: SparsePauliOp) -> SparsePauliOp:
+        """Utility to invert the order of Pauli operators in each term of a SparsePauliOp."""
+        inverted_labels = [label[::-1] for label in sparse_pauli_op.paulis.to_labels()]
+        # Create a new SparsePauliOp with the inverted labels but same coefficients
+        inverted_sparse_pauli_op = SparsePauliOp(inverted_labels, sparse_pauli_op.coeffs)
+        return inverted_sparse_pauli_op
```
ludmilaasb:
```suggestion        
        return SparsePauliOp(inverted_labels, sparse_pauli_op.coeffs)
```

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
```diff
@@ -44,5 +46,65 @@ def _pauli_table(register_length: int) -> list[tuple[PauliType, PauliType]]:
...
+                    ferm_op_pauli = FermionicOp({f"+_{i}": -1j, f"-_{i}": 1j})
+                    target_pauli_op = Pauli("I" * (i + 1) + "Z" * (num_qubits - i - 1)).compose(
+                        target_pauli_op
+                    )
+                else:
+                    ferm_op_pauli = FermionicOp.one()
+                ferm_term_ops.append(ferm_op_pauli)
+            term_fermionic_op = FermionicOp.one()
+            for op in ferm_term_ops:
+                term_fermionic_op = term_fermionic_op @ op
```
ludmilaasb, line 94: If I understood it correctly, you're are dividing the logic of this loop (up to the line I am commenting) to 3 parts:

1.  Instantiating a coefficient and a list containing the fermionic terms mapped from the qubit operator
2. Populate the list with the fermionic operators
3. Multiply all the elements of this list by each other

I think this can be simplified. Instead of having a list as intermediate step where you append each time a new operator, you could just do `term_fermionic_op = term_fermionic_op @ op` as you're doing at line 90. What do you think?

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
```diff
@@ -44,5 +46,65 @@ def _pauli_table(register_length: int) -> list[tuple[PauliType, PauliType]]:
...
+            the Hamiltonian in the Fermionic space.
+        """
+        num_qubits = (
+            qubit_op.num_qubits
+        )  # get number of qubits from input second quantized operator
+        qubit_op = cls.__invert_pauli_terms(qubit_op)
+        total_fermionic_op = FermionicOp.zero()
+        for term in qubit_op:
+            coef_term = term.coeffs[0]
+            target_pauli_op = term.paulis[0]
```
ludmilaasb, line 72: Why just the first entries of coeffs and paulis? What is the rest?

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
```diff
@@ -44,5 +46,65 @@ def _pauli_table(register_length: int) -> list[tuple[PauliType, PauliType]]:
...
+                ferm_term_ops.append(ferm_op_pauli)
+            term_fermionic_op = FermionicOp.one()
+            for op in ferm_term_ops:
+                term_fermionic_op = term_fermionic_op @ op
+            if target_pauli_op.phase == 1:
+                coef_term *= -1j
+            elif target_pauli_op.phase == 2:
+                coef_term *= -1
+            elif target_pauli_op.phase == 3:
+                coef_term *= 1j
```
ludmilaasb, line 100: You don't need to use flow control for this, a lookup table is enough. 

```suggestion
            coef_term *= ( -1j, -1, 1j)[target_pauli_op.phase]
```

test/second_q/mappers/test_jordan_wigner_mapper.py
```diff
@@ -142,6 +142,19 @@ def test_mapping_overwrite_reg_len(self):
         mapper = JordanWignerMapper()
         self.assertEqual(mapper.map(op, register_length=3), mapper.map(expected))
 
+    @unittest.skipIf(not _optionals.HAS_PYSCF, "pyscf not available.")
+    def test_reverse_map(self):
+        """Test reverse mapping from qubit operator back to fermionic operator."""
+        driver = PySCFDriver()
+        driver_result = driver.run()
+        fermionic_op, _ = driver_result.second_q_ops()
```
ludmilaasb, line 150: Maybe this could be a fixture? I don't think the driver or the fermionic operator are been modified.

Mind you that just lines 153 and 156 are the actual test. Everything else is just preparing the data around it. So if something happens with any of those lines, your test will fail before reaching what you really want to check.

test/second_q/mappers/test_jordan_wigner_mapper.py
```diff
@@ -142,6 +142,19 @@ def test_mapping_overwrite_reg_len(self):
         mapper = JordanWignerMapper()
         self.assertEqual(mapper.map(op, register_length=3), mapper.map(expected))
 
+    @unittest.skipIf(not _optionals.HAS_PYSCF, "pyscf not available.")
+    def test_reverse_map(self):
```
ludmilaasb, line 146: I would like to see more test cases. Here you just taking one operator and doing a round-trip. 

What happens if you test an empty operator? An identity? Are sure the phase of the new operator is correct? Things like this.

ludmilaasb (changes requested):
Hello @MarcoBarroca !

Thank you so much for your PR! This is a very nice and useful feature, having map that goes from the Qubit space to Fermionic space!

I just left few comments, mostly nit-picks. I think the tests can be improve a bit more, so we can catch more edge cases and be sure the new method you wrote is fine.

Bom trabalho! 🚀

test/second_q/mappers/test_jordan_wigner_mapper.py
```diff
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2021, 2023.
+# (C) Copyright IBM 2021, 2024.
```
ludmilaasb, line 3:
```suggestion
# (C) Copyright IBM 2021, 2026.
```

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
ludmilaasb: Or this (even better!!!)

```suggestion
(-1j) ** phase
```

MarcoBarroca:
Hi! I'll take a look at these suggestions this week. Thanks for the feedback!

qiskit_nature/second_q/mappers/jordan_wigner_mapper.py
```diff
@@ -44,5 +46,64 @@ def _pauli_table(register_length: int) -> list[tuple[PauliType, PauliType]]:
...
+                pauli_char = one_pauli.to_label()
+                if pauli_char == "Z":  # dealing Pauli Z op
+                    ferm_op_pauli = FermionicOp({"": 1, f"+_{i} -_{i}": -2})
+                elif pauli_char == "X":  # dealing Pauli X op
+                    ferm_op_pauli = FermionicOp({f"+_{i}": 1, f"-_{i}": 1})
+                    target_pauli_op = Pauli("I" * (i + 1) + "Z" * (num_qubits - i - 1)).compose(
+                        target_pauli_op
+                    )
+                elif one_pauli.to_label() == "Y":  # dealing Pauli Y op
+                    ferm_op_pauli = FermionicOp({f"+_{i}": -1j, f"-_{i}": 1j})
```
adamglos92: Hi! largely all looks great (though I agree with comments from @ludmilaasb) the only quick question I have is that if I'm not mistaken this ferm_op_pauli should be negated? Otherwise we will be getting `-YZZZZ`

MarcoBarroca:
Thanks @ludmilaasb and @adamglos92 for the comments! I pushed some changes addressing the feedback. After including some unit tests, I found that there were some issues with the original implementation that should now be corrected. (Among them, the sign-flipping issue you pointed out.)

Implementation changes:

- Iterate directly over `qubit_op.paulis` and `qubit_op.coeffs`, removing the ambiguous `[0]` accesses.
- Multiply the fermionic factors inline instead of collecting them in an intermediate list.
- Remove the Pauli-label inversion helper. Qiskit's native `Pauli[i]` indexing now identifies each mode directly, while processing modes from high to low handles the Jordan–Wigner parity ordering.
- Simplify phase handling to `(-1j) ** target_pauli_op.phase`.
- Preserve `num_spin_orbitals`, including for zero and identity operators.
- Correct the inverse `Y` mapping. With this mapper's convention,
  `Y_j Z_{<j} = i(a†_j - a_j)`, so the creation coefficient must be `+i` and the annihilation coefficient `-i`. The previous signs produced `-Y`.

Test changes:

- Move the PySCF/H₂ preparation into a shared class fixture used by both the forward-mapping and reverse-mapping H₂ tests.
- Retain the H₂ fermion → qubit → fermion round-trip test.
- Add exact tests for `I`, `X`, `Y`, and `Z`.
- Add zero and identity tests, including register-length preservation.
- Add multi-qubit and complex-coefficient cases exercising all four internal Pauli phases.
- Add creation-operator and orbital-ordering round trips.
- Update the copyright year to 2026.

adamglos92 (approved):
Thanks @MarcoBarroca nothing to add in here, currently it looks very good to me

ludmilaasb (approved):
LGTM!

PR #1371, opened 2024-12-22
Replace np.alltrue() with np.all()

iyanmv:
### Summary

`numpy.alltrue()` was deprecated in NumPy 1.25.0 and removed in NumPy 2.0. Because of this, `test_h2` fails on recent versions of NumPy.

This PR simply replaces `np.alltrue()` with `np.all()` as recommended by the [NumPy 2.0 migration guide](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#main-namespace). Test will continue to work on older versions of NumPy.

### Details and comments

Fixes: https://github.com/qiskit-community/qiskit-nature/issues/1370

### DIFF
--- a/test/second_q/mappers/test_bksf_mapper.py
+++ b/test/second_q/mappers/test_bksf_mapper.py
@@ -1,6 +1,6 @@
 # This code is part of a Qiskit project.
 #
-# (C) Copyright IBM 2021, 2023.
+# (C) Copyright IBM 2021, 2025.
 #
 # This code is licensed under the Apache License, Version 2.0. You may
 # obtain a copy of this license in the LICENSE.txt file in the root directory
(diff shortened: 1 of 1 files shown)


ftroisi (commented):
It looks good to me, thanks!

@mrossinek @woodsp-ibm it looks like the pipelines are failing for three reasons:
1. The macos pipelines are failing because the Conda environment cannot be created correctly. On Windows and Ubuntu that step works.
2. For all pipelines, it looks like the package `networkx` is using `np.float_`, which is not supported in `numpy>=2.0.0`. From the pipeline logs it looks like we are using `networkx==2.5.1`, whereas the last version is the 3.4.2:
```
Collecting networkx<2.6,>=2.2 (from -r requirements-dev.txt (line 19))
  Downloading networkx-2.5.1-py3-none-any.whl.metadata (4.9 kB)
```
This error is likely to be fixed by changing the `requirements-dev.txt`.
3. Both pylint and mypy are failing due to errors unrelated to this PR. I can take care of that in another PR

Issue #1370, opened 2024-12-22
TestBravyiKitaevSuperFastMapper::test_h2 fails with NumPy >= 2.0


iyanmv:
### Environment

- **Qiskit Nature version**:  0.7.2
- **Python version**:  3.13.1
- **Operating system**:  Arch Linux

### What is happening?

The test `TestBravyiKitaevSuperFastMapper::test_h2` from `test/second_q/mappers/test_bksf_mapper.py` fails because it uses `np.alltrue()`, which was deprecated in [NumPy 1.25.0](https://numpy.org/doc/stable/release/1.25.0-notes.html#deprecations) and removed in [NumPy 2.0](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#main-namespace).

### How can we reproduce the issue?

Try to run the tests with a recent version of NumPy.

### What should happen?

Test works on supported versions of NumPy

### Any suggestions?

Replace `np.alltrue()` with `np.all()` as explained in the NumPy 2.0 migration guidelines.

PR #1374, opened 2025-01-15
Pylint and Mypy fixes to solve CI errors. Remove Python 3.8 support

ftroisi:
### Summary
This PR originates from #1371, and fixes `pylint` and `mypy` errors in the CI. Also, the package `networkx` had to be updated to support `numpy 2.x`.
Since tests must pass, PR #1371 has been incorporated in this PR. Thanks to @iyanmv for opening that PR.

### Details and comments
Fixes: #1370 

Note that since MacOS 14, github-hosted runners do not have Conda bundled. Thus, on the `macos-latest` pipeline, we have to install Conda (https://github.com/conda-incubator/setup-miniconda/issues/344). A sincere thanks to @EndBug for the discussion on this topic.

Python 3.8 reached EOL last year. Since there were `pylint` errors due to this old version and `qiskit` no longer supports Python 3.8, this PR deprecates Python 3.8 support


.github/workflows/main.yml
```diff
@@ -149,7 +149,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8, 3.9, '3.10', 3.11, 3.12]
+        python-version: [3.8, 3.9, 3.10, 3.11, 3.12]
```
woodsp-ibm, line 152: As you can see from the failure the `s around 3.10 were intentional otherwise its interpreted as 3.1 something none of the other numbers have that issue with, While I guess they all could have been quoted it was only needed for 3.10.

woodsp-ibm:
Looks fine to me overall. One comment I have about the title and text is that this does not deprecate 3.8 support it removes it completely. Deprecation keeps support but issues warnings about the intent to remove it in an upcoming version. Since its changed the package from the user perspective in that regard the PR **should have a release note** stating that 3.8 support has been removed such that when a release is generated, including this, that this comes out in the release notes documentation. See https://github.com/qiskit-community/qiskit-nature/blob/main/CONTRIBUTING.md#release-notes for creating release notes (you can see examples in past PRs too if you need any).

I see you noted networkx requirement needed updating for numpy 2. I cannot recall any longer why it was pinned - I assume some compatibility related problem. Just noting that I think CI will install the latest unless constrained otherwise so maybe the entry in requirements-dev is not strictly needed. Commenting as I see the main requirements.txt file whose contents affect a normal install state numpy >=1.17 such that I guess if a user has an older numpy than 2.0 it still works. (I checked qiskit just because its a major dependency and it has the same min version)

woodsp-ibm:
I will note the branch rules here will need updating such that new jobs are made **Required** as needed and older jobs, no longer run, get removed from being Required. This can be done just ahead of this being merged so the rules match what is needed to pass this given the CI changes here.

ftroisi:
@woodsp-ibm I changed the name to reflect the removal of Python 3.8 and added a release note. I have also removed `networkx` from `requirements-dev`, as it gets the latest version from other dependencies.

It should be all good now, I am just waiting for approval and your update on the branch rules.

woodsp-ibm (approved):

woodsp-ibm:
@ftroisi FYI I see the nightly scheduled CI, that was run after this was merged, failed in unbuntu 3.12 with psi4, though it had passed here. You can see the CI runs under "Actions"  its here https://github.com/qiskit-community/qiskit-nature/actions/runs/12960770104 Not sure if that was just an intermittent and/or related to the psi4 errors you had before.

ftroisi:
@woodsp-ibm Thanks for the message, the error was about a temp folder which could be found (`FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/qiskit-nature/qiskit-nature/psi.3594.clean'`). 
Nothing related to the errors I fixed in the PR. Anyways, as you said let's wait and see if it happens again

PR #1398, opened 2026-05-22
Pass explicit matrix shape to `sparse.COO` initialization

matteoacrossi:
### Summary

Pass matrix shape explicitly to `sparse.COO` initialization.

### Details and comments

Newer versions of `sparse` require explicitly setting the shape of COO matrices.


qiskit_nature/second_q/operators/vibrational_integrals.py
matteoacrossi, line 80: @robertodr could you double check that this is correct?

robertodr (dismissed):

robertodr (approved):

